PR #9, opened 2018-10-02
PXF Building jars and configurations

divyabhargov:
As part of this PR, PXF does not need to have any hadoop client installed. The configuration files that we are shipping are minimal. User impersonation is still enabled, so the users would have to modify the configuration on the server side hadoop before using PXF.


shivzone:
```make clean``` unlike before doesn't seem to cleanup the top level build directory anymore now.

File: server/build.gradle [hunk 1]
```diff
@@ -640,12 +353,17 @@ task install(type: Copy, dependsOn: [subprojects.build, tomcatGet]) {
...
     baseName = "${project.name}"
     compression = Compression.GZIP
     extension = 'tar.gz'
-    from install.outputs.files
-    into "${project.name}"
+    destinationDir = file("${buildDir}/dist")
+    from assemble.outputs.files
+}
+
+task install (type: Copy, dependsOn: [tar]) {
```
shivzone: install should only depend on ```assembly``` task. It seems to be an overkill to create a tar and again extract the tar into the destination deploy path.

PR #15, opened 2018-10-09
Allow PXF_USER_IMPERSONATION to be overridden by env

lavjain:
Also, let the default hadoop configurations override pxf confs


benchristel:
:+1: Looks good!

lavjain:
@shivzone I will create a new story to get rid of the gradle profiles.

PR #19, opened 2018-10-11
Remove version plugin in build.gradle

frankgh:
The version plugin breaks the PXF pipelines

### DIFF
--- a/concourse/scripts/compile_pxf.bash
+++ b/concourse/scripts/compile_pxf.bash
@@ -12,7 +12,6 @@ _main() {
 	export JAVA_TOOL_OPTIONS=-Dfile.encoding=UTF8
 	pushd pxf_src/server
 		make install
-		make version > ${PXF_HOME}/version
 	popd
 	# Create tarball for PXF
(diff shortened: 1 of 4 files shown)


frankgh:
green pipeline; https://ud.ci.gpdb.pivotal.io/teams/main/pipelines/dev:pxf_fguerrero-remove-pxf-version_5x

frankgh:
I think it was breaking pipelines because of the way it was setting the version number in the generated jars. I think we should spend some more time on integrating the plugin and testing it through our pipelines.

benchristel:
@lavjain is it okay if we try to go back to using this plugin, which generates the version number from git tags?

Basically, the way it works is:
- if you build from a commit that is tagged, the version number will be the name of the tag
- if you build from a commit that is not tagged, the plugin searches back through the git history for the most recent tag, and bases the version number on that using a format like:

  ```
  4.0.3-12-43a3bf90
  ```

  That is: the name of the tag, the number of commits since that tag, and the current commit SHA.

This may affect our release process so I wanted to keep you in the loop.

PR #25, opened 2018-10-16
Use loginUser to access HiveMetaStore

frankgh:
When Kerberos is enabled we need to access the HiveMetaStore as the loginUser instead of the proxyUser.

### DIFF
--- a/server/build.gradle
+++ b/server/build.gradle
@@ -235,6 +235,7 @@ project('pxf-hdfs') {
 project('pxf-hive') {
     dependencies {
         compile(project(':pxf-hdfs'))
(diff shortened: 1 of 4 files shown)


server/build.gradle
```diff
@@ -235,6 +235,7 @@ project('pxf-hdfs') {
 project('pxf-hive') {
     dependencies {
         compile(project(':pxf-hdfs'))
+        compile(project(':pxf-service'))
```
denalex: we should avoid doing this, no plugin should depend on pxf-service, all common dependencies should go into pxf-api package.

PR #28, opened 2018-10-17
Remove gradle profiles

frankgh:
- Move gpdb properties to gradle properties
- Use jcentral as the main repository


server/build.gradle
```diff
@@ -356,20 +339,20 @@ task assemble(type: Copy, dependsOn: [subprojects.jar, project(':pxf-service').w
         rename('pxf-service', 'pxf')
         filter(ReplaceTokens,
                 tokens: [
-                        'pxfPortNum'    : databaseProperties.pxfPortNum,
-                        'pxfDefaultUser': (databaseProperties.pxfDefaultUser == null ? "" : databaseProperties.pxfDefaultUser)])
+                        'pxfPortNum'    : pxfPortNum,
+                        'pxfDefaultUser': pxfDefaultUser])
```
denalex: should we remove these tokens and hardcode the default values (5888, gpadmin - respectively) ?

server/build.gradle
```diff
@@ -356,20 +339,20 @@ task assemble(type: Copy, dependsOn: [subprojects.jar, project(':pxf-service').w
...
-                tokens: ['pxfLogDir'     : databaseProperties.pxfLogDir,
-                         'pxfRunDir'     : databaseProperties.pxfRunDir,
-                         'pxfPortNum'    : databaseProperties.pxfPortNum,
-                         'pxfDefaultUser': (databaseProperties.pxfDefaultUser == null ? "" : databaseProperties.pxfDefaultUser),
-                         'pxfDefaultUserImpersonation': (databaseProperties.pxfDefaultUserImpersonation == null ? "" : databaseProperties.pxfDefaultUserImpersonation)]) into 'conf'
+                tokens: ['pxfLogDir'     : pxfLogDir,
+                         'pxfRunDir'     : pxfRunDir,
+                         'pxfPortNum'    : pxfPortNum,
+                         'pxfDefaultUser': pxfDefaultUser,
+                         'pxfImpersonation': pxfImpersonation]) into 'conf'
```
denalex: same here, remove tokens, hardcode defaults ?

server/gradle.properties
```diff
@@ -21,11 +21,14 @@ hadoopVersion=2.8.5
...
 osFamily=el6
 parquetVersion=1.9.0
 awsJavaSdk=1.11.416
 org.gradle.daemon=true
-#org.gradle.parallel=true
+pxfLogDir=${PXF_HOME}/logs
+pxfRunDir=${PXF_HOME}/run
+pxfPortNum=5888
+pxfDefaultUser=
+pxfImpersonation=true
```
denalex: why parallel gradle has been commented and removed ?

server/gradle.properties
```diff
@@ -21,11 +21,14 @@ hadoopVersion=2.8.5
 htraceVersion=4.0.1-incubating
 hiveVersion=1.2.1
 hbaseVersionJar=1.1.2
-hbaseVersionRPM=1.1.2
 tomcatVersion=7.0.62
 pxfProtocolVersion=v15
 osFamily=el6
```
denalex: this property is also probably RPM artifact

server/build.gradle
frankgh: In that case, would it make more sense to hardcode it in `pxf-env.sh`? and we can get rid of `filter(ReplaceTokens(...))`

server/gradle.properties
frankgh: I believe it was hanging in one of the pipelines, @lavjain might have more details

server/build.gradle
denalex: yes, that's what I meant, hardcode in the source, remove gradle props.

server/gradle.properties
frankgh: are you referring to the `osFamily` property?

denalex: yes

frankgh: I will search for references, and will remove if it is safe to remove

PR #29, opened 2018-10-18
Refactor build.gradle to fix intellij support

frankgh:
Allows for better intellij integration.

### DIFF
--- a/server/build.gradle
+++ b/server/build.gradle
@@ -19,6 +19,10 @@
 
 buildscript {
 
+    repositories {
+        jcenter()
+    }
+
     dependencies {
         classpath "com.netflix.nebula:gradle-ospackage-plugin:2.2.6"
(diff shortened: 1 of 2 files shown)


server/build.gradle
```diff
@@ -356,7 +387,4 @@ task tar(type: Tar, dependsOn: [assemble]) {
 task install (type: Copy, dependsOn: [assemble]) {
     from assemble.outputs.files
     into System.properties['deployPath']
-}
-
-buildDir = '.'
-apply plugin: 'nebula-aggregate-javadocs'
+}
```
denalex: good stuff. I also think project 'pxf' does not feel proper as it is very artificial, I don't know if there's a way to produce overall artifact at the root level without declaring this dummy project.

frankgh: I think we could have a task instead of a dummy project.

PR #34, opened 2018-10-26
Start/stop the PXF cluster via `pxf cluster {start,stop}` command

benchristel:
One thing I would like feedback on is the way `build.gradle` invokes the make targets for the CLI. The existing code works but seems a bit kludgey. The CLI tests should probably run as part of `gradlew test` but I couldn't see where/how Gradle is running the Java tests.

We also need to decide how to get Go utilities (`dep`, `ginkgo`, and `gomega`) into our test containers. Maybe we can update the container image to include them?


benchristel:
@denalex and I talked about it yesterday and we agreed that it's harmless (if a bit confusing) for Gradle to invoke `make` as long as it's not calling the same Makefile that was used to invoke the Gradle task in the first place. In this case, the actual dependency graph looks like this:

```
server/Makefile
  -> server/build.gradle
    -> go/src/pxf-cluster/Makefile
      -> {dep, ginkgo, go build}
```

I considered having the `server/Makefile` invoke targets on the CLI's Makefile, like this:

```
server/Makefile
  -> go/src/pxf-cluster/Makefile
    -> {dep, ginkgo, go build}
```

But the problem with this approach is that I can't specify task dependencies in the `server/Makefile` as precisely as I would like. The constraint I want to enforce is that the CLI must be built and packaged with PXF whenever we package PXF for installation. Concretely, that means that whenever `gradle assemble` runs, the CLI needs to be built beforehand. Setting up the right target dependencies in the `server/Makefile` to ensure this is indirect and error-prone, because there's no make target that corresponds 1:1 to `gradle assemble`. I would have to add the CLI dependency to many different targets, each of which invokes `assemble`.

AFAICT, the most straightforward fix would be to refactor the `server` build system so all task dependencies are specified in the Makefile, not in Gradle. The current difficulty is that dependency declarations are spread out between both build tools.

Can we defer that work to a separate refactoring chore, or do we want to do it as part of this PR? @denalex @frankgh @lavjain

lavjain:
@benchristel I am OK with deferring it to a separate story as well.

README.md
```diff
@@ -52,14 +73,11 @@ make
 make unittest
 ```
 
-## Prerequisites
-In order to demonstrate end to end functionality you will need JDK, GPDB and Hadoop installed. 
-
-### JDK
-JDK version 1.8+ is recommended.
+## Demonstrating Hadoop Integration
+In order to demonstrate end to end functionality you will need GPDB and Hadoop installed.
```
benchristel: We moved the remark about needing JDK to the top of this file, under `Install Dependencies`, since I don't think you could build PXF without the JDK. Is that right @lavjain ?

benchristel: For the record: it is not possible to build PXF without the JDK.

PR #35, opened 2018-10-29
Cache Configuration

frankgh:
- Creates the configuration one-time, subsequent requests use the same configuration
- Introduces the concept of server for configurations
- Update all references to use ConfigurationCache instead of instantiating their own Configuration objects
- Moves UserGroupInformation functionality to the hdfs project
- pxf-service now depends on pxf-hdfs
- Reenable gradle parallel compilation


server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HdfsSplittableDataAccessor.java
```diff
@@ -41,11 +41,11 @@
...
 public abstract class HdfsSplittableDataAccessor extends Plugin implements
         ReadAccessor {
-    protected Configuration conf = null;
+    protected Configuration conf;
     protected RecordReader<Object, Object> reader = null;
-    protected InputFormat<?, ?> inputFormat = null;
+    protected InputFormat<?, ?> inputFormat;
     protected ListIterator<InputSplit> iter = null;
-    protected JobConf jobConf = null;
+    protected JobConf jobConf;
```
denalex: remove other explicit assignments as well ?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ConfigurationCache.java
```diff
@@ -0,0 +1,125 @@
...
+                try {
+                    secureConfigurationIfAvailable(config);
+                } catch (Exception e) {
+                    LOG.error("PXF Configuration Failed: " + e.getMessage());
+                    throw new RuntimeException(e);
+                }
+            }
+        }
+
+        return instance.cache.get(serverName);
```
denalex: I think we should be doing a clone of the cached entry as configuration can be modified by individual requests and those modifications should be local to those threads and not impact others.

server/pxf-api/src/main/java/org/greenplum/pxf/api/utilities/InputData.java
```diff
@@ -370,4 +378,19 @@ public void setNumAttrsProjected(int numAttrsProjected) {
         this.numAttrsProjected = numAttrsProjected;
     }
 
+    /**
+     * Returns the name of the server in a multi-HCFS setup
+     * @return the name of the server, or "default" if null
+     */
+    public String getServerName() {
+        return StringUtils.isBlank(serverName) ? "default" : serverName;
```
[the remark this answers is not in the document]
benchristel: Can the magic string `"default"` be extracted to a constant? It looks like the configuration cache is also using it.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ConfigurationCache.java
```diff
@@ -0,0 +1,125 @@
...
+    private static final String CONFIG_KEY_SERVICE_KEYTAB = "pxf.service.kerberos.keytab";
+
+    private static ConfigurationCache INSTANCE;
+    private final Map<String, Configuration> cache;
+
+    /**
+     * Prevent class instantiation
+     */
+    private ConfigurationCache() {
+        cache = new HashMap<>();
```
benchristel: Why use a `HashMap` with our own synchronization code instead of a `ConcurrentHashMap`?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ConfigurationCache.java
```diff
@@ -0,0 +1,125 @@
...
+                    // config.addResource("server1/core-site.xml");
+                    // We can create a configuration without loading
+                    // defaults = new Configuration(loadDefaults=false);
+
+                    instance.cache.put(serverName, config);
+                    configCreated = true;
+                }
+            }
+
+            if (configCreated && StringUtils.equals("default", serverName)) {
```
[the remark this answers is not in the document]
benchristel: There might also be a concurrency bug here. Once the config object is in the cache and the `synchronized` block ends, any thread can access it, right? So it might be possible for a thread to use a config object that has not had its kerberos stuff initialized.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ConfigurationCache.java
```diff
@@ -0,0 +1,125 @@
...
+        return instance.cache.get(serverName);
+    }
+
+    /**
+     * Adds the principal and keytab to the configuration if Kerberos is enabled.
+     *
+     * @param config the configuration
+     */
+    private static void secureConfigurationIfAvailable(Configuration config) throws IOException {
+        if (UserGroupInformation.isSecurityEnabled()) {
```
benchristel: Will `isSecurityEnabled` or the kerberos system properties ever change while PXF is running? If not, we could get this information just once instead of for each config object.

That might also allow us to make config objects immutable, dispelling the concern Alex mentioned above about requests modifying them.

frankgh: This should only be invoked once, during creation of the `default` configuration.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ConfigurationCache.java
```diff
@@ -0,0 +1,125 @@
...
+
+            config.set(CONFIG_KEY_SERVICE_PRINCIPAL, principal);
+            config.set(CONFIG_KEY_SERVICE_KEYTAB, keytabFilename);
+
+            if (LOG.isDebugEnabled()) {
+                LOG.debug("Kerberos principal: " + config.get(CONFIG_KEY_SERVICE_PRINCIPAL));
+                LOG.debug("Kerberos keytab: " + config.get(CONFIG_KEY_SERVICE_KEYTAB));
+            }
+
+            SecurityUtil.login(config, CONFIG_KEY_SERVICE_KEYTAB, CONFIG_KEY_SERVICE_PRINCIPAL);
```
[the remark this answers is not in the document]
frankgh: I agree, I was just lifting and shifting code from one place to another. I was not sure where to put it, so I left it there for now.

server/build.gradle
```diff
@@ -207,14 +207,9 @@ project('pxf-service') {
     }
     dependencies {
         providedCompile(project(':pxf-api'))
-        providedCompile "org.apache.hadoop:hadoop-common:$hadoopVersion"
-        providedCompile "org.apache.htrace:htrace-core4:$htraceVersion"
-        providedCompile "org.apache.hadoop:hadoop-hdfs:$hadoopVersion"
-        providedCompile "org.apache.hadoop:hadoop-hdfs-client:$hadoopVersion"
+        providedCompile(project(':pxf-hdfs'))
```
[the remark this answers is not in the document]
frankgh: I agree we should move code in that direction. But IMO we should have a PR that addresses the code reorg.

PR #36, opened 2018-10-29
Install Go and dev dependencies in base Dockerfile

benchristel:
This is blocking https://github.com/greenplum-db/pxf/pull/34, which uses Go to build and test a CLI executable.

### DIFF
--- a/concourse/docker/pxf-dev-base/Dockerfile
+++ b/concourse/docker/pxf-dev-base/Dockerfile
@@ -12,3 +12,8 @@ RUN cd /tmp/pxf_src/server && make tar && \
     chown -R 1000:1000 /home/gpadmin/.m2 && \
     ln -s /home/gpadmin/.gradle /root/.gradle && \
     rm -rf /tmp/pxf_src
+
+RUN yum install -y go && \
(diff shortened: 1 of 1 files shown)


concourse/docker/pxf-dev-base/Dockerfile
```diff
@@ -12,3 +12,8 @@ RUN cd /tmp/pxf_src/server && make tar && \
     chown -R 1000:1000 /home/gpadmin/.m2 && \
     ln -s /home/gpadmin/.gradle /root/.gradle && \
     rm -rf /tmp/pxf_src
+
+RUN yum install -y go && \
+    GOPATH=/home/gpadmin/go go get github.com/golang/dep/cmd/dep && \
+    GOPATH=/home/gpadmin/go go get github.com/onsi/ginkgo/ginkgo && \
```
denalex: does it matter which versions of these we're baking into the image ? otherwise we make repeatable builds potentially create different images

benchristel: Unfortunately, it seems `go get` does not let you specify a version.

Go's opinion is that it's up to package maintainers to not introduce breaking changes to a package. Given that Onsi maintains Ginkgo, at least we have an inside track to get it fixed if it breaks :)

PR #38, opened 2018-11-02
Add `pxf help` and `pxf version` commands to the CLI

benchristel:
This is a work in progress.


oliverralbertini:
We have decided to use cobra instead of the flags library.

PR #46, opened 2018-11-14
Change unzip to jar as it is part of JDK and already present 

lavjain:

### DIFF
--- a/server/pxf-service/src/scripts/pxf-service
+++ b/server/pxf-service/src/scripts/pxf-service
@@ -212,7 +212,7 @@ function instanceExists()
 }
 
 #
-# configureWebapp patches the webapp with pxf and user overriden configs
+# configureWebapp patches the webapp with pxf and user overridden configs
 # applied only if PXF_HOME is defined
 #
 function configureWebapp()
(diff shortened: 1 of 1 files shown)


server/pxf-service/src/scripts/pxf-service
```diff
@@ -225,7 +225,7 @@ function configureWebapp()
     rm -rf pxf
     mkdir pxf
     cd pxf
-    unzip -q ../pxf.war
+    jar -xf ../pxf.war
```
lisakowen: @lavjain - quick question.  we have the user "sudo yum -y install java-1.8.0-openjdk-1.8.0*".  will the "jar" executable be available in this package?  i noticed that jar originated from the openjdk-devel package in a centos7 docker container.  is this executable available elsewhere, or do we need the user into install the -devel* package?  thanks!

shivzone: this is worth checking

lavjain: Hi @lisakowen, Thanks a lot for catching this. The users would need to install `java-1.8.0-openjdk-devel` package. I am surprised to note that it is not part of `java-1.8.0-openjdk`

PR #47, opened 2018-11-14
Create a template for the release and PXF pipelines

frankgh:
The PXF and release pipelines are very similar. When changing one pipeline we had the risk of the other pipeline becoming out of sync. For that reason, we introduce a template with a single source of truth for both pipelines.
We introduce jinja2, following a similar approach from GPDB prod pipeline compilation. We also combine the logic in the existing PXF `deploy` script and the new pipeline generation scripts.

Templating the pipeline and integrating it to the deploy script also allows us to provide an optional flag to generate a pipeline with a compile job for GPDB.

Finally, we add the option to deploy an acceptance pipeline.

[#160429794] [#158565980]


concourse/README.md
```diff
@@ -100,21 +78,35 @@ fly -t ud set-pipeline \
...
 # Deploy development PXF pipelines
 Dev pipelines can be deployed with an optional feature name
 ```
-./deploy dev master
+./deploy -t dev -b master
 ```
 
 ```
-./deploy dev master feature-foo
+./deploy -t dev -b master -f feature-foo
```
denalex: I personally hate to remember all this option shortcuts, for me `git pull origin master` looks much better than `git -o pull -s origin -b master`, that's why we tried to refrain from  using - options unless necessary. Seems more options are added now, so maybe it's time, but I'll still ask a question -- can we do without them ?

lavjain: One compromise could be to always interpret first parameter as -t, second parameter as -b and the third parameter -f (if no options are present). This way it would be consistent with the previous deploy script. I don't know how much work would be required as the code has already become too complicated.

lavjain: On second thoughts, the deploy can be a shell script wrapper that can call deploy python scripts with the parameters. This way no changes would be required to the python script.

concourse/deploy
```diff
@@ -1,73 +1,234 @@
...
-eval ${cmd}
+#!/usr/bin/env python
+
+import argparse
+import datetime
+import os
+import os.path
+import sys
+
+from jinja2 import Environment, FileSystemLoader
```
lavjain: The readme needs to be updated to reflect jinja2 dependency (refer: https://github.com/greenplum-db/gpdb/blob/master/concourse/pipelines/README.md)

concourse/README.md
frankgh: It's possible to do it for required parameters. I have updated the PR with this change. The new syntax for the script is the following: `./deploy dev master -f feature-foo`

concourse/deploy
frankgh: I pushed an updated README.md to reflect the jinja2 dependency

concourse/README.md
frankgh: @denalex let me know if this syntax will suffice. We will be able to do `./deploy prod master` and `./deploy prod 5x` as we used to in the past.

frankgh: I was able to have the `feature` as an optional positional parameter like we previously had it: `./deploy dev master feature-foo`

concourse/deploy
```diff
@@ -1,73 +1,230 @@
...
+        description='Generate Concourse Pipeline utility',
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+
+    pxf_parser.add_argument('deploy_env',
+                            action='store',
+                            choices=['prod', 'dev'],
+                            help='Deploy environment ("prod" or "dev")')
+
+    pxf_parser.add_argument('gpdb_branch',
+                            action='store',
```
denalex: do we need to mention choices here ?

concourse/deploy
```diff
@@ -1,73 +1,230 @@
...
+                            help='GPDB Branch ("master", "5x")')
```
denalex: do we still support reading a custom yaml file, or we now only take 2 params for gpdb branch and pxf branch ? 5x and master were to decide which bucket to take icw_green from, I believe, not technically which gpdb branch to use. Is it still like that ?

concourse/pipelines/pxf_pipeline-generated.yml
```diff
@@ -78,9 +95,9 @@ resources:
   type: s3
   source:
     access_key_id: {{bucket-access-key-id}}
-    bucket: {{icw_green_bucket}}
     region_name: {{aws-region}}
     secret_access_key: {{bucket-secret-access-key}}
+    bucket: {{icw_green_bucket}}
```
denalex: if we want to keep arguments in alphabetic order, bucket has to go back up top

concourse/deploy
denalex: read code till the end and believe the old mechanisms is still in place

PR #48, opened 2018-11-14
Refactor PXF Configuration layout

frankgh:
- Remove the option of loading configuration files from /etc/{hadoop,hive,etc}/conf
- Do not ship any template files under `server/default`
- Provide sample template files under `$PXF_CONF/template` directory
- Remove GP_EXTENSIONS_HOME option from CLI
- Update CI
- Blacklist file:// protocol with HDFS access.

[#161932965]


server/pxf-api/src/main/java/org/greenplum/pxf/api/FileSystemFragmenter.java
```diff
@@ -0,0 +1,4 @@
+package org.greenplum.pxf.api;
+
+public interface FileSystemFragmenter {
+}
```
lavjain: What is the logic behind an empty interface?

frankgh: We intend to annotate classes that access hadoop systems. That way we can prevent them from accessing `file://`. I refactored the code to use annotations instead of an interface

server/pxf-api/src/main/java/org/greenplum/pxf/api/Fragmenter.java
```diff
@@ -41,6 +43,12 @@
     public Fragmenter(InputData metaData) {
         super(metaData);
         fragments = new LinkedList<>();
+
+        // TODO: we will add a test for this case when it's simpler to mock Configuration
+        if (this instanceof FileSystemFragmenter &&
+                StringUtils.startsWith(new Configuration().get("fs.defaultFS"), "file:")) {
```
shivzone: Sorry for the rather late review. We shouldn't have any specific logic in Fragmenter class. This is meant to be generic. Would be better if we introduce config/input data validations as part of other subclasses of Plugin

PR #51, opened 2018-11-26
Refactor PXF cli to use gp-common-go-libs and Cobra for parsing flags

oliverralbertini:


automation/pom.xml
```diff
@@ -38,8 +38,8 @@
 				<artifactId>maven-compiler-plugin</artifactId>
 				<version>3.1</version>
 				<configuration>
-					<source>1.7</source>
-					<target>1.7</target>
+					<source>1.8</source>
+					<target>1.8</target>
```
denalex: I hope all our docker images in CI (on all platforms) provision Java 1.8

server/Makefile
```diff
@@ -55,34 +55,32 @@ help:
 	@echo	"  - tar - bundle PXF along with tomcat into a single tarball"
 	@echo	"  - doc - creates aggregate javadoc under docs"
```
denalex: does help need to be updated, any changes in behavior  ?

server/pxf-cli/go/src/pxf-cli/Gopkg.lock
```diff
@@ -203,23 +217,27 @@
...
     "github.com/greenplum-db/gp-common-go-libs/gplog",
+    "github.com/greenplum-db/gp-common-go-libs/operating",
+    "github.com/greenplum-db/gp-common-go-libs/testhelper",
     "github.com/onsi/ginkgo",
-    "github.com/onsi/ginkgo/ginkgo",
     "github.com/onsi/gomega",
-    "golang.org/x/tools/cmd/goimports",
+    "github.com/onsi/gomega/gbytes",
+    "github.com/pkg/errors",
+    "github.com/spf13/cobra",
```
denalex: any license info update needed ?

server/pxf-cli/go/src/pxf-cli/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
denalex: why do we need Apache license here ?

server/pxf-cli/go/src/pxf-cli/Makefile
```diff
@@ -0,0 +1,36 @@
...
+	go install ${LDFLAGS}
+
+tar: build
+	mkdir -p ${PXF_STAGING_DIR}
+	cp ${PXF_SERVICE_SCRIPTS}/pxf-service ${PXF_STAGING_DIR}/pxf
+	cp ${GOPATH}/bin/pxf-cli ${PXF_STAGING_DIR}
+
+install: build
+	mkdir -p ${PXF_HOME}/bin
+	cp ${PXF_SERVICE_SCRIPTS}/pxf-service ${PXF_HOME}/bin/pxf
```
denalex: hmm ... maybe move the script to this package, even if it is bash for now, it will become GO at some point ? Or otherwise leave its packaging to server ? It would be better to avoid crossing into other system's build output, if possible.

server/pxf-service/src/scripts/pxf-service
```diff
@@ -373,8 +366,8 @@ doHelp() {
 	  init      initialize the local PXF server instance
 	  start     start the local PXF server instance
 	  stop      stop the local PXF server instance
-	  restart   restart the local PXF server instance
-	  status    show the status of the local PXF server instance
+	  restart   restart the local PXF server instance (not supported for cluster)
```
denalex: the message is a bit confusing, since other commands are described to be applicable to only local PXF server instances as well and do not claim cluster support.

server/pxf-service/src/scripts/pxf-service
```diff
@@ -482,8 +474,8 @@ case "${pxf_script_command}" in
 	'help' | '-h' | '--help' )
 		doHelp
 		;;
-	'version' )
-		doVersion
+	'version' | '--version' | '-v' )
+		doCluster --version
```
denalex: only cluster-level version ? how is it implemented on the local level ?

oliverralbertini: This is root level version, not cluster version. We moved it into Go side for easier building process (you can pass arbitrary strings into the cli binary when building with a command line flag).

server/pxf-cli/go/src/pxf-cli/Makefile
oliverralbertini: This sounds like a good idea, and @lavjain and I made an attempt at this last week. It turned out to be complicated because of the presence of some 'templates' in conjunction with the script?

server/pxf-cli/go/src/pxf-cli/LICENSE
oliverralbertini: Artifact of building the cli using cobra. It tries to be helpful by giving a license for the cli tool that it is building.

server/pxf-cli/go/src/pxf-cli/Gopkg.lock
oliverralbertini: Possibly. Let's make sure. The `github.com/onsi/**` stuff is OK because it's just for testing and doesn't ship in final binary.

server/Makefile
oliverralbertini: I just made a commit that adds help to the Makefile at the CLI level. Maybe we should also copy this to the `server/` level?

automation/pom.xml
lavjain: We have been using JDK 1.8 in docker for a while.

lavjain:
@denalex @shivzone I would like to merge this in today if there are no objections.

server/pxf-service/src/scripts/pxf-service
lisakowen: hey, haven't tried "pxf cluster help" out.  but if this is the help output for "pxf cluster" as opposed to "pxf", maybe we should:
- remove the "pxf <command> [-y]" line
- change "pxf {-h | --help}" to "pxf cluster ..."
- and fix the command name in the last sentence and refer to command consistently:  i.e.:
     Use "pxf cluster <command> --help for..."

lavjain: @lisakowen Let me commit this PR as it has been dragging. I will make the changes that you are asking in a separate commit.

PR #55, opened 2018-11-29
Exclude guava from hive-exec

frankgh:
- Guava is conflicting with other dependencies for google cloud support

### DIFF
--- a/server/build.gradle
+++ b/server/build.gradle
@@ -251,6 +251,7 @@ project('pxf-hive') {
         compile("org.apache.hive:hive-exec:$hiveVersion") {
             exclude module: 'calcite-core'
             exclude module: 'calcite-avatica'
+            exclude module: 'guava'
         }
         bundleJars "org.apache.hive:hive-metastore:$hiveVersion"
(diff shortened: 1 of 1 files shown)


frankgh:
Green pipeline here: https://ud.ci.gpdb.pivotal.io/teams/main/pipelines/dev:pxf_fguerrero-hive-exec-exclude-guava_5X_STABLE

server/build.gradle
```diff
@@ -251,6 +251,7 @@ project('pxf-hive') {
         compile("org.apache.hive:hive-exec:$hiveVersion") {
             exclude module: 'calcite-core'
             exclude module: 'calcite-avatica'
+            exclude module: 'guava'
```
denalex: does this mean we can not package and use guava or is it also available from a different transitive dependency (perhaps with a different version ?)

frankgh: we bundle the guava jar explicitly here: https://github.com/greenplum-db/pxf/blob/0dfca8eadf42108e234aeda358bdad3726640a74/server/build.gradle#L220 
hive-exec seems to be bundling guava as well

PR #56, opened 2018-12-04
Move CLI from server/pxf-cli to cli

lavjain:


Makefile
```diff
@@ -0,0 +1,34 @@
+PXF_VERSION := $(shell grep '^version=' ./server/gradle.properties | cut -d "=" -f2)
```
denalex: wonder if we can do it vice-versa -- have the version in the version file at the top level and have GRadel read it from there. Perhaps for the next iteration ?

Makefile
```diff
@@ -0,0 +1,34 @@
...
+
+ifneq "$(PXF_HOME)" ""
+	BUILD_PARAMS+= -DdeployPath="$(PXF_HOME)"
+else ifneq "$(GPHOME)" ""
+	PXF_HOME= "$(GPHOME)/pxf"
+	BUILD_PARAMS+= -DdeployPath="$(PXF_HOME)"
```
denalex: this line can be moved out of the statement and have if statement only set PXF_HOME if it is not yet set

Makefile
```diff
@@ -0,0 +1,34 @@
...
+server:
+	make -C server
+
+clean:
+	make -C cli/go/src/pxf-cli clean
+	make -C server clean
+
+tar:
+	make -C cli/go/src/pxf-cli tar
+	make -C server tar
```
denalex: so this will produce 2 different tarballs ? How and when will they be combined to produce a releasable artifact ?

oliverralbertini: This is a two-step process that will make one tarball. The cli steps makes the go artifact and renames the shell script (pxf-service). Then the server tar step uses gradle to make the releasable artifact.

PR #62, opened 2018-12-10
PXF command line sync configs to all hosts from master

lavjain:
Using RSYNC to sync for now.


server/pxf-service/src/scripts/pxf-service
```diff
@@ -395,6 +395,21 @@ function doStatus()
...
+    local source_host=$1
+	if [[ -z ${source_host} ]]; then
+        fail "PXF config source host must be set to use this option"
+	fi
+    instanceExists
+    if (( $? != 0 )); then
+        fail "Can't find PXF instance, maybe call init?"
+    fi
+    get_environment
+    SSH_OPTS="ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o LogLevel=error"
```
denalex: I am not sure these unsecure options are acceptable for production environments.

lavjain: I changed the options to just `-o StrictHostKeyChecking=no` as that is what gp-common-go-libs is using.

PR #63, opened 2018-12-11
Bundle Google Cloud Storage jar

frankgh:
- Run performance tests for google cloud storage

[#162315299]

### DIFF
--- a/concourse/scripts/pxf-perf-multi-node.bash
+++ b/concourse/scripts/pxf-perf-multi-node.bash
@@ -258,13 +258,11 @@ function create_adl_external_tables() {
         LOCATION('pxf://${ADL_ACCOUNT}.azuredatalakestore.net/adl-profile-test/output/${SCALE}/?PROFILE=adl:text&server=adlbenchmark') FORMAT 'CSV'"
 }
 
-function create_gcs_extension_external_tables() {
+function create_gcs_external_tables() {
     psql -c "CREATE EXTERNAL TABLE lineitem_gcs_read (LIKE lineitem)
-        LOCATION('pxf://${SCALE}/lineitem_data/?PROFILE=HdfsTextSimple')
-        FORMAT 'CSV' (DELIMITER '|');"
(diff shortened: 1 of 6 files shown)


concourse/scripts/pxf-perf-multi-node.bash
```diff
@@ -272,6 +281,59 @@ function assert_count_in_table {
...
+        <value>com.google.cloud.hadoop.fs.gcs.GoogleHadoopFileSystem</value>
+    </property>
+    <property>
+        <name>fs.AbstractFileSystem.gs.impl</name>
+        <value>com.google.cloud.hadoop.fs.gcs.GoogleHadoopFS</value>
+    </property>
+</configuration>
+EOF
+     # Make a backup of core-site and update it with the S3 core-site
+    gpscp -u centos -f /tmp/segment_hosts /tmp/core-site.xml =:/tmp/core-site-patch.xml
```
lavjain: Reminder here to use the new `pxf cluster sync` commands once it is merged to master.

concourse/scripts/pxf-perf-multi-node.bash
```diff
@@ -341,6 +350,59 @@ EOF
...
+      <name>google.cloud.auth.service.account.enable</name>
+      <value>true</value>
+    </property>
+    <property>
+      <name>google.cloud.auth.service.account.json.keyfile</name>
+      <value>/tmp/gsc-ci-service-account.key.json</value>
+    </property>
+    <property>
+        <name>fs.gs.impl</name>
+        <value>com.google.cloud.hadoop.fs.gcs.GoogleHadoopFileSystem</value>
```
shivzone: Why should fs impl and filesystem class be explicitly set when defaultFS is already set ? Doesn't core-default internally map them accordingly ?

concourse/scripts/pxf-perf-multi-node.bash
frankgh: sync is being used now

concourse/scripts/pxf-perf-multi-node.bash
frankgh: Yes it does. It has been removed. Thanks for catching it!

PR #65, opened 2018-12-15
Cleanup build.gradle (get rid of stuff we do not use anymore)

lavjain:


lavjain:
@denalex We are currently using pxf: top level project to bundle cloud related jars.

PR #71, opened 2019-01-03
Azure Blob Storage Support

frankgh:
- Add jars to support azure blob storage
- Add `wasbs-site.xml` template
- Add profiles to support Azure Blob Storage (wasb(s) - only SSL version is supported)
- Add performance test for Azure Blob Storage


shivzone:
Missed adding support for wasbs protocol when we test hdfs functionality using ```PROTOCOL= wasbs && make GROUP=wasbs```

frankgh:
@shivzone added support for Azure Blob Storage automation testing `PROTOCOL=wasbs && make GROUP=hcfs`

PR #72, opened 2019-01-04
Added minio docker image

denalex:


lavjain:
@denalex We can add minio to pxf-dev-base image so, it is available to all configurations (similar to singlecluster).

denalex:
implemented suggestion in a separate PR: https://github.com/greenplum-db/pxf/pull/73/files

PR #76, opened 2019-01-09
Support Minio storage, minio certification job, script for dev docker image

denalex:


README.md
```diff
@@ -183,6 +184,14 @@ pushd ~/workspace/singlecluster/bin
 ./start-hbase.sh
 popd
 ```
+### Setup Minio (optional)
```
shivzone: Can you move this setup minio section below Setup PXF section as minio is not a pre-requisite for PXF if someone simply wants to simply run installcheck or run HdfsSmokeTest

denalex: I think it is ok where it is, in the pre-requisite steps, it's marked "optional". If we move it between PXF and Tests, then you come across it every time you want to setup PXF / run tests. Sometimes you don't want to install / start Hadoop, but Hadoop is also in the pre- section.

denalex:
well, I have included minio-site.xml in the user templates, what else there is to do (other than documentation) for a product integration ? Customers will install Minio on their own, then create a server (s3 or otherwise), copy the template there and edit URL and credentials. That's all supported and tested now.

shivzone:
I see, I was thinking we would have to introduce additional profiles for minio in pxf-profile. If thats not needed we should be good.

automation/pom.xml
```diff
@@ -203,6 +203,43 @@
             <version>1.7.3</version>
         </dependency>
 
+        <!-- AWS Dependencies -->
```
frankgh: Interesting! The automation AWS jars that we use for S3 test were not sufficient for minio?

README.md
```diff
@@ -183,6 +184,14 @@ pushd ~/workspace/singlecluster/bin
...
+Minio is an S3-API compatible local storage solution. The development docker image comes with Minio software pre-installed. To start the Minio server, run the following script:
+```bash
+~/workspace/pxf/dev/start_minio.bash
+```
+After the server starts, you can access Minio UI at `http://localhost:9000` from the host OS. Use `admin` for the access key and `password` for the secret key when connecting to your local Minio instance.
+
+The script also sets `PROTOCOL=minio` so that the automation framework will use the local Minio server when running S3 automation tests.
```
frankgh: Probably unrelated to this PR, but we need to tell devs that they need to `unset PROTOCOL` to run regular HDFS tests.

dev/start_minio.bash
```diff
@@ -0,0 +1,16 @@
...
+
+export MINIO_ACCESS_KEY=admin
+export MINIO_SECRET_KEY=password
+echo "Minio credentials: accessKey=${MINIO_ACCESS_KEY} secretKey=${MINIO_SECRET_KEY}"
+
+echo 'Starting Minio ...'
+sudo /opt/minio/bin/minio server /opt/minio/data &
+
+# set variables used by automation
+export PROTOCOL=minio
```
oliverralbertini: This script has to be sourced in order for these exports to impact the current user shell. In the README, it says to run it as a subprocess.

denalex: good point, thanks for the catch, will update to source command

README.md
denalex: will do

automation/pom.xml
denalex: we needed a newer version of AWS APIs than what was bundled with Hadoop 2.8 due to bug in AWS-SDK that impacts Minio.

PR #77, opened 2019-01-10
Add MapR Support for HDFS

frankgh:
- Runs `hcfs` group in automation for MapR
- Adds CI job for MapR
- We add a fatjar, which is not used in non-MapR tests, but it bundles MapR jars that we need for running PXF as well as automation.


automation/Makefile
```diff
@@ -62,6 +62,9 @@ symlink_pxf_jars:
 		for X in $(PXF_HOME)/lib/pxf-*-[0-9]*.jar; do \
 			ln -sf $$X `echo \`basename $$X\` | sed -e 's/-[a-zA-Z0-9.]*.jar/.jar/'`; \
 		done; \
+		if [ ! -f "$(PXF_HOME)/tmp/pxf-extras.jar" ]; then \
+			touch $(PXF_HOME)/tmp/pxf-extras.jar; \
+		fi \
```
denalex: you don't really need a check condition, as you can touch a file that already exists. btw, why is creation of the empty file needed here, seems hacky

automation/pom.xml
```diff
@@ -99,6 +99,15 @@
             <version>LATEST</version>
         </dependency>
 
+        <dependency>
```
denalex: please put some comment here what is this dependency for and how / when it is created

concourse/pipelines/templates/pxf-tpl.yml
```diff
@@ -44,11 +48,13 @@ groups:
 
   - name: secure
     jobs:
+      - compile_pxf
       - test_pxf_hdp_secure_with_impersonation
       - test_pxf_hdp_secure_no_impersonation
 
   - name: no_impersonation
     jobs:
+      - compile_pxf
```
denalex: (not on this PR specifically, but still) -- I think minio also runs with impersonation off, maybe add it here too ?

concourse/scripts/test_pxf.bash
```diff
@@ -9,10 +9,15 @@ export GPHOME=${GPHOME:-"/usr/local/greenplum-db-devel"}
 export PXF_HOME="${GPHOME}/pxf"
 export JAVA_HOME="${JAVA_HOME}"
 export JAVA_TOOL_OPTIONS=-Dfile.encoding=UTF8
+if [[ ${HADOOP_CLIENT} == "MAPR" ]]; then
```
denalex: this will fail if HADOOP_CLIENT variable is not defined, if we surround it with quotes, it will not. Not sure if this is desired behavior.

concourse/scripts/test_pxf.bash
```diff
@@ -53,6 +58,11 @@ function setup_hadoop() {
 }
 
 function _main() {
+	if [[ ${HADOOP_CLIENT} == MAPR ]]; then
+		# start mapr services before installing GPDB
+		/root/init-script
```
denalex: name of the script is not very indicative of its purpose, smth like start_mapr.sh would've been better.

concourse/scripts/test_pxf.bash
```diff
@@ -71,20 +81,45 @@ function _main() {
 		cat << EOF > /tmp/gsc-ci-service-account.key.json
 ${GOOGLE_CREDENTIALS}
 EOF
-	elif [[ -z "${PROTOCOL}" ]]; then
+	elif [[ -z "${PROTOCOL}" ]] && [[ ! ${HADOOP_CLIENT} == MAPR ]]; then
 		# Setup Hadoop before creating GPDB cluster to use system python for yum install
-		setup_hadoop /singlecluster
+		setup_hadoop ${GPHD_ROOT}
```
denalex: can this logical branch be merged with top-level `if` that checks for MAPR ?

concourse/scripts/test_pxf.bash
```diff
@@ -71,20 +81,45 @@ function _main() {
...
+		cp ${HADOOP_COMMON}/hadoop-common-2.7.0-mapr-1707.jar ${PXF_CONF_DIR}/lib/
+		# Copy *-site.xml files
+		cp /opt/mapr/hadoop/hadoop-2.7.0/etc/hadoop/*-site.xml ${PXF_CONF_DIR}/servers/default/
+		# Copy mapred-site.xml for recursive hdfs directories test
+		# We need to do this step after PXF Server init
+		cp ${PXF_CONF_DIR}/templates/mapred-site.xml ${PXF_CONF_DIR}/servers/default/recursive-site.xml
+		# Set mapr port to 7222 in default.xml (sut)
+		pushd pxf_src/automation
+			sed -i 's|<port>8020</port>|<port>7222</port>|' src/test/resources/sut/default.xml
+		popd
```
denalex: too much logic in main(), should we extract this into a local function ?

concourse/scripts/test_pxf.bash
lavjain: The script was named after a similar script that was shipped with MapR dev container.

concourse/scripts/test_pxf.bash
lavjain: All the jobs in the dev pipeline are green

frankgh: Double brackets allow you to do away without quotes

automation/Makefile
frankgh: I agree that this is hacky. We tried using a profile for MapR in `pom.xml` and we had "extras" dependencies  there, but we were not very successful with that approach. Do you have another suggestion how to achieve the same result without creating an empty file?

concourse/scripts/test_pxf.bash
frankgh: I did a few changes. Let me know if that's sufficient

automation/Makefile
```diff
@@ -62,6 +62,7 @@ symlink_pxf_jars:
 		for X in $(PXF_HOME)/lib/pxf-*-[0-9]*.jar; do \
 			ln -sf $$X `echo \`basename $$X\` | sed -e 's/-[a-zA-Z0-9.]*.jar/.jar/'`; \
 		done; \
+		touch $(PXF_HOME)/tmp/pxf-extras.jar; \
```
shivzone: Please change the name to something more indicative of whats inside the jar like pxf-mapr-extras.jar

PR #81, opened 2019-01-15
Upgrade to hadoop version 2.9.2

frankgh:
By upgrading to hadoop version 2.9.2, we are seeing reduced warn messages for parquet files (i.e)

WARN tomcat-http--32 com.amazonaws.services.s3.internal.S3AbortableInputStream - Not all bytes were read from the S3ObjectInputStream, aborting HTTP connection. This is likely an error and may result in sub-optimal behavior. Request only the bytes you need via a ranged GET or drain the input stream after use.

This introduces new dependencies for Hadoop.

- [ ] Request approval for new jars


shivzone:
Did we test any other negative impact with using 2.9.2 on the other profile tests including Hive, etc

PR #82, opened 2019-01-17
Move GPDBWritable class from pxf-service to pxf-api

lavjain:


server/pxf-service/src/test/java/org/greenplum/pxf/service/io/BufferWritableTest.java
```diff
@@ -1,42 +0,0 @@
-package org.greenplum.pxf.service.io;
```
denalex: why were the tests removed ?

server/pxf-api/src/main/java/org/greenplum/pxf/api/model/BasePlugin.java
```diff
@@ -56,6 +57,21 @@ public void initialize(RequestContext requestContext) {
...
+     */
+    GPDBWritable makeGPDBWritableOutput() {
+        int num_actual_fields = context.getColumns();
+        int[] schema = new int[num_actual_fields];
+        for (int i = 0; i < num_actual_fields; i++) {
+            schema[i] = context.getColumn(i).columnTypeCode();
+        }
+        return new GPDBWritable(schema);
+    }
+
```
denalex: This is essentially a factory method and this class does not feel like the right place to put this logic -- BasePlugin doesn't know of any IO, formats etc, it only has things common for all plugins, like thread-safety and base initialization logic. What was the motivation to move GPDBWritable here ?

Also, where is this supposed to be used from ? BridgeOutputBuilder still builds GPDBWritable in its own methods.

server/pxf-service/src/test/java/org/greenplum/pxf/service/io/BufferWritableTest.java
lavjain: Thanks for catching. I forgot to do a git add to the tests.

PR #83, opened 2019-01-18
Remove parquet data fragmenter and use GPDBWritable directly

lavjain:


lisakowen:
@lavjain - i noticed that this PR modifies pxf-profiles-default.xml and removes the ParquetDataFragmenter class.  

1.  are there any implications to the end user if they upgrade to a version of PXF with this change? would they be able to read from or insert into an external table referencing parquet data that they previously created with a parquet-related profile?
2.  as far as API/SDK changes, what if someone created a PXF connector that extended ParquetDataFragmenter?  since this PR removes the class, they would need to change their code, right?
3.  if someone created a custom profile that referenced ParquetDataFragmenter, they would need to update the profile definition - could they just replace that reference with HdfsDataFragmenter?

do the changes in either this PR, or the moving of GPDBWritable class from pxf-service to pxf-api in PR #82, require a PXF major version increment (i.e. 6.0.0) for next release?

lavjain:
Hi @lisakowen
1. This affects runtime only. The customers would be able to read the data from external table that was created using the previous PXF profile.
2. The custom profiles should not be affected. Those customers would need to package the ParquetDataFragmenter as part of their jar, if they do not want to switch to the new accessor.
3. Also, moving GPDBWritable to pxf-api should not warrant a PXF major version change since it is only one class. Somehow, it was missed in the previous refactor.

lavjain:
There will be a separate PR that will address this. Some parts of this PR are already in master.

PR #85, opened 2019-01-24
Move download logic for tomcat from gradle to makefile

lavjain:


server/Makefile
```diff
@@ -33,7 +37,7 @@ help:
 	@echo	"  - doc - creates aggregate javadoc under docs"
 
 all:
-	./gradlew clean tar $(BUILD_PARAMS)
+	./gradlew clean tomcat tar $(BUILD_PARAMS)
```
denalex: both grade install and tar tasks depend on bundle which depends on tomcat being installed, so I guess we need to add dependency on tomcat task at Makefile level to install and tar Makefile tasks (and remove it from all)

server/Makefile
```diff
@@ -53,6 +57,23 @@ distclean maintainer-clean: clean
 doc:
 	./gradlew aggregateJavadoc
 
+.PHONY: tomcat
+tomcat:
+	@if [[ -f "${TOMCAT_DIR}/${TOMCAT_TAR}" ]]; then \
+		echo "${TOMCAT_TAR} already exists, nothing to do"; \
+    else \
```
denalex: indentation or just github ?

PR #87, opened 2019-01-30
Updated library versions with security issues

shivzone:
Updated jackson libaries to 2.9.8

Updated hive libraries to 1.2.2

Not in a position to update the hive-{jdbc,exec and service} libraries to > 2.3.3 as recommended by https://github.com/greenplum-db/pxf/network/alerts the as pxf only is certified with hive 1.1 version.


shivzone:
@denalex yes the dev branch pipeline is green. Do we need to request new OSL certification even for version changes ? @kongyew

PR #89, opened 2019-02-06
Implement optimized version of isTextForm

lavjain:
Also, add getObject/setObject to GPDBWritable

### DIFF
--- a/server/pxf-api/src/main/java/org/greenplum/pxf/api/io/DataType.java
+++ b/server/pxf-api/src/main/java/org/greenplum/pxf/api/io/DataType.java
@@ -59,9 +59,7 @@ public enum DataType {
     UNSUPPORTED_TYPE(-1);
 
     private static final Map<Integer, DataType> lookup = new HashMap<>();
-
     static {
-
         INT2ARRAY.typeElem = SMALLINT;
         INT4ARRAY.typeElem = INTEGER;
         INT8ARRAY.typeElem = BIGINT;
@@ -71,6 +69,17 @@ public enum DataType {
         for (DataType dt : EnumSet.allOf(DataType.class)) {
             lookup.put(dt.getOID(), dt);
         }
+    }
+
+    private static final Map<Integer, Integer> notText = new HashMap<>();
+    static {
+        notText.put(BIGINT.OID, 1);
+        notText.put(BOOLEAN.OID, 1);
+        notText.put(BYTEA.OID, 1);
+        notText.put(FLOAT8.OID, 1);
+        notText.put(INTEGER.OID, 1);
+        notText.put(REAL.OID, 1);
+        notText.put(SMALLINT.OID, 1);
     }
 
     private final int OID;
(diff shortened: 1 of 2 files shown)


server/pxf-api/src/main/java/org/greenplum/pxf/api/io/DataType.java
```diff
@@ -73,6 +71,17 @@
         }
     }
 
+    private static final Map<Integer, Integer> notText = new HashMap<>();
```
frankgh: Another option would be to use a set 
```suggestion
    private static final Set<Integer> notText = new HashSet<>();
```

server/pxf-api/src/main/java/org/greenplum/pxf/api/io/DataType.java
```diff
@@ -88,16 +97,18 @@
...
+        return type == null ? UNSUPPORTED_TYPE : type;
     }
 
     public static boolean isArrayType(int OID) {
         DataType type = lookup.get(OID);
         return type != null && type.typeElem != null;
     }
 
+    public static boolean isTextForm(int OID) {
+        return notText.get(OID) == null;
```
frankgh: And check whether the item is in the set
```suggestion
        return !notText.contains(OID);
```

server/pxf-api/src/main/java/org/greenplum/pxf/api/io/DataType.java
```diff
@@ -73,6 +71,17 @@
         }
     }
 
+    private static final Map<Integer, Integer> notText = new HashMap<>();
+    static {
+        notText.put(BIGINT.OID, 1);
```
frankgh:
```suggestion
        notText.add(BIGINT.OID);
```

server/pxf-api/src/main/java/org/greenplum/pxf/api/io/DataType.java
```diff
@@ -73,6 +71,17 @@
         }
     }
 
+    private static final Map<Integer, Integer> notText = new HashMap<>();
+    static {
```
frankgh: is there any reason we wanted two separate static blocks?

lavjain:
@frankgh Thanks for the review. I have incorporated your suggestions.

server/pxf-api/src/main/java/org/greenplum/pxf/api/io/DataType.java
```diff
@@ -59,6 +61,7 @@
     UNSUPPORTED_TYPE(-1);
 
     private static final Map<Integer, DataType> lookup = new HashMap<>();
+    private static final Set<Integer> notText = new HashSet<>();
```
denalex: another alternative is to implement EnumSet which should probably be even better as it uses bitsets for implementation: https://docs.oracle.com/javase/7/docs/api/java/util/EnumSet.html

frankgh: does EnumSet work with Integers? It seems its geared for enums

PR #91, opened 2019-02-09
Create new groups for hive and hbase tests

lavjain:
We do not need to run hive and hbase tests for gpdb group unless explicitly stated

### DIFF
--- a/automation/src/test/java/org/greenplum/pxf/automation/features/hive/HiveTextTest.java
+++ b/automation/src/test/java/org/greenplum/pxf/automation/features/hive/HiveTextTest.java
@@ -111,7 +111,7 @@ public class HiveTextTest extends HiveBaseTest {
      *
      * @throws Exception if test fails to run
      */
-    @Test(groups = { "features", "gpdb" })
+    @Test(groups = { "hive" })
     public void mismatchedTypes() throws Exception {
 
         // Hive column is SMALLINT, expected GPDB type is SMALLINT(int2), but actual is INTEGER(int4)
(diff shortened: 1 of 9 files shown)


lavjain:
@shivzone Removing from GPDB group does not mean that we will not be running those tests. The pipeline job explicitly needs to specify HDFS, HIVE, HBASE etc for the groups that it wants to run.

frankgh:
I would probably keep the `gpdb` group and add the specific subgroup to the test (i.e `hbase`, `hive`, etc).

automation/src/test/java/org/greenplum/pxf/automation/features/hive/HiveTextTest.java
```diff
@@ -93,7 +93,7 @@ void prepareData() throws Exception {
      *
      * @throws Exception if test fails to run
      */
-    //@Test(groups = "features")
+    @Test(groups = "features")
```
denalex: should this be "hive" now ?

denalex:
I would drop `gpdb` group altogether as this framework can no longer work against HAWQ. If we also want to deprecate `features` group, then we need to group by tested subsystem (such as here : `hive`, `hbase`, `hdfs`, `s3`, `jdbc` etc) and then either provide group mapping in some test artifact or specify groups to run for pipelines. I have not seen any of these 2 artifacts in this PR.

automation/src/test/java/org/greenplum/pxf/automation/features/hive/HiveTextTest.java
lavjain: No, because it does not work with hive on gpdb yet!

lavjain:
@denalex Is your suggestion that we replace `gpdb` group with `feature` group and delete/comment out the current feature tests that do not work with `gpdb`.

lavjain:
I will be merging this if there are no objections

PR #94, opened 2019-02-13
Column projection support changes

frankgh:
We remove "HAS-FILTER" header for the moment because it didn't add any additional information that cannot be induced from the "FILTER" header. 
Column projection information will not be available for queries with filters that pxf doesn't support.

Pending:
1. Will be updating the FilterPushDownTest to test for column projection information

Co-authored: Francisco Guerrero [aguerrero@pivotal.io](mailto:aguerrero@pivotal.io)
Co-authored: Shivram Mani [smani@pivotal.io](mailto:smani@pivotal.io)

### DIFF
--- a/server/pxf-service/src/main/java/org/greenplum/pxf/service/HttpRequestParser.java
+++ b/server/pxf-service/src/main/java/org/greenplum/pxf/service/HttpRequestParser.java
@@ -250,53 +250,16 @@ public class HttpRequestParser implements RequestParser<HttpHeaders> {
      * Attribute Projection information is optional
      */
     private void parseFilterAndProjectionInformationAndTupleDescription(RequestMap params, RequestContext context) {
-        Set<Integer> attrsProjected = new HashSet<>();
+        Set<Integer> attrsProjected;
 
         String filterString = params.removeOptionalProperty("FILTER");
         if (filterString != null) {
             context.setFilterString(filterString);
             context.setFilterStringValid(true);
 
-            // Parse filterString and add attrs to attrsProjected set
-            FilterParser parser = new FilterParser(new FilterParser.FilterBuilder() {
-                @Override
-                public Object build(FilterParser.Operation operation, Object left, Object right) {
-                    addAttributes(left);
-                    addAttributes(right);
-                    return null;
-                }
-
-                @Override
-                public Object build(FilterParser.Operation operation, Object operand) {
-                    addAttributes(operand);
-                    return null;
-                }
-
-                @Override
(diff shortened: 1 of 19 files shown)


server/pxf-service/src/main/java/org/greenplum/pxf/service/HttpRequestParser.java
```diff
@@ -250,38 +254,40 @@ private boolean parseBooleanValue(String threadSafeStr) {
      * Attribute Projection information is optional
      */
     private void parseTupleDescription(RequestMap params, RequestContext context) {
+        Set<Integer> attrsProjected = new HashSet<>();
```
denalex: good to use set to speedup lookups, even better to use BitSet here instead of HashSet -- this is a classic example where BitSets are useful -- binary index-based 0..n decisions. Initialize it after figuring out how many columns there are.

server/pxf-service/src/main/java/org/greenplum/pxf/service/HttpRequestParser.java
```diff
@@ -90,11 +92,14 @@ public RequestContext parseRequest(HttpHeaders request) {
...
-            context.setFilterString(params.removeProperty("FILTER"));
+        String filterString = params.removeOptionalProperty("FILTER");
+        String hasFilter = params.removeProperty("HAS-FILTER");
+        if (filterString != null) {
+            context.setFilterString(filterString);
+            context.setFilterStringValid(true);
+        } else if ("1".equals(hasFilter)) {
+            LOG.info("Original query has filter, but it was not propagated to PXF");
         }
-        context.setFilterStringValid(isFilterPresent);
```
denalex: so, nowhere in the whole project isFilterStringValid is called ? Let's remove it from context then.

frankgh: we use it in the RequestContext
`public boolean hasFilter() { return filterStringValid; }`
we can return true if the filter string is not null

denalex: I think we need to preserve the notion of hasFilter as the value of HAS_FILTER header coming in, orthogonal of the actual presence of filter string. But this is confusing to us, will be also confusing for plugin writers ...

frankgh: we decided to introduce the notion when we require it. for now we are only checking for filterString null

PR #95, opened 2019-02-15
Enable license generation for PXF

shivzone:
Authored-by: Shivram Mani <smani@pivotal.io>

Scope restricted to bundleJars to avoid pulling dependancies that don't make it into the gpdb binary

### DIFF
--- a/server/build.gradle
+++ b/server/build.gradle
@@ -17,6 +17,13 @@
  * under the License.
  */
 
+plugins {
+  id "com.github.hierynomus.license-report" version"0.15.0"
+}
+
+downloadLicenses {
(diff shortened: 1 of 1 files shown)


server/build.gradle
```diff
@@ -17,6 +17,13 @@
  * under the License.
  */
 
+plugins {
+  id "com.github.hierynomus.license-report" version"0.15.0"
+}
+
+downloadLicenses {
+    dependencyConfiguration = 'bundleJars'
```
frankgh: should we explicitly set `includeProjectDependencies` to `false`
```suggestion
    dependencyConfiguration = 'bundleJars'
    includeProjectDependencies = false
```

PR #98, opened 2019-02-26
PXF-JDBC: quote column names

leskin-in:
Implement automatic column names quotation. A [new method](https://github.com/greenplum-db/pxf/pull/98/files#diff-6cb1f96f1a9d4f9efddfd39c83aa7973R149) uses `DatabaseMetaData` object and column descriptors to define whether columns should be surrounded by quotation string. If so, such string is taken from `DatabaseMetaData` and applied to all SQL queries generated by PXF-JDBC plugin.

This enables PXF to query external tables with special (mixed-case or containing special symbols) column names.

At the moment, GPDB always sends column names without quotes (even if they were provided with quotes by user when defining `EXTERNAL TABLE`).

Changes:
* Introduce a new class `SQLQueryBuilder`. It incorporates relevant code previously located in `JdbcAccessor` and `WhereSQLBuilder`.
* Make `DbProduct` an `enum`. Move code from its child classes into `DbProduct` definition.
* Add unit tests for `DbProduct`
* Add extra unit tests for `SQLQueryBuilder`


server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/utils/MicrosoftProduct.java
```diff
@@ -20,16 +20,11 @@
  */
 
 /**
- * Implements methods for the Microsoft SQL server database
+ * Implements methods for Microsoft SQL server.
  */
-public class MicrosoftProduct extends DbProduct {
+public class MicrosoftProduct extends PostgresProduct {
```
denalex: this feels weird, why Postgres is now the base ?

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/utils/DbProduct.java
```diff
@@ -19,58 +19,54 @@
...
      * @param dbName A full name of the database
      * @return a DbProduct of the required class
      */
     public static DbProduct getDbProduct(String dbName) {
-        if (dbName.toUpperCase().contains("MYSQL"))
+        if (LOG.isDebugEnabled()) {
+            LOG.debug("Database product name is '" + dbName + "'");
+        }
+        if (dbName.toUpperCase().contains("POSTGRES"))
+            return new PostgresProduct();
```
denalex: since there's a limited number of subclasses and they are stateless and only contain methods, I'd propose to refactor this into an enum DBProduct { PostgresDB, MicrosoftDB, OracleDB .. } so that we don't have to create new instances of them on each request. Refer to `server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetTypeConverter.java` for an example of this pattern.

server/pxf-jdbc/README.md
```diff
@@ -82,11 +82,16 @@ The **`<plugin_parameters>`** are **optional**:
...
+&QUOTE_COLUMNS=<any_value>
+]
 ```
 
 The meaning of `BATCH_SIZE` is given in section [batching of INSERT queries](#Batching).
 
-The meaning of `POOL_SIZE` is given in section [using thread pool for INSERT queries](#Thread_pool)
+The meaning of `POOL_SIZE` is given in section [using thread pool for INSERT queries](#Thread_pool).
+
+`QUOTE_COLUMNS` parameter makes PXF surround all column names with a quotation string defined by JDBC driver. Use this option when these names require special treatment (e.g. they contain both upper- and lowercase letters and the external database is case-sensitive).
```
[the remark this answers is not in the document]
leskin-in: Thank you for this proposal. JDBC allows to make a similar check using [this method in `DatabaseMetaData`](https://docs.oracle.com/javase/7/docs/api/java/sql/DatabaseMetaData.html#supportsMixedCaseIdentifiers()). https://github.com/greenplum-db/pxf/pull/98/commits/2a4e14b872c1dc564a884abd0a46c05ec9ff4f57 implements this.

There is one case that is not covered in current implementation - when all columns in external database are quoted and their names are in lower case. Then GPDB request to PXF is indistinguishable from the one it would make if columns were not quoted, and PXF will not quote column names.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/utils/MicrosoftProduct.java
leskin-in: It has been base implicitly since the very introduction of `DbProduct` (`BaseProduct` had the same implementations of methods as `PostgresProduct`).

leskin-in:
@denalex https://github.com/greenplum-db/pxf/pull/98#discussion_r262496733 refers to an underlying problem: GPDB sends column names without quotation marks even if these columns were quoted by user.

On the other hand, if GPDB starts to quote column names, this may break other PXF plugins.

If GPDB could send some extra parameter (like it sends column id) to tell PXF column name is quoted, then the solution of quoted column names would be significantly easier.

Another problem of similar kind is illustrated by [this code](https://github.com/greenplum-db/pxf/blob/master/server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java#L226). If GPDB sent XID in a separate field, we would not need to match table name against that `Pattern`.

I believe the only difficulty we face here is the need to coordinate changes in PXF and GPDB.

What do you think?

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/utils/DbProduct.java
```diff
@@ -19,58 +19,97 @@
...
+    MICROSOFT {
+        @Override
+        public String wrapDate(Object val){
+            return "'" + val + "'";
+        }
+
+        @Override
+        public String wrapTimestamp(Object val) {
+            return "'" + val + "'";
+        }
```
denalex: seems only ORACLE has different implementation for this method, so maybe make it not abstract and push to the class, remove overridden methods from all enums other than ORACLE.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcFilterParser.java
```diff
@@ -23,42 +23,31 @@
...
-            ) {
-                throw new Exception("'" + filterString + "' could not be resolved to a filter");
-            }
-            return result;
-        }
-        catch (Exception e) {
-            throw new ParseException(e.getMessage(), 0);
-        }
+    public static List<BasicFilter> parseFilters(String filterString) throws UnsupportedOperationException, ParseException {
+        return convertBasicFilterList(getFilterObject(filterString), null);
```
denalex: what is the benefit of this one-liner ? can just copy the body of that method in here ?

denalex:
I wonder what if we always quote the column names in JDBC query created by the plugin so that the target Database always accepts the query ? Then we don't have to consider multiple edge cases and the system becomes more reliable.

For that to work, do the column names have to correspond in case to what the target database thinks they are ? 
Does GPDB always send column names in lower case no matter how they were defined in external table DDL ? In that case maybe we can lookup target db column name from table metadata in JDBC and use that name ?

Also, we might want to quote the schema and table name itself: `"schema"."table"`. For this, will probably have to parse on dot `.` table name as it is coming from GPDB and reconstruct with quotes.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcFilterParser.java
leskin-in: Both those methods are private. 

`convertBasicFilterList` can be called recursively, but when it is not, its `returnList` argument *must* be `null`. To encapsulate this and the fact that `getFilterObject` must be called on filter string first, `parseFilters` method is introduced.

Both `convertBasicFilterList` and `getFilterObject` come from older versions of PXF-JDBC.

denalex:
@leskin-in -- thanks for taking care of minor comments. Now what to you think about always quoting the columns ? I checked GPDB will preserve the table / column name case only when the external table is created with quoted names:
```
template1=# create external table "bAr" ("low" text, "mixED" text, "UPPER" text) LOCATION ('pxf://bar?PROFILE=Jdbc') FORMAT 'TEXT';
CREATE EXTERNAL TABLE
template1=# \d+ bar;
Did not find any relation named "bar".
template1=# \d+ bAr;
Did not find any relation named "bAr".
template1=# \d+ "bAr";
                    External table "public.bAr"
 Column | Type | Modifiers | Storage  | Stats target | Description
--------+------+-----------+----------+--------------+-------------
 low    | text |           | extended |              |
 mixED  | text |           | extended |              |
 UPPER  | text |           | extended |              |
Type: readable
Encoding: UTF8
Format type: text
Format options: delimiter '	' null '\N' escape '\'
External options: {}
External location: pxf://bar?PROFILE=Jdbc
Execute on: all segments
```
So, if we don't want to do extra mapping, using the quotes in GPDB table definition will likely result in case preservation all the way GPDB -> PXF -> Target DB if PXF always quotes the names in targetDB SQL. Is it too much to ask a GPDB user to follow this rule ?

If we always quote in PXF, then we also will loose a case of case insensitivity, where lowered case names  from GPDB (default) can be sent to UPPER cased Oracle table (default) that would've worked otherwise, right ?

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/utils/DbProduct.java
leskin-in: This is sensible, implemented in https://github.com/greenplum-db/pxf/pull/98/commits/46e996281014acac230c02f1948bc3486823079e. 

My initial thought here was that if we do what you proposed, we have two methods, both of which are meant to be abstract, but only one is declared so. The reason for such lack of `abstract` declaration is code reuse.

If more products with other `wrapTimestamp` are introduced, this method should be declared `abstract` again. But as this *may* happen in the future, I think it is OK to have its implementation in enum class for now.

leskin-in:
@denalex exactly; [this] is correct. We lose "implicit" case-insensitivity provided by many external databases for unquoted column names.

leskin-in:
The problem is, GPDB never preserves quotes (as PostgreSQL does (not), too).

As for schema and table names, they are passed in `LOCATION` clause, and there the user can pass quotes, special symbols, etc., so this is not a problem.

denalex:
Right, and so you will check if `specialCharactersNamePresent ||
            (mixedCaseNamePresent && !databaseMetaData.supportsMixedCaseIdentifiers())` and then ONLY you will quote, which incidentally will happen only if they were also quoted in GPDB, otherwise they would've come in lower case. I think I get it now... will explain to the team here.

Still, for schema / table name in LOCATION field, why would we ask a user to quote there ? They can just enter mixed case, for example, but then it will need to be quoted in dynamic SQL, right ? Should we apply the same logic -- if it comes not all in lower case, then it perhaps intentional and needs automatic quoting ? If all lower case, it is probably because target DB is case-insensitive and will consume lower case just fine.

leskin-in:
@denalex ,
> They can just enter mixed case, for example, but then it will need to be quoted in dynamic SQL, right ? 

PXF uses the data source name exactly as user provided it in `LOCATION` clause. No quotes or other symbols are omitted. I think it is better for PXF to act exactly as user asked.

> if it comes not all in lower case

It is harder do deal with such cases, as the data source name may contain special characters, like dots, to denote schema to use. Then, we need to parse data source name and quote its parts. In addition, some database may use other special characters to denote schema (or some other entity).

denalex:
@leskin-in -- so you're saying if the user defined "MyTable" in Oracle, the table name would need to be quoted in SQL: `SELECT .. FROM "MyTable"` and for this to work the location will need to be quoted as well: `LOCATION ('pxf://"MyTable"?PROFILE...')` , correct ? And if there is a schema, then it will look like `LOCATION ('pxf://"Schema"."MyTable"?PROFILE...')` ? I guess this will work, not sure if that's the best user experience. @lavjain @kongyew @frankgh - what do you think ?

leskin-in:
@denalex , https://github.com/greenplum-db/pxf/pull/98#issuecomment-470013562 is correct.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/SQLQueryBuilder.java
```diff
@@ -0,0 +1,309 @@
...
+        StringBuilder sb = new StringBuilder();
+
+        sb.append("SELECT ");
+
+        // Insert columns' names
+        String columnDivisor = "";
+        for (ColumnDescriptor column : columns) {
+            sb.append(columnDivisor);
+            columnDivisor = ", ";
+            sb.append(quoteString + column.columnName() + quoteString);
```
frankgh: Maybe this is an opportunity to start supporting column projection? I think we still need to return the column with null values to the Greenplum side. This doesn't need to go in this PR
```suggestion
            if (column.isProjected()) {
                sb.append(quoteString + column.columnName() + quoteString);
            } else {
                sb.append("null");
            }
```

denalex: @frankgh -- we will take this as a separate feature after this PR is committed, as it will require an integration test.

lavjain:
@leskin-in Thanks a lot for submitting the PR. We had a discussion on backward compatibility of the SQL statements that this PR affects and having an option to turn the quoting ON/OFF in the DDL options might help us with some edge cases.

Also, it would be really appreciated if you could add a couple of tests for the new behavior to https://github.com/greenplum-db/pxf/blob/master/automation/src/test/java/org/greenplum/pxf/automation/features/jdbc/JdbcTest.java

denalex:
The option that @lavjain was asking for would overwrite the existing auto-detect logic, if specified, so:
- if option is `QUOTE_COLUMNS=true` then columns should be quoted no matter what
- if option is `QUOTE_COLUMNS=false` then columns should be NOT quoted no matter what
- if option is not provided by the user, then current auto-detect logic is to be utilized

The idea is having the system to be smart enough for 90% of cases but giving the user an option for manual override if the situation requires it.

denalex:
@leskin-in -- thank you for addressing the comments and contributing the feature. We'll be merging this PR now. There's a typo in the integration test which makes it fail as written, but we'll fix it immediately after the merge in a separate commit.

PR #99, opened 2019-02-26
PXF-JDBC: Implement query-preceding SQL command and configuration file support

leskin-in:
Allow user to pass an arbitrary SQL command to be executed before execution of "main" query generated by PXF (`SELECT` or `INSERT`). This can be useful, for example, when "main" query requires some options to be `SET` before execution.

There was a similar issue in [Apache HAWQ](https://github.com/apache/hawq) repository (some time ago GitHub issues were disabled there, and I was unable to find it). A request there was to execute arbitrary SQL queries *and* get their results (in text form). This PR solves an easier task (discarding the results) in order to enable "on-the-fly" external database environment customization.

As setting such arbitrary SQL in external table DDL is not safe, configuration file support is implemented. README is partially rewritten to fit for this change.

A configuration parameter, `jdbc.pre_query.sql` is introduced in configuration file. It contains plain SQL query executed by PXF right before making "main" query (queries) to external database.

A new option, `STOP_IF_PRE_FAILS` is introduced. If it is set, PXF will not execute "main" query if query-preceding SQL command fails.

Changes:
* Introduce `JdbcPluginSettings` class. It contains configuration file parameters and context options names, parses them and is used to output them in error messages sent to user.
* Introduce `utils.SimpleSQLQueryExecutor` class and unit tests for it. This class executes given SQL queries, discarding their results, and processes exceptions inside
* Update `JdbcPlugin` and `JdbcAccessor`
* Update README.md
* Add example configuration file (jdbc-site.xml)


pivotalguru:
We shouldn't encourage a solution that promotes excessive DDL statements nor should we store SQL in the DDL of an external table.  DDL for all objects are available to all users so it would be readable by everyone.  This could encourage compromising the remote server with SQL injection.

The SQL for the statement should be instead located in a file to avoid these two problems and the external table should reference it.  This should be done in the LOCATION similarly to how gpfdist works with TRANSFORM.

denalex:
What are the use cases that require executing a SQL statement before the main query ?

I am concerned with having plain SQL statements in the LOCATION field -- security issues, SQL injections, user syntax mistakes come to mind. I'd prefer this to be defined in a file.

We now have a feature where options can be configured per configured "server" in the file located in file system. For example, let's say you have an Oracle instance, you would create a "server" directory, let's call it myoracle in `PXF_CONF/servers/myoracle` and place some `*-site.xml` there, let's say `jdbc-site.xml`. Now you can provide properties for DB connection, user, password etc, instead of placing them into LOCATION field. Then from GPDB you mention `?SERVER=myoracle` in LOCATION field to refer to this special configuration.

JDBC Connector will need to get enhanced slightly to support this, but this will eliminate the need to provide connection information in clear text in LOCATION field.

And I would propose once the above is done, we should move the PRE_SQL to that file as well. It will be the SQL that will be executed before each query that uses that server runs. But if each query requires separate pre-sql, then it will be problematic.

leskin-in:
@denalex @pivotalguru thanks for your comments. [This code from latest commit](https://github.com/greenplum-db/pxf/pull/99/commits/b8dc0e42179bf0e85ae92c556076cb9359d69e16#diff-a7a41c3581cd04b216b24a3cc63e1c09) implements retrieval of plugin setting from configuration files. I would be glad to receive comments and remarks about this implementation.

I have chosen [parameter names](https://github.com/greenplum-db/pxf/pull/99/commits/b8dc0e42179bf0e85ae92c556076cb9359d69e16#diff-a7a41c3581cd04b216b24a3cc63e1c09R31) so that they do not differ much from old ones, defined in `LOCATION`. I would be grateful for proposals on renaming them, as I am unaware of any convention for such names (e.g. [this javadoc on Hadoop Configuration class](https://hadoop.apache.org/docs/r2.6.4/hadoop-project-dist/hadoop-common/core-default.xml) does not specify them).

leskin-in:
It is not clear why CI build fails. I can build the plugin and successfully pass all unit tests.

Unfortunately, I cannot access CI results. Request to `https://ud.ci.gpdb.pivotal.io/api/v1/builds/15671` fails with code 403, although I am authorized via GitHub.

leskin-in:
@denalex the use case we faced is shortly described [in README](https://github.com/greenplum-db/pxf/pull/99/files#diff-d2f69cfc2b417ca813bc26aabca0b97fR317).

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
```diff
@@ -201,28 +137,126 @@ public static void closeStatement(Statement statement) {
...
 
+    /**
+     * Load plugin settings from configuration and context.
+     * 'this.context' and 'this.configuration' are used as sources.
+     *
+     * This method also parses non-string arguments.
+     */
+    private void loadPluginSettings() throws IllegalArgumentException {
+        jdbcDriver = JdbcPluginSettings.jdbcDriver.loadFromContextOrConfiguration(context, configuration);
+        jdbcUrl = JdbcPluginSettings.jdbcUrl.loadFromContextOrConfiguration(context, configuration);
```
denalex: this is not quite unnecessary. All properties in `*-site.xml` files will be loaded and available as configuration properties. The options provided by the user are available as `context.getOption()` context options. In case a connector wants to automatically apply context options as configuration properties, it can whitelist them in pxf-profiles-default.xml, see example for s3:
```
   <profile>
        <name>s3:text:multi</name>
        <plugins>
...
        </plugins>
        <protocol>s3a</protocol>
        <optionMappings>
            <mapping option="accesskey" property="fs.s3a.access.key"/>
            <mapping option="secretkey" property="fs.s3a.secret.key"/>
        </optionMappings>
    </profile>
```
So if you would like to allow users to specify some things as both options or properties, whitelist them for the JDBC profile in a similar manner and always access as configuration properties.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcPluginSettings.java
```diff
@@ -0,0 +1,103 @@
...
+import org.greenplum.pxf.api.model.RequestContext;
+
+import org.apache.hadoop.conf.Configuration;
+
+/**
+ * PXF-JDBC Plugin settings' definitions.
+ *
+ * Note settings' values are stored separately.
+ */
+public enum JdbcPluginSettings {
```
denalex: as mentioned earlier, I don't think this class is needed, just whitelist these properties in profiles.xml

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcPluginSettings.java
```diff
@@ -0,0 +1,103 @@
...
+    jdbcDriver("JDBC_DRIVER", "jdbc.driver"),
+    jdbcUrl("DB_URL", "jdbc.url"),
+    jdbcUser("USER", "jdbc.user"),
+    jdbcPassword("PASS", "jdbc.pass"),
+    preQuerySql(null, "jdbc.pre_query.sql"),
+    stopIfPreQueryFails("STOP_IF_PRE_FAILS", "jdbc.pre_query.stop_if_fails"),
+    batchSize("BATCH_SIZE", "jdbc.batch_size"),
```
denalex: maybe `jdbc.batch.size` and `jdbc.pool.size` for consistency ? (dots instead of underscores)

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcPluginSettings.java
```diff
@@ -0,0 +1,103 @@
...
+/**
+ * PXF-JDBC Plugin settings' definitions.
+ *
+ * Note settings' values are stored separately.
+ */
+public enum JdbcPluginSettings {
+    jdbcDriver("JDBC_DRIVER", "jdbc.driver"),
+    jdbcUrl("DB_URL", "jdbc.url"),
+    jdbcUser("USER", "jdbc.user"),
+    jdbcPassword("PASS", "jdbc.pass"),
```
denalex: maybe `jdbc.password` ?

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcPluginSettings.java
```diff
@@ -0,0 +1,103 @@
...
+    preQuerySql(null, "jdbc.pre_query.sql"),
+    stopIfPreQueryFails("STOP_IF_PRE_FAILS", "jdbc.pre_query.stop_if_fails"),
+    batchSize("BATCH_SIZE", "jdbc.batch_size"),
+    poolSize("POOL_SIZE", "jdbc.pool_size"),
+    partitionBy("PARTITION_BY", "jdbc.partition.by"),
+    partitionRange("RANGE", "jdbc.partition.range"),
+    partitionInterval("INTERVAL", "jdbc.partition.interval");
```
denalex: I am not sure PARTITION_BY, RANGE and INTERVAL belong in the file. Config file should contain per-server properties, not per-query ones. The following seem like per-server properties:
- jdbc.driver
- jdbc.url
- jdbc.user (until multi-user server impersonation is developed)
- jdbc.password
- jdbc.pre_query.sql
- pre_query.stop_if_fails

I am not so convinced about:
- jdbc.batch_size
- jdbc.pool_size
Did you have or anticipate cases where these will be set per query (per table) ?

Finally, partitioning properties are definitely per-table and should not be in the file.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/utils/SimpleSQLQueryExecutor.java
```diff
@@ -0,0 +1,69 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import java.sql.Connection;
+import java.sql.SQLException;
+import java.sql.Statement;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
```
denalex: please use SLF4J

denalex:
The PR pipeline was failing for some reason for your PR on compiling `pxf-cli` GO package:
```
+ make -C pxf_src test install
make: Entering directory `/tmp/build/04bb138f/pxf_src'
make -C cli/go/src/pxf-cli test
make[1]: Entering directory `/tmp/build/04bb138f/pxf_src/cli/go/src/pxf-cli'
(1/18) Failed to write github.com/pkg/errors@v0.8.0
(2/18) Failed to write golang.org/x/sys@master
(3/18) Failed to write golang.org/x/text@v0.3.0
(4/18) Failed to write github.com/spf13/cobra@v0.0.3
(5/18) Failed to write gopkg.in/tomb.v1@v1
(6/18) Failed to write golang.org/x/net@master
(7/18) Failed to write github.com/inconshreveable/mousetrap@v1.0
(8/18) Failed to write github.com/greenplum-db/gp-common-go-libs@master
(9/18) Failed to write github.com/onsi/ginkgo@v1.6.0
(10/18) Failed to write github.com/jackc/pgx@v3.2.0
(11/18) Failed to write github.com/onsi/gomega@v1.4.2
(12/18) Failed to write gopkg.in/yaml.v2@v2
(13/18) Failed to write gopkg.in/DATA-DOG/go-sqlmock.v1@v1.3.0
(14/18) Failed to write github.com/blang/semver@v3.5.1
(15/18) Failed to write github.com/spf13/pflag@v1.0.3
(16/18) Failed to write github.com/hpcloud/tail@v1.0.0
(17/18) Failed to write gopkg.in/fsnotify.v1 (from https://github.com/fsnotify/fsnotify.git)@v1.4.7
grouped write of manifest, lock and vendor: error while writing out vendor tree: failed to write dep tree: failed to export github.com/pkg/errors: 
	(1) failed to list versions for https://github.com/pkg/errors: fatal: bad config file line 15 in .git/config
: exit status 128
	(2) failed to list versions for ssh://git@github.com/pkg/errors: fatal: bad config file line 15 in .git/config
: exit status 128
	(3) failed to list versions for git://github.com/pkg/errors: fatal: bad config file line 15 in .git/config
: exit status 128
	(4) failed to list versions for http://github.com/pkg/errors: fatal: bad config file line 15 in .git/config
: exit status 128
```
We have it working fine for other PRs, not sure what's the issue there, but once you push more changes we will review once again.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
leskin-in: @denalex, I wonder whether it is possible to blacklist settings that should be provided only as options. I have not found such a mechanism [here](https://github.com/greenplum-db/pxf/blob/480bd60593535574556f9e34bd87724a9a3a4897/server/pxf-service/src/main/java/org/greenplum/pxf/service/profile/ProfilesConf.java#L121).

We could implement this feature (for example, introduce a new tag `<optionOnlyMappings>` and check that options listed there are not present in configuration); this would lead to changes in [pxf-service/.../profile/Profile.java](https://github.com/greenplum-db/pxf/blob/480bd60593535574556f9e34bd87724a9a3a4897/server/pxf-service/src/main/java/org/greenplum/pxf/service/profile/Profile.java). Do you find this reasonable?

Also, I suppose we would also need a method to provide user with correct `LOCATION` option name in case an exception happens (e.g. [here](https://github.com/greenplum-db/pxf/pull/99/files#diff-0d3419e9a6bf82afdc1af5f7d62020d2R128)).

pivotalguru:
What is a use case for pre_sql?

What is the point to stop_if_pre_fails?  Stop what?

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcPluginSettings.java
leskin-in: [This commit](https://github.com/greenplum-db/pxf/pull/99/commits/ffd0ba89174938d98a17decb095abd1c77dc6bce) moves partitioning properties to file.

I suppose we can allow user to control those settings both from query and from file. This can be useful for tests to figure out best values of those parameters (I believe it is faster to re-create a table than to distribute a new version of the file).

pivotalguru:
What is the use case for pre_sql?  stop_if_pre_fails is a boolean?

leskin-in:
@pivotalguru [use case](https://github.com/greenplum-db/pxf/pull/99/files#diff-d2f69cfc2b417ca813bc26aabca0b97fR318), meaning and [allowed](https://github.com/greenplum-db/pxf/pull/99/files#diff-d2f69cfc2b417ca813bc26aabca0b97fR159) [values](https://github.com/greenplum-db/pxf/pull/99/files#diff-d2f69cfc2b417ca813bc26aabca0b97fR168) of those parameters are described in README. Basically, it is for cases when some variables should be set in external database before execution of the query.

`stop_if_pre_fails` is an instruction for PXF to "stop" (its processing of current query) if "pre" (preceding command) fails. No doubt this can be changed to some other name.

pivotalguru:
This is a bad idea.  You shouldn't be mixing up writable and readable
tables like this.  A readable table should never have the ability to write
and vice versa.

The better choice is to use an writable external table to do your pre_sql
that writes to your database.  The next statement would be a select from
the external table that only reads data.  You can put this in a transaction
too so you either rollback or ignore the insert command.

pivotalguru:
LOCATION
('pxf://PUBLIC.T2?PROFILE=JDBC&JDBC_DRIVER=org.company.JdbcDriver&DB_URL=jdbc:company://1.2.3.4:12345&PRE_SQL=INSERT
INTO T2 VALUES ((SELECT MAX(K) + 1 FROM T2), 1);&STOP_IF_PRE_FAILS=1')

I'm also worried about this syntax.

SQL Injection much?  We shouldn't be passing and storing this information.
The profile should contain the external information (JDBC driver, URL,
username, password, read level, and extraProps) and you should pass in a
reference to a file that contains the SQL to execute.  So two references
only.

leskin-in:
@pivotalguru Alex Denissov pointed to that in comment https://github.com/greenplum-db/pxf/pull/99#issuecomment-467577690, and commit https://github.com/greenplum-db/pxf/pull/99/commits/b8dc0e42179bf0e85ae92c556076cb9359d69e16 disallowed user to set `PRE_SQL`; instead, that statement should be placed to a configuration file `<server>-site.xml`.

The problem with two separate statements is that every time we make a request to external database, PXF creates a new session to access it. Thus we need to "load" the state set by first request, which in turn requires to execute an additional query in the second session.

pivotalguru:
I don't like this idea but if you insist, you could put the statements in a
single file and parse it.

insert into t1 values ....;

insert into t2 values ....;

select * from customer;

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
denalex: @leskin-in -- I am not sure that is the need to blacklist options. Since the files are managed by sysadmin, we trust that person to provide correct properties. Having sysadmin manage the list of options he is not supposed to set properties for is counter-intuitive.

With Hadoop connector, hadoop clients get all their configurations from config files, managed by sysadmin. If connector authors believe some properties could be overwritten by a user in DDL, then they add them to the whitelist for PXF to automatically copy the values from request options into configuration properties.

Ideally we envision all connectors relying on Configuration object for their config information with per-query options either accessed directly, copied to configuration via whitelist or copied to configuration in connector's logic after additional validation.

lavjain:
I agree with @pivotalguru here. Running an INSERT as described in the README would be running the insert statement for each of the segments.

Ideally such statements should be run from the master node. (We have a design in progress that would allow PXF to just run the fragmenter call from the master node and rest of the segments can just use that data.)

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
leskin-in: @denalex this sounds good. But I see a problem with the implementation you propose.

Options for partitioning are complex, and user can easily make mistake there. Options are set per query. If `JdbcPluginSettings` is removed, we still would need to generate an appropriate error message in case those options are invalid. 

I can imagine two ways to deal with this:
* Hardcode those messages in multiple places (this is current implementation in `master` branch)
* Use a class to generate such messages. 
But the second option is what [`JdbcPluginSettings`](https://github.com/greenplum-db/pxf/pull/99/files/b8dc0e42179bf0e85ae92c556076cb9359d69e16#diff-a7a41c3581cd04b216b24a3cc63e1c09R96) does, and if we remove `loadFrom...` functions there, the class definition would shorten just a few lines, and we would still have to hardcode option names there.

Another thing we could do is to output configuration names for all settings. But this would make user confused: for example, instead of `'RANGE' is incorrect` one would see something like `'jdbc.partition.range' is incorrect`.

leskin-in:
@lavjain @pivotalguru @denalex ,
the current version supports configuration file parameter `jdbc.pre_query.sql`. The implementation has following features:
* Pre-query SQL is executed by each PXF segment in cluster that was given a fragment
* Pre-query SQL may be different for each PXF segment. It may also be absent on some segments
* Multiple SQL queries can be passed in `jdbc.pre_query.sql` configuration parameter
* Pre-query SQL is executed in the same session as "main" SQL query

The last feature enables user to `SET` variables in external database, which was the initial use case for the feature. This also makes it possible to "load the state" of external database before working with data in external database.

lavjain:
@leskin-in The PXF is common for all segments on the host. Whatever pre-query in the PXF configuration is specified would be executed for all segments on that host. It looks like we need a way to support creating and terminating sessions inside the JDBC plugin that reuse the same connection.

leskin-in:
@lavjain ,
I may have been unclear in my previous comment. The current implementation has following features:
* pre-query is executed once for every [`Fragment`](https://github.com/greenplum-db/pxf/blob/480bd60593535574556f9e34bd87724a9a3a4897/server/pxf-api/src/main/java/org/greenplum/pxf/api/model/Fragment.java#L26) (this implies that when partitioning is not used, execution happens exactly once, no matter how many PXF or GPDB segments the cluster has).
* PXF-JDBC creates a new JDBC connection for every `Fragment`. This is why we need to run pre-query for each of them.
* Every host runs a single PXF process. So there is a bijection between PXF segments and configuration files (`...-site.xml`).
* Distribution of fragments changes each time a query to `EXTERNAL TABLE` is made.

These imply that all fragments processed by one PXF segment use the same pre-query, and it is executed once for every one of them.

This solves the task of setting up an environment for "main" query execution. On the other hand, partitioning (controlled by user) defines how many times the pre-query will be executed.

**Reusage of JDBC connections** is a good idea, but its implementation requires changes in current API provided to PXF extensions (at the moment, all data they store is erased when fragment processing finishes, including `static` class members).

Another thing we could do is to implement partitioning of pre-queries. However, I suppose it would be quite bizarre (e.g., it is not clear where pre-query for each partition should be stored).

Shall we implement some kind of pre-query partitioning? If so, what would be the requirements for it and what use cases would it cover?

leskin-in:
Closing this in favour of #114 and #118 .

PR #100, opened 2019-02-26
PXF-Ignite: Use Ignite thin client for Java

leskin-in:
Use Ignite thin client for Java instead of REST in PXF-Ignite plugin. Ignite thin client uses custom Ignite binary protocol and is much faster than REST in most cases.

There are a few new parameters (passed in `LOCATION` clause of external table DDL) that Ignite thin connector supports. User can now set `lazy`, `replicated_only`, `tcp_nodelay` flags for Ignite thin connector. This can speed up queries' execution. `IGNITE_HOSTS` parameter allows to pass multiple hosts so that if one host failed, Ignite thin connector would try to use other ones.

Complete description of new parameters is given in README.

Changes:
* Introduce multiple new `LOCATION` clause parameters
* Remove `gson 2.8.2` dependency; introduce `ignite-core 2.6.0` dependency instead. Note this is not the only version of `ignite-core` supported.
* Update unit tests
* Update README.md

### DIFF
--- a/server/build.gradle
+++ b/server/build.gradle
@@ -78,7 +78,6 @@ allprojects {
             force "org.codehaus.jackson:jackson-mapper-asl:1.9.13"
             force "junit:junit:${junitVersion}"
             force "com.google.guava:guava:11.0.2" // because this version ships with HDP-2.5.3
-            force "com.google.code.gson:gson:2.8.2" // because HDP ships with an older version, but Ignite expects 2.8.2
         }
     }
 
@@ -207,8 +206,8 @@ project('pxf-hive') {
 project('pxf-ignite') {
     dependencies {
         compile(project(':pxf-api'))
-        compile "com.google.code.gson:gson:2.8.2"
         compile "org.apache.commons:commons-compress:1.16.1"
+        compile "org.apache.ignite:ignite-core:2.6.0"
     }
 }
 
--- a/server/pxf-ignite/README.md
+++ b/server/pxf-ignite/README.md
@@ -1,15 +1,13 @@
 # Accessing Ignite database using PXF
 
-The PXF Ignite plug-in enables to access the [Apache Ignite database](https://ignite.apache.org/) (both read and write operations are supported) via REST API.
+The PXF Ignite plugin provides access to [Apache Ignite database](https://ignite.apache.org/) (both `SELECT` and `INSERT` are supported) via Ignite thin connector for Java.
 
 
 ## Prerequisites
 
-Check the following before using the plug-in:
-
-* The Ignite plug-in is installed on all PXF nodes;
-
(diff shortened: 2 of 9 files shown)


frankgh:
This PR needs OSL approval for the new JAR `org.apache.ignite:ignite-core:2.6.0` prior to being merged.

denalex:
`org.apache.ignite:ignite-core:2.6.0` is licensed under ASLv2.

But it also transitively brings:
```
\- org.apache.ignite:ignite-core:jar:2.6.0:compile
   +- javax.cache:cache-api:jar:1.0.0:compile
   +- org.jetbrains:annotations:jar:13.0:compile
   \- org.gridgain:ignite-shmem:jar:1.0.0:compile
```
which would need to be shipped and OSL-compliance checked.
The first artifact's license: https://raw.github.com/jsr107/jsr107spec/master/LICENSE.txt and seems like ASLv2, the other 2 are direct ASLv2.

GroupId | ArtifactId | Version | Type | Licenses
-- | -- | -- | -- | --
javax.cache | cache-api | 1.0.0 | jar | JSR-000107 JCACHE 2.9 Public Review - Updated Specification License
org.gridgain | ignite-shmem | 1.0.0 | jar | The Apache Software License, Version 2.0
org.jetbrains | annotations | 13.0 | jar | The Apache Software License, Version 2.0

leskin-in:
I would like to close this PR in favour of using [Ignite JDBC thin driver](https://apacheignite-sql.readme.io/docs/jdbc-driver#jdbc-thin-driver) in PXF-JDBC plugin.

The reasons are as follows:

* Ignite thin client introduces multiple new dependencies. However, they are of no use for other plugins in PXF;

* Ignite Java thin client (introduced by this PR) is not much different from [Ignite JDBC thin driver](https://apacheignite-sql.readme.io/docs/jdbc-driver#jdbc-thin-driver). The latter implements the same underlying binary protocol. Also, it turns out the JDBC thin driver enables access to Ignite at a speed comparable to one of Java thin client (with appropriate settings);

* JDBC thin driver always implicitly sets `lazy` flag which prevents Ignite server node from failure due to out-of-memory error. Therefore, it can be considered as a safer alternative to Java thin client;

* JDBC thin driver does not require a separate PXF plugin. Thus, the user gets all features implemented for PXF-JDBC plugin (e.g. partitioning) and does not need to use other syntax for that. On the other hand, this dramatically simplifies the maintenance of PXF &mdash; Ignite interaction implementation.

Due to some of these reasons, we may want to remove PXF-Ignite plugin completely. The only use case when current (REST) implementation is as performant as JDBC thin driver is `SELECT` query with very small output (such that the packets of data from Ignite take less than 1 MTU).

denalex:
@leskin-in -- do you also want to remove the Ignite plugin from the codebase altogether and advise users to use JDBC plugin instead ? If so, please raise the appropriate PR.

leskin-in:
@denalex, removal of Ignite plugin seems rational to me now. See https://github.com/greenplum-db/pxf/pull/119.

PR #101, opened 2019-03-02
Update logging configuration to limit Hadoop INFO logging

denalex:


server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/FragmenterResource.java
```diff
@@ -93,6 +95,13 @@ public Response getFragments(@Context final ServletContext servletContext,
         fragments = AnalyzeUtils.getSampleFragments(fragments, context);
         FragmentsResponse fragmentsResponse = FragmentsResponseFormatter.formatResponse(fragments, path);
 
+        int numberOfFragments = fragments.size();
+        SessionId session = new SessionId(context.getSegmentId(), context.getTransactionId(), context.getUser());
+        long elapsedMillis = (System.nanoTime() - startTime) / 1000000;
+        LOG.info("{} returns {} fragment{} for path {} in {} ms for {} [profile {} filter is{} available]",
+                fragmenter.getClass().getSimpleName(), numberOfFragments, numberOfFragments == 1 ? "" : "s",
+                path, elapsedMillis, session, context.getProfile(), context.hasFilter() ? "" : " not");
```
frankgh: what about the case where profile is not used? Do we care about knowing the accessor and resolver?

server/pxf-service/src/main/java/org/greenplum/pxf/service/servlet/SecurityServletFilter.java
```diff
@@ -132,6 +132,9 @@ public void doFilter(final ServletRequest request, final ServletResponse respons
             } catch (Throwable t) {
                 LOG.error("Error releasing UGICache for session: {}", session, t);
             }
+            if (lastCallForSegment) {
+                LOG.info("Finished processing {}", session);
```
frankgh: Do we need the last fragment call debug log in lines 128-129? I can see some value, but may be redundant now

server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/FragmenterResource.java
```diff
@@ -82,6 +83,7 @@ public Response getFragments(@Context final ServletContext servletContext,
                                  @QueryParam("path") final String path)
             throws Exception {
 
+        long startTime = System.nanoTime();
```
lavjain: Any reason for not using System.currentTimeMillis() instead?

server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/FragmenterResource.java
denalex: I don't think we do, those cases are very rare and the current log line is already too long.

server/pxf-service/src/main/java/org/greenplum/pxf/service/servlet/SecurityServletFilter.java
denalex: it still might be useful for timing of ugiCache.release calls when DEBUG is enabled.

PR #103, opened 2019-03-05
Fixed race condition in Multinode CI job

denalex:


denalex:
Pipelines will have to be re-deployed as some tasks changed parameters

concourse/scripts/test_pxf_multinode.bash
```diff
@@ -60,6 +60,58 @@ function close_ssh_tunnels() {
...
+        echo 'export JAVA_HOME=/usr/lib/jvm/jre' | sudo tee -a ~gpadmin/.bashrc &&
+        echo 'export JAVA_HOME=/usr/lib/jvm/jre' | sudo tee -a ~centos/.bashrc &&
+        sudo tar -xzf pxf_tarball/pxf.tar.gz -C ${GPHOME} &&
+        sudo chown -R gpadmin:gpadmin ${GPHOME}/pxf"
+}
+
+function setup_pxf_on_cluster() {
+    # untar pxf on all nodes in the cluster
+    for node in ${gpdb_nodes}; do
+        setup_pxf_on_segment ${node} &
```
lavjain: `setup_pxf_on_segment` should be done in the previous task as it is installing the JDK as well, which is time-consuming. The rest looks good!

denalex: the reason I took it out is that it installs PXF into GPHOME/pxf and by that time we are not guaranteed gpdb would be even installed. JDK install doesn't take long anyways.

PR #105, opened 2019-03-07
Add descriptive message when JAVA_HOME is not set

frankgh:
JAVA_HOME needs to be set in the interactive non-login shell file (i.e.
~/.bashrc) or in the default environment script on all greenplum hosts.
Also, init function should fail when re-init fails because
${instance_root}/pxf-service file exists

#163201852

### DIFF
--- a/server/pxf-service/src/scripts/pxf-service
+++ b/server/pxf-service/src/scripts/pxf-service
@@ -103,7 +103,7 @@ function get_environment()
 
     # validate JAVA_HOME
     if [[ ! -x ${JAVA_HOME}/bin/java ]]; then
-        fail "\$JAVA_HOME is invalid"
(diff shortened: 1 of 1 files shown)


lisakowen:
the docs currently mention pxf-env-default.sh as a PXF internal config file.  we have not discussed modifying it in the docs up to this point.  should i add setting JAVA_HOME in pxf-env-default.sh to step 3b on this page:  https://gpdb.docs.pivotal.io/5170/pxf/install_java.html?

frankgh:
@lisakowen https://gpdb.docs.pivotal.io/5170/pxf/install_java.html documentation is sufficient `echo 'export JAVA_HOME=/usr/lib/jvm/java-1.8.0-openjdk-1.8.0.x86_64/jre' >> /home/gpadmin/.bashrc
g`. No need to add it in both pxf-env-default.sh and .bashrc

PR #106, opened 2019-03-07
Add commented out dfs.client.use.datanode.hostname option in core-site

frankgh:
We add the commented out option dfs.client.use.datanode.hostname in
template for core-site.xml for datanodes where the IP address is not
reachable by the client, and a domain name is provided instead, letting
clients perform their own DNS resolution. This was useful when setting
up a dataproc cluster and attempting to access it from a different
network.

[#164339537]


frankgh:
Will move this information to the troubleshooting document.

PR #108, opened 2019-03-08
Enabled column projection pushdown for JDBC profile

frankgh:


frankgh:
@leskin-in -- will appreciate if you can take a look at this PR.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -59,48 +59,51 @@
...
                     break;
                 default:
-                    throw new UnsupportedOperationException("Field type '" + DataType.get(oneField.type).toString() + "' (column '" + column.toString() + "') is not supported");
+                    throw new UnsupportedOperationException(
+                            String.format("Field type '%s' (column '%s') is not supported",
+                                    DataType.get(oneField.type),
+                                    column));
             }
 
             oneField.val = value;
```
leskin-in: I notice a bug not related to this PR here.

When `result` value of a column is `null`, `result.getSomeType()` still returns non-null value. We should call `result.wasNull()` to check if value was `null`.

So, this line should be changed to `oneField.val = result.wasNull() ? null : value;` or something similar.

I believe there was such check before; otherwise it seems unnecessary to use `value` variable while we could fill `oneField.val` directly.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -59,48 +59,51 @@
...
+                    value = result.getDate(colIndex);
                     break;
                 case TIMESTAMP:
-                    value = result.getTimestamp(colName);
+                    value = result.getTimestamp(colIndex);
                     break;
                 default:
-                    throw new UnsupportedOperationException("Field type '" + DataType.get(oneField.type).toString() + "' (column '" + column.toString() + "') is not supported");
+                    throw new UnsupportedOperationException(
+                            String.format("Field type '%s' (column '%s') is not supported",
```
leskin-in: Note that you introduce `String.format()` to this file here. There are numerous places where string concatenation is used (not only in JdbcResolver.java). If `.format()` is better, then we should replace concatenation everywhere, at least in this file.

frankgh: True, I tend to change it as I see them. I will go ahead and change all of the concats in this class

frankgh: done

PR #109, opened 2019-03-11
Enable Integration sanity tests to run on CentOS 7

benchristel:
w/ @sambitesh @soumyadeep2007 

This PR involves changes to support sanity tests on CentOS 7:

1. `service` is not installed by default in CentOS 7. So we replaced commands to start/restart sshd using the service command with direct invocations of sshd and kill. 
Note: We tried installing the `initscripts` package (which contains service). However, we ran into the following error:
```
Redirecting to /bin/systemctl start sshd.service
    Failed to get D-Bus connection: Operation not permitted
```
This happens even when running the container with --privileged.

2. We had to allow ping to run as a non-root user for CentOS 7 by setting capabilities. This is to satisfy gpinitsystem.

### DIFF
--- a/concourse/scripts/install_hadoop.bash
+++ b/concourse/scripts/install_hadoop.bash
@@ -9,12 +9,16 @@ SSH_OPTS="-i cluster_env_files/private_key.pem"
 GPHD_ROOT="/singlecluster"
 
 function install_hadoop_single_cluster() {
-
     local hadoop_ip=${1}
+    # We can't use service sshd restart as service is not installed on CentOS 7.
+    # Also, we got "Operation not permitted" errors even running service in a
+    # privileged container.
+    # So we have to kill and start the sshd process directly.
     ssh ${SSH_OPTS} centos@edw0 "sudo mkdir -p /root/.ssh &&
         sudo cp /home/centos/.ssh/authorized_keys /root/.ssh &&
         sudo sed -i 's/PermitRootLogin no/PermitRootLogin yes/' /etc/ssh/sshd_config &&
-        sudo service sshd restart"
+        sudo cat /var/run/sshd.pid | xargs kill -9 &&
+        { sudo /usr/sbin/sshd -D & }"
 
     tar -xzf pxf_tarball/pxf.tar.gz -C /tmp
     cp /tmp/pxf/lib/pxf-hbase-*.jar /singlecluster/hbase/lib
--- a/concourse/scripts/install_hadoop.bash
+++ b/concourse/scripts/install_hadoop.bash
@@ -17,7 +17,7 @@ function install_hadoop_single_cluster() {
     ssh ${SSH_OPTS} centos@edw0 "sudo mkdir -p /root/.ssh &&
         sudo cp /home/centos/.ssh/authorized_keys /root/.ssh &&
         sudo sed -i 's/PermitRootLogin no/PermitRootLogin yes/' /etc/ssh/sshd_config &&
-        sudo cat /var/run/sshd.pid | xargs kill -9 &&
+        sudo cat /var/run/sshd.pid | xargs kill -9 ;
         { sudo /usr/sbin/sshd -D & }"
 
     tar -xzf pxf_tarball/pxf.tar.gz -C /tmp
--- a/concourse/scripts/install_hadoop.bash
+++ b/concourse/scripts/install_hadoop.bash
@@ -10,15 +10,10 @@ GPHD_ROOT="/singlecluster"
 
 function install_hadoop_single_cluster() {
     local hadoop_ip=${1}
-    # We can't use service sshd restart as service is not installed on CentOS 7.
-    # Also, we got "Operation not permitted" errors even running service in a
-    # privileged container.
-    # So we have to kill and start the sshd process directly.
     ssh ${SSH_OPTS} centos@edw0 "sudo mkdir -p /root/.ssh &&
         sudo cp /home/centos/.ssh/authorized_keys /root/.ssh &&
         sudo sed -i 's/PermitRootLogin no/PermitRootLogin yes/' /etc/ssh/sshd_config &&
-        sudo cat /var/run/sshd.pid | xargs kill -9 ;
-        { sudo /usr/sbin/sshd -D & }"
+        sudo service sshd restart"
 
     tar -xzf pxf_tarball/pxf.tar.gz -C /tmp
     cp /tmp/pxf/lib/pxf-hbase-*.jar /singlecluster/hbase/lib
--- a/concourse/scripts/pxf-perf-multi-node.bash
+++ b/concourse/scripts/pxf-perf-multi-node.bash
@@ -58,7 +58,7 @@ function writable_external_table_parquet_query() {
 ###########################################
 
 function setup_sshd() {
-    /usr/sbin/sshd -D &
+    service sshd start
     passwd -u root
 
     if [[ -d cluster_env_files ]]; then
--- a/concourse/scripts/pxf-perf-multi-node.bash
+++ b/concourse/scripts/pxf-perf-multi-node.bash
@@ -58,7 +58,7 @@ function writable_external_table_parquet_query() {
 ###########################################
 
 function setup_sshd() {
-    service sshd start
+    /usr/sbin/sshd -D &
     passwd -u root
 
     if [[ -d cluster_env_files ]]; then
--- a/concourse/scripts/pxf_common.bash
+++ b/concourse/scripts/pxf_common.bash
@@ -50,7 +50,7 @@ function install_gpdb_binary() {
 	fi
 
 	if [[ ${TARGET_OS} == "centos" ]]; then
-		service sshd start
+		/usr/sbin/sshd -D &
 		psi_dir=$(find /usr/lib64 -name psi | sort -r | head -1)
 	elif [[ ${TARGET_OS} == "ubuntu" ]]; then
         # Adjust GPHOME if the binary expects it to be /usr/local/gpdb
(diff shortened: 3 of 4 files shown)


concourse/scripts/install_hadoop.bash
```diff
@@ -9,12 +9,16 @@ SSH_OPTS="-i cluster_env_files/private_key.pem"
...
+    # We can't use service sshd restart as service is not installed on CentOS 7.
+    # Also, we got "Operation not permitted" errors even running service in a
+    # privileged container.
+    # So we have to kill and start the sshd process directly.
     ssh ${SSH_OPTS} centos@edw0 "sudo mkdir -p /root/.ssh &&
         sudo cp /home/centos/.ssh/authorized_keys /root/.ssh &&
         sudo sed -i 's/PermitRootLogin no/PermitRootLogin yes/' /etc/ssh/sshd_config &&
-        sudo service sshd restart"
+        sudo cat /var/run/sshd.pid | xargs kill -9 ;
+        { sudo /usr/sbin/sshd -D & }"
```
bradfordboyle: If we are going to background `sshd` anyways, why include `-D`? Why not just `/usr/bin/sbin/sshd`?

concourse/scripts/test_pxf.bash
```diff
@@ -86,7 +86,20 @@ function setup_hadoop() {
 	fi
 }
 
+function allow_ping_to_run_as_non_root() {
+	if [ -f /etc/redhat-release ] && [[ "$(cat /etc/redhat-release)" =~ 'CentOS Linux release 7' ]]; then
+		# On CentOS 7, the setuid bit is not set on /bin/ping (as it is in CentOS
+		# 6), so we need to add these capabilities to let non-root users use it. The
+		# capabilities/setuid are needed because ping uses raw sockets.
+		setcap cap_net_admin,cap_net_raw+p /bin/ping
```
bradfordboyle: Out of curiosity, would `chmod u+s /bin/ping` have worked here?

concourse/scripts/install_hadoop.bash
```diff
@@ -9,12 +9,16 @@ SSH_OPTS="-i cluster_env_files/private_key.pem"
...
-
     local hadoop_ip=${1}
+    # We can't use service sshd restart as service is not installed on CentOS 7.
+    # Also, we got "Operation not permitted" errors even running service in a
+    # privileged container.
+    # So we have to kill and start the sshd process directly.
     ssh ${SSH_OPTS} centos@edw0 "sudo mkdir -p /root/.ssh &&
         sudo cp /home/centos/.ssh/authorized_keys /root/.ssh &&
         sudo sed -i 's/PermitRootLogin no/PermitRootLogin yes/' /etc/ssh/sshd_config &&
-        sudo service sshd restart"
```
bradfordboyle: This is being run on a remote VM (and not in a container), right? We could use `systemctl restart sshd`. If we need this script to handle CentOS 6 and CentOS 7 you could use the following

```sh
command -v service && sudo service sshd restart # CentOS 6
command -v systemctl && sudo systemctl restart sshd # CentOS 7
```

Not saying that we should do it this way, but the comment about the containers combined w/ the SSH has me confused.

concourse/scripts/install_hadoop.bash
benchristel: We thought it was exiting because we saw `[1]+  Done                 /usr/sbin/sshd` when we didn't use `-D`. But I just checked, and apparently, there's an sshd process that lives on even when you don't use `-D`.

concourse/scripts/install_hadoop.bash
benchristel: The comment is probably wrong. `systemctl` seems like a good way to go.

concourse/scripts/test_pxf.bash
benchristel: Not sure. We can try it!

concourse/scripts/install_hadoop.bash
soumyadeep2007: If we run systemctl in privileged mode, it still yields the same error:

```
Failed to get D-Bus connection: Operation not permitted
```

benchristel: But Bradford is right that this particular sshd is running in a Google Cloud VM, not in a container. We just did a find/replace all for `service`, and it may not be necessary to do the replacement here.

However, this raises another concern in my mind: the VMs we're starting are probably running CentOS 6, even in the job that is supposed to test CentOS 7. We should rework this so the VMs are using the appropriate platform.

concourse/scripts/test_pxf.bash
soumyadeep2007: It works!

concourse/scripts/install_hadoop.bash
soumyadeep2007: My apologies: I missed the point that this runs in a VM - systemctl is going to work in that context.

soumyadeep2007: w/ @sambitesh 
We agree that we are not running these on a CentOS 7 VM. Making those changes however, is outside the scope of this PR, since this PR is concerned chiefly with the integration testing pipeline. We will undo this change, as this piece of code is part of another pipeline(s).

concourse/scripts/pxf-perf-multi-node.bash
```diff
@@ -58,7 +58,7 @@ function writable_external_table_parquet_query() {
 ###########################################
 
 function setup_sshd() {
-    service sshd start
+    /usr/sbin/sshd &
```
soumyadeep2007: w/ @sambitesh 
This is also out of scope for this PR as we are not concerned with the performance testing pipeline for PXF, we are concerned only with the integration testing pipeline.
We should revert this.

concourse/scripts/pxf_common.bash
```diff
@@ -50,7 +50,10 @@ function install_gpdb_binary() {
 	fi
 
 	if [[ ${TARGET_OS} == "centos" ]]; then
-		service sshd start
+		# We can't use service sshd restart as service is not installed on CentOS 7.
+		# Also, we got "Operation not permitted" errors even running service in a
+		# privileged container.
+		/usr/sbin/sshd &
```
lavjain: This should be the only change needed to support centos7

concourse/scripts/test_pxf.bash
```diff
@@ -87,6 +87,10 @@ function setup_hadoop() {
 }
 
 function _main() {
+	# kill the sshd background process when this script exits. Otherwise, the
+	# concourse build will run forever.
+	trap "pkill sshd" EXIT
+
```
lavjain: The above should not be needed (we don't use it for centos6 either)

concourse/scripts/test_pxf.bash
```diff
@@ -102,6 +106,9 @@ EOF
 		/root/init-script
 	fi
 
+  # Ping is called by gpinitsystem, which must be run by gpadmin
+	chmod u+s /bin/ping
+
```
lavjain: Same here. Should not be needed.

soumyadeep2007: This is actually required in CentOS 7. The gpinitsystem script calls PING_HOST, which tries to run the ping command and this happens:
```
ping: socket: Operation not permitted
```
which causes gpinitsystem to echo:
```
20190312:19:20:51:003409 gpinitsystem:8ec8bddb-f791-4033-439b-a672e00ec796:gpadmin-[INFO]:-Commencing multi-home checks, please wait...
.20190312:19:20:51:003409 gpinitsystem:8ec8bddb-f791-4033-439b-a672e00ec796:gpadmin-[FATAL]:-Unknown host 8ec8bddb-f791-4033-439b-a672e00ec796 Script Exiting!
========================================
gpinitsystem returned: 2
========================================
```
chmodding enables gpinitsystem to run ping as gpadmin.

concourse/scripts/test_pxf.bash
soumyadeep2007: w/ @sambitesh 
This needs to be incorporated now since we are starting sshd explicitly in the background in pxf_common.bash (above) in place of the service command. In case, there is any error encountered by the script, this needs to execute to ensure that Concourse doesn't freeze. (Concourse waits indefinitely if there is something running in the background)

concourse/scripts/pxf_common.bash
soumyadeep2007: w/ @sambitesh 
Please refer to replies to review comments below.

lavjain:
Thanks for taking a stab at this. The changes should be made to pxf_common.sh instead (as it is being used by all pxf tests):

```
--- a/concourse/scripts/pxf_common.bash
+++ b/concourse/scripts/pxf_common.bash
@@ -50,7 +50,7 @@ function install_gpdb_binary() {
        fi

        if [[ ${TARGET_OS} == "centos" ]]; then
-               service sshd start
+               /usr/sbin/sshd &
                psi_dir=$(find /usr/lib64 -name psi | sort -r | head -1)
        elif [[ ${TARGET_OS} == "ubuntu" ]]; then
         # Adjust GPHOME if the binary expects it to be /usr/local/gpdb
@@ -135,6 +135,8 @@ function setup_gpadmin_user() {
        echo -e "gpadmin soft nproc 131072" >> /etc/security/limits.d/gpadmin-limits.conf
        echo -e "gpadmin soft nofile 65536" >> /etc/security/limits.d/gpadmin-limits.conf
        echo -e "export JAVA_HOME=${JAVA_HOME}" >> /home/gpadmin/.bashrc
+       # Ping is called by gpinitsystem, which must be run by gpadmin
+       chmod u+s /bin/ping
        if [[ -d gpdb_src/gpAux/gpdemo ]]; then
                chown -R gpadmin:gpadmin gpdb_src/gpAux/gpdemo
        fi
```

benchristel:
@soumyadeep2007 and I talked with @lavjain and concluded that since PXF is a jar and should have no dependencies on the OS, it's okay to move ahead with Beta without testing on CentOS 7. We may want to revisit this PR later, but for now we are closing it.

PR #110, opened 2019-03-12
Parquet column projection

frankgh:
- Enable column projection for the parquet profile.
- Allows for greenplum schema and parquet schema to be in different order (but names need to match).
- Supports parquet schemas with mixed case i.e. two columns with the same name in different case "column" and "CoLuMn"
- Supports a subset of columns in the target schema

### DIFF
--- a/server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/ParquetResolverTest.java
+++ b/server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/ParquetResolverTest.java
@@ -212,7 +212,7 @@ public class ParquetResolverTest {
         assertField(fields, 2, 1, DataType.INTEGER);
         assertField(fields, 3, 6.0d, DataType.FLOAT8);
         assertField(fields, 4, BigDecimal.valueOf(1234560000000000000L, 18), DataType.NUMERIC);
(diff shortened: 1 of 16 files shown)


server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/ParquetResolverTest.java
```diff
@@ -424,4 +505,24 @@ private MessageType getParquetSchemaForPrimitiveTypes(Type.Repetition repetition
...
+        return schema.getFields()
+                .stream()
+                .map(f -> {
+                    ParquetTypeConverter converter = ParquetTypeConverter.from(f.asPrimitiveType());
+                    return new ColumnDescriptor(f.getName(), converter.getDataType(f).getOID(), 1, "", new Integer[]{});
+                })
+                .collect(Collectors.toList());
+    }
+
+    private MessageType buildReadSchema(MessageType originalSchema) {
```
lavjain: Why not use the buildReadSchema method from parquetFileAccessor instead of re-defining it here?

frankgh: they are doing different things, and the logic is different. buildReadSchema here is only for testing ParquetResolver

PR #111, opened 2019-03-19
Fixed propagation of write exception from the JDBC plugin

denalex:


denalex:
@leskin-in : please take a look, we fixed a bug where JDBC plugin would not fail the INSERT query if there was backend problem. It manifested when POOL_SIZE was not enabled, callable was called directly and the returned exception was ignored (callable return exceptions and do not throw them).

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/writercallable/BatchWriterCallable.java
```diff
@@ -71,6 +72,10 @@ public SQLException call() throws IOException, SQLException, ClassNotFoundExcept
         try {
             statement.executeBatch();
         }
+        catch (BatchUpdateException bue) {
+            SQLException cause = bue.getNextException();
```
leskin-in: It is unclear why this is necessary. `BatchUpdateException.printStackTrace()` includes a detailed explanation of error cause; it is also a subclass of `SQLException`.

denalex: This was necessary since the error message presented to the user in GPDB was not informative, it was saying that `BatchUpdate failed, use getNextException() method to find the cause`. With this code we will give the actual problem's message to the end user, if there was a cause reported.

PR #88, opened 2019-02-05
Support Parquet repeated primitive types serialized into JSON

denalex:

PR #112, opened 2019-03-19
Support additional parquet write config options

lavjain:
As correctly pointed out by @lisakowen we missed merging https://github.com/greenplum-db/pxf/pull/88/commits/db30a656d8b7639e452e6d3400d5b52d933e06c3 into master because the code was refactored.

### DIFF
--- a/server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetFileAccessor.java
+++ b/server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetFileAccessor.java
@@ -82,16 +82,13 @@ public class ParquetFileAccessor extends BasePlugin implements Accessor {
     private RecordReader<Group> recordReader;
     private GroupRecordConverter groupRecordConverter;
     private GroupWriteSupport groupWriteSupport;
+    private WriterVersion parquetVersion;
     private FileSystem fs;
     private Path file;
     private String filePrefix;
-    private int fileIndex;
-    private long rowsRead;
-    private long rowsWritten;
-    private long totalRowsRead;
-    private long totalRowsWritten;
-    private long rowsInRowGroup;
-    private long rowGroupsReadCount;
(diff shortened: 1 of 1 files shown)


server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetFileAccessor.java
```diff
@@ -210,6 +207,17 @@ public boolean openForWrite() throws IOException {
         String compressCodec = context.getOption("COMPRESSION_CODEC");
         codecName = getCodec(compressCodec);
 
+        // Options for parquet write
+        String pageSizeStr = context.getOption("PAGE_SIZE");
+        pageSize = pageSizeStr != null ? Integer.parseInt(pageSizeStr) : DEFAULT_PAGE_SIZE;
+        String rowgroupSizeStr = context.getOption("ROWGROUP_SIZE");
+        rowgroupSize = rowgroupSizeStr != null ? Integer.parseInt(rowgroupSizeStr) : DEFAULT_ROWGROUP_SIZE;
+        String dictSizeStr = context.getOption("DICTIONARY_PAGE_SIZE");
+        dictionarySize = dictSizeStr != null ? Integer.parseInt(dictSizeStr) : DEFAULT_DICTIONARY_PAGE_SIZE;
```
denalex: maybe introduce a method `Integer context.getIntOption(String name, Integer default)` to make all above one-liners ?

denalex:
Need some tests here (unit / integration) to make sure we pass data from context to Parquet, especially proof that we can write v2 Parquet format. Can we even read v2 format ?

lavjain:
@denalex We would need to delay this feature as currently there are no tests for ParquetAccessor. This PR was just for the changes that were missed.

denalex:
I'm ok with row group size, etc, but we are opening this up to support v2 without any tests. Either we have to add a roundtrip test for writing/reading v2, or remove v2 support from this PR. The rest of params is ok to go.

PR #113, opened 2019-03-20
added PARQUET_VERSION parameter and tests

denalex:


frankgh:
can you briefly describe why this is needed?

denalex:
This was part of the original changes that we pulled back due to lack of the test. Parquet v2 has better performance due to additional available encodings: https://www.waitingforcode.com/apache-parquet/schema-versions-parquet/read

We know some GPHDFS users are writing Parquet v2, so we need to be on-par with PXF.

PR #114, opened 2019-03-22
enable file-based configuration for JDBC plugin

denalex:
This PR will enable file-based configuration for the JDBC plugin, where there can be a server created under `PXF_CONF/servers` and in `jdbc-site.xml` configuration file there the `driver`, `url`, `user` and `password` can now be specified instead of providing them in a table DDL. They can be still provided in DDL and should override the values specified in the configuration file.

There are a few tasks outstanding, but want to get an early feedback as we would want to merge this by Mon 3/25. Still to be done:
- doc updates
- write test case
- override test case
- make sure multi-node test works

Also, the overall code design needs to be refactored, bunch of logic in `JdbcBasePlugin` that is only relevant to `Accessor`, SQL connection management will also need to get abstracted, etc ... perhaps for a separate PR.

@leskin-in -- will really appreciate your review and feedback. Also, if you can please test with your environment, especially with others DBs. In our test setup user and password are not required to connect back to GPDB, would be great to know they work as expected when used against Oracle, etc.


automation/Makefile
```diff
@@ -82,6 +90,16 @@ clean-logs:
 dev: symlink_pxf_jars
 	$(MVN) $(MAVEN_TEST_OPTS) package -DskipTests=true
 
+sync_jdbc_config:
```
frankgh: this logic in the makefile... :)

automation/src/main/java/org/greenplum/pxf/automation/structures/tables/utils/TableFactory.java
```diff
@@ -367,12 +367,17 @@ private static ExternalTable getPxfJdbcReadableTable(String tableName,
                                                          String[] fields, String dataSourcePath, String driver,
                                                          String dbUrl, boolean isPartitioned,
                                                          Integer partitionByColumnIndex, String rangeExpression,
-                                                         String interval, String user, EnumPartitionType partitionType) {
+                                                         String interval, String user, EnumPartitionType partitionType,
```
frankgh: This is getting ugly, maybe an opportunity to introduce the builder pattern here?

server/pxf-api/src/main/java/org/greenplum/pxf/api/model/RequestContext.java
```diff
@@ -117,6 +117,29 @@ public String getOption(String option, String defaultValue) {
         return options.getOrDefault(option, defaultValue);
     }
 
+    public int getOption(String option, int defaultValue) {
+        return getOption(option, defaultValue, false);
+    }
+
+    public int getOption(String option, int defaultValue, boolean naturalOnly) {
```
frankgh: we need javadocs here

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
```diff
@@ -42,29 +44,73 @@
  */
 public class JdbcBasePlugin extends BasePlugin {
 
+    /*
+        At the moment, when writing into some table, the table name is concatenated with a special string
+        that is necessary to write into HDFS. However, a raw table name is necessary in case of JDBC.
+        This Pattern allows to extract the correct table name from the given RequestContext.dataSource
+     */
+    private static final Pattern tableNamePattern = Pattern.compile("/(.*)/[0-9]*-[0-9]*_[0-9]*");
```
frankgh: we should be able to get rid of this pretty soon

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
```diff
@@ -129,19 +148,19 @@ public void initialize(RequestContext context) {
      */
     public Connection getConnection() throws ClassNotFoundException, SQLException, SQLTimeoutException {
         Connection connection;
-        if (user != null) {
+        if (jdbcUser != null) {
             LOG.debug("Open JDBC connection: driver={}, url={}, user={}, pass={}, table={}",
-                    jdbcDriver, dbUrl, user, pass, tableName);
+                    jdbcDriver, jdbcUrl, jdbcUser, maskPassword(jdbcPassword), tableName);
```
frankgh: +1 for masking the password

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
```diff
@@ -42,29 +44,73 @@
...
+        jdbcUrl = configuration.get(JDBC_URL_PROPERTY_NAME);
+        assertMandatoryParameter(jdbcUrl, JDBC_URL_PROPERTY_NAME, JDBC_URL_OPTION_NAME);
 
-        dbUrl = context.getOption("DB_URL");
-        if (dbUrl == null) {
-            throw new IllegalArgumentException("DB_URL is a required parameter");
+        // authentication information is not required
+        jdbcUser = configuration.get(JDBC_USER_PROPERTY_NAME);
+        if (jdbcUser != null) {
+            jdbcPassword = configuration.get(JDBC_PASSWORD_PROPERTY_NAME);
```
frankgh: `getpassword` for jceks support?
```suggestion
            jdbcPassword = configuration.getPassword(JDBC_PASSWORD_PROPERTY_NAME);
```

server/pxf-service/src/main/resources/pxf-profiles-default.xml
```diff
@@ -186,6 +186,12 @@ under the License.
             <accessor>org.greenplum.pxf.plugins.jdbc.JdbcAccessor</accessor>
             <resolver>org.greenplum.pxf.plugins.jdbc.JdbcResolver</resolver>
         </plugins>
+        <optionMappings>
+            <mapping option="jdbc_driver" property="jdbc.driver"/>
```
frankgh: I would add a short version of some of these params, for simplicity of use. And keep the old ones for backwards compatibility.
```suggestion
            <mapping option="jdbc_driver" property="jdbc.driver"/>
            <mapping option="driver" property="jdbc.driver"/>
```

server/pxf-service/src/main/resources/pxf-profiles-default.xml
```diff
@@ -186,6 +186,12 @@ under the License.
...
+            <mapping option="db_url" property="jdbc.url"/>
```
frankgh: Same here?
```suggestion
            <mapping option="db_url" property="jdbc.url"/>
            <mapping option="url" property="jdbc.url"/>
```

automation/Makefile
denalex: yea, I know, well, that's the way it is for now ...

automation/src/main/java/org/greenplum/pxf/automation/structures/tables/utils/TableFactory.java
denalex: good idea, but perhaps in a different PR ?

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
denalex: this would require converting returned byte[] to String and handling possible exception -- let's do it when we officially support the password encryption feature ?

server/pxf-service/src/main/resources/pxf-profiles-default.xml
denalex: we should not be, IMO, promoting DDL use of these parameters, so maybe the current names are ok ?

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
frankgh: yeah, makes sense

server/pxf-service/src/main/resources/pxf-profiles-default.xml
frankgh: good point

leskin-in:
@denalex,
I find the new implementation of options and context parsing (introduced [here](https://github.com/greenplum-db/pxf/pull/114/files#diff-7485e17f411fe4b144fb6415cbbf1092R58)) well-designed and more clear than in #99.

However, we may have to introduce significant extra changes to it. [`JdbcPartitionFragmenter`](https://github.com/greenplum-db/pxf/blob/jdbc-server-config/server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcPartitionFragmenter.java#L52) uses options from context to generate partitions. The new mechanism of settings "parsing" deals only with `JdbcBasePlugin` class, leaving the old implementation in `JdbcPartitionFragmenter` unchanged. Obviously, this creates code inconsistency between those two classes.

I believe we should avoid this, so that we use the same mechanism to extract settings in all classes that are part of PXF-JDBC plugin. (Also, this was one of the reasons why `JdbcPluginSettings` was introduced in #99 ).

Note that `JdbcPartitionFragmenter` may be constructed before (and separately from) `JdbcBasePlugin` to only generate partitions. Due to this, deriving `JdbcPartitionFragmenter` from `JdbcBasePlugin` does not look like a good idea (though this is a possible solution to the problem I mentioned above).

server/pxf-service/src/templates/user/templates/jdbc-site.xml
```diff
@@ -0,0 +1,23 @@
...
+        <name>jdbc.user</name>
+        <value>YOUR_DATABASE_JDBC_USER</value>
+        <description>User name for connecting to the database (e.g. postgres)</description>
+    </property>
+    <property>
+        <name>jdbc.password</name>
+        <value>YOUR_DATABASE_JDBC_PASSWORD</value>
+        <description>Password for connecting to the database (e.g. postgres)</description>
+    </property>
+</configuration>
```
lisakowen: a few questions @denalex :

1.  suppose user1 has access permissions to databases db1 and db2.  the `jdbc.url` as described above includes the database name.  if the user wants to access both databases, must they define a separate jdbc server config for each database?
2.  if the user specifies a DB_URL in the CREATE EXTERNAL TABLE, will that override a jdbc.url setting in the server config they are using?

denalex: @lisakowen -- yes on both questions:
1. It will be the best practice to define 2 different configuration servers for 2 different databases. Potentially the end-user can just override the db url in DDL and reuse driver / username / password, but I would not recommend doing this.
2. Yes, all parameters specified in DDL will override those defined in the server configuration file. These parameters need to be whitelisted in profiles.xml file (which they are for JDBC) and the code needs to be written to fetch them from Configuration object (which it is for JDBC).

denalex:
@leskin-in -- thank you for your feedback. Let me address a few of your points:

`JdbcPartitionFragmenter` -- I reviewed it and decided to leave it intact because:
- it parses the options it needs from the context just fine
- it doesn't use any configuration properties, only DDL options provided by the user in the request
- the set of options it processes is mutually exclusive from the set of options that `JdbcBasePlugin` deals with, so there is no common parsing logic that is repeated
- and that is fine, since the design for Plugins is that they get whatever they need from the context

Parsing that happens during initialization of a plugin (fragmenter, accessor and resolver each would have their initialize() method called during a request) should be relevant to what particular plugin needs. Having the plugin parse information it does not need is not efficient, but trying to optimize and parse it once and pass between plugins would violate plugin independence.

That brings me to another point, where I agree with you -- we should not be extending `JdbcPartitionFragmenter` from `JdbcBasePlugin`, since we do not need all that additional parsing for the Fragmenter. To say more, all that parsing (other than `columns` information) is not needed for `JdbcResolver` either (which currently extends `JdbcBasePlugin`). So if the parsing is not needed for Fragmenter and Resolver, it is only then really needed for `JdbcAccessor`, where we should move it in the near future under a separate PR.

To take it one step further, after we move the parsing logic to `JdbcAccessor`, then only `SQLConnection / PreparedStatement` management methods would remain in `JdbcBasePlugin`, which is also only needed by the `JdbcAccessor`, so they should be moved there as well (or perhaps to a dedicated utility). This might make `JdbcBasePlugin` empty and moot and we might as well delete it, and nothing pleases me more than deleting code we don't have to maintain :-)

Those were my ideas and motivations when I mentioned in this PR original comment `"... Also, the overall code design needs to be refactored, bunch of logic in JdbcBasePlugin that is only relevant to Accessor, SQL connection management will also need to get abstracted, etc ... perhaps for a separate PR."`

I hope I managed to convince you, but I might be missing some nuances, so please let me know what you think about it.

leskin-in:
@denalex ,
I agree with your points. It also seems logical to remove inheritance of `JdbcResolver` from `JdbcPlugin`. The PR overall looks good to me.

leskin-in:
I have tested current version on our cluster, and found the case that does not work.

Configuration:
```
<?xml version="1.0" encoding="UTF-8"?>
<configuration>
    <property>
        <name>jdbc.driver</name>
        <value>org.apache.ignite.IgniteJdbcThinDriver</value>
    </property>
</configuration>
```

DDL:
```
CREATE EXTERNAL TABLE ext_ignite_2(k INT, val INT)
LOCATION ('pxf://PUBLIC.T2?PROFILE=JDBC&SERVER=JDBC&DB_URL=sample')
FORMAT 'CUSTOM' (formatter='pxfwritable_import');
```

(first, `DB_URL` was valid, I changed it to exclude the case of incorrect parsing).

Error:
```
psql:5_jdbc_file.sql:9: ERROR:  remote component error (500) from '127.0.0.1:5888':  type  Exception report   message   Required parameter jdbc.url is missing or empty in jdbc-site.xml and option DB_URL is not specified in table definition.    description   The server encountered an internal error that prevented it from fulfilling this request.    exception   java.lang.IllegalArgumentException: Required parameter jdbc.url is missing or empty in jdbc-site.xml and option DB_URL is not specified in table definition. (libchurl.c:944)  (seg0 slice1 10.2.58.3:20000 pid=20003) (cdbdisp.c:254)
DETAIL:  External table ext_ignite_2
```

denalex:
@leskin-in -- I tried to reproduce the issue you're reporting:
- created `$PXF_CONF/servers/JDBC` directory
- put `jdbc-site.xml` in it (used Postgres driver, but it should not matter):
```
$ cat ~/pxf/servers/JDBC/jdbc-site.xml
<?xml version="1.0" encoding="UTF-8"?>
<configuration>
    <property>
        <name>jdbc.driver</name>
        <value>org.postgresql.Driver</value>
    </property>
</configuration>
```
- defined external table:
```
template1=# CREATE EXTERNAL TABLE ext_ignite_2(k INT, val INT)
LOCATION ('pxf://PUBLIC.T2?PROFILE=JDBC&SERVER=JDBC&DB_URL=sample')
FORMAT 'CUSTOM' (formatter='pxfwritable_import');
```
- issued SELECT statement:
```
template1=# select * from ext_ignite_2;
ERROR:  remote component error (500) from '127.0.0.1:5888':  type  Exception report   message   No suitable driver found for sample    description   The server encountered an internal error that prevented it from fulfilling this request.    exception   java.io.IOException: No suitable driver found for sample (libchurl.c:944)  (seg2 slice1 172.17.0.2:25437 pid=23201) (libchurl.c:944)
CONTEXT:  External table ext_ignite_2
```

I think I am getting correct error `No suitable driver found for sample`, as there is no JDBC driver that would process `sample` URL. I am not getting the error you indicated.

Can you maybe debug and see where / how you encounter the error ? Have you recompiled / reinstalled PXF ? Do you have updated `$GPHOME/pxf/conf/pxf-profiles-default.xml` that has newly introduced options mappings ?

leskin-in:
@denalex ,
I have found the cause of error in my case: `$GPHOME/pxf/conf/pxf-profiles-default.xml` was not updated. After update of this file, new PXF works as expected.

Thank you for implementing this feature!

PR #115, opened 2019-03-22
Update the write path to include transaction and segments id

frankgh:
- Only required for the HCFS profiles. There is a corresponding change on the c-side

### DIFF
--- a/server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HcfsType.java
+++ b/server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HcfsType.java
@@ -4,6 +4,8 @@ import org.apache.commons.lang.StringUtils;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileSystem;
 import org.greenplum.pxf.api.model.RequestContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.net.URI;
 import java.util.Arrays;
@@ -44,6 +46,8 @@ public enum HcfsType {
     // We prefer WASBS over WASB for Azure Blob Storage,
     // as it uses SSL for communication to Azure servers
     WASBS;
+
+    protected static final Logger LOG = LoggerFactory.getLogger(HcfsType.class.getName());
 
     private static final String FILE_SCHEME = "file";
     private String prefix;
@@ -103,6 +107,21 @@ public enum HcfsType {
                     String.format("profile protocol (%s) is not compatible with server filesystem (%s)",
                             schemeFromContext, defaultFSScheme));
         }
+    }
+
+    /**
+     * Returns a fully resolved URI including the protocol for write
+     *
+     * @return an absolute data path for write
+     */
+    public String getUriForWrite(Configuration configuration, RequestContext context) {
+        String fileName = StringUtils.removeEnd(getDataUri(configuration, context), "/") +
+                "/" +
+                context.getTransactionId() +
+                "_" +
+                context.getSegmentId();
+        LOG.debug("File name for write: {}", fileName);
+        return fileName;
     }
 
     /**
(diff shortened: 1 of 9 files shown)


server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/LineBreakAccessor.java
```diff
@@ -73,16 +73,14 @@ protected Object getReader(JobConf jobConf, InputSplit split)
...
         // get compression codec
-        if (compressCodec != null) {
-            codec = HdfsUtilities.getCodec(configuration, compressCodec);
-            String extension = codec.getDefaultExtension();
-            fileName += extension;
+        CompressionCodec codec = compressCodec != null ?
+                HdfsUtilities.getCodec(configuration, compressCodec) : null;
+        if (codec != null) {
+            // append codec extension to the filename
+            fileName += codec.getDefaultExtension();
```
denalex: this code gets repeated in multiple places (even in this PR), maybe move to HdfsUtilities or some other common place ?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HcfsType.java
```diff
@@ -45,6 +47,8 @@ public String normalizeDataSource(String dataSource) {
     // as it uses SSL for communication to Azure servers
     WASBS;
 
+    protected static final Logger LOG = LoggerFactory.getLogger(HcfsType.class.getName());
```
denalex: To make is subclass specific:
```suggestion
    protected Logger LOG = LoggerFactory.getLogger(this.getClass());
```

denalex:
There is currently a difference between `ReadBridge` and `WritableBridge` in the way how `context.setDataSource()` is handled. Will this still be required after this change ? Can we make that distinction go away, if possible ?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/LineBreakAccessor.java
frankgh: Yeah, I had moved it to HcfsType.java, but then it did not feel okay for HcfsType to know anything about codecs, maybe I will move it again, or maybe we can find a better place for it

frankgh: I will move it down for now.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/SequenceFileAccessor.java
```diff
@@ -79,11 +78,12 @@ protected Object getReader(JobConf jobConf, InputSplit split) throws IOException
...
-        String filename = hcfsType.getDataUri(configuration, context);
-        LOG.debug("Filename for write without updated file extension: {}", filename);
+        String filename = hcfsType.getUriForWrite(configuration, context);
         getCompressionCodec(context);
-        filename = updateFileExtension(filename, codec);
-
+        if (codec != null) {
+            // append extension to the filename
+            filename += codec.getDefaultExtension();
+        }
```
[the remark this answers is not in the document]
frankgh: it's been added back to the common place where we build the write uri

frankgh:
@denalex yes, I removed it in commit c327aed

frankgh:
> What about JDBC and Ignite plugins, do they still work with corresponding C-side change ?

Integration tests are green with the corresponding C-side change. I think we should address JDBC and Ignite plugins with integration tests after merging this PR.

frankgh:
@denalex commit ba32cc6 updates support for jdbc and ignite.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HcfsType.java
```diff
@@ -105,6 +111,30 @@ private static void checkForConfigurationMismatch(String defaultFSScheme, String
         }
     }
 
+    /**
+     * Returns a fully resolved URI including the protocol for write
+     *
```
shivzone: javadoc would complain in the absence of parameters. Also would be useful to have a short note on how the write filename is constructed in the comments

PR #117, opened 2019-03-26
Add support for 6x pipelines in deploy script

frankgh:


concourse/deploy
```diff
@@ -169,13 +176,19 @@ if __name__ == "__main__":
     if ARGS.gpdb_branch == '5x':
         ARGS.gpdb_branch = '5X_STABLE'
         ARGS.ICW_GREEN_BUCKET = 'gpdb5-stable-concourse-builds'
+    elif ARGS.gpdb_branch == '6x':
+        ARGS.gpdb_branch = '6X_STABLE'
+        ARGS.ICW_GREEN_BUCKET = 'gpdb5-assert-concourse-builds'
```
denalex: doesn't this need come out of google storage now ? at least for 6x / master ?

concourse/deploy
```diff
@@ -169,13 +176,19 @@ if __name__ == "__main__":
...
     elif ARGS.gpdb_branch == 'master':
         ARGS.ICW_GREEN_BUCKET = 'gpdb5-assert-concourse-builds'
-    elif not ARGS.compile_gpdb:
-        sys.stderr.write('error: Only master and 5x branches are supported for GPDB.'
+    else:
+        sys.stderr.write('error: Only master, 6x and 5x branches are supported for GPDB.'
                          'Include --compile_gpdb flag if you intend to compile GPDB from a different branch.\n')
```
denalex: is --compile_gpdb still used, I see you introduced --branch parameter

frankgh: yeah, both are used, I was thinking of combining them into a single parameter, but for now I have decided to keep them separate. (i.e.) In case you want to compile 5X_STABLE and don't want to type 5X_STABLE in the console

concourse/deploy
frankgh: that's what I am trying to figure out. They have icw_green artifacts in amazon, but they have RC in google cloud storage

frankgh: icw_green still goes to S3

PR #118, opened 2019-03-27
PXF-JDBC: Enable external database configuration and connection settings modification

leskin-in:
This PR continues #99 . Closes https://github.com/greenplum-db/pxf/issues/129.

Enable user to modify:
* External database session configuration (GUC variables in PostgreSQL and equivalent entities in other databases. These are modified by execution of `SET`-like DBMS-specific queries for each partition `Fragment`, before "main" `SELECT` or `INSERT` query is executed);
* JDBC connection information (`info` object passed to [`JdbcDriver.getConnection()`](https://docs.oracle.com/javase/8/docs/api/java/sql/DriverManager.html#getConnection-java.lang.String-java.util.Properties-)).

To do that, two new configuration parameters (with common syntax) are introduced:
* `jdbc.env`. Controls external database session configuration
* `jdbc.info`. Controls JDBC connection information object

The use case solved by this feature is described in #99 (`SET` "environment" variables in external database before executing "main" SQL query) and in https://github.com/greenplum-db/pxf/issues/129 (set JDBC driver connection parameter).

Other changes:
* Update README.md
* Update template `jdbc-site.xml`
* Add unit tests for the new feature


server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/SQLQueryBuilder.java
```diff
@@ -134,6 +132,25 @@ public String buildInsertQuery() throws SQLException {
...
+     *
+     * @return Complete SQL query
+     * @return null if {@link DbProduct} disallows env queries
+     *
+     * @throws SQLException
+     */
+    public String buildEnvQuery(Properties envs) throws SQLException {
+        String result = envs.entrySet().stream()
+                .map(env -> dbProduct.buildEnvQuery(env))
+                .filter(query -> query != null && !query.isEmpty())
```
frankgh: Will this work?
```suggestion
                .filter(StringUtils::isNotEmpty)
```

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/SQLQueryBuilder.java
```diff
@@ -134,6 +132,25 @@ public String buildInsertQuery() throws SQLException {
...
+     * @param envs {@link Properties} object from {@link JdbcBasePlugin}
+     *
+     * @return Complete SQL query
+     * @return null if {@link DbProduct} disallows env queries
+     *
+     * @throws SQLException
+     */
+    public String buildEnvQuery(Properties envs) throws SQLException {
+        String result = envs.entrySet().stream()
+                .map(env -> dbProduct.buildEnvQuery(env))
```
frankgh:
```suggestion
                .map(dbProduct::buildEnvQuery)
```

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/SQLQueryBuilder.java
```diff
@@ -134,6 +132,25 @@ public String buildInsertQuery() throws SQLException {
         return sb.toString();
     }
 
+    /**
+     * Build query to set environment variables in external database.
```
frankgh: is this statement more accurate?
```suggestion
     * Build query to set session-level variables in external database.
```

server/pxf-jdbc/src/test/java/org/greenplum/pxf/plugins/jdbc/utils/DbProductTest.java
```diff
@@ -105,6 +128,21 @@ public void testOracleTimestamps() {
         }
     }
 
+    @Test
+    public void testOracleBuildEnvQuery() {
+        final Properties EXPECTED = new Properties();
```
frankgh: I think these tests should be simplified to the following
```suggestion
        DbProduct dbProduct = DbProduct.getDbProduct(DB_NAME_ORACLE);
        assertEquals("ALTER SESSION SET key = value;", dbProduct.buildEnvQuery(entry));
```

leskin-in: I would prefer these tests to stay unchanged. 

There is [a cycle](https://github.com/greenplum-db/pxf/pull/118/files#diff-698f3ad022ae10ac10c6618a799970ceR140) over `ENVS` (which may contain multiple values). In addition, due to the fact that `dbProduct.buildEnvQuery()` receives `Entry` (which can only be retrieved from `Properties`), some kind of [iteration](https://github.com/greenplum-db/pxf/pull/118/files#diff-698f3ad022ae10ac10c6618a799970ceR138) would still be required even in case we move to the proposed simplified comparison.

leskin-in:
Thank you for the comments, @frankgh !

Example with variables is now included in `jdbc-site.xml`.

server/pxf-service/src/templates/user/templates/jdbc-site.xml
```diff
@@ -20,4 +20,9 @@
         <value>YOUR_DATABASE_JDBC_PASSWORD</value>
         <description>Password for connecting to the database (e.g. postgres)</description>
     </property>
+    <property>
```
frankgh:
```suggestion

    <!-- 
       -- Uncomment to set session-level variables for queries accessing the database
    <property>
```

server/pxf-service/src/templates/user/templates/jdbc-site.xml
```diff
@@ -20,4 +20,9 @@
...
+        <name>jdbc.env</name>
+        <value>YOUR_VAR_1=YOUR_VALUE_1, YOUR_VAR_2=YOUR_VALUE_2</value>
+        <description>Session-level variables to set in external database before query execution</description>
+    </property>
```
frankgh: -->

leskin-in:
@frankgh ,
a few days ago, https://github.com/greenplum-db/pxf/issues/129 was opened, which is quite close to what the current PR is about. Due to this, I have implemented a patch to address the mentioned issue.

We could open an extra PR for that, but I suppose both instruments (session configuration and JDBC connection info management) complement each other, so introducing them both would cover more use cases. For example, PostgreSQL JDBC driver was unable to `SET` arbitrary GUC variables passed in JDBC connection info until recently.

`jdbc.env` and `jdbc.info` (the newly-introduced settings) now have the same syntax.

I will be glad to receive any comments on this.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
```diff
@@ -117,34 +129,60 @@ public void initialize(RequestContext context) {
...
-                    jdbcDriver, jdbcUrl, tableName);
-        }
-        Class.forName(jdbcDriver);
-        if (jdbcUser != null) {
-            connection = DriverManager.getConnection(jdbcUrl, jdbcUser, jdbcPassword);
+    public Connection getConnection() throws ClassNotFoundException, SQLException {
+        if (LOG.isDebugEnabled()) {
+            LOG.debug(String.format(
+                "New JDBC connection. Driver: '%s'; URL: '%s'",
+                jdbcDriver, jdbcUrl
```
denalex: no need to String.format with SLF4J which does it automatically, like in the replaced syntax

denalex: why replace the original logging message ? now table name is missing

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
```diff
@@ -227,7 +265,14 @@ private void assertMandatoryParameter(String value, String paramName, String opt
...
+    private static String maskPassword(String password) {
         return password == null ? "" : StringUtils.repeat("*", password.length());
     }
+
+    /**
+     * Retrieve password from data available to this plugin
+     */
+    private String retrievePassword() {
+        return configuration.get(JDBC_PASSWORD_PROPERTY_NAME);
+    }
```
denalex: not sure why this one-liner is needed

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/SQLQueryBuilder.java
```diff
@@ -134,6 +132,25 @@ public String buildInsertQuery() throws SQLException {
...
+     * @return Complete SQL query
+     * @return null if {@link DbProduct} disallows env queries
+     *
+     * @throws SQLException
+     */
+    public String buildEnvQuery(Properties envs) throws SQLException {
+        String result = envs.entrySet().stream()
+                .map(dbProduct::buildEnvQuery)
+                .filter(StringUtils::isNotEmpty)
+                .collect(Collectors.joining(" "));
```
denalex: do we need to check that the value does not contain semi-column to avoid SQL-injection vulnerability ? For example, someone can do this: `"foo=bar;delete * from table"` for a property

server/pxf-jdbc/src/test/java/org/greenplum/pxf/plugins/jdbc/utils/PropertiesParserTest.java
```diff
@@ -0,0 +1,210 @@
...
+        PropertiesParser.parse(toParse);
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testThrowExceptionOnEmptyValue() {
+        final String toParse = "a=1,b=,c=3";
+
+        PropertiesParser.parse(toParse);
+    }
+}
```
denalex: nice tests. Do you guys think you can create an integration test for this against greenplum ?

server/pxf-jdbc/README.md
```diff
@@ -56,162 +53,155 @@ FRAGMENTER=org.greenplum.pxf.plugins.jdbc.JdbcPartitionFragmenter
...
 
 
-## Examples
+#### JDBC connection information
+*Can be set only in configuration file*
 
-The following example shows how to access a MySQL table via JDBC.
+Connection information passed to JDBC driver when opening connection to external database. See [Java documentation](https://docs.oracle.com/javase/8/docs/api/java/sql/DriverManager.html#getConnection-java.lang.String-java.util.Properties-) for the meaning of this object. See external database documentation for information on supported key-value pairs.
+
+* **Configuration parameter**: `jdbc.info`
```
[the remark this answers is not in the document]
shivzone: I agree. Things become a lot more straightforward during troubleshooting or debugging if we use clear namespacing for properties and simpley key value pairs and avoid command separated properties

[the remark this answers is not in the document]
leskin-in: Thanks a lot for the comment! I thought of something like this, too, but missed a method for that in Apache Hadoop `Configuration` object API.

server/pxf-jdbc/README.md
```diff
@@ -56,162 +53,155 @@ FRAGMENTER=org.greenplum.pxf.plugins.jdbc.JdbcPartitionFragmenter
...
 
-`QUOTE_COLUMNS` controls whether PXF should quote column names when constructing SQL query to the external database. Set this parameter to `true` (case-insensitive) to force PXF to quote all column names; set to any other value to prevent PXF from quoting column names. By default (`QUOTE_COLUMNS` not present), PXF automatically checks whether some column name should be quoted, and if so, it quotes all column names in the query.
+* Configuration file located at `$PXF_CONF/servers/<server_name>-site.xml` (on every PXF segment), where `<server_name>` is an arbitrary name (the file is intended to include options specific for each external database server). Hereinafter setting set in configuration file is referred to as "configuration parameter".
 
-The meaning of other parameters is given in section [partitioning](#Partitioning).
+PXF will not pull any configuration parameters if `SERVER` option is not set correctly in external table DDL. A warning is added to PXF log file if `SERVER` is set to incorrect value, and PXF is unable to read configuration file.
```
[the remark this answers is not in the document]
leskin-in: Thank you for the correction. I have to note that a warning is added to PXF logs, though.

server/pxf-jdbc/README.md
```diff
@@ -56,162 +53,155 @@ FRAGMENTER=org.greenplum.pxf.plugins.jdbc.JdbcPartitionFragmenter
...
 
+If an option is set, it overrides configuration parameter with the same name. Some settings can be set only in configuration file or only in `LOCATION` clause of external table DDL.
```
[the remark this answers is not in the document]
leskin-in: I would be glad to know your opinion on [current version](https://github.com/greenplum-db/pxf/pull/118/commits/e7ed43dba5b8fc5832b955d71be2b01af5b29bd7#diff-d2f69cfc2b417ca813bc26aabca0b97fR71) of this line.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
leskin-in: I suppose we could use some sophisticated mechanism to retrieve password here. There were comments on that in other PRs, so by intoducing this function, we could prevent future merge conflicts when such mechanism is introduced.

We could return to the old implementation (when this line was inside `initalize()`). I'm really not sure which is better. Do you find the current version useful for the listed reasons?

server/pxf-jdbc/README.md
```diff
@@ -257,3 +238,152 @@ Finally, a query to a GPDB external table is made:
...
+        <name>jdbc.driver</name>
+        <value>org.example.JdbcDriver</value>
+    </property>
+    <property>
+        <name>jdbc.url</name>
+        <value>jdbc:example://1.2.3.4:12345</value>
+    </property>
+    <property>
+        <name>jdbc.env</name>
+        <value>key1=value1, key2=value2</value>
```
denalex: documentation need to change to correspond to updated design with individual property names.

server/pxf-jdbc/README.md
```diff
@@ -56,162 +53,151 @@ FRAGMENTER=org.greenplum.pxf.plugins.jdbc.JdbcPartitionFragmenter
...
-The PXF JDBC plugin will generate two fragments  for a query `SELECT * FROM sales`. Then GPDB will assign each of them to a separate PXF segment. Each segment will perform the SELECT query, and the first one will get tuples with `cdate` values for year `2008`, while the second will get tuples for year `2009`. Then each PXF segment will send its results back to GPDB, where they will be "concatenated" and returned.
 
+#### JDBC connection properties
+*Can be set only in configuration file*
 
-## Examples
+Connection properties (`java.util.Properties`) passed to JDBC driver when opening a connection to external database. See [Java documentation](https://docs.oracle.com/javase/8/docs/api/java/sql/DriverManager.html#getConnection-java.lang.String-java.util.Properties-) for the meaning of this object. See external database documentation for information on supported key-value pairs.
 
-The following example shows how to access a MySQL table via JDBC.
+* **Configuration parameter**: `jdbc.connection.property.<name>`, where `<name>` is the name of a session-level variable
```
denalex: maybe "... name of connection level variable ..."

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcAccessor.java
```diff
@@ -78,7 +78,15 @@ else if (quoteColumns) {
         queryRead = sqlQueryBuilder.buildSelectQuery();
         LOG.trace("Select query: {}", queryRead);
 
+        // Build env query
+        String queryEnv = sqlQueryBuilder.buildEnvQuery(sessionConfiguration);
```
denalex: I would've called this `sessionQuery` and the method `buildSessionQuery()`

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcAccessor.java
```diff
@@ -136,6 +144,15 @@ else if (quoteColumns) {
         queryWrite = sqlQueryBuilder.buildInsertQuery();
         LOG.trace("Insert query: {}", queryWrite);
 
+        // Build and execute env query
+        String queryEnv = sqlQueryBuilder.buildEnvQuery(sessionConfiguration);
```
denalex: same here for the name of the variable

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
```diff
@@ -50,13 +53,15 @@
     private static final String JDBC_URL_PROPERTY_NAME = "jdbc.url";
     private static final String JDBC_USER_PROPERTY_NAME = "jdbc.user";
     private static final String JDBC_PASSWORD_PROPERTY_NAME = "jdbc.password";
+    private static final String JDBC_SESSCONF_PROPERTY_PREFIX = "jdbc.session.property.";
```
denalex: I'd call it `JDBC_SESSION_PROPERTY_PREFIX` for clarity and consistency, same for the constant below.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
```diff
@@ -117,34 +129,53 @@ public void initialize(RequestContext context) {
         if (quoteColumnsRaw != null) {
             quoteColumns = Boolean.parseBoolean(quoteColumnsRaw);
         }
+
+        // This parameter is not required. The default value is empty map
+        Map<String, String> sessionConfigurationRaw = configuration.getPropsWithPrefix(JDBC_SESSCONF_PROPERTY_PREFIX);
+        sessionConfiguration.putAll(sessionConfigurationRaw);
+        LOG.debug("Session configuration: {}",
```
denalex: this needs to get wrapped in ` if Log.isDebugEnabled() { ... }` since the log method does intensive computation to come up with parameter values.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
```diff
@@ -117,34 +129,53 @@ public void initialize(RequestContext context) {
...
+        sessionConfiguration.putAll(sessionConfigurationRaw);
+        LOG.debug("Session configuration: {}",
+            sessionConfiguration.entrySet().stream()
+                    .map(entry -> "'" + entry.getKey().toString() + "'='" + entry.getValue() + "'")
+                    .collect(Collectors.joining(", "))
+        );
+
+        // This parameter is not required. The default value is empty map
+        Map<String, String> connectionConfigurationRaw = configuration.getPropsWithPrefix(JDBC_CONNCONF_PROPERTY_PREFIX);
+        connectionConfiguration.putAll(connectionConfigurationRaw);
```
denalex: 2 lines above can be combined as connectionConfigurationRaw variable is never referred to anywhere else.
```suggestion
        connectionConfiguration.putAll(configuration.getPropsWithPrefix(JDBC_CONNECTION_PROPERTY_PREFIX));
```

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
```diff
@@ -117,34 +129,53 @@ public void initialize(RequestContext context) {
...
         }
-        else {
-            connection = DriverManager.getConnection(jdbcUrl);
+
+        if (jdbcPassword != null) {
+            connectionConfiguration.setProperty("password", jdbcPassword);
         }
-        return connection;
+
+        Class.forName(jdbcDriver);
```
denalex: why all of the above is done here and not in `initialize()` along with `user` property ? Connection parameters will not change from request to request, at least not until we enable impersonation and user mapping in JDBC connector. This way we do not need to store `jdbcDriver` and `jdbcPassword` as member variables. Once `connectionConfiguration` is created and populated in `initialize()` method, it can be reused for subsequent calls to get connection.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
```diff
@@ -214,7 +245,7 @@ public static void closeStatement(Statement statement) {
      * @param paramName parameter name
      * @param optionName name of the option for a given parameter
      */
-    private void assertMandatoryParameter(String value, String paramName, String optionName) {
+    private static void assertMandatoryParameter(String value, String paramName, String optionName) {
```
denalex: what is the benefit of declaring them as static if these are private methods and are not called from outside of the class ? Yes, they don't use any member fields and can be static, but I am not sure they should.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
denalex: yes, we will make password retrieval more complex in the future, but for the current state it is not really needed. I don't really care one way or another, but usually stick to the principal "we will make it such when we really need it".

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/SQLQueryBuilder.java
denalex: I think you agreed with my argument, but I still don't see the logic to screen for the semicolons, maybe I missed it ?

server/pxf-service/src/templates/user/templates/jdbc-site.xml
```diff
@@ -20,4 +20,18 @@
         <value>YOUR_DATABASE_JDBC_PASSWORD</value>
         <description>Password for connecting to the database (e.g. postgres)</description>
     </property>
+    <!-- Uncomment to set session-level variables for queries to external database
+    <property>
+        <name>jdbc.env</name>
+        <value>YOUR_VAR_1=YOUR_VALUE_1,YOUR_VAR_2=YOUR_VALUE_2</value>
```
denalex: the template needs to change to correspond to updated design with individual property names.

server/pxf-service/src/templates/user/templates/jdbc-site.xml
```diff
@@ -20,4 +20,18 @@
...
+        <description>Session-level variables to set in external database before query execution</description>
+    </property>
+    -->
+    <!-- Uncomment to set JDBC connection information
+    <property>
+        <name>jdbc.info</name>
+        <value>YOUR_VAR_1=YOUR_VALUE_1,YOUR_VAR_2=YOUR_VALUE_2</value>
```
denalex: the template needs to change to correspond to updated design with individual property names.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/SQLQueryBuilder.java
leskin-in: @denalex ,
indeed, no change to logic was made in https://github.com/greenplum-db/pxf/pull/118/commits/e7ed43dba5b8fc5832b955d71be2b01af5b29bd7.

[This document](https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/SQL_Injection_Prevention_Cheat_Sheet.md) states we should avoid sanitization if possible. That is why I tried to use `PreparedStatement` to execute SET query. However, that does not work (at least with PostgreSQL as external database). Other methods described there do not apply to our case.

As far as I remember, current "agreement" is to trust database administrator. Due to this, complete removal of protection against injection might be considered.

For now, though, I followed your advice and implemented [a check](https://github.com/greenplum-db/pxf/pull/118/commits/717a1b3348ae95ced25cc00257b69899874a56ae#diff-7485e17f411fe4b144fb6415cbbf1092R137) for forbidden characters. Note this implementation limits the abilities of database administrator.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
```diff
@@ -205,38 +220,107 @@ public static void closeConnection(Connection connection) {
...
+     */
+    private static void closeConnection(Connection connection) throws SQLException {
+        if (connection == null) {
+            return;
+        }
+        if (!connection.isClosed()) {
+            if ((connection.getMetaData().supportsTransactions()) && (!connection.getAutoCommit())) {
+                connection.commit();
+            }
+            connection.close();
```
denalex: would this result in non-closed connection if the commit fails ? Usually statements / connections closures happen in `finally` blocks

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
```diff
@@ -205,38 +220,107 @@ public static void closeConnection(Connection connection) {
...
+
+        // Prepare session (process sessionConfiguration)
+        if (!sessionConfiguration.isEmpty()) {
+            DbProduct dbProduct = DbProduct.getDbProduct(connection.getMetaData().getDatabaseProductName());
+
+            try (Statement statement = connection.createStatement()) {
+                for (Map.Entry<Object,Object> e : sessionConfiguration.entrySet()) {
+                    statement.execute(
+                        dbProduct.buildSessionQuery(
+                            String.class.cast(e.getKey()), String.class.cast(e.getValue())
```
denalex: why casts, can't sessionConfiguration be `Map<String, String>` ?

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
```diff
@@ -90,111 +92,124 @@
...
+        // Optional parameter. The default value is empty map
+        sessionConfiguration.putAll(configuration.getPropsWithPrefix(JDBC_SESSION_PROPERTY_PREFIX));
+        // Check forbidden symbols
+        // Note: PreparedStatement tnables us to skip this check: its values are distinct from its SQL code
+        // However, SET queries cannot be executed this way. This is why we do this check
+        if (sessionConfiguration.entrySet().stream()
+                .anyMatch(
+                    entry ->
+                        StringUtils.containsAny(
+                            String.class.cast(entry.getKey()), FORBIDDEN_SESSION_PROPERTY_CHARACTERS
```
denalex: why casts, can't sessionConfiguration be `Map<String, String>` ?

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
```diff
@@ -90,111 +92,124 @@
...
+                        ) ||
+                        StringUtils.containsAny(
+                            String.class.cast(entry.getValue()), FORBIDDEN_SESSION_PROPERTY_CHARACTERS
```
denalex: why casts, can't sessionConfiguration be `Map<String, String>` ?

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
```diff
@@ -90,111 +92,124 @@
...
+                )
+        ) {
+            throw new IllegalArgumentException("Some session configuration parameter contains forbidden characters");
+        }
         if (LOG.isDebugEnabled()) {
-            LOG.debug("New JDBC connection. URL: '{}'; Driver: '{}'; Table: '{}'",
-                jdbcUrl, jdbcDriver, tableName
+            LOG.debug("Session configuration: {}",
+                sessionConfiguration.entrySet().stream()
+                        .map(entry -> "'" + entry.getKey().toString() + "'='" + entry.getValue() + "'")
```
denalex: no need for `toString()`

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
```diff
@@ -90,111 +92,124 @@
...
+     * Open a new JDBC connection
      *
-     * @param connection connection to close
+     * @throws SQLException if a database access or connection error occurs
+     *
+     * @return {@link Connection}
      */
-    public static void closeConnection(Connection connection) {
+    public Connection getConnection() throws SQLException {
+        if (LOG.isDebugEnabled()) {
```
denalex: no need for `isLogDebugEnabled()` check here as parameters are already pre-computed, cost of function call is negligible

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
leskin-in: Thanks a lot for this comment.

leskin-in:
@denalex ,
I have addressed your latest comments. To avoid mistakes with connection closure in the future, a few unit tests were implemented.

Some unit tests exclusively for `initialize()` call of `JdbcBasePlugin` were also added, as that method is now quite complex.

PR #125, opened 2019-03-29
LICENSE: Remove reference to hadoop binaries dependencies

frankgh:
Since we are bundling the libraries, we no longer require the jars
described in the LICENSE file


shivzone:
Do we have to list the dependancies in the license file ? Yet another resource to maintain :( We don't seem to be listing them in other greenplum-db projects

frankgh:
@shivzone totally agree with you that we should not list dependencies in the license file. We should discuss further

PR #131, opened 2019-04-03
Support optional MASTER_HOSTNAME env var

oliverralbertini:
This env var would override the hostname that the operating system
reports, for users who have, for example, an internal network/hostname
for GPDB and external network/hostname for other things. If the hostname
that the system reports is not known to the GPDB, the user can avoid
errors by setting `MASTER_HOSTNAME` to their GPDB-facing hostname.

### DIFF
--- a/cli/go/src/pxf-cli/operating/operating_test.go
+++ b/cli/go/src/pxf-cli/operating/operating_test.go
@@ -1,0 +1,46 @@
+package operating_test
+
+import (
+	"os"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	. "pxf-cli/operating"
+
+	"github.com/greenplum-db/gp-common-go-libs/operating"
+)
+
+var _ = Describe("Operating", func() {
+	BeforeEach(func() {
+		operating.System.Hostname = func() (string, error) {
+			return "system_hostname", nil
+		}
+	})
+
+	Context("SystemHostname", func() {
+		Context("When PXF_MASTER_HOSTNAME env var is unset", func() {
+			It("Reports the system hostname", func() {
+				Expect(SystemHostname()).To(Equal("system_hostname"))
+			})
+		})
+
+		Context("When PXF_MASTER_HOSTNAME env var is set to empty string", func() {
+			BeforeEach(func() {
+				_ = os.Setenv("PXF_MASTER_HOSTNAME", "")
+			})
+			It("Reports the system hostname", func() {
+				Expect(SystemHostname()).To(Equal("system_hostname"))
+			})
+		})
+
+		Context("When PXF_MASTER_HOSTNAME env var is set to something", func() {
+			BeforeEach(func() {
+				_ = os.Setenv("PXF_MASTER_HOSTNAME", "my-real-hostname")
+			})
+			It("Reports the hostname from PXF_MASTER_HOSTNAME", func() {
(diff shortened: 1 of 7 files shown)


oliverralbertini:
Here are the results of trying this on a perf pipeline (a multinode env):

```
[gpadmin@mdw ~]$ MASTER_HOSTNAME="fake-name" pxf cluster stop
ERROR: pxf cluster commands should only be run from Greenplum master
[gpadmin@mdw ~]$ MASTER_HOSTNAME="mdw" pxf cluster stop
Stopping PXF on 3 segment hosts...
PXF stopped successfully on 3 out of 3 hosts
```

When `MASTER_HOSTNAME` matches what GPDB knows (`mdw`) we can run the command, but if it doesn't match, we get a failure.

oliverralbertini:
@larham @benchristel : would you mind taking a look at this PR? Would like some more folks with Go experience to give their opinion on the changes.

lavjain:
Please check with @vraghavan78 on the environment variable name. I would prefer it to be called PXF_MASTER_HOSTNAME to comply with other PXF variables.

vraghavan78:
I agree with @lavjain `PXF_MASTER_HOSTNAME` is a better option

oliverralbertini:
I have renamed the env var. Thanks for the suggestion.

larham:
It seems odd to have PXF-specific code in the middle of a common library.

Consider a different approach:  when you initialize these function pointers, early in running your specific application, you can *override* the given value with your PXF custom logic.  So something like:

```
// early initialization  
operating.System = operating.InitializeSystemFunctions()
operating.System.Hostname = func()  (name string, err error) { ... }
```

oliverralbertini:
> Consider a different approach: when you initialize these function pointers, early in running your specific application, you can _override_ the given value with your PXF custom logic. So something like:
> 
> ```
> // early initialization  
> operating.System = operating.InitializeSystemFunctions()
> operating.System.Hostname = func()  (name string, err error) { ... }
> ```

This was going to be the initial approach, but since this is needed in multiple packages (`pxf`, `cmd`, `pxf_test`, `cmd_test`) I believe I would need to do this for each package. So I opted to make a separate package called `operating`.

larham:
@oliverralbertini Ah, I see. sorry, I confused your package name with that of the common library. Nevermind!

oliverralbertini:
Yeah, unfortunately we share a lot of the same concepts and thus names ;-)

cli/go/src/pxf-cli/operating/operating_test.go
```diff
@@ -0,0 +1,46 @@
...
+		Context("When PXF_MASTER_HOSTNAME env var is set to empty string", func() {
+			BeforeEach(func() {
+				_ = os.Setenv("PXF_MASTER_HOSTNAME", "")
+			})
+			It("Reports the system hostname", func() {
+				Expect(SystemHostname()).To(Equal("system_hostname"))
+			})
+		})
+
+		Context("When PXF_MASTER_HOSTNAME env var is set to something", func() {
```
frankgh: just minor comment here
```suggestion
		Context("When PXF_MASTER_HOSTNAME env var is set to some value", func() {
```

oliverralbertini:
We decided to take a different approach: we will no longer be verifying that the user is on the master host. If this turns out to be a problem, we can bring it back.

PR #134, opened 2019-04-08
pxf cluster: stop checking that hostname is master

oliverralbertini:
Checking that the user is running `pxf cluster` from master is helpful
for ensuring that the source of truth for PXF configs lives on master.
We don't want users doing cluster configuration from segment-only hosts.

However, this causes issues when the hostname reported by the OS doesn't
match what Greenplum knows about, so we are foregoing this check for
now.

### DIFF
--- a/cli/go/src/pxf-cli/cmd/cluster.go
+++ b/cli/go/src/pxf-cli/cmd/cluster.go
@@ -7,8 +7,6 @@ import (
 	"pxf-cli/pxf"
 	"strings"
 
-	"github.com/greenplum-db/gp-common-go-libs/operating"
-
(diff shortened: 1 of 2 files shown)


oliverralbertini:
By the way, here's what happens when we try to run `pxf cluster sync` from a segment-only host:
```
[gpadmin@sdw1 ~]$ pxf cluster sync
ERROR: Could not connect to GPDB.
could not connect to server: Connection refused
        Is the server running on host "sdw1" and accepting
        TCP/IP connections on port 5432?
Please make sure that your Greenplum database is running and you are on the master node.
```

PR #135, opened 2019-04-09
Cache Fragmenter calls

frankgh:
- A simple approach to cache fragmenter calls, keyed by transactionId


frankgh:
This PR needs testing.

server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/FragmenterResource.java
```diff
@@ -87,12 +111,19 @@ public Response getFragments(@Context final ServletContext servletContext,
...
 
-        List<Fragment> fragments = fragmenter.getFragments();
-        fragments = AnalyzeUtils.getSampleFragments(fragments, context);
+        List<Fragment> fragments = fragmentCache.get(context.getTransactionId(),
+                new Callable<List<Fragment>>() {
+                    @Override
+                    public List<Fragment> call() throws Exception {
+                        LOG.debug("Caching fragments for transactionId={} from segmentId={}",
+                                context.getTransactionId(), context.getSegmentId());
+                        return AnalyzeUtils.getSampleFragments(fragmenter.getFragments(), context);
```
shivzone: We are we invoking getSampleFragments ? This was introduce for the purpose of stats with Analyze command

shivzone: AnalyzeUtils.getSampleFragments was introduced in PXF for ANALYZE command for the purpose of fetching stats. This functionality hasn't been implemented in gpdb. Invoke fragmenter.getFragments() directly.

server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/FragmenterResource.java
```diff
@@ -87,12 +111,19 @@ public Response getFragments(@Context final ServletContext servletContext,
         LOG.debug("FRAGMENTER started for path \"{}\"", path);
 
         RequestContext context = parseRequest(headers);
-
         /* Create a fragmenter instance with API level parameters */
         final Fragmenter fragmenter = fragmenterFactory.getPlugin(context);
 
-        List<Fragment> fragments = fragmenter.getFragments();
-        fragments = AnalyzeUtils.getSampleFragments(fragments, context);
+        List<Fragment> fragments = fragmentCache.get(context.getTransactionId(),
```
shivzone: To be safe, shouldn't we key in on the transactionId + dataSource ?

lavjain:
We need to make sure that the transaction id is unique for different slices when a parallel execution is going on. There was only one entry per segment id according to the heap dump.

server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/FragmenterResource.java
denalex: if multiple slices of the same query (think `select distinct a, b from table where a=1 union all select distinct a,b from table where a=2) use the same transaction id, then this will not work, since the first slice, using partition pruning, will fetch only fragments for a=1 and cache them and the second slice (a=2) will get it as a result. This will cause the whole slice of a=2 be missing from the result. This need be thought through in more details.

shivzone: We will need to key based on {XID,FILEPATH,FILTERSTRING}

frankgh:
I added a new flag to enable/disable Fragmenter Cache

server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/FragmenterResource.java
frankgh: We have some integration tests that rely on AnalyzeUtils:
```
HdfsAnalyzeTest.statsMaxFragmentsParameter...OK
HdfsAnalyzeTest.statsSampleRatioParameter...OK
```
Should we consider them disabling them?

server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/FragmenterResource.java
frankgh: New Key is based on `XID:FILEPATH:FILTERSTRING`

server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/FragmenterResource.java
shivzone: The analyze functionality tests have not been enabled for gpdb.
These 2 tests just test for couple of additional headers added to the fragmenter api to set limits for the #of fragments. This functionality is not exposed via gpdb though, so we can as well disable them

lavjain:
Why would someone need to turn off the fragment cache?

frankgh:
@lavjain we want to provide ways to turn off the cache in case there was some issue during the caching

server/pxf-api/src/main/java/org/greenplum/pxf/api/utilities/FragmenterCacheFactory.java
```diff
@@ -0,0 +1,44 @@
...
+
+    /**
+     * Singleton instance of the FragmenterCacheFactory
+     */
+    private static final FragmenterCacheFactory instance = new FragmenterCacheFactory();
+
+    private final Cache<String, List<Fragment>> fragmenterCache = CacheBuilder.newBuilder()
+            .expireAfterAccess(10, TimeUnit.SECONDS)
+            .removalListener((RemovalListener<String, List<Fragment>>) notification ->
+                    LOG.debug("Remove fragmenterCache entry for transactionId {}", notification.getKey()))
```
denalex: If this happens after the removal has occurred, maybe change to past tense "Removed.." and list how many fragments were cached.
```suggestion
                    LOG.debug("Removed fragmenterCache entry for transactionId {} with {} fragments", notification.getKey(), notification.getValue().size()))
```
Or even better yet to use extended API to check for the reason: https://google.github.io/guava/releases/23.0/api/docs/com/google/common/cache/RemovalNotification.html

Also, not clear from the refs above whether the values / keys can be nulls, in which case, finding size would need to be guarded by checking for nulls.

server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/FragmenterResource.java
```diff
@@ -139,4 +175,17 @@ public Response getFragmentsStats(@Context final ServletContext servletContext,
         return Response.ok(response, MediaType.APPLICATION_JSON_TYPE).build();
     }
 
+    private List<Fragment> getFragmentsFromContext(RequestContext context) throws Exception {
```
denalex: naming -- we are getting fragments not from the context, but from Fragmenter, otherwise a reader can assume that the context already has them stored somewhere, just call it `getFragments(RequestContext context)` or `loadFragments(RequestContext context)`

server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/FragmenterResource.java
```diff
@@ -139,4 +175,17 @@ public Response getFragmentsStats(@Context final ServletContext servletContext,
...
+        /* Create a fragmenter instance with API level parameters */
+        final Fragmenter fragmenter = fragmenterFactory.getPlugin(context);
+        return AnalyzeUtils.getSampleFragments(fragmenter.getFragments(), context);
```
denalex: can be 1-lined ?

server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/FragmenterResource.java
```diff
@@ -139,4 +175,17 @@ public Response getFragmentsStats(@Context final ServletContext servletContext,
...
+    }
+
+    private String getFragmenterCacheKey(RequestContext context) {
+        return String.format("%s:%s:%s",
+                context.getTransactionId(),
+                context.getDataSource(),
+                context.getFilterString());
```
denalex: might need to provide some comments why filter string is included in the key

server/pxf-service/src/test/java/org/greenplum/pxf/service/rest/FragmenterResourceTest.java
```diff
@@ -0,0 +1,335 @@
...
+    @Before
+    public void setup() {
+        parser = mock(RequestParser.class);
+        fragmenterFactory = mock(FragmenterFactory.class);
+        fragmenterCacheFactory = mock(FragmenterCacheFactory.class);
+        servletContext = mock(ServletContext.class);
+        headersFromRequest1 = mock(HttpHeaders.class);
+        headersFromRequest2 = mock(HttpHeaders.class);
+        fragmenter1 = mock(Fragmenter.class);
+        fragmenter2 = mock(Fragmenter.class);
```
denalex: if you use @RunWith(MockitoJUnit4Runner.class) then you can just add @Mock annotation to variable definitions and remove all the lines above.

server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/FragmenterResource.java
```diff
@@ -81,25 +96,46 @@ public FragmenterResource() {
...
+        if (Utilities.isFragmenterCacheEnabled()) {
+            try {
+                // We can't support lambdas here because asm version doesn't support it
+                fragments = fragmenterCacheFactory.getCache()
+                        .get(fragmenterCacheKey, new Callable<List<Fragment>>() {
+                            @Override
+                            public List<Fragment> call() throws Exception {
+                                LOG.debug("Caching fragments for transactionId={} from segmentId={} with key={}",
+                                        context.getTransactionId(), context.getSegmentId(), fragmenterCacheKey);
+                                return getFragmentsFromContext(context);
```
denalex: context doesn't need to be declared final in line 104 to be used here from the closure ?

frankgh: This is a java 8 change, prior to Java 8, we cannot use a non-final local variable in an anonymous class.

server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/FragmenterResource.java
frankgh: yes

PR #136, opened 2019-04-11
pxf-cli: Support sync and init on standby master

oliverralbertini:
This is a temporary solution until [this issue on `gp-common-go-libs` is
addressed](https://github.com/greenplum-db/gp-common-go-libs/issues/38).

This issue would allow us to add standby/mirrors to the hosts we
want to execute on. Instead, in this commit, we are using a direct query
to the database to find out standby master (`contentID`: -1, not primary).
With the standby, we add it to our list of segments early in execution,
but only if we are doing sync or init.

This change also includes adding the ability to add extra nodes in the
perf pipeline.

### DIFF
--- a/cli/go/src/pxf-cli/cmd/cluster.go
+++ b/cli/go/src/pxf-cli/cmd/cluster.go
@@ -14,10 +14,10 @@ import (
 )
 
 type ClusterData struct {
-	Cluster        *cluster.Cluster
-	Output         *cluster.RemoteOutput
-	NumHosts       int
-	connectionPool *dbconn.DBConn
+	Cluster    *cluster.Cluster
+	Output     *cluster.RemoteOutput
+	NumHosts   int
+	connection *dbconn.DBConn
 }
 
 var (
@@ -30,9 +30,10 @@ var (
 		Use:   "init",
 		Short: "Initialize the local PXF server instance",
 		Run: func(cmd *cobra.Command, args []string) {
-			clusterData, err := doSetup()
+			command := &pxf.InitCommand
+			clusterData, err := doSetup(command)
 			if err == nil {
-				err = clusterRun(&pxf.Init, clusterData)
+				err = clusterRun(command, clusterData)
 			}
 			exitWithReturnCode(err)
 		},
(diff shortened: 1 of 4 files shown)


lisakowen:
a few questions:
1.  what sort of error is displayed when the sync operation fails on a host?  can the user distinguish when a sync failed to the standby (probably not fatal) vs. when the sync failed to a segment host (affects pxf functioning)?
2.  can we claim HA support with this feature?  i.e. if master fails over to standby, will PXF continue to work?

lisakowen:
a few questions:
1.  what sort of error is displayed when the sync operation fails on a host?  can the user distinguish when a sync failed to the standby (probably not fatal) vs. when the sync failed to a segment host (affects pxf functioning)?
2.  can we claim HA support with this feature?  i.e. if master fails over to standby, will PXF continue to work?

oliverralbertini:
> 1. what sort of error is displayed when the sync operation fails on a host? can the user distinguish when a sync failed to the standby (probably not fatal) vs. when the sync failed to a segment host (affects pxf functioning)?

I created an artificial situation where the standby master host, named `edw0`, is known as `edw0-botched-username` by GPDB. Not a great example, but I couldn't think of an easier way to fake a failure.

In this situation, syncing to standby master will fail because the real hostname is `edw0`. As you can see, the error is prefixed by the name of the server as understood by GPDB, so the user should be able to figure out that it's just standby master, not a segment host.
``` 
[gpadmin@mdw ~]$ pxf cluster sync
Syncing PXF configuration files to 4 hosts...
ERROR: PXF configs failed to sync on 1 out of 4 hosts
edw0-botched-hostname ==> ssh: Could not resolve hostname edw0-botched-hostname: Name or service not known
rsync: connection unexpectedly closed (0 bytes received so far) [sender]...
```

> 2. can we claim HA support with this feature?  i.e. if master fails over to standby, will PXF continue to work?

I don't have enough context to know this, but that's a great question, which I hope we can answer!

oliverralbertini:
@lisakowen Since PXF doesn't run on master host anyway, we think that having the files synced to standby master is all we need in case of master failure.

PR #137, opened 2019-04-17
test_pxf: Run sshd directly w/out `service`

oliverralbertini:
CentOS 7 moved to systemd and so to have a solution that works for both
CentOS 6 and 7, it's best to just start sshd directly.

Also:
* Kill sshd when the test script exits to keep concourse from
hanging
* Allow gpadmin to run ping on CentOS 6/7 by giving ping the `setuid`
flag
* Stop providing TARGET_OS_VERSION default in the task file, since this
should come from outside the task (pipeline, etc.)

Co-authored-by: Ben Christel <bchristel@pivotal.io>
Co-authored-by: Sambitesh Dash <sdash@pivotal.io>
Co-authored-by: Soumyadeep Chakraborty <sochakraborty@pivotal.io>

### DIFF
--- a/concourse/scripts/pxf_common.bash
+++ b/concourse/scripts/pxf_common.bash
@@ -67,7 +67,8 @@ function install_gpdb_binary() {
 	fi
 
 	if [[ ${TARGET_OS} == "centos" ]]; then
-		service sshd start
+		# We can't use service sshd restart as service is not installed on CentOS 7.
+		/usr/sbin/sshd &
 		psi_dir=$(find /usr/lib64 -name psi | sort -r | head -1)
 	elif [[ ${TARGET_OS} == "ubuntu" ]]; then
         # Adjust GPHOME if the binary expects it to be /usr/local/gpdb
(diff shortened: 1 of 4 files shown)


oliverralbertini:
Please see https://github.com/greenplum-db/pxf/pull/109 for the previous attempt by members of the gp-releng team and further discussion.

concourse/scripts/pxf_common.bash
```diff
@@ -67,7 +67,8 @@ function install_gpdb_binary() {
 	fi
 
 	if [[ ${TARGET_OS} == "centos" ]]; then
-		service sshd start
+		# We can't use service sshd restart as service is not installed on CentOS 7.
+		/usr/sbin/sshd &
```
frankgh: will this work for both centos 6 and 7?

frankgh: can we do `systemctl restart sshd.service` for 7?

oliverralbertini: Yes, but then we would need some logic to detect whether we are on 6 or 7.

oliverralbertini: This solution works for both 6 & 7.

PR #141, opened 2019-04-22
pxf-api: Fix BaseConfigurationFactory logging

leskin-in:
Fix logging of server configuration file directory in `BaseConfigurationFactory.initConfiguration()`.

Three issues with the mentioned logging are fixed by this PR:
* `serverDirectoryName` does not include a path separator;
* Instead of logging actual server configuration file directory, server name is logged (while the former can differ from server name, as the comparison between them is case-insensitive);
* When server configuration file directory cannot be read by PXF (due to lack of permissions), a logged message states the directory does not exist.

No changes to unit tests are required.

### DIFF
--- a/server/pxf-api/src/main/java/org/greenplum/pxf/api/model/BaseConfigurationFactory.java
+++ b/server/pxf-api/src/main/java/org/greenplum/pxf/api/model/BaseConfigurationFactory.java
@@ -46,16 +46,20 @@ public class BaseConfigurationFactory implements ConfigurationFactory {
                         f.canRead() &&
                         StringUtils.equalsIgnoreCase(serverName, f.getName()));
 
-        String serverDirectoryName = serversConfigDirectory + serverName;
-
         if (serverDirectories == null || serverDirectories.length == 0) {
-            LOG.warn("Directory {} does not exist, no configuration resources are added for server {}", serverDirectoryName, serverName);
+            LOG.warn("Directory {} does not exist or cannot be read by PXF, no configuration resources are added for server {}",
+                serversConfigDirectory + File.separator + serverName,
+                serverName
+            );
         } else if (serverDirectories.length > 1) {
             throw new IllegalStateException(String.format(
-                    "Multiple directories found for server %s. Server directories are expected to be case-insensitive.", serverName));
+                    "Multiple directories found for server %s. Server directories are expected to be case-insensitive.", serverName
+            ));
         } else {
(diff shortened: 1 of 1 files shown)


server/pxf-api/src/main/java/org/greenplum/pxf/api/model/BaseConfigurationFactory.java
```diff
@@ -46,16 +46,20 @@ public Configuration initConfiguration(String serverName, Map<String, String> ad
                         f.canRead() &&
                         StringUtils.equalsIgnoreCase(serverName, f.getName()));
 
-        String serverDirectoryName = serversConfigDirectory + serverName;
-
         if (serverDirectories == null || serverDirectories.length == 0) {
-            LOG.warn("Directory {} does not exist, no configuration resources are added for server {}", serverDirectoryName, serverName);
+            LOG.warn("Directory {} does not exist or cannot be read by PXF, no configuration resources are added for server {}",
```
frankgh:
```suggestion
            LOG.warn("Directory {}{}{} does not exist or cannot be read by PXF, no configuration resources are added for server {}",
```

server/pxf-api/src/main/java/org/greenplum/pxf/api/model/BaseConfigurationFactory.java
```diff
@@ -46,16 +46,20 @@ public Configuration initConfiguration(String serverName, Map<String, String> ad
...
+                serversConfigDirectory + File.separator + serverName,
```
frankgh:
```suggestion
                serversConfigDirectory, File.separator, serverName,
```

server/pxf-api/src/main/java/org/greenplum/pxf/api/model/BaseConfigurationFactory.java
leskin-in: @frankgh ,
thank you for the comment. 

I wonder why this is a better way to deal with LOG. The concatenated path is a single entity, thus it seems logical to log it as a single entity.

frankgh: @leskin-in We have decided to move away from string concatenation as much as possible for LOG statements. String concatenation adds some cost, and if LOG level is lower than WARN, we will avoid doing any unnecessary work when using `org.slf4j.Logger`

Issue #130, opened 2019-04-03
Add isolation level to JDBC Server configuration file


pivotalguru:
Databases like DB2 and SQL Server use locking to provide read consistency so long running queries can block other sessions from updating a table.  To prevent this, many users of these databases will use dirty reads to prevent blocking locks.  This is achieved in JDBC by setting the transaction isolation level to read uncommitted.  

Example:
```
int transactionIsolationLevel = 0;

if (readCommitted) 
  transactionIsolationLevel = Connection.TRANSACTION_READ_COMMITTED;
else
  transactionIsolationLevel = Connection.TRANSACTION_READ_UNCOMMITTED;

conn.setTransactionIsolation(transactionIsolationLevel);
```
So add property to JDBC server configuration file so that the read isolation level can be set.  The default should be read committed.

frankgh:
This will be released in PXF version 5.3.0

PR #142, opened 2019-04-22
JDBC Transaction Isolation Level

oliverralbertini:
Adds ability for user to set the transaction isolation level, which would allow e.g. dirty reads if permitted by the external database.

### DIFF
--- a/server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
+++ b/server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
@@ -28,6 +28,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.sql.Connection;
+import java.sql.DatabaseMetaData;
 import java.sql.DriverManager;
 import java.sql.PreparedStatement;
 import java.sql.SQLException;
@@ -40,7 +41,7 @@ import java.util.stream.Collectors;
 
 /**
  * JDBC tables plugin (base class)
- *
+ * <p>
  * Implemented subclasses: {@link JdbcAccessor}, {@link JdbcResolver}.
  */
 public class JdbcBasePlugin extends BasePlugin {
@@ -57,11 +58,36 @@ public class JdbcBasePlugin extends BasePlugin {
     private static final String JDBC_SESSION_PROPERTY_PREFIX = "jdbc.session.property.";
     private static final String JDBC_CONNECTION_PROPERTY_PREFIX = "jdbc.connection.property.";
 
+    // connection parameter names
+    private static final String JDBC_CONNECTION_TRANSACTION_ISOLATION = "jdbc.connection.transactionIsolation";
+
     // DDL option names
     private static final String JDBC_DRIVER_OPTION_NAME = "JDBC_DRIVER";
     private static final String JDBC_URL_OPTION_NAME = "DB_URL";
 
     private static final String FORBIDDEN_SESSION_PROPERTY_CHARACTERS = ";\n\b\0";
+
+    private enum TransactionIsolation {
+        READ_UNCOMMITTED(1),
+        READ_COMMITTED(2),
+        REPEATABLE_READ(4),
+        SERIALIZABLE(8),
+        NOT_PROVIDED(-1);
+
+        private int isolationLevel;
+
+        TransactionIsolation(int transactionIsolation) {
+            isolationLevel = transactionIsolation;
+        }
+
+        public int getLevel() {
+            return isolationLevel;
+        }
+
+        public static TransactionIsolation typeOf(String str) {
+            return valueOf(str);
+        }
+    }
 
     // JDBC parameters from config file or specified in DDL
     private String jdbcUrl = null;
@@ -84,6 +110,9 @@ public class JdbcBasePlugin extends BasePlugin {
     // Properties object to pass to JDBC Driver when connection is created
     protected Properties connectionConfiguration = new Properties();
 
+    // Transaction isolation level that a user can configure
+    private TransactionIsolation transactionIsolation = TransactionIsolation.NOT_PROVIDED;
+
     // Columns description
     protected List<ColumnDescriptor> columns = null;
 
@@ -99,8 +128,7 @@ public class JdbcBasePlugin extends BasePlugin {
         try {
             LOG.debug("JDBC driver: '{}'", jdbcDriver);
             Class.forName(jdbcDriver);
-        }
-        catch (ClassNotFoundException e) {
+        } catch (ClassNotFoundException e) {
             throw new RuntimeException(e);
         }
 
@@ -140,27 +168,31 @@ public class JdbcBasePlugin extends BasePlugin {
         // However, SET queries cannot be executed this way. This is why we do this check
         if (sessionConfiguration.entrySet().stream()
                 .anyMatch(
-                    entry ->
-                        StringUtils.containsAny(
-                            entry.getKey(), FORBIDDEN_SESSION_PROPERTY_CHARACTERS
-                        ) ||
-                        StringUtils.containsAny(
-                            entry.getValue(), FORBIDDEN_SESSION_PROPERTY_CHARACTERS
-                        )
+                        entry ->
+                                StringUtils.containsAny(
+                                        entry.getKey(), FORBIDDEN_SESSION_PROPERTY_CHARACTERS
+                                ) ||
+                                        StringUtils.containsAny(
+                                                entry.getValue(), FORBIDDEN_SESSION_PROPERTY_CHARACTERS
+                                        )
                 )
         ) {
             throw new IllegalArgumentException("Some session configuration parameter contains forbidden characters");
         }
         if (LOG.isDebugEnabled()) {
             LOG.debug("Session configuration: {}",
-                sessionConfiguration.entrySet().stream()
-                        .map(entry -> "'" + entry.getKey() + "'='" + entry.getValue() + "'")
-                        .collect(Collectors.joining(", "))
+                    sessionConfiguration.entrySet().stream()
+                            .map(entry -> "'" + entry.getKey() + "'='" + entry.getValue() + "'")
+                            .collect(Collectors.joining(", "))
             );
         }
 
         // Optional parameter. The default value is empty map
         connectionConfiguration.putAll(configuration.getPropsWithPrefix(JDBC_CONNECTION_PROPERTY_PREFIX));
+
+        // Optional parameter. The default value depends on the database
+        String transactionIsolationString = configuration.get(JDBC_CONNECTION_TRANSACTION_ISOLATION, "NOT_PROVIDED");
+        transactionIsolation = TransactionIsolation.typeOf(transactionIsolationString);
 
         // Optional parameter. By default, corresponding connectionConfiguration property is not set
         String jdbcUser = configuration.get(JDBC_USER_PROPERTY_NAME);
@@ -170,11 +202,12 @@ public class JdbcBasePlugin extends BasePlugin {
 
         if (LOG.isDebugEnabled()) {
             LOG.debug("Connection configuration: {}",
-                connectionConfiguration.entrySet().stream()
-                        .map(entry -> "'" + entry.getKey() + "'='" + entry.getValue() + "'")
-                        .collect(Collectors.joining(", "))
+                    connectionConfiguration.entrySet().stream()
+                            .map(entry -> "'" + entry.getKey() + "'='" + entry.getValue() + "'")
+                            .collect(Collectors.joining(", "))
             );
         }
+
 
         // This must be the last parameter parsed, as we output connectionConfiguration earlier
         // Optional parameter. By default, corresponding connectionConfiguration property is not set
@@ -190,20 +223,16 @@ public class JdbcBasePlugin extends BasePlugin {
     /**
      * Open a new JDBC connection
      *
-     * @throws SQLException if a database access or connection error occurs
-     *
      * @return {@link Connection}
+     * @throws SQLException if a database access or connection error occurs
      */
     public Connection getConnection() throws SQLException {
-        LOG.debug("New JDBC connection. URL: '{}'; Table: '{}'",
-            jdbcUrl, tableName
-        );
+        LOG.debug("New JDBC connection. URL: '{}'; Table: '{}'", jdbcUrl, tableName);
 
         Connection connection = DriverManager.getConnection(jdbcUrl, connectionConfiguration);
         try {
             prepareConnection(connection);
-        }
-        catch (SQLException e) {
+        } catch (SQLException e) {
             closeConnection(connection);
             throw e;
         }
@@ -215,10 +244,8 @@ public class JdbcBasePlugin extends BasePlugin {
      * Prepare a JDBC PreparedStatement
      *
      * @param connection connection to use for creating the statement
-     * @param query query to execute
-     *
+     * @param query      query to execute
      * @return PreparedStatement
-     *
      * @throws SQLException if a database access error occurs
      */
     public PreparedStatement getPreparedStatement(Connection connection, String query) throws SQLException {
@@ -232,7 +259,6 @@ public class JdbcBasePlugin extends BasePlugin {
      * Close a JDBC statement and underlying {@link Connection}
      *
      * @param statement statement to close
-     *
      * @throws SQLException
      */
     public static void closeStatementAndConnection(Statement statement) throws SQLException {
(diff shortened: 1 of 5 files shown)


server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
```diff
@@ -190,20 +197,16 @@ public void initialize(RequestContext context) {
...
-            jdbcUrl, tableName
-        );
+        LOG.debug("New JDBC connection. URL: '{}'; Table: '{}'", jdbcUrl, tableName);
 
         Connection connection = DriverManager.getConnection(jdbcUrl, connectionConfiguration);
         try {
             prepareConnection(connection);
-        }
-        catch (SQLException e) {
+        } catch (Exception e) {
```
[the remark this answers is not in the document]
oliverralbertini: Paging @denalex. 

I believe we needed to open this up to `Exception` because now `prepareConnection()` can also throw another type of exception.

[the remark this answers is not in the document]
frankgh: Yeah, I was wondering about this, `prepareConnection` seems to be throwing a SQLException only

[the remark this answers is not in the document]
denalex: I thought we modified it to throw RuntimeException if the user has provided value of Transaction Isolation that is not supported by the database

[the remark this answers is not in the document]
oliverralbertini: Yeah, but I changed that to be a specific type of exception declared in `JdbcBasePlugin` class. This made it easier to do testing. Here's the new [exception](https://github.com/greenplum-db/pxf/blob/f55c20fc1906ad768d47efb9413c69978f1fdcfc/server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java#L318).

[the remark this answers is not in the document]
oliverralbertini: And here's the [declaration](https://github.com/greenplum-db/pxf/blob/f55c20fc1906ad768d47efb9413c69978f1fdcfc/server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java#L341).

frankgh:
This PR is related to this issue: https://github.com/greenplum-db/pxf/issues/130

frankgh:
Can we move the log4j to it's own PR or commit outside of this PR?

server/pxf-jdbc/src/test/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePluginTest.java
```diff
@@ -272,12 +279,127 @@ public void testCloseConnectionWithExceptionStatementOnGetConnection() throws Ex
...
+        additionalProps.put("jdbc.url", "test-url");
+        additionalProps.put("jdbc.connection.transactionIsolation", "foobarValue");
+        context.setAdditionalConfigProps(additionalProps);
+
+        Connection mockConnection = mock(Connection.class);
+        DatabaseMetaData fakeMetaData = mock(DatabaseMetaData.class);
+        when(mockConnection.getMetaData()).thenReturn(fakeMetaData);
+        when(DriverManager.getConnection(anyString(), anyObject())).thenReturn(mockConnection);
+
+
```
leskin-in: Probably one blank line here?

server/pxf-jdbc/src/test/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePluginTest.java
```diff
@@ -272,12 +279,127 @@ public void testCloseConnectionWithExceptionStatementOnGetConnection() throws Ex
...
+        additionalProps.put("jdbc.url", "test-url");
+        additionalProps.put("jdbc.connection.transactionIsolation", "999");
+        context.setAdditionalConfigProps(additionalProps);
+
+        Connection mockConnection = mock(Connection.class);
+        DatabaseMetaData fakeMetaData = mock(DatabaseMetaData.class);
+        when(mockConnection.getMetaData()).thenReturn(fakeMetaData);
+        when(DriverManager.getConnection(anyString(), anyObject())).thenReturn(mockConnection);
+
+
```
leskin-in: Probably one blank line here?

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
```diff
@@ -140,28 +147,31 @@ public void initialize(RequestContext context) {
...
+                            .map(entry -> "'" + entry.getKey() + "'='" + entry.getValue() + "'")
+                            .collect(Collectors.joining(", "))
             );
         }
 
         // Optional parameter. The default value is empty map
         connectionConfiguration.putAll(configuration.getPropsWithPrefix(JDBC_CONNECTION_PROPERTY_PREFIX));
 
+        // Optional parameter. The default value depends on the database
+        connectionTransactionIsolationLevel = configuration.getInt(JDBC_CONNECTION_TRANSACTION_ISOLATION, ISOLATION_LEVEL_NOT_PROVIDED);
```
leskin-in: I think we could introduce string parsing (e.g. like [here](https://github.com/greenplum-db/pxf/blob/483753b7c3137ff652d90cd515b4447f4cc3da15/server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcPartitionFragmenter.java#L294-L302)) here:
* [Java platform API specification](https://docs.oracle.com/javase/8/docs/api/java/sql/Connection.html#setTransactionIsolation-int-) states that only four constants are supported by `setTransactionIsolation`. I believe we should refer those constants instead of an integer (which itself has no meaning, but is a convention);
* Passing string (like `READ_COMMITTED`) is more user-friendly than passing an integer, IMHO.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
```diff
@@ -84,6 +89,9 @@
     // Properties object to pass to JDBC Driver when connection is created
     protected Properties connectionConfiguration = new Properties();
 
+    // Transaction isolation level that a user can configure
+    protected int connectionTransactionIsolationLevel = ISOLATION_LEVEL_NOT_PROVIDED;
```
shivzone: Shouldn't this be private since the base class itself handles the connection object

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
```diff
@@ -285,15 +285,19 @@ private static void closeConnection(Connection connection) throws SQLException {
...
             }
-        }
-        finally {
-            connection.close();
+        } finally {
+            try {
+                connection.close();
+            } catch (Exception e) {
+                // ignore
+                LOG.warn("Failed to close JDBC connection, ignoring the error.");
```
shivzone: Why warn instead of error ?

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
denalex: we were actually going back-and-forth on this. Ideally JDBC should've introduced this enum themselves. Otherwise we would need to:
1. introduce that enum, as you were suggesting
2. have users type long string values, which is error-prone
3. handle the case where entered string does not have corresponding enum value

We felt that using the numbers and documenting what the numbers mean in the template is less error-prone and gives better usability. We also consider this functionality to be exercised very rarely.

@leskin-in -- what do you think ? can we trust users handling long strings vs numbers ?

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
denalex: previously, if the operation is done and committed successfully, but there was an error during `connection.close()`, that error would result in an exception and aborting of GPDB transaction. We changed it here so that a failure to close a connection will be just a warning (we ignore the Exception).

@leskin-in -- any concerns with this change of behavior ? 
@oliverralbertini -- we should probably log an exception message if not the whole stack trace.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
```diff
@@ -307,29 +311,46 @@ private void prepareConnection(Connection connection) throws SQLException {
...
-                    statement.execute(
-                        dbProduct.buildSessionQuery(e.getKey(), e.getValue())
-                    );
+                    statement.execute(dbProduct.buildSessionQuery(e.getKey(), e.getValue()));
                 }
             }
         }
     }
 
+    class TransactionLevelNotSupportedException extends RuntimeException {
```
denalex: usually there is no value in creating strongly types RuntimeExceptions unless you are catching them specifically somewhere in the logic, which I do not see in this PR. I suggest we remove it altogether and use regular RuntimeException or IllegalStateException

server/pxf-jdbc/src/test/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePluginTest.java
```diff
@@ -272,12 +279,127 @@ public void testCloseConnectionWithExceptionStatementOnGetConnection() throws Ex
...
+        JdbcBasePlugin plugin = new JdbcBasePlugin();
+        plugin.initialize(context);
+        plugin.getConnection();
+    }
+
+    @Test(expected = TransactionLevelNotSupportedException.class)
```
denalex: when you get rid of that `TransactionLevelNotSupportedException`, you can use `@Rule` way of checking exception type and message, which is now recommended anyways.

server/pxf-jdbc/src/test/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePluginTest.java
```diff
@@ -272,12 +279,127 @@ public void testCloseConnectionWithExceptionStatementOnGetConnection() throws Ex
...
+        context.setDataSource("test-table");
+        Map<String, String> additionalProps = new HashMap<>();
+        additionalProps.put("jdbc.driver", "org.greenplum.pxf.plugins.jdbc.FakeJdbcDriver");
+        additionalProps.put("jdbc.url", "test-url");
+        additionalProps.put("jdbc.connection.transactionIsolation", "999");
+        context.setAdditionalConfigProps(additionalProps);
+
+        Connection mockConnection = mock(Connection.class);
+        DatabaseMetaData fakeMetaData = mock(DatabaseMetaData.class);
+        when(mockConnection.getMetaData()).thenReturn(fakeMetaData);
```
denalex: don't we need to train `fakeMetaData` to return `false` on `isSupported` call ?

server/pxf-jdbc/src/test/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePluginTest.java
```diff
@@ -272,12 +279,127 @@ public void testCloseConnectionWithExceptionStatementOnGetConnection() throws Ex
...
+
+        Connection mockConnection = mock(Connection.class);
+        doThrow(new SQLException("")).when(mockConnection).getMetaData();
+        when(DriverManager.getConnection(anyString(), anyObject())).thenReturn(mockConnection);
+
+        JdbcBasePlugin plugin = new JdbcBasePlugin();
+        plugin.initialize(context);
+        Connection conn = plugin.getConnection();
+
+        verify(conn).setTransactionIsolation(2);
```
denalex: hmm ? So do we expect an exception if we can't get metadata or do we expect to set the level no matter what ? If the former, than this line can be removed, as we will never get here, it's confusing.

server/pxf-service/src/templates/user/templates/jdbc-site.xml
```diff
@@ -20,6 +20,18 @@
...
     </property>
+    <!-- Uncomment to set transaction isolation level
+        Set to one of these integers if supported by external database:
+        int TRANSACTION_READ_UNCOMMITTED = 1;
+        int TRANSACTION_READ_COMMITTED   = 2;
+        int TRANSACTION_REPEATABLE_READ  = 4;
+        int TRANSACTION_SERIALIZABLE     = 8;
+    <property>
+        <name>jdbc.connection.transactionIsolation</name>
+        <value>some_integer_1</value>
```
denalex:
```suggestion
        <value>YOUR_CHOSEN_TRANSACTION_ISOLATION_CODE_FROM_ABOVE</value>
```

server/pxf-service/src/templates/user/templates/jdbc-site.xml
```diff
@@ -20,6 +20,18 @@
...
+        <description>Connection-level variable to set in external database which might allow for dirty reads</description>
```
denalex:
```suggestion
        <description>JDBC connection transaction isolation level. Make sure your database supports the level you define here.</description>
```

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
oliverralbertini: Switched to using an enum.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
frankgh: I think we have been using IllegalStateException in cases where values are not valid. We should try to stick to IllegalStateException for consistency

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
frankgh: My preference would be to not introduce a new enum just for this purpose. The feature does not seem to justify it. I would prefer to stick with integers.

server/pxf-service/src/templates/user/templates/jdbc-site.xml
oliverralbertini: @denalex : Please take a look at the updated format, since we switched from ints to strings.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
leskin-in: As far as I understand, error on `close()` would not lead to an exception being propagated to GPDB.  [This code](https://github.com/greenplum-db/pxf/blob/148676a7f6441bf10c0b6e3e7aae1e6ae43ef04f/server/pxf-service/src/main/java/org/greenplum/pxf/service/bridge/ReadBridge.java#L139-L146) logs the exception (and I suppose, if we want to change log level, we should do it there), and [this](https://github.com/greenplum-db/pxf/blob/148676a7f6441bf10c0b6e3e7aae1e6ae43ef04f/server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/BridgeResource.java#L152-L156) prevents exceptions from propagation to GDPB.

That said, I think we should either left this unchanged or modify the log level in `BridgeResource`.

server/pxf-jdbc/src/test/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePluginTest.java
oliverralbertini: Not needed, but good for clarity.

oliverralbertini:
> Can we move the log4j changes to it's own PR or commit outside of this PR?

I removed it from this branch, we can add it in another PR.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
oliverralbertini: @leskin-in thanks for the input. In this case the concern is not so much that an exception bubbles out, but that an `SQLException` that we care about gets masked by an exception from closing the connection.

@denalex @shivzone I am switching to logging the entire stack trace here.

PR #144, opened 2019-04-23
Remove support for Logical operator NOT with Hive Partition Filtering

shivzone:
Removed log. op. NOT from hive partition filter
Logical operators are now upper case with hive partition filtering string
!= is now <>

Predicates with NOT logical operator eg: NOT(a='foo') are not supported with hive partition filtering. 
https://cwiki.apache.org/confluence/display/Hive/Partition+Filter+Syntax
Will need a follow up PR to handle this case. One possibility we could reverse the underlying predicate in order to avoid NOT. eg: NOT(a='foo') -> (a!='foo)

### DIFF
--- a/server/pxf-hive/src/main/java/org/greenplum/pxf/plugins/hive/HiveFilterBuilder.java
+++ b/server/pxf-hive/src/main/java/org/greenplum/pxf/plugins/hive/HiveFilterBuilder.java
@@ -167,7 +167,7 @@ public class HiveFilterBuilder implements FilterParser.FilterBuilder {
                 logicalOperator = " or ";
                 break;
             case HDOP_NOT:
-                logicalOperator = " not ";
+                logicalOperator = "not ";
                 break;
             default:
                 logicalOperator = "";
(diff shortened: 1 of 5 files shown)


server/pxf-hive/src/main/java/org/greenplum/pxf/plugins/hive/HiveFilterBuilder.java
```diff
@@ -183,10 +180,7 @@ private String buildCompoundFilter(LogicalFilter filter) throws Exception {
             }
             if (serializedFilter != null) {
                 if (filter.getOperator() == FilterParser.LogicalOperation.HDOP_NOT) {
-                    filterString
-                            .append("NOT(")
-                            .append(serializedFilter)
-                            .append(")");
+                    return null;
```
denalex: we can move this check way up and not go into filter builder / recursion if we see NOT operator

shivzone: good point, i've added this check earlier up now

PR #146, opened 2019-04-24
A fragmenter that returns a list of files instead of splits

frankgh:
A fragmenter that returns a list of files instead of splits


shivzone:
I guess this is intended to be a WIP PR as opposed to a PR

shivzone:
I don't think we need a separate fragmenter for this as this would bloat the # of profiles. If we go with a user defined property, we can add this functionality in existing fragmenters which can support this

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HdfsDataFragmenter.java
```diff
@@ -65,18 +65,19 @@ public void initialize(RequestContext context) {
     public List<Fragment> getFragments() throws Exception {
         Path path = new Path(hcfsType.getDataUri(configuration, context));
         List<InputSplit> splits = getSplits(path);
+        String[] hosts = {"foo"};
```
shivzone: We should just set it to null. FragmentsResponseFormatters.convertHostsToIPs handles null hosts appropriately

frankgh:
Closing in favor of https://github.com/greenplum-db/pxf/pull/158

PR #147, opened 2019-04-25
Reverse direction of rsync in pxf sync command

oliverralbertini:
`pxf sync` is documented as being run from master, i.e. `pxf sync
remote` should sync $PXF_CONF/{conf,lib,servers} from master to remote.
Instead, it syncs from remote to local. This is a vestige of the
original implementation of `pxf cluster sync`, which has now been
changed.

Authored-by: Oliver Albertini <oalbertini@pivotal.io>

PR #148, opened 2019-04-25
Distinguish master and standby master when installing PXF

oliverralbertini:

### DIFF
--- a/concourse/scripts/install_pxf.bash
+++ b/concourse/scripts/install_pxf.bash
@@ -4,7 +4,7 @@ set -euxo pipefail
 
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 GPHOME="/usr/local/greenplum-db-devel"
-MASTER_HOSTNAME=$( < cluster_env_files/etc_hostfile grep "mdw.*" | awk '{print $2}')
+MASTER_HOSTNAME=$(grep -E '\bmdw.*' cluster_env_files/etc_hostfile | awk '{print $2}')
 HADOOP_HOSTNAME="ccp-$(cat terraform_dataproc/name)-m"
 PXF_CONF_DIR="/home/gpadmin/pxf"
 


oliverralbertini:
This should only needs to be merged in when https://github.com/pivotal/gp-concourse-cluster-provisioner/pull/63 is merged.

The above PR will allow us to use CCP to standup an env with standby master called `smdw`.

frankgh:
@oliverralbertini this seems to be part of this PR as well: https://github.com/greenplum-db/pxf/pull/147

oliverralbertini:
Yeah, it is needed to get the standby master successfully on the GPDB cluster for the pipeline. It won't be merged as part of the other PR.

oliverralbertini:
OK, it's time to merge this, since I have merged the PR for CCP: https://github.com/pivotal/gp-concourse-cluster-provisioner/pull/63

concourse/scripts/install_pxf.bash
```diff
@@ -4,7 +4,7 @@ set -euxo pipefail
 
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 GPHOME="/usr/local/greenplum-db-devel"
-MASTER_HOSTNAME=$( < cluster_env_files/etc_hostfile grep "mdw.*" | awk '{print $2}')
+MASTER_HOSTNAME=$(grep -E '\bmdw.*' cluster_env_files/etc_hostfile | awk '{print $2}')
```
frankgh: what is the hostname of the standby master?

frankgh:
are there any other scripts that use the master hostname in a similar way?

concourse/scripts/install_pxf.bash
oliverralbertini: standby master is called `smdw` (and master is just `mdw`). So without this fix, the variable `MASTER_HOSTNAME` gets set to something like:
```
'mdw
smdw'
```

which includes the newline. With the word boundary in the regex, it makes sure we _only_ get `mdw`.

oliverralbertini:
> are there any other scripts that use the master hostname in a similar way?

I will check on that.

PR #150, opened 2019-04-26
Add debug statements for the JDBC connection

frankgh:
Log JDBC connection lifecycle to make sure connections are being properly closed.

Co-authored-by: Alexander Denissov (adenissov@pivotal.io)

### DIFF
--- a/server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
+++ b/server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
@@ -227,9 +227,12 @@ public class JdbcBasePlugin extends BasePlugin {
      * @throws SQLException if a database access or connection error occurs
      */
     public Connection getConnection() throws SQLException {
-        LOG.debug("New JDBC connection. URL: '{}'; Table: '{}'", jdbcUrl, tableName);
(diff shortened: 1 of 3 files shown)


server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
```diff
@@ -277,6 +281,7 @@ public static void closeStatementAndConnection(Statement statement) throws SQLEx
         }
 
         try {
+            LOG.debug("Closing statement for connection {}", connection);
             statement.close();
         } catch (SQLException e) {
             LOG.error("Exception when closing Statement", e);
```
shivzone: didn't we want to bring this down to WARN level log

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
```diff
@@ -286,7 +291,7 @@ public static void closeStatementAndConnection(Statement statement) throws SQLEx
         try {
             closeConnection(connection);
         } catch (SQLException e) {
-            LOG.error("Exception when closing Connection", e);
+            LOG.error(String.format("Exception when closing connection %s", connection), e);
```
shivzone: same here

PR #153, opened 2019-04-28
Minor code cleanup

frankgh:
A small code cleanup

### DIFF
--- a/server/pxf-hive/src/main/java/org/greenplum/pxf/plugins/hive/HiveORCAccessor.java
+++ b/server/pxf-hive/src/main/java/org/greenplum/pxf/plugins/hive/HiveORCAccessor.java
@@ -50,11 +50,11 @@ public class HiveORCAccessor extends HiveAccessor implements StatsAccessor {
 
     private static final Log LOG = LogFactory.getLog(HiveORCAccessor.class);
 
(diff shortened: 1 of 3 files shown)


server/pxf-hive/src/main/java/org/greenplum/pxf/plugins/hive/HiveORCAccessor.java
```diff
@@ -50,11 +50,11 @@
 
     private static final Log LOG = LogFactory.getLog(HiveORCAccessor.class);
 
-    private final String READ_COLUMN_IDS_CONF_STR = "hive.io.file.readcolumn.ids";
-    private final String READ_ALL_COLUMNS = "hive.io.file.read.all.columns";
-    private final String READ_COLUMN_NAMES_CONF_STR = "hive.io.file.readcolumn.names";
-    private final String SARG_PUSHDOWN = "sarg.pushdown";
-    protected Reader orcReader;
+    private static final String READ_COLUMN_IDS_CONF_STR = "hive.io.file.readcolumn.ids";
```
frankgh: making this static, because we saw the heap dump had hundreds of HiveORCAccessors, each one had created these String constants that used up memory

frankgh:
I will be merging this if there are no objections

PR #155, opened 2019-04-30
WIP Add support to read multiline files as a single row

frankgh:
Introduce support for reading multiline files as a single row in GPBD using PXF. Each file represents a single row in the database. You can create a external table like this:

```
CREATE EXTERNAL TABLE ext_json (
  json_email_msg json
)
LOCATION('pxf://some-location/file.json?PROFILE=gs:text&SERVER=google&FILE_AS_ROW=true')
FORMAT 'CSV';
```
And you can use GPDB's JSON to query data like this:

```
    SELECT  
       ext_json_blob->>'someKey' AS key,
       ext_json_blob->>'someText' AS someText,
       ext_json_blob->>'name' AS name,
       ext_json_blob->>'lastname' AS lastname,
       ext_json_blob->>'address' AS address,
       ext_json_blob->>'phone' as phone
   FROM ext_json;

```

### DIFF
--- a/server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HdfsSplittableDataAccessor.java
+++ b/server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HdfsSplittableDataAccessor.java
@@ -164,18 +164,11 @@ public abstract class HdfsSplittableDataAccessor extends BasePlugin implements A
             }
         }
 
-        if (data instanceof Text) {
-            if (fileAsRow) {
-                // When fileAsRow is true, we need to return a quoted
-                // string, and we need to escape quotes for a valid
-                // multiline CSV to be interpreted by GPDB
-                escapeCSV((Text) data);
-            }
-
-            if (!fileAsRow || reader.getProgress() != 1.0F) {
-                // Add a new line to the end of the row
-                ((Text) data).set(data.toString() + "\n");
-            }
+        if (fileAsRow && data instanceof Text) {
+            // When fileAsRow is true, we need to return a quoted
+            // string, and we need to escape quotes for a valid
+            // multiline CSV to be interpreted by GPDB
+            escapeCSV((Text) data);
         }
 
         /*
--- a/server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HdfsSplittableDataAccessor.java
+++ b/server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HdfsSplittableDataAccessor.java
@@ -20,6 +20,8 @@ package org.greenplum.pxf.plugins.hdfs;
  */
 
 
+import org.apache.commons.lang.StringUtils;
+import org.apache.hadoop.io.Text;
 import org.apache.hadoop.mapred.FileSplit;
 import org.apache.hadoop.mapred.InputFormat;
 import org.apache.hadoop.mapred.InputSplit;
@@ -47,7 +49,9 @@ public abstract class HdfsSplittableDataAccessor extends BasePlugin implements A
     protected InputFormat<?, ?> inputFormat;
     protected JobConf jobConf;
     protected Object key, data;
-    protected HcfsType hcfsType;
+    protected boolean isFileBlob;
+    private boolean firstLine, lastLine;
+    HcfsType hcfsType;
 
     private ListIterator<InputSplit> iter;
 
@@ -64,6 +68,9 @@ public abstract class HdfsSplittableDataAccessor extends BasePlugin implements A
     public void initialize(RequestContext requestContext) {
         super.initialize(requestContext);
 
+        // true if the files are read as a single blob, false otherwise
+        isFileBlob = StringUtils.equalsIgnoreCase("true", context.getOption("BLOB"));
+
         // variable required for the splits iteration logic
         jobConf = new JobConf(configuration, HdfsSplittableDataAccessor.class);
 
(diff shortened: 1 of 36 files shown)


server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HdfsSplittableDataAccessor.java
```diff
@@ -64,6 +68,9 @@ protected HdfsSplittableDataAccessor(InputFormat<?, ?> inFormat) {
     public void initialize(RequestContext requestContext) {
         super.initialize(requestContext);
 
+        // true if the files are read as a single blob, false otherwise
+        isFileBlob = StringUtils.equalsIgnoreCase("true", context.getOption("BLOB"));
```
denalex: not sure `BLOB` is the best option name we can come up with, in ORCL BLOB==BinaryLargeOBject. Maybe `FILE_AS_ROW` ?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HdfsSplittableDataAccessor.java
```diff
@@ -128,21 +136,34 @@ protected boolean getNextSplit() throws IOException {
...
+            if (!hasMoreSplits ||
+                    !reader.next(key, data)) { // read the first record of the new split
+                lastLine = true;
+                // for file blobs return a quote, otherwise return null
+                return isFileBlob ? new OneRow(null, "\"") : null;
             }
         }
 
+        if (isFileBlob && data instanceof Text) {
+            data = new Text(data.toString().replaceAll("\"", "\"\""));
```
denalex: can we avoid creating a new Text object here, instead try to operate on the data within by either `data.setValue(data.getValue().transform())` or by defining a method on `Text` object ?

denalex: would be nice to explain why we are turning single quotes into double quotes here.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HdfsSplittableDataAccessor.java
```diff
@@ -128,21 +136,34 @@ protected boolean getNextSplit() throws IOException {
...
+        if (isFileBlob && firstLine) {
+            firstLine = false;
+            return new OneRow(null, "\"");
+        }
+
+        if (isFileBlob && lastLine) {
+            // return null after the last line
+            return null;
+        }
+
```
denalex: this is only applicable to blobs, so maybe roll into 1 top-level if:
```suggestion
if (isFileBlob) {
    if (firstLine) {
            firstLine = false;
            return new OneRow(null, "\"");
    }
    if (lastLine) {
            // return null after the last line
            return null;
    }
}
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HdfsSplittableDataAccessor.java
frankgh: agreed. BLOB was just a placeholder for the moment. Thanks for the suggestion

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HdfsSplittableDataAccessor.java
frankgh: makes sense

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HdfsSplittableDataAccessor.java
frankgh: This is `org.apache.hadoop.io.Text` object. Not sure what would be the best approach to add a method, maybe a wrapper class?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HdfsSplittableDataAccessor.java
frankgh: done

server/pxf-service/src/main/java/org/greenplum/pxf/service/BridgeOutputBuilder.java
```diff
@@ -281,7 +281,7 @@ void fillText(List<OneField> recFields) throws BadRecordException {
             }
         } else { // from QuotedLineBreakAccessor
             String textRec = (String) val;
-            output = new Text(textRec + "\n");
+            output = new Text(textRec);
```
frankgh: I decided to remove the line break from the BridgeOutputBuilder, because the Bridge should not be concerned about adding line breaks, or having to deal with line breaks

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HdfsSplittableDataAccessor.java
```diff
@@ -152,6 +180,14 @@ public OneRow readNextObject() throws IOException {
         return new OneRow(key, data);
     }
 
+    /**
+     * Returns a escaped CSV Text
+     * @param data the original data
+     */
+    private void escapeCSV(Text data) {
```
shivzone: Can we instead use StringEscapeUtils.escapeCsv

shivzone:
The concern i have is the splits that we can get from hdfs come with specific offsets. Will there be a scenario where a file is broken into two different splits and which might mess up this logic

frankgh:
Closing in favor of https://github.com/greenplum-db/pxf/pull/158

PR #156, opened 2019-05-02
Set JDBC statement properties including fetch size

shivzone:
1. Additional statement properties for jdbc profile
- fetch_size for read (visible in jdbc-site and DDL)
- query_timeout for read and write (not visible in jdbc-site.xml as it doesn't work with older versions of jdbc driver. eg: postgres 9.0+ mysql 5.0+ etc)
2. batch_size for write can be configured in jdbc-site as well


server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
```diff
@@ -47,8 +47,10 @@
 public class JdbcBasePlugin extends BasePlugin {
 
     // '100' is a recommended value: https://docs.oracle.com/cd/E11882_01/java.112/e16548/oraperf.htm#JJDBC28754
-    private static final int DEFAULT_BATCH_SIZE = 100;
+    private static final int DEFAULT_WRITE_SIZE = 100;
```
denalex: I don't know if we want to rename `batch` to `write` everywhere as batch is pretty well-known JDBC concept for updates

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
```diff
@@ -61,6 +63,11 @@
     // connection parameter names
     private static final String JDBC_CONNECTION_TRANSACTION_ISOLATION = "jdbc.connection.transactionIsolation";
 
+    // statement properties
+    private static final String JDBC_STATEMENT_WRITE_SIZE = "jdbc.statement.writeSize";
+    private static final String JDBC_STATEMENT_FETCH_SIZE = "jdbc.statement.fetchSize";
+    private static final String JDBC_STATEMENT_QUERY_TIMEOUT = "jdbc.statement.queryTimeout";
```
denalex: please add `_PROPERTY_NAME` to the end of all these names to be consistent

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
```diff
@@ -47,8 +47,10 @@
 public class JdbcBasePlugin extends BasePlugin {
 
     // '100' is a recommended value: https://docs.oracle.com/cd/E11882_01/java.112/e16548/oraperf.htm#JJDBC28754
-    private static final int DEFAULT_BATCH_SIZE = 100;
+    private static final int DEFAULT_WRITE_SIZE = 100;
+    private static final int DEFAULT_FETCH_SIZE = 1000;
```
denalex: didn't we want to set it to 10000 ?

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
```diff
@@ -145,16 +159,20 @@ public void initialize(RequestContext context) {
...
-        batchSizeIsSetByUser = context.getOption("BATCH_SIZE") != null;
-        batchSize = context.getOption("BATCH_SIZE", DEFAULT_BATCH_SIZE, true);
-        if (batchSize == 0) {
-            batchSize = 1; // if user set to 0, it is the same as batchsize of 1
+        // Optional parameters
+        writeSizeIsSetByUser = configuration.get(JDBC_STATEMENT_WRITE_SIZE) != null;
+        writeSize = configuration.getInt(JDBC_STATEMENT_WRITE_SIZE, DEFAULT_WRITE_SIZE);
+
+        if (writeSize <= 0) {
+            writeSize = 1; // if user set to 0, it is the same as writeSize of 1
```
denalex: we used to error out on negative values, will they now be allowed ?

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
```diff
@@ -145,16 +159,20 @@ public void initialize(RequestContext context) {
...
         }
 
-        // Optional parameter. The default value is 1
+        fetchSize = configuration.getInt(JDBC_STATEMENT_FETCH_SIZE, DEFAULT_FETCH_SIZE);
```
denalex: what is user sets negative or 0 ? should we validate, error out ?

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
```diff
@@ -145,16 +159,20 @@ public void initialize(RequestContext context) {
...
+
         poolSize = context.getOption("POOL_SIZE", DEFAULT_POOL_SIZE);
 
+        queryTimeout = configuration.getInt(JDBC_STATEMENT_QUERY_TIMEOUT, DEFAULT_QUERY_TIMEOUT);
```
denalex: what is user sets negative? should we validate, error out ?

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
frankgh: yeah, we talked about 10000

shivzone: I thought about 10k, but then when i noticed we are very conservative on writes at 100, i wanted to reduce it to 1000 for reads. Moreover with oracle the default is 10.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
shivzone: my concern was tomorrow if another developer picks this up it might be misleading as it doesn't quite convey if its for reads or writes. If we think it is coherent using batch can be refactored

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
shivzone: I wanted to leave standard statement properties handling to the driver implementation

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
shivzone: Will revert this change

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
shivzone: will be handled by the driver

denalex:
Please also update `README.md` in `pxf-jdbc` module with the info about the new feature

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
shivzone: i've reverted it to batch size

PR #157, opened 2019-05-03
JDBC profile can execute a query to read data from an external DB

denalex:
This PR implements a feature where a user can define a query in a SQL file under sever's configuration directory. This query will be run by the JDBC plugin and the data returned by the query will be sent to GPDB.
 
The following example shows how to execute an aggregation query in MySQL and return results via PXF JDBC plugin.

The following tables are created in MySQL:
```
use demodb;
create table dept(
    id int(4) not null primary key,
    name varchar(20) not null
);

create table emp(
    dept_id int(4) not null,
    name varchar(20) not null,
    salary int(8)
);
```
Then some data is inserted into MySQL tables:
```
insert into dept values(1, 'sales');
insert into dept values(2, 'finance');
insert into dept values(3, 'it');

insert into emp values(1, 'alice', 11000);
insert into emp values(2, 'bob', 10000);
insert into emp values(3, 'charlie', 10500);
```
Then a complex aggregation query is created and placed in a file, say report.sql. The file needs to be placed in the server configuration directory under `$PXF_CONF/servers/`. So, let's assume we have created a `mydb` server configuration directory, then this file will be `$PXF_CONF/servers/mydb/report.sql`. Jdbc driver name and connection parameters should be configured in `$PXF_CONF/servers/mydb/jdbc-site.xml` for this server.
```
SELECT dept.name AS name, count(*) AS count, max(emp.salary) AS max
FROM demodb.dept JOIN demodb.emp
ON dept.id = emp.dept_id 
GROUP BY dept.name;
```
This query returns a name of the department, count of employees, and the maximal salary in each department.

The MySQL JDBC driver files (JAR) are copied to `$PXF_CONF/lib` on all hosts with PXF. After this, all PXF segments are restarted.

Then a table in GPDB is created with the schema corresponding to the results returned by the aggregation query. It is important that the table column names and types correspond to those returned by the aggregation query.
```
CREATE EXTERNAL TABLE dept_report (
    name text,
    count int,
    max int
)
LOCATION (
    'pxf://query:report?PROFILE=JDBC&SERVER=mydb'
)
FORMAT 'CUSTOM' (
    FORMATTER='pxfwritable_import'
);
```
Finally, a query to a GPDB external table is made:
```
SELECT * FROM dept_report;
SELECT name, count FROM dept_report WHERE max > 10000;
```

### DIFF
--- a/automation/src/test/java/org/greenplum/pxf/automation/features/jdbc/JdbcTest.java
+++ b/automation/src/test/java/org/greenplum/pxf/automation/features/jdbc/JdbcTest.java
@@ -363,6 +363,22 @@ public class JdbcTest extends BaseFeature {
         pxfJdbcNamedQuery.setHost(pxfHost);
         pxfJdbcNamedQuery.setPort(pxfPort);
         gpdb.createTableAndVerify(pxfJdbcNamedQuery);
+
+        pxfJdbcNamedQuery = TableFactory.getPxfJdbcReadablePartitionedTable(
+                "pxf_jdbc_read_named_query_partitioned",
+                NAMED_QUERY_FIELDS,
+                "query:report",
+                null,
+                null,
+                1,
+                "1:5",
+                "1",
+                null,
+                EnumPartitionType.INT,
+                "database");
+        pxfJdbcNamedQuery.setHost(pxfHost);
+        pxfJdbcNamedQuery.setPort(pxfPort);
+        gpdb.createTableAndVerify(pxfJdbcNamedQuery);
     }
 
     @Test(groups = {"features", "gpdb"})
--- a/automation/src/test/java/org/greenplum/pxf/automation/features/jdbc/JdbcTest.java
+++ b/automation/src/test/java/org/greenplum/pxf/automation/features/jdbc/JdbcTest.java
@@ -50,6 +50,11 @@ public class JdbcTest extends BaseFeature {
             "\"does_not_exist_on_source\" text",
             "\"num 1\" int",
             "\"n@m2\" int"};
+    private static final String[] NAMED_QUERY_FIELDS = new String[]{
+            "name  text",
+            "count int",
+            "max  int"};
+
     private ExternalTable pxfJdbcSingleFragment;
     private ExternalTable pxfJdbcMultipleFragmentsByInt;
     private ExternalTable pxfJdbcMultipleFragmentsByDate;
@@ -62,11 +67,13 @@ public class JdbcTest extends BaseFeature {
     private ExternalTable pxfJdbcColumns;
     private ExternalTable pxfJdbcColumnProjectionSubset;
     private ExternalTable pxfJdbcColumnProjectionSuperset;
+    private ExternalTable pxfJdbcNamedQuery;
 
     private static final String gpdbTypesDataFileName = "gpdb_types.txt";
     private static final String gpdbColumnsDataFileName = "gpdb_columns.txt";
     private Table gpdbNativeTableTypes, gpdbNativeTableColumns, gpdbWritableTargetTable;
     private Table gpdbWritableTargetTableNoBatch, gpdbWritableTargetTablePool;
+    private Table gpdbDeptTable, gpdbEmpTable;
 
     @Override
     protected void beforeClass() throws Exception {
@@ -86,6 +93,7 @@ public class JdbcTest extends BaseFeature {
         prepareColumnProjectionSubsetInDifferentOrder();
         prepareColumnProjectionSuperset();
         prepareFetchSizeZero();
+        prepareNamedQuery();
     }
 
     private void prepareTypesData() throws Exception {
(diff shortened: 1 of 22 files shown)


server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/SQLQueryBuilder.java
```diff
@@ -206,13 +236,7 @@ private void buildWhereSQL(StringBuilder query) throws ParseException {
...
 
         try {
-            StringBuilder prepared = new StringBuilder();
-            if (!query.toString().contains("WHERE")) {
-                prepared.append(" WHERE ");
-            }
-            else {
-                prepared.append(" AND ");
-            }
+            StringBuilder prepared = new StringBuilder(" WHERE ");
```
denalex: @leskin-in -- I had to remove this logic as the query can now contain WHERE clause from the internal query, so `contains()` check will no longer work. Was there any reason it was there in the first place ? What would've been a case with the original code where we would already find WHERE statement once we just start building filters part ?

leskin-in: The change is correct. 

This code was useful when `buildWhereSQL()` was called directly from `JdbcAccessor`. It allowed reordering of calls to `Fragmenter.buildFragmenterSQL()` and `buildWhereSQL()`.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcAccessor.java
```diff
@@ -49,6 +53,22 @@
...
+
+    private static final Logger LOG = LoggerFactory.getLogger(JdbcAccessor.class);
+
+    // Read variables
+    private String queryRead = null;
+    private Statement statementRead = null;
+    private ResultSet resultSetRead = null;
+
+    // Write variables
+    private String queryWrite = null;
```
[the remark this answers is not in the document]
denalex: yes, will remove it after master rebase, thanks

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcAccessor.java
```diff
@@ -270,19 +294,34 @@ public void closeForWrite() throws Exception {
...
+        // read the contents of the file holding the text of the query with a given name
+        String serverDirectory = configuration.get(ConfigurationFactory.PXF_CONFIG_SERVER_DIRECTORY_PROPERTY);
+        if (StringUtils.isBlank(serverDirectory)) {
+            throw new IllegalStateException("No server configuration directory found for server " + context.getServerName());
+        }
+
+        String queryText;
+        try {
+            File queryFile = new File(serverDirectory, queryName + ".sql");
+            if (LOG.isDebugEnabled()) {
```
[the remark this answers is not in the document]
denalex: yes, getting the path involves some calculations we want to avoid if debugging is off.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcAccessor.java
```diff
@@ -270,19 +294,34 @@ public void closeForWrite() throws Exception {
...
+                LOG.debug("Reading text of query={} from {}", queryName, queryFile.getCanonicalPath());
+            }
+            queryText = FileUtils.readFileToString(queryFile);
+        } catch (IOException e) {
+            throw new RuntimeException(String.format("Failed to read text of query %s : %s", queryName, e.getMessage()), e);
+        }
+        return queryText;
```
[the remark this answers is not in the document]
denalex: sure we can add the logic for empty file. IF we do, then we will throw an exception that will be shown to users. I am not sure integration test will be of a lot of value as having empty file would be rather unusual. We do not have any other validation of the SQL provided (including checking for absence of semi-colons, presence of SELECT word, etc).

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcAccessor.java
```diff
@@ -270,19 +294,34 @@ public void closeForWrite() throws Exception {
...
-    // Write variables
-    private String queryWrite = null;
-    private PreparedStatement statementWrite = null;
-    private WriterCallableFactory writerCallableFactory = null;
-    private WriterCallable writerCallable = null;
-    private ExecutorService executorServiceWrite = null;
-    private List<Future<SQLException> > poolTasks = null;
+    /**
+     * Gets the text of the query by reading the file from the server configuration directory. The name of the file
+     * es expected to be the same as the name of the query provided by the user and have extension ".sql"
```
oliverralbertini: es -> is

automation/src/test/java/org/greenplum/pxf/automation/features/jdbc/JdbcTest.java
```diff
@@ -116,6 +124,37 @@ private void prepareTypesData() throws Exception {
...
+        String[][] deptRows = new String[][] {
+                { "sales", "1"},
+                { "finance", "2"},
+                { "it", "3"}};
+        Table dataTable = new Table("data", deptTableFields);
+        dataTable.addRows(deptRows);
+        gpdb.insertData(dataTable, gpdbDeptTable);
+
+        String[] empTableFields = new String[]{"name text", "dept_id int", "salary int"};
+        gpdbEmpTable = new Table("gpdb_emp", empTableFields);
```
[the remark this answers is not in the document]
denalex: it will have to be named "query:..." as you not only need to have colon, but have the path start with "query:". Extreme edge case that would not work in current implementation. If it ever occurs we would need to introduce an override option in user options, but doing it now seems premature.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcAccessor.java
denalex: that's a valid Spanish no ? :) thanks for noticing

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcAccessor.java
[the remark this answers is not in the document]
denalex: added logic and unit test case

server/pxf-jdbc/src/test/java/org/greenplum/pxf/plugins/jdbc/JdbcAccessorTest.java
```diff
@@ -0,0 +1,184 @@
...
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+import org.junit.runner.RunWith;
+import org.mockito.ArgumentCaptor;
+import org.powermock.api.mockito.PowerMockito;
+import org.powermock.core.classloader.annotations.PrepareForTest;
+import org.powermock.modules.junit4.PowerMockRunner;
+
+import java.io.File;
+import java.sql.*;
```
frankgh: single-line imports are preferred

denalex: tried to change it, but IntelliJ automatically collapses imports to this if there are 5+ imports from the same package, will leave it like this.

denalex: @oliverralbertini suggested to use Alt+ENTER on that line which expanded the imports, thanks !

PR #158, opened 2019-05-06
Add support to read multiline files as a single row

frankgh:
Introduce support for reading multiline files as a single row in GPBD using PXF. Each file represents a single row in the database. You can create a external table like this:

    CREATE EXTERNAL TABLE ext_json (
      json_email_msg json
    )
    LOCATION('pxf://some-location/file.json?PROFILE=gs:text:multi&SERVER=google&FILE_AS_ROW=true')
    FORMAT 'CSV';

And you can use GPDB's JSON to query data like this:

        SELECT  
           ext_json_blob->>'someKey' AS key,
           ext_json_blob->>'someText' AS someText,
           ext_json_blob->>'name' AS name,
           ext_json_blob->>'lastname' AS lastname,
           ext_json_blob->>'address' AS address,
           ext_json_blob->>'phone' as phone
       FROM ext_json;


automation/src/test/java/org/greenplum/pxf/automation/features/hcfs/FileAsRowTest.java
```diff
@@ -0,0 +1,94 @@
...
+    private static final String emptyTextFile = "empty";
+    private static final String twoLineTextFile = "twoline";
+    private static final String singleLineTextFile = "singleline";
+    private static final String multiLineTextFile = "multiline";
+    private static final String multiLineJsonFile = "tweets-pp.json";
+
+    private static final String[] PXF_MULTILINE_COLS = {"text_blob text"};
+
+    @Override
+    protected void beforeClass() throws Exception {
```
denalex: usually you just need to annotate with @Before and the framework would call super-class' before method and then this one.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HdfsSplittableDataAccessor.java
```diff
@@ -131,14 +131,10 @@ public OneRow readNextObject() throws IOException {
...
-                if (!reader.next(key, data)) {
-                    // make sure we return nulls
-                    return null;
-                }
-            } else {
-                // make sure we return nulls
+            boolean hasMoreSplits = getNextSplit();
+
+            if (!hasMoreSplits ||
+                    !reader.next(key, data)) { // read the first record of the new split
```
denalex: minor: I particularly find this shifted formatting confusing, prefer them to be on one line

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/QuotedLineBreakAccessor.java
```diff
@@ -20,56 +20,110 @@
...
+
         return new OneRow(null, next_line);
     }
 
+    /**
+     * Read one line ahead, to determine when the last line occurs
+     *
+     * @return the next line
+     */
+    String readLine() throws IOException {
```
denalex: since all this queue business is only needed for "file_as_row" feature, should we pass a param here that will do it conditionally ?

automation/tincrepo/main/pxf/features/hcfs/file_as_row/json/expected/query01.ans
```diff
@@ -0,0 +1,106 @@
...
+-[ RECORD 1 ]-----------------------------------------------------------------------------------------------------------------------------------------
+json_blob | {
+          |   "root":[
+          |     {
+          |       "record":{
+          |         "created_at":"Fri Jun 07 22:45:02 +0000 2013",
+          |         "id":343136547115253761,
+          |         "id_str":"343136547115253761",
+          |         "text":"text1",
+          |         "source":"<a href=\"http:\/\/twitter.com\/download\/iphone\" rel=\"nofollow\">Twitter for iPhone<\/a>",
```
denalex: will this look like a valid link with forward slash "/" being escaped ? I think only back slashes need escaping: https://www.freeformatter.com/json-escape.html

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/QuotedLineBreakAccessor.java
```diff
@@ -20,56 +20,110 @@
...
+        String next_line = readLine();
         if (next_line == null) /* EOF */ {
             return null;
         }
 
+        if (fileAsRow) {
+            // Wrap text around quotes, and escape single quotes
+            next_line = (firstLine ? CSV_QUOTE : StringUtils.EMPTY) +
+                    next_line.replace(CSV_QUOTE, CSV_ESCAPED_QUOTE) +
+                    (lastLine ? CSV_QUOTE : StringUtils.EMPTY);
```
denalex: I think this is the essence of this PR -- we are wrapping all file in quotes and escape internal quotes, correct ? Who/how will unwrap them ?

Also, this whole trick will only work for external tables with 1 column, right ? What will happen if there are more than 1 column in the definition ? Should we provide validation (ideally on C-side, but at least on Java side) checking how many columns are defined if this option is set ?

denalex:
Does the format needs to be CSV in the external table definition ? That would be weird for JSON files, can it be 'TEXT' ?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/QuotedLineBreakAccessor.java
frankgh: @denalex this is correct. The unwrapping happens at the GPDB level. GPDB will interpret it as a multiline CSV record. 

For your second point, you are correct. I will add validation on the java side to check that only a single column is provided

automation/tincrepo/main/pxf/features/hcfs/file_as_row/json/expected/query01.ans
frankgh: well, the backslashes are escaped in the original file: https://github.com/greenplum-db/pxf/blob/master/server/pxf-json/src/test/resources/tweets-pp.json#L9 
So, we just preserve the escaping from the original file

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/QuotedLineBreakAccessor.java
shivzone: i don't think we need extra validation for columns. We don't do column validation for other profiles right ?

shivzone:
> Does the format needs to be CSV in the external table definition ? That would be weird for JSON files, can it be 'TEXT' ?

frankgh:
@denalex unfortunately I don't think it's possible to use TEXT. I tried doing it, but I am seeing the following error: `quote available only in CSV mode`. We use GPDB's Copy functionality in the backend, which only supports multiline rows when using the CSV format as described here: https://gpdb.docs.pivotal.io/5170/ref_guide/sql_commands/CREATE_EXTERNAL_TABLE.html

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/QuotedLineBreakAccessor.java
frankgh: @shivzone, it makes sense to validate that we have a single column for this feature, otherwise, there might be some unexpected behavior on the GPDB side

denalex:
where exactly in that link is specifies this (what to search for ?) . Can we truly serve CSV in this case, let's say we want 2 columns : (file_name, content) and point the URL to a directory containing bunch of files, can we read them all as rows into the table ? Can we do it now with 1 column (we will not know which file it comes from) ? Or do we have to limit table DDL to 1 external file only ?

frankgh:
@denalex ` [QUOTE [AS] 'quote'] ` is what we are looking for in the link. If you point to a directory, every file is going to be a different row. (Maybe we should add an integration test to validate that functionality). I'm not sure if we should support multicolumn, not sure exactly how that would work. The DDL can take a path, and the path can contain multiple files, each file would represent a row.

denalex:
yes, no need to do multi-column now, I was just wondering.

Also I wonder if we declare some weird character , say `0x00` or `0x01` as a quote character and use it as a wrapper, then we wouldn't need to escape regular quotes in the text, which might give us speed up and less memory consumption for JSON files. Users will have to specify it as such in DDL as GPDB will need to be aware of it. In fact, we maybe will be able to use StringPassResolver in this case, just adding weird chars in the beginning and at the end ?

In the future we can pass this char in the headers by parsing table DDL in C, for now we can use hardcoded one ? What do you think ?

frankgh:
For now, like you mentioned, we don't pass any of the `FORMAT` parameters from the C-side to Java side. We should probably think about how we will integrate this into FDW, maybe something to consider for the longer term.

Whenever we think about C-side changes, I would also like to start thinking about how these changes will integrate to FDW.

frankgh:
@denalex added a test for reading multiple files from a directory

PR #160, opened 2019-05-08
JDBC: Optimize resolver for INSERT queries

leskin-in:
Optimize `JdbcResolver`:
* Use `java.util.EnumSet` instead of switch-case sequence to define whether the provided `OneRow` field type is supported;
* Stop using `ThreadLocal` objects holding `SimpleDateFormat`, move the latter to instances of `JdbcResolver` instead;
* Remove type incompatibility warning log call (called when the provided field is of type TEXT, but table DDL states the field has other type, so `JdbcResolver` tries to make type conversion anyway).

All changes deal only with INSERT queries logic.

### Perfomance improvement
This patch was tested in the test case described below.

The perfomance improvement is as follows:

|  | [Before patch](https://github.com/greenplum-db/pxf/tree/9c28afa07a8e953dc5d54a13cfd4a7829e53f9e9) | [Before patch](https://github.com/greenplum-db/pxf/tree/9c28afa07a8e953dc5d54a13cfd4a7829e53f9e9) with log level `ERROR` | After patch |
| --- | --- | --- | --- |
| Execution time | 64±1 s | 29±0.5 s | 5.1±0.1 s |

#### Test case
For the test, the following table **DDL** was used:
```
CREATE WRITABLE EXTERNAL TABLE ch_t1_wpart (a INT, b INT, c TIMESTAMP)
LOCATION ('pxf://r1.t1?profile=jdbc&SERVER=ch&BATCH_SIZE=500000&POOL_SIZE=2')
FORMAT 'custom' (formatter='pxfwritable_export');
```
To our best knowledge, such DDL is the most effective one for our test case.

The dataset consisted of **five million rows** with all fields set to random values.

GPDB cluster consisted of **four segments** on two hosts. **Two PXF instances** were used (one on each GPDB host).

The host CPUs were `Intel(R) Xeon(R) CPU E5-4650 v2 @ 2.40GHz` (4 cores).

[Clickhouse](https://clickhouse.yandex) was used as an external database. It was run on two separate hosts, each connected to GPDB segment hosts with (reliable) 100Mbps connection.

Clickhouse table DDL (executed on each of two hosts):
```
create table r1.t1 (a UInt32 codec(ZSTD(1)), b UInt32 codec(ZSTD(1)), c DateTime codec(ZSTD(1))) 
engine ReplicatedMergeTree('/clickhouse/tables/s1/t1', 'r1') 
order by (a);
```

### DIFF
--- a/server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
+++ b/server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
@@ -24,6 +24,8 @@ import org.greenplum.pxf.api.OneRow;
 import org.greenplum.pxf.api.io.DataType;
 import org.greenplum.pxf.api.model.Resolver;
 import org.greenplum.pxf.api.utilities.ColumnDescriptor;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.math.BigDecimal;
@@ -49,6 +51,33 @@ import java.util.Set;
  * JDBC tables resolver
  */
 public class JdbcResolver extends JdbcBasePlugin implements Resolver {
+    private static final Set<DataType> DATATYPES_SUPPORTED = EnumSet.of(
+            DataType.VARCHAR,
+            DataType.BPCHAR,
+            DataType.TEXT,
+            DataType.BYTEA,
+            DataType.BOOLEAN,
+            DataType.INTEGER,
+            DataType.FLOAT8,
+            DataType.REAL,
+            DataType.BIGINT,
+            DataType.SMALLINT,
+            DataType.NUMERIC,
+            DataType.TIMESTAMP,
+            DataType.DATE
+    );
+
+    /// DateTimeFormatter to parse timestamp
+    private static final DateTimeFormatter TIMESTAMP_FORMATTER =
+            new DateTimeFormatterBuilder().appendPattern("yyyy-MM-dd HH:mm:ss")
+                    // Parsing nanos in strict mode, the number of parsed digits must be between 0 and 6 (millisecond support)
+                    .appendFraction(ChronoField.NANO_OF_SECOND, 0, 6, true).toFormatter();
+
+    /// SimpleDateFormat to parse date
+    private final SimpleDateFormat DATE_FORMAT = new SimpleDateFormat("yyyy-MM-dd");
+
+    private static final Logger LOG = LoggerFactory.getLogger(JdbcResolver.class);
+
     /**
      * getFields() implementation
      *
@@ -133,10 +162,10 @@ public class JdbcResolver extends JdbcBasePlugin implements Resolver {
      */
     @Override
     public OneRow setFields(List<OneField> record) throws UnsupportedOperationException, ParseException {
-        int column_index = 0;
+        int columnIndex = 0;
 
         for (OneField oneField : record) {
-            ColumnDescriptor column = columns.get(column_index++);
+            ColumnDescriptor column = columns.get(columnIndex++);
 
             DataType oneFieldType = DataType.get(oneField.type);
             DataType columnType = DataType.get(column.columnTypeCode());
@@ -145,6 +174,11 @@ public class JdbcResolver extends JdbcBasePlugin implements Resolver {
                 throw new UnsupportedOperationException(
                         String.format("Field type '%s' (column '%s') is not supported",
                                 oneFieldType, column));
+            }
+
+            if (LOG.isDebugEnabled() && DataType.get(oneField.type) == DataType.BYTEA) {
+                String converted = (oneField.val != null) ? new String((byte[]) oneField.val) : "null";
+                LOG.debug("OneField content (conversion from BYTEA): '{}'", converted);
             }
 
             // Convert TEXT columns into native data types
@@ -187,7 +221,7 @@ public class JdbcResolver extends JdbcBasePlugin implements Resolver {
                         oneField.val = new Timestamp(LocalDateTime.parse(rawVal, TIMESTAMP_FORMATTER).toInstant(ZoneOffset.UTC).toEpochMilli());
                         break;
                     case DATE:
-                        oneField.val = new Date(dateSDF.parse(rawVal).getTime());
+                        oneField.val = new Date(DATE_FORMAT.parse(rawVal).getTime());
                         break;
                     default:
                         throw new UnsupportedOperationException(
@@ -299,28 +333,4 @@ public class JdbcResolver extends JdbcBasePlugin implements Resolver {
             }
         }
     }
-
-    private static final Set<DataType> DATATYPES_SUPPORTED = EnumSet.of(
-            DataType.VARCHAR,
-            DataType.BPCHAR,
-            DataType.TEXT,
-            DataType.BYTEA,
-            DataType.BOOLEAN,
-            DataType.INTEGER,
-            DataType.FLOAT8,
-            DataType.REAL,
-            DataType.BIGINT,
-            DataType.SMALLINT,
-            DataType.NUMERIC,
-            DataType.TIMESTAMP,
-            DataType.DATE
-    );
-
-    private static final DateTimeFormatter TIMESTAMP_FORMATTER =
-            new DateTimeFormatterBuilder().appendPattern("yyyy-MM-dd HH:mm:ss")
-                    // Parsing nanos in strict mode, the number of parsed digits must be between 0 and 6 (millisecond support)
-                    .appendFraction(ChronoField.NANO_OF_SECOND, 0, 6, true).toFormatter();
-
-    // SimpleDateFormat objects to parse TEXT into DATE
-    private final SimpleDateFormat dateSDF = new SimpleDateFormat("yyyy-MM-dd");
 }
--- a/server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
+++ b/server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
@@ -36,12 +36,6 @@ import java.sql.SQLException;
 import java.sql.Timestamp;
 import java.sql.Types;
 import java.text.ParseException;
-import java.text.SimpleDateFormat;
-import java.time.LocalDateTime;
-import java.time.ZoneOffset;
-import java.time.format.DateTimeFormatter;
-import java.time.format.DateTimeFormatterBuilder;
-import java.time.temporal.ChronoField;
 import java.util.EnumSet;
 import java.util.LinkedList;
 import java.util.List;
@@ -66,15 +60,6 @@ public class JdbcResolver extends JdbcBasePlugin implements Resolver {
             DataType.TIMESTAMP,
             DataType.DATE
     );
-
-    /// DateTimeFormatter to parse timestamp
-    private static final DateTimeFormatter TIMESTAMP_FORMATTER =
-            new DateTimeFormatterBuilder().appendPattern("yyyy-MM-dd HH:mm:ss")
-                    // Parsing nanos in strict mode, the number of parsed digits must be between 0 and 6 (millisecond support)
-                    .appendFraction(ChronoField.NANO_OF_SECOND, 0, 6, true).toFormatter();
-
-    /// SimpleDateFormat to parse date
-    private final SimpleDateFormat DATE_FORMAT = new SimpleDateFormat("yyyy-MM-dd");
 
     private static final Logger LOG = LoggerFactory.getLogger(JdbcResolver.class);
 
@@ -177,20 +162,18 @@ public class JdbcResolver extends JdbcBasePlugin implements Resolver {
             }
 
             if (LOG.isDebugEnabled()) {
+                String valDebug;
                 if (oneField.val == null) {
-                    LOG.debug("Column {} OneField: type {}, content null", columnIndex, oneFieldType);
+                    valDebug = "null";
+                }
+                else if (oneFieldType == DataType.BYTEA) {
+                    valDebug = String.format("'{}'", new String((byte[]) oneField.val));
                 }
                 else {
-                    String converted;
-                    switch(oneFieldType) {
-                        case BYTEA:
-                            converted = (oneField.val != null) ? new String((byte[]) oneField.val) : "null";
-                            break;
-                        default:
-                            converted = (oneField.val != null) ? oneField.val.toString() : "null";
-                    }
-                    LOG.debug("Column {} OneField: type {}, content '{}'", columnIndex, oneFieldType, converted);
+                    valDebug = String.format("'{}'", oneField.val.toString());
                 }
+
+                LOG.debug("Column {} OneField: type {}, content {}", columnIndex, oneFieldType, valDebug);
             }
 
             // Convert TEXT columns into native data types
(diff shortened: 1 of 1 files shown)


server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -335,19 +306,27 @@ public static void decodeOneRowToPreparedStatement(OneRow row, PreparedStatement
...
-            new SimpleDateFormat("yyyy-MM-dd HH:mm:ss.SSSSS"),
-            new SimpleDateFormat("yyyy-MM-dd HH:mm:ss.SSSS"),
-            new SimpleDateFormat("yyyy-MM-dd HH:mm:ss.SSS"),
-            new SimpleDateFormat("yyyy-MM-dd HH:mm:ss.SS"),
-            new SimpleDateFormat("yyyy-MM-dd HH:mm:ss.S"),
-            new SimpleDateFormat("yyyy-MM-dd HH:mm:ss"),
-            new SimpleDateFormat("yyyy-MM-dd")
-    });
+    // SimpleDateFormat objects to parse TEXT into DATE
+    private final SimpleDateFormat dateSDF = new SimpleDateFormat("yyyy-MM-dd");
```
frankgh: You can use the following instead, not sure if it will impact performance though. Maybe try it?
```suggestion
    private static final String DATE_FORMATTER_BASE_PATTERN = "yyyy-MM-dd HH:mm:ss";
    static final DateTimeFormatter DATE_FORMATTER =
            new DateTimeFormatterBuilder().appendPattern(DATE_FORMATTER_BASE_PATTERN)
                    // Parsing nanos in strict mode, the number of parsed digits must be between 0 and 6 (millisecond support)
                    .appendFraction(ChronoField.NANO_OF_SECOND, 0, 6, true).toFormatter();
```

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -130,109 +129,81 @@
...
-                default:
-                    throw new UnsupportedOperationException(
-                            String.format("Field type '%s' (column '%s') is not supported",
-                                    DataType.get(oneField.type),
-                                    column));
-            }
+            DataType oneFieldType = DataType.get(oneField.type);
+            DataType columnType = DataType.get(column.columnTypeCode());
 
-            if (LOG.isDebugEnabled() &&
```
frankgh: Can we preserve logging?

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -130,109 +129,81 @@
     @Override
     public OneRow setFields(List<OneField> record) throws UnsupportedOperationException, ParseException {
         int column_index = 0;
```
frankgh: Rename to columnIndex?
```suggestion
        int columnIndex = 0;
```

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
leskin-in: Thank you. I tested the perfomance of the fix; no changes were noted.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -130,109 +129,81 @@
     @Override
     public OneRow setFields(List<OneField> record) throws UnsupportedOperationException, ParseException {
         int column_index = 0;
+
         for (OneField oneField : record) {
-            ColumnDescriptor column = columns.get(column_index);
-            if (LOG.isWarnEnabled() && DataType.get(column.columnTypeCode()) != DataType.get(oneField.type)) {
```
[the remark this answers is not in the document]
leskin-in: This log was introduced because there is no explicit guarantee that tuples will not be disordered in PXF or GPDB docs. However, GPDB always sends tuples ordered the same way as external table DDL.

On the other hand, some data types are represented as TEXT when received from GPDB (e.g. timestamp), thus they cause this log call for normal tuples.

To sum up, this log call is made during normal tuple processing and is of no use.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
leskin-in: For sure. The logging makes sense here.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -129,110 +162,76 @@
...
-                                            DataType.get(oneField.type),
-                                            column));
-                    }
+                if (oneField.val == null) {
+                    continue;
                 }
-            }
 
-            column_index += 1;
+                String rawVal = (String) oneField.val;
```
frankgh: this log might still be useful for debugging purposes.
```suggestion
                String rawVal = (String) oneField.val;
                LOG.debug("OneField content (conversion from TEXT): '{}'", rawVal);
```

leskin-in: Thanks for this comment. I have changed the logging logic so that it includes output for fields of all types. This increases the size of log file, but can be disabled without rebuilding PXF.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -129,110 +162,88 @@
...
-                                    // pass
-                                }
-                            }
-                            if (!isConversionSuccessful) {
-                                throw new ParseException(rawVal, 0);
-                            }
-                            break;
-                        case DATE:
-                            oneField.val = new Date(dateSDF.get().parse(rawVal).getTime());
+                            converted = (oneField.val != null) ? new String((byte[]) oneField.val) : "null";
```
denalex: this is inside else, so we know oneField.val != null here and can drop extra check.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -129,110 +162,88 @@
...
                             break;
                         default:
-                            throw new UnsupportedOperationException(
-                                    String.format("Field type '%s' (column '%s') is not supported",
-                                            DataType.get(oneField.type),
-                                            column));
+                            converted = (oneField.val != null) ? oneField.val.toString() : "null";
```
denalex: same here

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -129,110 +162,88 @@
...
-                        case VARCHAR:
-                        case BPCHAR:
-                        case TEXT:
+            if (LOG.isDebugEnabled()) {
+                if (oneField.val == null) {
+                    LOG.debug("Column {} OneField: type {}, content null", columnIndex, oneFieldType);
+                }
+                else {
+                    String converted;
+                    switch(oneFieldType) {
```
denalex: switch with 1 check and default can be better handled by if / else statement.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -37,13 +37,46 @@
...
+    );
+
+    /// DateTimeFormatter to parse timestamp
+    private static final DateTimeFormatter TIMESTAMP_FORMATTER =
+            new DateTimeFormatterBuilder().appendPattern("yyyy-MM-dd HH:mm:ss")
+                    // Parsing nanos in strict mode, the number of parsed digits must be between 0 and 6 (millisecond support)
+                    .appendFraction(ChronoField.NANO_OF_SECOND, 0, 6, true).toFormatter();
+
+    /// SimpleDateFormat to parse date
+    private final SimpleDateFormat DATE_FORMAT = new SimpleDateFormat("yyyy-MM-dd");
```
denalex: why not use the new-style thread safe formatter for dates as well ?

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -129,110 +162,88 @@
...
+                        oneField.val = Long.parseLong(rawVal);
+                        break;
+                    case SMALLINT:
+                        oneField.val = Short.parseShort(rawVal);
+                        break;
+                    case NUMERIC:
+                        oneField.val = new BigDecimal(rawVal);
+                        break;
+                    case TIMESTAMP:
+                        oneField.val = new Timestamp(LocalDateTime.parse(rawVal, TIMESTAMP_FORMATTER).toInstant(ZoneOffset.UTC).toEpochMilli());
```
denalex: we are potentially losing microsecond precision here, is it OK ? i know it is similar as before the change, but since we are looking at this piece of code now, it is the right time to ask the question.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -129,110 +162,88 @@
...
+                        break;
+                    case DATE:
+                        oneField.val = new Date(DATE_FORMAT.parse(rawVal).getTime());
```
frankgh: From LocalDate.parse it will parse text such as "2007-12-03"
```suggestion
                        oneField.val = Date.valueOf(LocalDate.parse(rawVal));
```

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
frankgh: true, Timestamp only provides millisecond precision

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
frankgh: not needed if you use `LocalDate.parse`

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
denalex: well, Timestamp seems to store nanos separately from seconds and there's a way to get nanos from it using `getNanos()`: https://courses.cs.washington.edu/courses/cse341/98au/java/jdk1.2beta4/docs/api/java/sql/Timestamp.html
So it will depend on JDBC driver on how it pulls nanos out of this datatype, our obligation is to provide as precise info as possible, no ?

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
denalex: @leskin-in -- great performance improvement ! Coming down from 30s to 5s is a factor of 6 !! What was the bottleneck -- threadlocal or the old formatter or both ?

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
denalex: what is not needed, the formatter altogether ? @frankgh -- do you have a proposed code change ?

frankgh: I do, I added my comment below
`oneField.val = Date.valueOf(LocalDate.parse(rawVal));`

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
denalex: wouldn't just java.sql.Date.valueOf(rawVal) work ? https://docs.oracle.com/javase/8/docs/api/java/sql/Date.html#valueOf-java.lang.String-

frankgh: it looks like it would :)

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -129,110 +162,88 @@
...
-                if (oneField.val != null) {
-                    String rawVal = (String) oneField.val;
-
-                    LOG.debug("OneField content (conversion from TEXT): '{}'", rawVal);
-                    switch (DataType.get(column.columnTypeCode())) {
-                        case VARCHAR:
-                        case BPCHAR:
-                        case TEXT:
+            if (LOG.isDebugEnabled()) {
+                if (oneField.val == null) {
```
frankgh: Maybe something like this:
```suggestion
                String converted;
                if (oneField.val == null) {
                    converted = "null";
                } else if (oneFieldType == DataType.BYTEA) {
                    converted = "'" + new String((byte[]) oneField.val) + "'";
                } else {
                    converted = "'" + oneField.val.toString() + "'";
                }
                LOG.debug("Column {} OneField: type {}, content {}", columnIndex, oneFieldType, converted);
```

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
leskin-in: Both. The old parsing logic could also throw up to six exceptions (and it did in case of our dataset) that should have been processed.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
leskin-in: The [`Timestamp` constructor accepting year, month, etc. separately](https://docs.oracle.com/javase/8/docs/api/java/sql/Timestamp.html#Timestamp-int-int-int-int-int-int-int-) (all these data can be acquired from `LocalDateTime`) is deprecated. On the other hand, neither `LocalDateTime` nor `Instant` has a method to get UNIX epoch *seconds*, so in case we used [`Timestamp(long)` constructor](https://docs.oracle.com/javase/8/docs/api/java/sql/Timestamp.html#Timestamp-long-), we would have to conduct division to get rid of milliseconds and then add them again, this time with nanosecond precision.

Fortunately, `java.sql.Timestamp` can be constructed using its `valueOf()` static method. This is implemented in https://github.com/greenplum-db/pxf/pull/160/commits/572cec1252bd30b2884db0fc25a4d827c1e50d1b.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
leskin-in: Thanks for this comment. As mentioned [here](https://github.com/greenplum-db/pxf/pull/160#discussion_r284143900), the same applies to `Timestamp`.

PR #161, opened 2019-05-10
Fix Oracle timestamp wrapping

frankgh:
When a timestamp is coming from GPDB like '2017-12-01
14:01:02'::timestamp, Oracle is reporting an error "ORA-01849: hour must
be between 1 and 12".

Oracle requires 24 hour format like this:
to_timestamp('...', 'YYYY-MM-DD HH24:MI:SS.FF'), where HH24 represents
the Hour of day from 0-23.

This PR fixes timestamp wrapping for Oracle.

### DIFF
--- a/server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/utils/DbProduct.java
+++ b/server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/utils/DbProduct.java
@@ -53,7 +53,7 @@ public enum DbProduct {
 
         @Override
         public String wrapTimestamp(Object val) {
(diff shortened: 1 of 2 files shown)


frankgh:
@leskin-in This issue was reported to us. Can you take a look and tell us if you have seen something similar or if this fix does not makes sense?

leskin-in:
According to [this document](https://docs.oracle.com/cd/B19306_01/server.102/b14200/sql_elements004.htm#i34924), the fix is correct. The old version works for hours in AM/PM format, which is not the one GPDB / PXF uses.

Probably when timestamp was introduced to PXF (then part of HAWQ), the hours part for test did not exceed 12, thus this problem was not revealed.

frankgh:
@leskin-in thanks!

PR #162, opened 2019-05-10
Kill JVM/Tomcat on OutOfMemoryError

oliverralbertini:
After an OutOfMemoryError, the JVM and Tomcat instance are considered
unreliable, and it's best to kill the process. In this commit, we
introduce a script `kill-pxf.sh` in the pxf-service/bin that handles
this situation.

A sleep is used to allow the JVM to log and bubble up to the user the
OutOfMemoryError.

Co-authored-by: Shivram Mani <smani@pivotal.io>


server/pxf-service/src/scripts/pxf-service
```diff
@@ -428,40 +406,37 @@ silent=false
...
         ;;
-    'cluster' )
+    'cluster')
         doCluster "$@"
         ;;
-    * )
+    *)
         printUsage
         exit 2
         ;;
```
denalex: let's limit this PR to only task at hand. The overall cleanup of this script should be a separate PR.

automation/tincrepo/main/pxf/features/general/outOfMemory/expected/query01.ans
```diff
@@ -0,0 +1,33 @@
...
+-- wait until the JVM has been killed
+SELECT pg_sleep(5);
+ pg_sleep 
+----------
+ 
+(1 row)
+
+-- We expect to see a Failed to connect error
+SELECT * from test_out_of_memory;
+psql:/home/gpadmin/workspace/pxf/automation/tincrepo/main/pxf/features/general/outOfMemory/output/query01.sql:13: ERROR:  remote component error (500) from '127.0.0.1:5888':  type  Exception report   message   javax.servlet.ServletException: com.google.common.util.concurrent.ExecutionError: java.lang.OutOfMemoryError: Java heap space    description   The server encountered an internal error that prevented it from fulfilling this request.    exception   javax.servlet.ServletException: javax.servlet.ServletException: com.google.common.util.concurrent.ExecutionError: java.lang.OutOfMemoryError: Java heap space (libchurl.c:920)  (seg0 slice1 172.17.0.2:25435 pid=22230) (libchurl.c:920)
```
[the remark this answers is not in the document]
oliverralbertini: OK

server/pxf-service/src/scripts/pxf-service
```diff
@@ -23,16 +23,16 @@
 parent_script_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd )"
 
 # establish PXF_HOME and global vars used by all commands
-if [[ -z ${PXF_HOME} ]]; then
-    export PXF_HOME=${parent_script_dir}
+if [[ -z $PXF_HOME ]]; then
```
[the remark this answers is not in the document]
oliverralbertini: brackets are only needed when variable is touching some other text, so that bash knows where the variable ends.

```
$ myVar=hello
$ echo "$myVar_world"

$ echo "${myVar}_world"
hello_world
```

server/pxf-service/src/scripts/pxf-service
```diff
@@ -23,16 +23,16 @@
...
 default_env_script=${PXF_HOME}/conf/pxf-env-default.sh
 tomcat_root=${PXF_HOME}/apache-tomcat
 tomcat_templates=${PXF_HOME}/templates/tomcat
-instance_root=${PXF_HOME}
-instance_name=pxf-service
+instance_root=$PXF_HOME
+instance=${instance_root}/pxf-service
 
-curl=`which curl`
+curl=$(command -v curl)
```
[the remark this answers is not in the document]
oliverralbertini: yes, it's more portable than `which`, since it's built-in to bash

server/pxf-service/src/scripts/pxf-service
```diff
@@ -23,16 +23,16 @@
...
-    export PXF_HOME=${parent_script_dir}
+if [[ -z $PXF_HOME ]]; then
+    export PXF_HOME=$parent_script_dir
 fi
 default_env_script=${PXF_HOME}/conf/pxf-env-default.sh
 tomcat_root=${PXF_HOME}/apache-tomcat
 tomcat_templates=${PXF_HOME}/templates/tomcat
-instance_root=${PXF_HOME}
-instance_name=pxf-service
+instance_root=$PXF_HOME
```
frankgh: Do you think it would it be hard to separate the bash changes to a different PR, and only keep the changes relevant to this PR for pxf-service?

frankgh: at the very least they should be different commits, so it's easy to review the changes relevant to the PR.

oliverralbertini: Yeah I agree, sorry for the mess. I will separate out the non-essential bash changes.

oliverralbertini:
@lavjain mentioned that we can put some of the commonly used matches/substitutions in
`automation/tinc/main/tinctest/lib/global_init_file`, and I tried it and it still works.

Maybe we want to think about putting some of the widely used regexes in
there?

automation/tincrepo/main/pxf/features/general/outOfMemory/expected/query01.ans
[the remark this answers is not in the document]
oliverralbertini: @denalex Please take a look at the second commit of this PR, should have the proper templating now.

server/pxf-service/src/scripts/pxf-service
oliverralbertini: OK, here's a new PR with all the bash changes: https://github.com/greenplum-db/pxf/pull/163. New PR is best merged in after this one.

Sorry about that!

server/pxf-service/src/scripts/pxf-service
oliverralbertini: Here's the new PR: https://github.com/greenplum-db/pxf/pull/163.

server/pxf-service/src/scripts/pxf-service
```diff
@@ -214,19 +215,19 @@ function instanceExists()
...
-    command=$1
-    if [[ ${PXF_DEBUG} == "true" && ${command} == "start" ]]; then
-        command="jpda $command"
+    local catalina_flags=()
+    if [[ $PXF_DEBUG == true && $1 == start ]]; then
+        catalina_flags+=(jpda)
     fi
-    pushd ${instance_root} > /dev/null
+    catalina_flags+=("$@")
+    pushd "$instance_root" > /dev/null || return 1
```
frankgh: Why is the return 1 needed here?

server/pxf-service/src/templates/tomcat/bin/setenv.sh
```diff
@@ -24,7 +24,7 @@ JAVA_AGENTS=""
 JAVA_LIBRARY_PATH=""
 
 # DO NOT EDIT VALUES FOR THE VARIABLES BELOW -- they are generated by the start script
-PXF_OPTS="-Dconnector.https.port=8443 -Dpxf.home=${PXF_HOME} -Dpxf.conf=${PXF_CONF} -Dconnector.http.port=${PXF_PORT} -Dbase.shutdown.port=${PXF_SHUTDOWN_PORT} -Dpxf.log.dir=${PXF_LOGDIR} -Dpxf.service.user.impersonation.enabled=${PXF_USER_IMPERSONATION} -Dpxf.service.fragmenter.cache.enabled=${PXF_FRAGMENTER_CACHE} -Dpxf.service.kerberos.keytab=${PXF_KEYTAB} -Dpxf.service.kerberos.principal=${PXF_PRINCIPAL}"
+PXF_OPTS="-XX:OnOutOfMemoryError='${PXF_HOME}/pxf-service/bin/kill-pxf.sh %p &' -Dconnector.https.port=8443 -Dpxf.home=${PXF_HOME} -Dpxf.conf=${PXF_CONF} -Dconnector.http.port=${PXF_PORT} -Dbase.shutdown.port=${PXF_SHUTDOWN_PORT} -Dpxf.log.dir=${PXF_LOGDIR} -Dpxf.service.user.impersonation.enabled=${PXF_USER_IMPERSONATION} -Dpxf.service.fragmenter.cache.enabled=${PXF_FRAGMENTER_CACHE} -Dpxf.service.kerberos.keytab=${PXF_KEYTAB} -Dpxf.service.kerberos.principal=${PXF_PRINCIPAL}"
```
frankgh: I was thinking maybe we should have a way to disable killing the JVM?

automation/tincrepo/main/pxf/features/general/outOfMemory/sql/query01.sql
```diff
@@ -0,0 +1,32 @@
...
+-- s/line \d* of //g
+-- m/Failed to connect to/
+-- s/Failed to connect to.*/Failed to connect to server, must be down/
+--
+-- end_matchsubs
+
+SELECT * from test_out_of_memory;
+
+-- wait until the JVM has been killed
+SELECT pg_sleep(5);
```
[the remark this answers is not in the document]
oliverralbertini: Three seconds seems to be enough, but two seconds is not enough. With two seconds, we get a `libchurl` error other than the `Failed to connect...`. Here's the diff:

```
44c42
< ERROR:  remote component error (0) from 'SOME_IP:SOME_PORT': Recv failure: Connection reset by peer (libchurl.c:xxx)
---
> ERROR:  remote component error (0) from ':SOME_PORT': Failed to connect to server, must be down
```

server/pxf-service/src/scripts/pxf-service
oliverralbertini: In case something goes wrong with `pushd`, it's safer to fail than to continue working in the wrong directory. Same goes for `popd` or even `cd`. But we really should just be setting `set -e` at the top of the script.

automation/tincrepo/main/pxf/features/general/outOfMemory/expected/query01.ans
```diff
@@ -0,0 +1,38 @@
...
+--
+-- m/CONTEXT:.*line.*/
+-- s/line \d* of //g
+-- m/Failed to connect to/
+-- s/Failed to connect to.*/Failed to connect to server, must be down/
+--
+-- end_matchsubs
+SELECT * from test_out_of_memory;
+psql:/home/gpadmin/workspace/pxf/automation/tincrepo/main/pxf/features/general/outOfMemory/output/query01.sql:25: ERROR:  remote component error (500) from 'SOME_IP:SOME_PORT':  type  Exception java.lang.OutOfMemoryError: Java heap space
+CONTEXT:  External table test_out_of_memory, line 1 of file pxf://dummy_path?FRAGMENTER=org.greenplum.pxf.automation.testplugin.OutOfMemoryFragmenter&ACCESSOR=org.greenplum.pxf.plugins.hdfs.LineBreakAccessor&RESOLVER=org.greenplum.pxf.plugins.hdfs.StringPassResolver&delimiter=,
```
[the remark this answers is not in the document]
oliverralbertini: I corrected these, and the above. Thanks.

automation/tincrepo/main/pxf/features/general/outOfMemory/sql/query01.sql
```diff
@@ -0,0 +1,32 @@
...
+--
+-- m/Exception report.*(java.lang.OutOfMemoryError: Java heap space).*/
+-- s/report.*/java.lang.OutOfMemoryError: Java heap space/
+--
+-- m/DETAIL/
+-- s/DETAIL/CONTEXT/
+--
+-- m/CONTEXT:.*line.*/
+-- s/line \d* of //g
+-- m/Failed to connect to/
```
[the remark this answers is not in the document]
oliverralbertini: Added some `--` lines in between all match/substitution pairs.

server/pxf-service/src/templates/tomcat/bin/kill-pxf.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
```
shivzone: Why not move this into the pxf service script instead of maintaining yet another script.
Invoke pxf kill or pxf stop --force command

server/pxf-service/src/templates/user/conf/pxf-env.sh
```diff
@@ -30,3 +30,9 @@ PXF_CONF="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd )"
 
 # Fragmenter cache, set to false to disable
 # export PXF_FRAGMENTER_CACHE=true
+
+# Kill PXF on OutOfMemoryError, set to false to disable
+# export PXF_OOM_KILL=true
+
+# Dump heap on OutOfMemoryError, set to dump path to enable
```
shivzone: Will need to warn or state that the provided path will need to have sufficient space to log heap dump

server/pxf-service/src/templates/tomcat/bin/setenv.sh
oliverralbertini: We have enabled a flag that allows the user to not kill the JVM on the OOM error, and another environment variable that allows the HeapDump to be collected.

server/pxf-service/src/templates/tomcat/bin/kill-pxf.sh
oliverralbertini: We really need the sleep to allow the JVM enough time to bubble up the exception and create a heap dump if required.

server/pxf-service/src/scripts/pxf-service
```diff
@@ -214,19 +215,19 @@ function instanceExists()
...
 
     # Run command as a current user
-    ${instance_root}/${instance_name}/bin/catalina.sh ${command}
+    "${instance_root}/${instance_name}/bin/catalina.sh" "${catalina_flags[@]}"
+    rc=$?
```
denalex: should this be `local rc=$?`

server/pxf-service/src/templates/tomcat/bin/kill-pxf.sh
```diff
@@ -0,0 +1,17 @@
...
+
+log() {
+  echo "=====> $(date) $* <======"
```
denalex: what does $* show ?

server/pxf-service/src/templates/tomcat/bin/kill-pxf.sh
```diff
@@ -0,0 +1,17 @@
...
+}
+
+_main() {
+  sleep 1
+  log 'Stopping PXF'
+  "${PXF_HOME}/bin/pxf" stop -force
```
denalex: does pxf stop have a -force option ?

server/pxf-service/src/templates/tomcat/bin/setenv.sh
```diff
@@ -25,6 +25,12 @@ JAVA_LIBRARY_PATH=""
 
 # DO NOT EDIT VALUES FOR THE VARIABLES BELOW -- they are generated by the start script
 PXF_OPTS="-Dconnector.https.port=8443 -Dpxf.home=${PXF_HOME} -Dpxf.conf=${PXF_CONF} -Dconnector.http.port=${PXF_PORT} -Dbase.shutdown.port=${PXF_SHUTDOWN_PORT} -Dpxf.log.dir=${PXF_LOGDIR} -Dpxf.service.user.impersonation.enabled=${PXF_USER_IMPERSONATION} -Dpxf.service.fragmenter.cache.enabled=${PXF_FRAGMENTER_CACHE} -Dpxf.service.kerberos.keytab=${PXF_KEYTAB} -Dpxf.service.kerberos.principal=${PXF_PRINCIPAL}"
+if [[ $PXF_OOM_KILL == true ]]; then
+  PXF_OPTS+=" -XX:OnOutOfMemoryError='${PXF_HOME}/pxf-service/bin/kill-pxf.sh %p &'"
```
denalex: do we need & at the end of the command ? I thought we do backgrounding in the script itself.

server/pxf-service/src/scripts/pxf-service
oliverralbertini: yes

server/pxf-service/src/templates/tomcat/bin/kill-pxf.sh
oliverralbertini: we are passing all options given by the caller of `pxf` command to `catalina.sh`. `catalina.sh` supports this flag, and will use `kill -9` to bring down the JVM if needed.

oliverralbertini: Right now, this only applies for `pxf start` and `pxf stop`.

server/pxf-service/src/templates/tomcat/bin/setenv.sh
oliverralbertini: Great catch.

server/pxf-service/src/templates/tomcat/bin/kill-pxf.sh
oliverralbertini: all arguments passed to `log()`.

server/pxf-service/src/templates/user/conf/pxf-env.sh
oliverralbertini: Not sure where or how to warn about this. Another possible thing to warn about is that you cannot have a file already there at that path, or it won't get overwritten. The JVM will refuse to overwrite a file that's already there, as discussed in [this comment on StackOverflow](https://stackoverflow.com/questions/542979/using-heapdumponoutofmemoryerror-parameter-for-heap-dump-for-jboss#comment29801217_543037).

server/pxf-service/src/templates/tomcat/bin/kill-pxf.sh
frankgh: @denalex support for the -force option was introduced in this PR

server/pxf-service/src/templates/tomcat/bin/setenv.sh
```diff
@@ -24,8 +25,14 @@ JAVA_AGENTS=""
 JAVA_LIBRARY_PATH=""
 
 # DO NOT EDIT VALUES FOR THE VARIABLES BELOW -- they are generated by the start script
-PXF_OPTS="-Dconnector.https.port=8443 -Dpxf.home=${PXF_HOME} -Dpxf.conf=${PXF_CONF} -Dconnector.http.port=${PXF_PORT} -Dbase.shutdown.port=${PXF_SHUTDOWN_PORT} -Dpxf.log.dir=${PXF_LOGDIR} -Dpxf.service.user.impersonation.enabled=${PXF_USER_IMPERSONATION} -Dpxf.service.fragmenter.cache.enabled=${PXF_FRAGMENTER_CACHE} -Dpxf.service.kerberos.keytab=${PXF_KEYTAB} -Dpxf.service.kerberos.principal=${PXF_PRINCIPAL}"
-JAVA_OPTS="${PXF_JVM_OPTS} $AGENT_PATHS $JAVA_AGENTS $JAVA_LIBRARY_PATH $PXF_OPTS"
+PXF_OPTS="-Dconnector.https.port=8443 -Dpxf.home=$PXF_HOME -Dpxf.conf=$PXF_CONF -Dconnector.http.port=$PXF_PORT -Dbase.shutdown.port=$PXF_SHUTDOWN_PORT -Dpxf.log.dir=$PXF_LOGDIR -Dpxf.service.user.impersonation.enabled=$PXF_USER_IMPERSONATION -Dpxf.service.fragmenter.cache.enabled=$PXF_FRAGMENTER_CACHE -Dpxf.service.kerberos.keytab=$PXF_KEYTAB -Dpxf.service.kerberos.principal=$PXF_PRINCIPAL"
+if [ "$PXF_OOM_KILL" = true ]; then
+  PXF_OPTS="$PXF_OPTS -XX:OnOutOfMemoryError='${PXF_HOME}/pxf-service/bin/kill-pxf.sh %p'"
+fi
+if [ -n "$PXF_OOM_DUMP" ]; then
```
frankgh: @oliverralbertini for Shivram's comment, you could log it here. Not sure if it will be visible for users

server/pxf-service/src/templates/tomcat/bin/kill-pxf.sh
lisakowen: will force be exposed to end users?  if so, maybe we should make -it a double-dash option -i.e. `--force`, like we do for `--help`.   (... single dash for single-character flags, double-dash for multicharacter(i.e. word) options.)

automation/src/main/java/org/greenplum/pxf/automation/testplugin/OutOfMemoryFragmenter.java
```diff
@@ -0,0 +1,19 @@
...
+import java.util.LinkedList;
+import java.util.List;
+
+public class OutOfMemoryFragmenter extends BaseFragmenter {
+
+    @Override
+    public List<Fragment> getFragments() throws Exception {
+        List<Object> objectList = new LinkedList<>();
+        while (true) {
+            byte[] a = new byte[1024 * 1024];
```
frankgh: Let's speed it up a little
```suggestion
            byte[] a = new byte[10 * 1024 * 1024];
```

server/pxf-service/src/templates/tomcat/bin/kill-pxf.sh
shivzone: we can still do the sleep within the kill function in pxf script right

oliverralbertini: yeah, but why should we delay the day to day stopping of the app just for this OOM situation. Another possibility is to add a flag to the `pxf stop` which stops with a sleep.

server/pxf-service/src/templates/tomcat/bin/kill-pxf.sh
oliverralbertini: We decided not to use the `pxf` command to kill PXF in this script. We are passing `-force` internally to `catalina.sh` for `pxf stop` and `pxf restart` (both call `catalina.sh stop -force`) now, but the user won't need to know about it.

server/pxf-service/src/scripts/pxf-env-default.sh
```diff
@@ -76,3 +76,9 @@ export PXF_DEBUG=${PXF_DEBUG:-false}
 
 # Fragmenter cache, set to false to disable
 export PXF_FRAGMENTER_CACHE=${PXF_FRAGMENTER_CACHE:-true}
+
+# Kill PXF on OutOfMemoryError, set to false to disable
+export PXF_OOM_KILL=${PXF_OOM_KILL:-true}
+
+# Dump heap on OutOfMemoryError, set to dump path to enable
+# export PXF_OOM_DUMP=
```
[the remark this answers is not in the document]
oliverralbertini: Good idea, thanks for the suggestion.

server/pxf-service/src/scripts/pxf-service
```diff
@@ -212,23 +213,6 @@ function instanceExists()
     return $?
 }
 
-function commandWebapp()
```
oliverralbertini: @frankgh Please take a look at this bash change that I snuck in.

oliverralbertini: (I noticed that we don't need to change directory to use catalina.sh).

server/pxf-service/src/templates/tomcat/bin/setenv.sh
```diff
@@ -24,8 +25,22 @@ JAVA_AGENTS=""
...
-JAVA_OPTS="${PXF_JVM_OPTS} $AGENT_PATHS $JAVA_AGENTS $JAVA_LIBRARY_PATH $PXF_OPTS"
+PXF_OPTS="-Dconnector.https.port=8443 -Dpxf.home=$PXF_HOME -Dpxf.conf=$PXF_CONF -Dconnector.http.port=$PXF_PORT -Dbase.shutdown.port=$PXF_SHUTDOWN_PORT -Dpxf.log.dir=$PXF_LOGDIR -Dpxf.service.user.impersonation.enabled=$PXF_USER_IMPERSONATION -Dpxf.service.fragmenter.cache.enabled=$PXF_FRAGMENTER_CACHE -Dpxf.service.kerberos.keytab=$PXF_KEYTAB -Dpxf.service.kerberos.principal=$PXF_PRINCIPAL"
+if [ "$PXF_OOM_KILL" = true ]; then
+  PXF_OPTS="$PXF_OPTS -XX:OnOutOfMemoryError='${PXF_HOME}/pxf-service/bin/kill-pxf.sh %p'"
+fi
+if [ -n "$PXF_OOM_DUMP_PATH" ]; then
+  cat <<EOF
+=========================================================================================
+  NOTICE: Setting Tomcat options HeapDumpOnOutOfMemoryError and HeapDumpPath
```
oliverralbertini: @frankgh Please review this NOTICE to the user.

server/pxf-service/src/scripts/pxf-service
frankgh: let's see what others have to say about this one. I don't have context on why we were `pushd`ing to the instance_root directory

server/pxf-service/src/templates/tomcat/bin/setenv.sh
frankgh: These are not tomcat options.
```suggestion
  NOTICE: Adding JVM options -XX:+HeapDumpOnOutOfMemoryError and -XX:+HeapDumpPath
```

server/pxf-service/src/templates/tomcat/bin/setenv.sh
```diff
@@ -24,8 +25,22 @@ JAVA_AGENTS=""
...
+  Please make sure you have enough disk space at $PXF_OOM_DUMP_PATH
```
frankgh:
```suggestion
  Please make sure you have enough disk space at '$PXF_OOM_DUMP_PATH' in case of an OutOfMemoryError
```

server/pxf-service/src/templates/tomcat/bin/setenv.sh
```diff
@@ -24,8 +25,22 @@ JAVA_AGENTS=""
...
+  Tomcat will not overwrite heap dump files, so you may need to rename files
```
frankgh:
```suggestion
  If using a directory, a new Java heap dump file will be generated on each OutOfMemoryError
```

server/pxf-service/src/templates/tomcat/bin/setenv.sh
```diff
@@ -24,8 +25,22 @@ JAVA_AGENTS=""
...
+  before starting PXF, or consider setting \$PXF_OOM_DUMP_PATH to a directory
```
frankgh:
```suggestion
  If using a file path, a single Java heap dump file will generated on the first OutOfMemoryError, and will not be overwritten on subsequent OutOfMemoryErrors
```

server/pxf-service/src/scripts/pxf-service
oliverralbertini: Here's the [commit[(Here's the commit that brought it in, still don't see the necessity for it.) that brought it in, still don't see the necessity for it.

PR #163, opened 2019-05-12
pxf-service: improve bash

oliverralbertini:
This follows various best practices and suggestions from [shellcheck](https://www.shellcheck.net/)
and other optimizations/readability improvements.

In particular:

* Only use `{}` around shell variables when necessary (when referring to
an array index or [parameter expansion](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html))
or when the end of the variable is directly before other characters in a
string
* Quote variables whenever necessary, (it's not necessary on
assignment, in `[[ ]]` tests, or in argument to `case`)
* prefer `$()` over backticks for subshells
* prefer single quotes if no variables are in the string
* prefer `if command; then do something; fi` over `command; rc=$?; if ((
rc == 0 )); then do something; fi`
* prefer bash built-in `command -v` over `which`
* prefer `-n` over `! -z`
* separate declaration of e.g. `local` vars from assignment, since
declaration will mask return value from subshell, e.g.:
`local var=$(false)` has a return code of 0


oliverralbertini:
This should be merged in after https://github.com/greenplum-db/pxf/pull/162.

frankgh:
this branch will get stale the longer we wait to merge. is this good to merge?

PR #166, opened 2019-05-17
Add PXF FDW document

frankgh:
After several iterations and discussions, we would like to propose the following for the PXF foreign data wrapper.

### DIFF
--- a/PXF_FDW.md
+++ b/PXF_FDW.md
@@ -358,7 +358,11 @@ follows:
 
      CREATE FOREIGN TABLE hdfs_parquet
           SERVER hdfs_hdp
-          OPTIONS ( resource '/hdfs/path/to/parque/files', format 'parquet' );
+          OPTIONS ( resource '/hdfs/path/to/parquet/files', format 'parquet' );
+
+     CREATE FOREIGN TABLE hdfs_parquet_gzip
+          SERVER hdfs_hdp
+          OPTIONS ( resource '/hdfs/path/to/parquet/gzip/files', format 'parquet', compression_codec 'gzip' );
 
 #### s3:parquet
 
--- a/PXF_FDW.md
+++ b/PXF_FDW.md
@@ -135,23 +135,23 @@ and Azure Blob Storage access. Users need to create a server for each cloud acce
 
 ### S3 Server Example
 
-     CREATE FOREIGN SERVER s3_server
+     CREATE SERVER s3_server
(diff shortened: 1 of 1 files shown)


shivzone:
Would be useful if you can add for each of the DDL statements (CREATE FOREIGN DATA WRAPPER, CREATE SERVER and CREATE FOREIGN TABLE) the generic template before you go head and discuss the mandatory properties for the DDL.

eg: For Foreign table
```
CREATE FOREIGN TABLE <tbl_name>
      SERVER <server_name>
      OPTIONS ( resource <external_resource>, [...] );
```

shivzone:
Compression codec is something that spans several data formats. Would useful to mention this in the list of supported table properties

frankgh:
@shivzone added compresion_codec example : https://github.com/greenplum-db/pxf/blob/pxf-fdw-d/PXF_FDW.md#hdfsparquet

PXF_FDW.md
```diff
@@ -0,0 +1,548 @@
...
+# PXF Foreign Data Wrapper
+
+For every PXF connector, we propose a PXF Foreign Data Wrapper. PXF will provide
+the following Foreign Data Wrappers.
+
+    CREATE FOREIGN DATA WRAPPER <connector>_pxf_fdw
+        HANDLER pxf_fdw_handler
+        VALIDATOR pxf_fdw_validator
+        OPTIONS ( protocol '<protocol_name>' [, ... ] );
+
```
dkrieger-pvtl: What is the lifecycle of a PXF FDW?  That is, once I CREATE one, how do I query to find which ones I have in use, how do I actually use it, and how do I "drop" it?  These might be covered in another document.

shivzone: \dew would list all the available FDWs. Once the given FDW say pxf_jdbc_fdw is loaded using `create extension`, you wouldn't need create a second jdbc fdw (if you do it would error out saying already available).  For more information on create/drop of FDWs you can refer to https://www.postgresql.org/docs/9.4/sql-createextension.html

frankgh: You cannot have two FDW with the same name, it will produce an error. Like @shivzone said, the FDWs are created when you `create extension pxf_fdw;`. All the PXF FDWs become available to the database where you created the extension. Once FDWs are available, you can create a SERVER, and with the SERVER you can create a FOREIGN TABLE. To allow users to access a specific SERVER, you need to create USER MAPPINGS, that map users to a server.

PR #168, opened 2019-05-19
Add jacoco for code coverage reports

frankgh:
Adding jacoco for code coverage reports. Later, we need to leverage the results of code coverage reports.

### DIFF
--- a/server/build.gradle
+++ b/server/build.gradle
@@ -18,7 +18,8 @@
  */
 
 plugins {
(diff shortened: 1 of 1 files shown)


shivzone:
I'm not entirely convinced that we need jacoco for the state of the code we have it might just get ignored ? Maybe when we get closer to 70-80% coverage it would make sense ?

frankgh:
will close for now. We need to think this one through

PR #169, opened 2019-05-20
HdfsFileFragmenter

frankgh:
HdfsFileFragmenter returns a Fragment per file (instead of a split) from HDFS, S3, or any other hadoop compatible filesystem.

This fragmenter is useful for accessors that process full files (instead of splits) such as the HdfsAtomicDataAccessor or the QuotedLineBreakAccessor

### DIFF
--- a/server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HdfsFileFragmenter.java
+++ b/server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HdfsFileFragmenter.java
@@ -39,7 +39,7 @@ public class HdfsFileFragmenter extends BaseFragmenter {
     public List<Fragment> getFragments() throws Exception {
         String fileName = hcfsType.getDataUri(configuration, context);
         Path path = new Path(fileName);
-        // The hostname is not used anymore, so we hardcode it to localhost
+        // The hostname is no longer used, hardcoding it to localhost
         String[] hosts = {"localhost"};
         byte[] dummyMetadata = HdfsUtilities
                 .prepareFragmentMetadata(0, Integer.MAX_VALUE, hosts);
--- a/server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HdfsFileFragmenter.java
+++ b/server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HdfsFileFragmenter.java
@@ -7,6 +7,8 @@ import org.apache.hadoop.fs.RemoteIterator;
 import org.greenplum.pxf.api.model.BaseFragmenter;
 import org.greenplum.pxf.api.model.Fragment;
 import org.greenplum.pxf.api.model.RequestContext;
+import org.greenplum.pxf.api.utilities.FragmentMetadata;
+import org.greenplum.pxf.plugins.hdfs.utilities.HdfsUtilities;
 
 import java.net.URI;
 import java.util.List;
@@ -54,6 +56,9 @@ public class HdfsFileFragmenter extends BaseFragmenter {
         Path path = new Path(fileName);
         // The hostname is not used anymore, so we hardcode it to localhost
         String[] hosts = {"localhost"};
+        byte[] dummyMetadata = HdfsUtilities
+                .writeBaseFragmentInfo(0, Integer.MAX_VALUE, hosts)
+                .toByteArray();
 
         FileSystem fs = FileSystem.get(URI.create(fileName), configuration);
         RemoteIterator<LocatedFileStatus> fileStatusListIterator =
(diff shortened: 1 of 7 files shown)


server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HdfsFileFragmenter.java
```diff
@@ -0,0 +1,76 @@
...
+     * Gets the fragments for a data source URI that can appear as a file name,
+     * a directory name or a wildcard. Returns the data fragments in JSON
+     * format.
+     */
+    @Override
+    public List<Fragment> getFragments() throws Exception {
+        long fragmenterAccessCountCurrent = fragmenterAccessCount.incrementAndGet();
+
+        if (fragmenterAccessCountCurrent % 100 == 0) {
+            LOG.debug("HdfsFileFragmenter has been invoked {} times during the lifetime of this application",
```
shivzone: What exactly is the value of logging this ?
This would also mean PXF has a limit(Long.MAX_VALUE) in terms of # of requests it can serve over the course of its life.

frankgh: yeah, I'm not sure this adds much value. I will probably remove it. I was exploring ways to expose stats about our fragmenter calls. But this is probably rather limited.

server/pxf-service/src/main/resources/pxf-profiles-default.xml
```diff
@@ -614,7 +614,7 @@ under the License.
             splittable (non parallel) and slower than HdfsTextSimple.
         </description>
         <plugins>
-            <fragmenter>org.greenplum.pxf.plugins.hdfs.HdfsDataFragmenter</fragmenter>
+            <fragmenter>org.greenplum.pxf.plugins.hdfs.HdfsFileFragmenter</fragmenter>
```
denalex: Do cloud-based filesystems (s3, gs, ...) support split-based fragmentation ? If not, we can set FileFragmenter as default for them for :text profiles as well.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HdfsFileFragmenter.java
```diff
@@ -0,0 +1,61 @@
...
+     * format.
+     */
+    @Override
+    public List<Fragment> getFragments() throws Exception {
+        String fileName = hcfsType.getDataUri(configuration, context);
+        Path path = new Path(fileName);
+        // The hostname is no longer used, hardcoding it to localhost
+        String[] hosts = {"localhost"};
+        byte[] dummyMetadata = HdfsUtilities
+                .prepareFragmentMetadata(0, Integer.MAX_VALUE, hosts);
```
denalex: if this is really dummy metadata, meaning it is not used, then maybe set it as null, if possible and update Accessors to consider it optional ? Or at least send 0 as second argument, otherwise systems might believe Integer.MAX is indeed the length of the file.

shivzone: I don't think we should use dummyMetadata as it seems completely unnecessary. We should simply send null back and in openForRead read from 0 to MAX_FILE_TUPLE_SIZE which we can either set to 1GB or Integer.MAX_VALUE. (Refer to the PR https://github.com/greenplum-db/pxf/pull/146/files)
Even the host information we send back is completely ignored

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HdfsFileFragmenter.java
```diff
@@ -0,0 +1,61 @@
...
+import java.net.URI;
+import java.util.List;
+
+/**
+ * Fragmenter class for file resources. This fragmenter
+ * adds support for profiles that require files without
+ * splits. The list of fragments will be the list of files
+ * at the storage layer.
+ */
+public class HdfsFileFragmenter extends BaseFragmenter {
```
shivzone: Better to extend HdfsDataFragmenter and override getFragments. initializeRequest is the same function as the base class

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HdfsFileFragmenter.java
frankgh: @shivzone , I didn't want to modify the accessors. The `HdfsAtomicDataAccessor` calls `HdfsUtilities.parseFileSplit(context)`, which will fail if the metadata is null.

server/pxf-service/src/main/resources/pxf-profiles-default.xml
frankgh: We only know about S3. I am not aware of the implementation of Azure. Should we only do it for S3 since we are confident about that service only?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HdfsFileFragmenter.java
shivzone: We can possibly make the changes in a separate PR to remove unnecessary members in fragment metadata (hosts, etc) and cleanup HdfsAtomicDataAccessor as well. As far as this PR is concerned we should be good to go then.

PR #171, opened 2019-05-23
Enable JDBC connection to Hive and JDBC-specific user impersonation per server 

denalex:


denalex:
@leskin-in -- FYI.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
```diff
@@ -233,9 +242,15 @@ public void initialize(RequestContext context) {
...
-        // Optional parameter. By default, corresponding connectionConfiguration property is not set
+        // Set optional user parameter, taking into account impersonation setting for the server.
         String jdbcUser = configuration.get(JDBC_USER_PROPERTY_NAME);
+        boolean impersonationEnabledForServer = Boolean.parseBoolean(configuration.get(PXF_IMPERSONATION_JDBC_PROPERTY_NAME));
+        LOG.debug("JDBC impersonation is {}enabled for server {}", impersonationEnabledForServer ? "" : "not ", context.getServerName());
+        if (impersonationEnabledForServer) {
+            jdbcUser = context.getUser();
+        }
         if (jdbcUser != null) {
+            LOG.debug("Setting JDBC user to {}", jdbcUser);
```
leskin-in: This is useful, but not required. This setting is included in connection configuration, which is put to log with `DEBUG` level [here](https://github.com/greenplum-db/pxf/pull/171/files#diff-7485e17f411fe4b144fb6415cbbf1092R258).

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcAccessor.java
```diff
@@ -156,15 +160,18 @@ public boolean openForWrite() throws SQLException, SQLTimeoutException, ParseExc
...
         }
         queryWrite = sqlQueryBuilder.buildInsertQuery();
         LOG.trace("Insert query: {}", queryWrite);
 
         statementWrite = super.getPreparedStatement(connection, queryWrite);
-        statementWrite.setQueryTimeout(queryTimeout);
+
+        if (queryTimeout != null) {
+            LOG.debug("Setting query timeout to {} seconds", queryTimeout);
+            statementWrite.setQueryTimeout(queryTimeout);
```
leskin-in: Unfortunately, this is not the only place where query timeout should be set. [`BatchWriterCallable`](https://github.com/greenplum-db/pxf/blob/6167233a0d52134c18907bfa41d64d285fc63717/server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/writercallable/BatchWriterCallable.java#L63) also creates new statements.

I would propose to move this call to [`JdbcPlugin#getPreparedStatement()`](https://github.com/greenplum-db/pxf/blob/6167233a0d52134c18907bfa41d64d285fc63717/server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java#L307) (note that is not a static method). This would also keep processing of user-defined JDBC settings (timeout in this case) in the class where most settings are processed and connections created.

denalex: @leskin-in -- thanks for noticing. I was basically fixing a bug where Hive Driver does not support queryTimeout, so this setting needs to be conditional, but the original commit of that feature missed the case you pointed out. I applied your suggestion and moved the logic to `getPreparedStatement()` method.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
```diff
@@ -233,8 +242,13 @@ public void initialize(RequestContext context) {
         String transactionIsolationString = configuration.get(JDBC_CONNECTION_TRANSACTION_ISOLATION, "NOT_PROVIDED");
         transactionIsolation = TransactionIsolation.typeOf(transactionIsolationString);
 
-        // Optional parameter. By default, corresponding connectionConfiguration property is not set
+        // Set optional user parameter, taking into account impersonation setting for the server.
         String jdbcUser = configuration.get(JDBC_USER_PROPERTY_NAME);
+        boolean impersonationEnabledForServer = Boolean.parseBoolean(configuration.get(PXF_IMPERSONATION_JDBC_PROPERTY_NAME));
```
shivzone: you can use configuration.getBoolean instead of this

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
```diff
@@ -233,8 +242,13 @@ public void initialize(RequestContext context) {
         String transactionIsolationString = configuration.get(JDBC_CONNECTION_TRANSACTION_ISOLATION, "NOT_PROVIDED");
         transactionIsolation = TransactionIsolation.typeOf(transactionIsolationString);
 
-        // Optional parameter. By default, corresponding connectionConfiguration property is not set
+        // Set optional user parameter, taking into account impersonation setting for the server.
         String jdbcUser = configuration.get(JDBC_USER_PROPERTY_NAME);
+        boolean impersonationEnabledForServer = Boolean.parseBoolean(configuration.get(PXF_IMPERSONATION_JDBC_PROPERTY_NAME));
+        LOG.debug("JDBC impersonation is {}enabled for server {}", impersonationEnabledForServer ? "" : "not ", context.getServerName());
```
shivzone: would be useful to also log the user that is used for the jdbc request along with the impersonation being set for clarity

server/build.gradle
```diff
@@ -223,6 +223,8 @@ project('pxf-jdbc') {
         compile(project(':pxf-api'))
 
         bundleJars "postgresql:postgresql:8.4-702.jdbc4"
+        bundleJars "org.apache.hive:hive-jdbc:1.1.0"     // 1.2.2 breaks on CDH-5.x
```
frankgh: what's the issue with CDH? I'm concerned about using an older version of the jar

server/pxf-jdbc/README.md
```diff
@@ -463,3 +463,55 @@ and `$PXF_CONF/servers/EXAMPLE/jdbc-site.xml` on every PXF segment:
 ```
 
 When `SELECT` from this external table is called, PXF executes `SET key1 = value1`, `SET key2 = value2` and `SELECT k, val FROM PUBLIC.T2` (all in the same session). PostgreSQL syntax is used because `Example` is a database unknown to PXF. All queries are executed once as partitioning is not set (one fragment is generated).
+
+## Using JDBC plugin to connect to Hive
+You can use the PXF JDBC connector to retrieve data from Hive. You can also use the "named query" feature to submit a custom SQL query to Hive and retrieve results using PXF JDBC connector.
```
frankgh: can we use the JDBC plugin to write data to hive?

server/pxf-jdbc/README.md
```diff
@@ -463,3 +463,55 @@ and `$PXF_CONF/servers/EXAMPLE/jdbc-site.xml` on every PXF segment:
...
+
+## Using JDBC plugin to connect to Hive
+You can use the PXF JDBC connector to retrieve data from Hive. You can also use the "named query" feature to submit a custom SQL query to Hive and retrieve results using PXF JDBC connector.
+
+While you can specify most of the properties in the EXTERNAL TABLE DDL, the instructions below assume you would use server-based configuration.
+Follow these steps to enable connectivity to Hive:
+1. Define a new PXF configuration server in `$PXF_CONF/servers/` directory
+2. Copy template from `$PXF_CONF/templates/jdbc-site.xml` to your server directory
+3. Edit the file and provide Hive JDBC driver and URL
+    ```angular2html
```
frankgh: should this be xml?

server/pxf-jdbc/README.md
```diff
@@ -463,3 +463,55 @@ and `$PXF_CONF/servers/EXAMPLE/jdbc-site.xml` on every PXF segment:
...
+            ```
+            <property>
+                <name>jdbc.user</name>
+                <value>gpadmin</value>
+            </property>
+            ```
+            2) enable PXF JDBC impersonation so that PXF will automatically use Greenplum's user name to connect to Hive:
+            ```
+            <property>
+                <name>pxf.impersonation.jdbc</name>
```
frankgh: should this property be `pxf.jdbc.impersonation` or just `pxf.impersonation`?

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
frankgh: why not configuration.getBoolean?

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcAccessor.java
```diff
@@ -101,7 +101,11 @@ else if (quoteColumns) {
         // Execute queries
         statementRead = connection.createStatement();
         statementRead.setFetchSize(fetchSize);
-        statementRead.setQueryTimeout(queryTimeout);
+
+        if (queryTimeout != null) {
```
frankgh: why is the timeout being set at the accessor level for read, but was moved for the write case to the JdbcBasePlugin? Should we also move setting the timeout to JdbcBasePlugin as well for the read case?

server/pxf-jdbc/src/test/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePluginTest.java
```diff
@@ -407,4 +408,57 @@ public void testTransactionIsolationSetByUserFailedToGetMetadata() throws SQLExc
...
+        when(DriverManager.getConnection(anyString(), anyObject())).thenReturn(mockConnection);
+
+        PreparedStatement mockStatement = mock(PreparedStatement.class);
+        when(mockConnection.prepareStatement(anyString())).thenReturn(mockStatement);
+
+        JdbcBasePlugin plugin = new JdbcBasePlugin();
+        plugin.initialize(context);
+        plugin.getPreparedStatement(mockConnection, "foo");
+
+        verify(mockStatement, never()).setQueryTimeout(173);
```
frankgh: Would it be more valuable to check for any int?
```suggestion
        verify(mockStatement, never()).setQueryTimeout(anyInt());
```

server/pxf-jdbc/src/test/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePluginTestInitialize.java
```diff
@@ -405,6 +435,93 @@ public void testUser() throws Exception {
...
+
+        // Initialize plugin
+        prepareBaseConfigurationFactory(configuration);
+        JdbcBasePlugin plugin = new JdbcBasePlugin();
+        plugin.initialize(context);
+
+        // Checks
+        Properties expected = new Properties();
+        expected.setProperty("user", "proxy");
+        assertEquals(expected.entrySet(), ((Properties)getInternalState(plugin, "connectionConfiguration")).entrySet());
```
frankgh: can we make `connectionConfiguration` package private, so we don't have to use `getInternalState` to access it?

server/build.gradle
denalex: CDH packages Hive 1.1.0 and if Hive Client 1.2.2 is used to connect, it fails with some wire-level error that is not happening when an older (same as server) version is used.

server/pxf-jdbc/README.md
denalex: I think so, even though that would not be the best method to load data into Hive. We do not yet have any tests for writing, though.

server/pxf-jdbc/README.md
denalex: I don't know, IntelliJ plugin automatically inserted these.

server/pxf-jdbc/README.md
denalex: I didn't want to have just `pxf.impersonation` not to confuse with the system-level property. When we do impersonation per profile the namespace will be easier to manage with the current scheme, where we define "PXF impersonation for profile --> pxf.impersonation.<profile>".

server/pxf-jdbc/src/test/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePluginTest.java
denalex: agree, for never() verification wider is better

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcAccessor.java
denalex: I think because for write we are using preparedStatement that is managed in the base class, while for read the statement is created here. Ideally, the whole connection / statement management logic needs some refactoring.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
denalex: sure, will change, it was done like this in other places in this method

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
denalex: it will be logged below as part of connection configuration (line 262). I originally had it here, but removed to avoid duplicate logging as requested by a reviewer.

PR #172, opened 2019-05-23
Support user xml resource in server

shivzone:
This PR adds support for user specific property overrides to be supported in the server configuration.

The user xml resource will be in the following location under server
`<servername>/`
`<servername>/<username>-user.xml`

The appropriate user configuration file will be picked based on the user running the query and will override the server configuration properties. DDL properties will continue to override both server and user properties


denalex:
Just reading the functional description, I'd say user-level should be the last one that overrides everything else, including DDL. That way a table might be defined once with DDL providing default settings and multiple users can reuse table definition and apply their own overrides, if specified. 

In FDW, however, user-mapping is defined on the server level, so potentially any options on the table level might override user mappings. I guess there the order would be:
FDW < SERVER < USER < TABLE

I don't think we can achieve similar paradigm with current External Table based design as all options are defined in LOCATION on table level, but most of them are potentially SERVER-level (like URL and credentials).

server/pxf-api/src/test/java/org/greenplum/pxf/api/model/BaseConfigurationFactoryTest.java
```diff
@@ -54,52 +54,70 @@ public void testInitConfigurationFailsWhenMultipleDirectoriesWithSameName() {
...
 
         assertEquals("blue", configuration.get("test.blue"));
         assertEquals("red", configuration.get("test.red"));
 
         // Should return null because the file name does not end in -site.xml
         assertNull(configuration.get("test.green"));
+
+        assertEquals("bluevaluefromuser", configuration.get("test.blue.key"));
+        assertEquals("redvaluefromuser", configuration.get("test.red.key"));
+        assertEquals("uservalue", configuration.get("test.user.key"));
```
denalex: I'd assume you've added `dummy-user.xml` file into test resources with these values, but I do not see it as a part of the commit.

shivzone: I missed committing that file. I just did

shivzone:
User level props overriding DDL for the case you mentioned does seem very reasonable. I'm fine with making this change. @frankgh can you confirm as well ?

frankgh:
@shivzone yes. Let's have user-level properties have precedence over any other settings.

server/pxf-api/src/main/java/org/greenplum/pxf/api/model/BaseConfigurationFactory.java
```diff
@@ -68,18 +69,18 @@ public Configuration initConfiguration(String serverName, Map<String, String> ad
...
-                configuration.addResource(resourceURL);
-
-                // store the path to the resource in the configuration in case plugins need to access the files again
-                String fileName = path.getFileName().toString();
-                configuration.set(String.format("%s.%s", PXF_CONFIG_RESOURCE_PATH_PROPERTY, fileName), resourceURL.toString());
+                addResource(configuration, path, serverName);
+            }
+            // add user config file as configuration resource
+            Path path = Paths.get(String.format("%s/%s-user.xml", directory.toPath(), userName));
+            if(Files.exists(path)) {
```
frankgh: formatting
```suggestion
            if (Files.exists(path)) {
```

server/pxf-api/src/test/java/org/greenplum/pxf/api/model/BaseConfigurationFactoryTest.java
```diff
@@ -54,52 +54,70 @@ public void testInitConfigurationFailsWhenMultipleDirectoriesWithSameName() {
         factory = new BaseConfigurationFactory(mockServersDirectory);
         when(mockServersDirectory.listFiles(any(FileFilter.class))).thenReturn(new File[]{new File("a"), new File("b")});
 
-        factory.initConfiguration("dummy", null);
+        factory.initConfiguration("dummy","dummy", null);
```
frankgh: formatting
```suggestion
        factory.initConfiguration("dummy", "dummy", null);
```

shivzone:
> It's looking good. we need integration tests, and it should be good to go.

Just did

automation/Makefile
```diff
@@ -104,14 +104,16 @@ sync_jdbc_config:
...
+		sed $(SED_OPTS) "s|pxfautomation|template1|" $(PXF_CONF_SERVERS)/database/testuser-user.xml; \
 	fi
 	@cp src/test/resources/report.sql $(PXF_CONF_SERVERS)/database
 
 	@mkdir -p $(PXF_CONF_SERVERS)/db-session-params
 	@if [ ! -f "$(PXF_CONF_SERVERS)/db-session-params/jdbc-site.xml" ]; then \
 		cp $(PXF_HOME)/templates/user/templates/jdbc-site.xml $(PXF_CONF_SERVERS)/db-session-params/; \
 		sed $(SED_OPTS) "s|YOUR_DATABASE_JDBC_DRIVER_CLASS_NAME|org.postgresql.Driver|" $(PXF_CONF_SERVERS)/db-session-params/jdbc-site.xml; \
-		sed $(SED_OPTS) "s|YOUR_DATABASE_JDBC_URL|jdbc:postgresql://$(JDBC_HOST):$(JDBC_PORT)/pxfautomation|" $(PXF_CONF_SERVERS)/db-session-params/jdbc-site.xml; \
+		sed $(SED_OPTS) "s|YOUR_DATABASE_JDBC_URL|jdbc:postgresql://$(JDBC_HOST):$(JDBC:q_PORT)/pxfautomation|" $(PXF_CONF_SERVERS)/db-session-params/jdbc-site.xml; \
```
frankgh: this looks like a typo

automation/tincrepo/main/pxf/features/multi_user/expected/query02.ans
```diff
@@ -0,0 +1,61 @@
...
+--
+-- m/You are now connected.*/
+-- s/.*//g
+--
+-- end_matchsubs
+GRANT ALL ON TABLE pxf_jdbc_readable_overrideddl TO PUBLIC;
+GRANT
+\set OLD_GP_USER :USER
+DROP ROLE IF EXISTS testuser;
+psql:/home/gpadmin/workspace/pxf/automation/tincrepo/main/pxf/features/multi_user/output/query02.sql:15: NOTICE:  role "testuser" does not exist, skipping
```
frankgh: this path seems hardcoded for the test environment where you were developing

shivzone: will fix this

automation/Makefile
shivzone: good catch must've been by accident and couldnt' yet verify this due to the red pipeline

server/pxf-service/src/main/java/org/greenplum/pxf/service/utilities/SecureLogin.java
```diff
@@ -70,7 +70,7 @@ public void login() {
             boolean isUserImpersonationEnabled = Utilities.isUserImpersonationEnabled();
             LOG.info("User impersonation is {}", (isUserImpersonationEnabled ? "enabled" : "disabled"));
 
-            Configuration configuration = configurationFactory.initConfiguration("default", null);
+            Configuration configuration = configurationFactory.initConfiguration("default", "dummy", null);
```
frankgh: maybe don't use `dummy` for actual code

denalex: yes, passing null here would be better and there should be a check for null inside the function that would disable user-level lookup.

server/pxf-api/src/main/java/org/greenplum/pxf/api/model/BaseConfigurationFactory.java
```diff
@@ -88,4 +93,22 @@ private void processServerResources(Configuration configuration, String serverNa
...
+        // add user config file as configuration resource
+        try {
+            Path path = Paths.get(String.format("%s/%s-user.xml", directory.toPath(), userName));
+            if (Files.exists(path)) {
+                Configuration userConfiguration = new Configuration(false);
+                URL resourceURL = path.toUri().toURL();
+                userConfiguration.addResource(resourceURL);
+                LOG.debug("Adding user properties for server {} from {}", serverName, resourceURL);
+                userConfiguration.forEach(entry -> configuration.set(entry.getKey(), entry.getValue()));
+                configuration.set(String.format("%s.%s", PXF_CONFIG_RESOURCE_PATH_PROPERTY, path.getFileName().toString()), resourceURL.toString());
```
frankgh: Is this necessary? Wouldn't this value be set here already: https://github.com/greenplum-db/pxf/pull/172/files#diff-acfe8bf39cb3e219a94b6ea2b3e7a53eR86

PR #173, opened 2019-05-25
Fix paramiko pipeline failure

oliverralbertini:
Paramiko is available on the CentOS concourse boxes, but not inside the
`PYTHONPATH` that is normally defined inside
`$GPHOME/greenplum_path.sh`. Because we source this [very late in the
Tinc run process](https://github.com/greenplum-db/pxf/blob/4be25a53e3e422d873b6d03999af9b706a1d74df/automation/src/main/java/org/greenplum/pxf/automation/components/tinc/Tinc.java#L32), we should try to make a change in `greenplum_path.sh`.

The alternative is to copy some files into `$GPHOME/lib/python` from the
system python libraries.

On CentOS 6, these are the dirs we would need:

```
/usr/lib/python2.6/site-packages/paramiko
/usr/lib64/python2.6/site-packages/Crypto
```

On CentOS 7, these are the files/dirs we would need:

```
/usr/lib/python2.7/site-packages/{pip{,/_vendor/pkg_resources},paramiko,pyasn1,idna,six.py,setuptools,enum,ipaddress.py,pycparser}
/usr/lib64/python2.7/site-packages/{cryptography,cffi}
```

### DIFF
--- a/concourse/docker/pxf-dev-base/Dockerfile
+++ b/concourse/docker/pxf-dev-base/Dockerfile
@@ -24,14 +24,18 @@ RUN useradd -s /sbin/nologin -d /opt/minio minio && \
 
 ADD pxf_src /tmp/pxf_src
 
-RUN cd /tmp/pxf_src && \
-    export BUILD_PARAMS='-Dorg.gradle.daemon=false' && \
-    GOPATH=/opt/go PATH=/opt/go/bin:/usr/local/go/bin:$PATH make tar && \
+RUN ["/bin/bash", "-c", "export BUILD_PARAMS=-Dorg.gradle.daemon=false && \
+    GOPATH=/opt/go PATH=/opt/go/bin:/usr/local/go/bin:$PATH make -C /tmp/pxf_src tar && \
     PXF_HOME=/tmp/pxf_src/server/build/stage make -C /tmp/pxf_src/automation dev && \
-    mkdir -p /home/gpadmin/.tomcat && \
-    mv /tmp/pxf_src/server/tomcat/build/apache-tomcat-*.tar.gz /home/gpadmin/.tomcat && \
-    mv /tmp/pxf_src/cli/go/pkg/dep/sources /home/gpadmin/.go-dep-cached-sources && \
-    mv /root/.gradle /root/.m2 /home/gpadmin && \
-    ln -s /home/gpadmin/.tomcat /home/gpadmin/.go-dep-cached-sources /home/gpadmin/.m2 /home/gpadmin/.gradle /root && \
-    chown -R gpadmin:gpadmin /home/gpadmin && \
-    rm -rf /tmp/pxf_src
+    mkdir -p ~gpadmin/.tomcat && \
+    mv /tmp/pxf_src/server/tomcat/build/apache-tomcat-*.tar.gz ~gpadmin/.tomcat && \
+    mv /tmp/pxf_src/cli/go/pkg/dep/sources ~gpadmin/.go-dep-cached-sources && \
+    mv /root/.{gradle,m2} ~gpadmin && \
+    ln -s ~gpadmin/.{tomcat,go-dep-cached-sources,m2,gradle} ~root && \
+    chown -R gpadmin:gpadmin ~gpadmin && \
+    rm -rf /tmp/pxf_src && \
+    if grep 'CentOS release 6' /etc/centos-release; then \
+      yum install -y python-paramiko && yum clean all; \
+    else \
+      pip install psi paramiko --no-cache-dir; \
+    fi"]
--- a/concourse/scripts/pxf_common.bash
+++ b/concourse/scripts/pxf_common.bash
@@ -3,6 +3,7 @@
 GPHOME="/usr/local/greenplum-db-devel"
 PXF_HOME="${GPHOME}/pxf"
 MDD_VALUE="/data/gpdata/master/gpseg-1"
+PXF_COMMON_SRC_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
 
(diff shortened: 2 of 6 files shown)


oliverralbertini:
Another alternative is to add some code like:

```
import sys
sys.path.append('/usr/lib/python2.6/site-packages')
sys.path.append('/usr/lib64/python2.6/site-packages')
```

somewhere in the Tinc code.

oliverralbertini:
Open question:

Where in our Docker/build/test process is `paramiko` getting added?

oliverralbertini:
Here's a [dev gpdb6-integration pipeline](https://dev.ci.gpdb.pivotal.io/teams/main/pipelines/oli:paramiko-fix).

concourse/docker/pxf-dev-base/Dockerfile
```diff
@@ -34,4 +34,5 @@ RUN cd /tmp/pxf_src && \
     mv /root/.gradle /root/.m2 /home/gpadmin && \
     ln -s /home/gpadmin/.tomcat /home/gpadmin/.go-dep-cached-sources /home/gpadmin/.m2 /home/gpadmin/.gradle /root && \
     chown -R gpadmin:gpadmin /home/gpadmin && \
-    rm -rf /tmp/pxf_src
+    rm -rf /tmp/pxf_src && \
+    yum install python-paramiko
```
frankgh:
```suggestion
    yum install -y python-paramiko
```

concourse/scripts/pxf_common.bash
```diff
@@ -63,30 +63,34 @@ function install_gpdb_binary() {
...
 		service ssh start
 		pip install psi
 		psi_dir=$(find /usr/local/lib -name psi | sort -r | head -1)
 	fi
 
-	# Copy PSI package from system python to GPDB as automation test requires it
-	if [[ ! -d ${GPHOME}/lib/python/psi ]]; then
-		cp -r ${psi_dir} ${GPHOME}/lib/python
-	fi
+	pythonpath+=:$(cd "$psi_dir/.." && pwd)
```
lavjain: PSI directory may not be required if it is located inside `/usr/lib` or `/usr/lib64`

concourse/scripts/pxf_common.bash
```diff
@@ -63,30 +63,34 @@ function install_gpdb_binary() {
...
+	echo "export PYTHONPATH=\$PYTHONPATH$pythonpath" >> "${GPHOME}/greenplum_path.sh"
```
lavjain: Could we also check setting the PYTHONPATH on line27 so that greenplum_path script does not have to be modified?

concourse/scripts/pxf_common.bash
oliverralbertini: It is located in `/usr/lib*` under CentOS, but not Ubuntu.

concourse/scripts/pxf_common.bash
oliverralbertini: Won't work there, since we source `greenplum_path.sh` in the Tinc process. But we discussed moving it to `automation/tinc/main/tinc_env.sh`.

concourse/scripts/pxf_common.bash
```diff
@@ -63,30 +63,34 @@ function install_gpdb_binary() {
 		mkdir -p ${GPHOME}
 		tar -xzf bin_gpdb/*.tar.gz -C ${GPHOME}
 	else
-	    build_install_gpdb
+		build_install_gpdb
 	fi
 
+	local psi_dir pythonpath
```
frankgh: don't you need a comma here?

concourse/scripts/pxf_common.bash
lavjain: Looks like pip installs stuff in /usr/local/lib (vs installing the package directly)
```
root@9918b34f2e24:/# find /usr -name paramiko
/usr/lib/python2.7/dist-packages/paramiko
/usr/local/lib/python2.7/dist-packages/paramiko
```
Should we install `paramiko` using `pip`?

concourse/scripts/pxf_common.bash
oliverralbertini: No, check it out:
```
# function test() { local var1 var2; var1=$1; var2=$2; echo var1: $var1 var2: $var2; }
# test one two
var1: one var2: two
# echo var1: $var1 var2: $var2;
var1: var2:
```

concourse/scripts/pxf_common.bash
oliverralbertini: Looks like Ubuntu and CentOS behave differently in that regard.

concourse/scripts/pxf_common.bash
oliverralbertini: Also, assignments are possible:
```
function test() { local var1=$1 var2=$2 var3=$var1$var2; }
```

concourse/scripts/pxf_common.bash
```diff
@@ -57,37 +58,41 @@ function build_install_gpdb() {
...
 		psi_dir=$(find /usr/local/lib -name psi | sort -r | head -1)
 	fi
 
-	# Copy PSI package from system python to GPDB as automation test requires it
-	if [[ ! -d ${GPHOME}/lib/python/psi ]]; then
-		cp -r ${psi_dir} ${GPHOME}/lib/python
-	fi
-
+	psi_dir=$(cd "$psi_dir/.." && pwd)
+	[[ $export_pythonpath =~ $psi_dir ]] || export_pythonpath+=:$psi_dir
```
frankgh: what does this do?

oliverralbertini: Adds `$psi_dir` to the `$export_pythonpath` string only if it is not already included.

PR #174, opened 2019-05-28
Update path to the ubuntu dockerfile

frankgh:
The Ubuntu image location will change in the GPDB repo from src/tools/docker/ubuntu16/Dockerfile to src/tools/docker/ubuntu/Dockerfile. This PR reflects that change in the docker images pipeline


lavjain:
This PR is dependent on https://github.com/greenplum-db/gpdb/pull/7778

PR #176, opened 2019-05-30
Upgrade jackson 2 version from 2.9.8 -> 2.9.9

frankgh:
From the github security report:
A Polymorphic Typing issue was discovered in FasterXML
jackson-databind 2.x before 2.9.9. When Default Typing
is enabled (either globally or for a specific property)
for an externally exposed JSON endpoint, the service
has the mysql-connector-java jar (8.0.14 or earlier)
in the classpath, and an attacker can host a crafted
MySQL server reachable by the victim, an attacker can
send a crafted JSON message that allows them to read
arbitrary local files on the server. This occurs
because of missing com.mysql.cj.jdbc.admin.MiniAdmin
validation.

### DIFF
--- a/server/build.gradle
+++ b/server/build.gradle
@@ -170,9 +170,9 @@ project('pxf-hdfs') {
     dependencies {
         compile(project(':pxf-api'))
         compile "org.apache.avro:avro-mapred:1.7.7"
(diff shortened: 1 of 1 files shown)


shivzone:
@frankgh would be useful to have the link to the security alert in your description for future reference

PR #177, opened 2019-05-30
Avoid non-POSIX shell in server/Makefile

oliverralbertini:
On Ubuntu, `dash` is used, which doesn't support `pushd` or `popd`.
In general, we can only rely on POSIX shell commands.


shivzone:
Aren't there more places where we use pushd ? that might be during the install path as opposed to comiplation

oliverralbertini:
I searched in all the Makefiles, didn't find any other occurrences of `pushd`. [Previous PR](https://github.com/greenplum-db/pxf/pull/173/commits/e6699f435d82342c4ce7fed7cba8f040fc6e0b06) had some other non-POSIX stuff (`[[` and `==`), but I just missed this one.

shivzone:
got it

PR #178, opened 2019-05-30
Make pxf threads configurable

shivzone:
Users can now configure maximum pxf threads by setting `pxf_threads` property in `${PXF_CONF}/conf/pxf-env.sh`
Also reduced the minimum threads default to 25 instead of 50 which is also the tomcat default

Authored-by: Shivram Mani <smani@pivotal.io>


server/pxf-service/src/scripts/pxf-env-default.sh
```diff
@@ -62,6 +62,9 @@ export PXF_SHUTDOWN_PORT=${PXF_SHUTDOWN_PORT:=5889}
 # Memory
 export PXF_JVM_OPTS=${PXF_JVM_OPTS:="-Xmx2g -Xms1g"}
 
+# Threads
+export PXF_THREADS=${PXF_THREADS:=300}
```
frankgh: we were considering reducing this number from 300 to 150 for out of the box settings. What do you think?

server/pxf-service/src/templates/tomcat/conf/server.xml
```diff
@@ -33,8 +33,8 @@ under the License.
                   type="org.apache.catalina.UserDatabase"/>
     </GlobalNamingResources>
     <Service name="Catalina">
-        <Executor maxThreads="300"
-                  minSpareThreads="50"
+        <Executor maxThreads="${pxf.threads}"
+                  minSpareThreads="25"
```
frankgh: do we also want to change this number?

shivzone:
@frankgh i thought about that, but from a users perspective i thought they shouldn't really be aware of min or max threads but instead only aware of # of pxf worker threads. Thats why i went with PXF_THREADS

server/pxf-service/src/scripts/pxf-env-default.sh
shivzone: My only concern with reducing this is to avoid any regression for existing customers. Also even if we reduce the maxThreads to 150 and used about 25MB/thread we would still need about 3.75GB of memory. If you are certain about this change i can do it.

server/pxf-service/src/templates/tomcat/conf/server.xml
shivzone: 25 is the tomcat default so set it to that number. I don't see any valid reason to have it as high as 50.

denalex: yes, 25 should be good enough

denalex:
both of you have valid points, I'm fine either way

shivzone:
@frankgh i've modified the env var and also the system property

Issue #181, opened 2019-06-06
Greenplum PXF EXTERNAL READABLE TABLE using pre-defined query


sree80:
As per the example listed for aggregation query, I did exactly the same -
MYSQL :-

use demodb;
create table dept(
    id int(4) not null primary key,
    name varchar(20) not null
);

create table emp(
    dept_id int(4) not null,
    name varchar(20) not null,
    salary int(8)
);

Some data is inserted into MySQL tables:

insert into dept values(1, 'sales');
insert into dept values(2, 'finance');
insert into dept values(3, 'it');

insert into emp values(1, 'alice', 11000);
insert into emp values(2, 'bob', 10000);
insert into emp values(3, 'charlie', 10500);

Then a complex aggregation query is created and placed in a file, say report.sql. The file needs to be placed in the server configuration directory under $PXF_CONF/servers/. So, let's assume we have created a mydb server configuration directory, then this file will be $PXF_CONF/servers/mydb/report.sql. Jdbc driver name and connection parameters should be configured in $PXF_CONF/servers/mydb/jdbc-site.xml for this server.

SELECT dept.name AS name, count(*) AS count, max(emp.salary) AS max
FROM demodb.dept JOIN demodb.emp
ON dept.id = emp.dept_id 
GROUP BY dept.name;

A table in GPDB is created with the schema corresponding to the results returned by the aggregation query.

CREATE EXTERNAL TABLE dept_report (
    name text,
    count int,
    max int
)
LOCATION (
    'pxf://query:report?PROFILE=JDBC&SERVER=mydb'
)
FORMAT 'CUSTOM' (
    FORMATTER='pxfwritable_import'
);

A query to a GPDB external table is made:

I'm getting this Error message :-

ERROR:  remote component error (500) from '127.0.0.1:5888':  type  Exception report   message   

You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ':report' at line 1    description   The server encountered an internal error that prevented it from fulfilling this request.    exception   java.io.IOException: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ':report' at line 1 (libchurl.c:944)  (seg1 slice1 172.17.0.2:50001 pid=7495) (libchurl.c:944)

MYSQL Version = 5.1
MYSQL JDBC Connector =  5.1

frankgh:
[...]

PR #182, opened 2019-06-06
Run a named query that ends with semicolon

frankgh:
Named queries should work correctly even when semicolon is present at the end of the query

### DIFF
--- a/server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcAccessor.java
+++ b/server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcAccessor.java
@@ -330,7 +330,8 @@ public class JdbcAccessor extends JdbcBasePlugin implements Accessor {
             throw new RuntimeException(String.format("Query text file is empty for query %s", queryName));
         }
 
-        // Remove semicolon at the end of the query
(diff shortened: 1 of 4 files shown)


frankgh:
This PR is inspired by https://github.com/greenplum-db/pxf/issues/181

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcAccessor.java
```diff
@@ -329,6 +329,15 @@ private String getQueryText() {
...
         }
+
+        int semicolonIndex = queryText.lastIndexOf(";");
+        // Remove semicolon at the end of the query
+        if (semicolonIndex > -1) {
+            if (StringUtils.isBlank(queryText.substring(semicolonIndex + 1))) {
+                queryText = queryText.substring(0, semicolonIndex);
+            }
+        }
+
```
oliverralbertini: I think I would prefer something using [str.replaceFirst](https://docs.oracle.com/javase/7/docs/api/java/lang/String.html#replaceFirst(java.lang.String,%20java.lang.String))
```
queryText.replaceFirst(";\s*$", "");
```

This will also handle the case where there is white space at the end.

PR #183, opened 2019-06-06
 Upgrade postgres driver to version 42.2.5

frankgh:
Our existing postgres driver is 8 years old. The new version of the
driver was moved to the org.postgres namespace and the latest released
version is 42.2.5 released in August 2018.

### DIFF
--- a/automation/pom.xml
+++ b/automation/pom.xml
@@ -172,9 +172,9 @@
         </dependency>
 
         <dependency>
-            <groupId>postgresql</groupId>
+            <groupId>org.postgresql</groupId>
             <artifactId>postgresql</artifactId>
-            <version>8.4-702.jdbc4</version>
+            <version>42.2.5</version>
         </dependency>
 
         <dependency>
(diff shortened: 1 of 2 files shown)


denalex:
Which problem does this upgrade solve ? Our driver version is in line with the Postgres versions Greenplum is based on. Using the latest one assumes a backward compatibility over a series of major versions.

shivzone:
I'm ok with using 42.2.5  it supports PostgreSQL 8.2 and above and java 8. Also it allows us to use statement properties like query timeout. In fact 42.2.5 is the driver @lisakowen has in our pxf jdbc [docs](https://gpdb.docs.pivotal.io/6-0Beta/pxf/jdbc_pxf.html)

The only question i have is why did we choose to bundle postgres driver with our product, and if there is a reason why not bundle the other well known database drivers as well.

frankgh:
@denalex  In my view, the use case is to access external postgres services, the most recent postgres version is postgres 11, and postgres 12 is coming out. So, it's better to be in line with newer version of postgres than Greenplum.

frankgh:
Related to https://github.com/greenplum-db/pxf/issues/184

PR #186, opened 2019-06-14
Enable JDBC Connection Pooling

denalex:


denalex:
@leskin-in -- fyi, please take a look -- introducing connection pool to limit number of connections to the target DB.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
```diff
@@ -73,6 +79,7 @@
     private static final String PXF_IMPERSONATION_JDBC_PROPERTY_NAME = "pxf.impersonation.jdbc";
 
     private static final String HIVE_URL_PREFIX = "jdbc:hive2://";
+    private static final String DATASOURCE_PROPERTY_PREFIX = "datasource.";
```
denalex: dataSource ? test case ?

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/utils/PoolDescriptor.java
```diff
@@ -0,0 +1,102 @@
...
+import java.util.Objects;
+import java.util.Properties;
+import java.util.Set;
+
+public class PoolDescriptor {
+
+    private static final String USER_PROPERTY_NAME = "user";
+    private static final String PASSWORD_PROPERTY_NAME = "password";
+    private static final Set<String> PROHIBITED_PROPERTIES =
+            Sets.newHashSet("username", "password", "dataSource.user", "dataSource.password", "dataSourceClassName", "jdbcUrl");
```
denalex: add comment

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
denalex: property is not used, removed it.

leskin-in:
@denalex ,
thank you for implementing this feature. I think this is the one many have been expecting for a long time.

Could you clarify whether the intent is to preserve JDBC connections to the external database between different queries (from GPDB) to PXF? If so, what mechanism allows us to make [this](https://github.com/greenplum-db/pxf/pull/186/files?file-filters%5B%5D=.java&file-filters%5B%5D=.md&file-filters%5B%5D=.xml#diff-b20a89115a5b15ad7821ba35bb57b529R32) or maybe some other object persistent (by this I mean that such a persistent object must be preserved across different queries)?

I tried something similar a few weeks ago and found that even the `static` members of some **class from a plugin folder** are not persistent. They are recreated for each incoming query to PXF.

denalex:
@leskin-in -- the main motivation for the connection pooling is to limit the number of connections to the target database from across the Greenplum cluster. Users can limit the number of connections per (server/url/user/params) to "X" and PXF will not establish more than "GP hosts * X" connections.

Out of the box we'll ship with setting allowing connection to be idle for 30 secs. If a single JVM is fetching multiple fragments of the same query one after the other, then the connection will be re-used, which should improve performance slightly as well. After being idle for more than 30sec, connections will be closed and even the whole connection pool might be removed 6 hours after the connection has last been requested from it.

I have not heard or experiences the problems with `static` members you mention. Ideally, this should not be happening, of course. But if it does, that would mean each request causes a class to be loaded in a new classloader, which would be less than efficient and sometimes plain wrong. Do you have a reproduction test case or scenario ?

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/utils/ConnectionManager.java
```diff
@@ -124,5 +141,49 @@ public static String maskPassword(String password) {
...
+    static class DataSourceFactory {
+
+        private static final DataSourceFactory dataSourceFactoryInstance = new DataSourceFactory();
+
+        /**
+         * Creates a new datasource instance based on parameters contained in PoolDescriptor.
+         * @param poolDescriptor descriptor containing pool parameters
+         * @return instance of HikariDataSource
+         */
+        HikariDataSource createDataSource(PoolDescriptor poolDescriptor) {
```
frankgh: Should we use the DataSource interface instead?

server/pxf-jdbc/src/test/java/org/greenplum/pxf/plugins/jdbc/utils/ConnectionManagerTest.java
```diff
@@ -139,5 +145,98 @@ public void testGetConnectionPoolEnabledWithPoolProps() throws SQLException {
...
+        ticker.advanceTime(ConnectionManager.CLEANUP_TIMEOUT_NANOS + 100000, TimeUnit.NANOSECONDS);
+
+        // wait again as cleaner needs to pick new ticker value
+        Uninterruptibles.sleepUninterruptibly(150, TimeUnit.MILLISECONDS);
+
+        verify(mockMBean, atLeast(3)).getActiveConnections();
+        verify(mockDataSource, times(1)).close(); // verify datasource is closed when evicted
+    }
+
+    class MockTicker extends Ticker {
```
frankgh: Can we reuse the `FakeTicker` class instead?
https://github.com/greenplum-db/pxf/blob/master/server/pxf-service/src/test/java/org/greenplum/pxf/service/FakeTicker.java

frankgh: Also, regarding naming. This is not really a Mock

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/utils/ConnectionManager.java
denalex: we did originally, but then it doesn't have close() method so we needed to cast it HikariDS everywhere we needed to do anything bu getting a connection. Since we don't support any others, we decided to be specific here.

server/pxf-jdbc/src/test/java/org/greenplum/pxf/plugins/jdbc/utils/ConnectionManagerTest.java
denalex: That would've introduced test dependency on pxf-service module and having test dependencies between modules is tricky and not worth the effort, IMO. Name is MockTicker to indicate it is to be used with mocks, but yes, might be confusing, could've called it `CustomTicker`. It is not "fake" either :)

leskin-in:
@denalex,
Thank you for the explanation.

> I have not heard or experienced the problems with `static` members you mention. Ideally, this should not be happening, of course. But if it does, that would mean each request causes a class to be loaded in a new classloader, which would be less than efficient and sometimes plain wrong. Do you have a reproduction test case or scenario ?

I have re-checked this and found the static members are preserved. So the problem I mentioned above does not reproduce.

PR #188, opened 2019-06-18
Support for FDW

frankgh:
This PR adds support for FDW:
- The first commit is code clean up in the fragmenter call. This is a backwards compatible change
- The second commit adds new profiles for csv format (which point to the same Fragmenter/Accessor/Resolver as their text counterpart).

### DIFF
--- a/server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/FragmenterResource.java
+++ b/server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/FragmenterResource.java
@@ -96,6 +96,7 @@ public class FragmenterResource extends BaseResource {
                                  @Context final HttpHeaders headers)
             throws Throwable {
 
+        LOG.debug("Received FRAGMENTER call");
         long startTime = System.currentTimeMillis();
         final RequestContext context = parseRequest(headers);
         final String path = context.getDataSource();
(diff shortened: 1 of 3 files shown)


server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/FragmenterResource.java
```diff
@@ -86,24 +86,23 @@ public FragmenterResource() {
...
             throws Throwable {
 
         long startTime = System.currentTimeMillis();
-        LOG.debug("FRAGMENTER started for path \"{}\"", path);
-
         final RequestContext context = parseRequest(headers);
+        final String path = context.getDataSource();
         final String fragmenterCacheKey = getFragmenterCacheKey(context);
 
+        LOG.debug("FRAGMENTER started for path \"{}\"", path);
```
denalex: the original placement of the log statement helped to debug arrival of request to PXF before any substantial logic could be called. We should preserve it, perhaps log a different statement

server/pxf-service/src/main/resources/pxf-profiles-default.xml
```diff
@@ -220,7 +220,18 @@ under the License.
     <profile>
         <name>hdfs:text</name>
         <description>This profile is suitable for using when reading delimited single line records
-            from plain text files on HDFS.
+            from plain text, tab-delimited, files on HDFS.
```
denalex: why tab-delimited ?

frankgh: This is making reference to tab-delimited text, which is the default for 'TEXT' format in the external table framework: https://gpdb.docs.pivotal.io/5190/ref_guide/sql_commands/CREATE_EXTERNAL_TABLE.html

server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/FragmenterResource.java
frankgh: added `Received FRAGMENTER call`

PR #189, opened 2019-06-20
Implement cluster reset command

oliverralbertini:
`pxf cluster` will call local `pxf-service` script with 'destroy' as a
command on each segment host.

Right now we are deleting `$PXF_CONF`, and we are not sure that's really necessary or a good idea.

### DIFF
--- a/cli/go/src/pxf-cli/cmd/cluster.go
+++ b/cli/go/src/pxf-cli/cmd/cluster.go
@@ -20,89 +20,41 @@ type ClusterData struct {
 	connection *dbconn.DBConn
 }
 
+func createCobraCommand(use string, short string, command *pxf.Command) *cobra.Command {
+	if command == nil {
+		return &cobra.Command{Use: use, Short: short}
+	}
+	return &cobra.Command{
+		Use:   use,
+		Short: short,
+		Run: func(cmd *cobra.Command, args []string) {
+			clusterData, err := doSetup(command)
+			if err == nil {
+				err = clusterRun(command, clusterData)
+			}
+			exitWithReturnCode(err)
+		},
+	}
+}
+
 var (
-	clusterCmd = &cobra.Command{
-		Use:   "cluster",
-		Short: "Perform <command> on each segment host in the cluster",
-	}
-
-	initCmd = &cobra.Command{
-		Use:   "init",
-		Short: "Initialize the PXF server instances on master, standby master, and the segment hosts",
-		Run: func(cmd *cobra.Command, args []string) {
-			command := &pxf.InitCommand
-			clusterData, err := doSetup(command)
-			if err == nil {
-				err = clusterRun(command, clusterData)
-			}
-			exitWithReturnCode(err)
-		},
-	}
-
-	startCmd = &cobra.Command{
-		Use:   "start",
-		Short: "Start the PXF server instances on the segment hosts",
-		Run: func(cmd *cobra.Command, args []string) {
-			command := &pxf.StartCommand
-			clusterData, err := doSetup(command)
-			if err == nil {
-				err = clusterRun(command, clusterData)
-			}
-
-			exitWithReturnCode(err)
-		},
-	}
-
-	stopCmd = &cobra.Command{
-		Use:   "stop",
-		Short: "Stop the PXF server instances on the segment hosts",
-		Run: func(cmd *cobra.Command, args []string) {
-			command := &pxf.StopCommand
-			clusterData, err := doSetup(command)
-			if err == nil {
-				err = clusterRun(command, clusterData)
-			}
-
-			exitWithReturnCode(err)
-		},
-	}
-
-	syncCmd = &cobra.Command{
-		Use:   "sync",
-		Short: "Sync PXF configs from master to standby master and the segment hosts",
-		Run: func(cmd *cobra.Command, args []string) {
-			command := &pxf.SyncCommand
-			clusterData, err := doSetup(command)
-			if err == nil {
-				err = clusterRun(command, clusterData)
-			}
-
-			exitWithReturnCode(err)
-		},
-	}
-
-	statusCmd = &cobra.Command{
-		Use:   "status",
-		Short: "Get status of PXF servers on the segment hosts",
-		Run: func(cmd *cobra.Command, args []string) {
-			command := &pxf.StatusCommand
-			clusterData, err := doSetup(command)
-			if err == nil {
-				err = clusterRun(command, clusterData)
-			}
-
-			exitWithReturnCode(err)
-		},
-	}
+	clusterCmd = createCobraCommand("cluster", "Perform <command> on each segment host in the cluster", nil)
+	initCmd    = createCobraCommand("init", "Initialize the PXF server instances on master, standby master, and all segment hosts", &pxf.InitCommand)
+	startCmd   = createCobraCommand("start", "Start the PXF server instances on all segment hosts", &pxf.StartCommand)
+	stopCmd    = createCobraCommand("stop", "Stop the PXF server instances on all segment hosts", &pxf.StopCommand)
+	statusCmd  = createCobraCommand("status", "Get status of PXF servers on all segment hosts", &pxf.StatusCommand)
+	syncCmd    = createCobraCommand("sync", "Sync PXF configs from master to standby master and all segment hosts", &pxf.SyncCommand)
+	resetCmd   = createCobraCommand("reset", "Reset PXF (undo initialization) on all segment hosts", &pxf.ResetCommand)
 )
 
 func init() {
 	rootCmd.AddCommand(clusterCmd)
 	clusterCmd.AddCommand(initCmd)
 	clusterCmd.AddCommand(startCmd)
-	clusterCmd.AddCommand(statusCmd)
 	clusterCmd.AddCommand(stopCmd)
+	clusterCmd.AddCommand(statusCmd)
 	clusterCmd.AddCommand(syncCmd)
+	clusterCmd.AddCommand(resetCmd)
 }
 
 func exitWithReturnCode(err error) {
@@ -194,6 +146,13 @@ func adaptContentIDToHostname(cluster *cluster.Cluster, f func(string) string) f
 
 func clusterRun(command *pxf.Command, clusterData *ClusterData) error {
 	defer clusterData.connection.Close()
+
+	err := command.Warn(os.Stdin)
+	if err != nil {
+		gplog.Info(fmt.Sprintf("%s", err))
+		return err
+	}
+
 	f, err := command.GetFunctionToExecute()
 	if err != nil {
 		gplog.Error(fmt.Sprintf("Error: %s", err))
--- a/cli/go/src/pxf-cli/cmd/cluster_test.go
+++ b/cli/go/src/pxf-cli/cmd/cluster_test.go
@@ -64,6 +64,11 @@ var _ = Describe("GenerateStatusReport()", func() {
 		_ = cmd.GenerateStatusReport(&pxf.StatusCommand, clusterData)
 		Expect(testStdout).Should(gbytes.Say("Checking status of PXF servers on 2 hosts..."))
 	})
+
+	It("reports the number of hosts that are getting reset", func() {
+		_ = cmd.GenerateStatusReport(&pxf.ResetCommand, clusterData)
+		Expect(testStdout).Should(gbytes.Say("Resetting PXF on master and 2 other hosts..."))
+	})
 })
 
 var _ = Describe("GenerateOutput()", func() {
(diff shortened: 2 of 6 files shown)


cli/go/src/pxf-cli/pxf/pxf.go
```diff
@@ -134,6 +135,16 @@ var (
 		envVars:    []EnvVar{Gphome},
 		whereToRun: cluster.ON_HOSTS,
 	}
+	DestroyCommand = Command{
+		commandName: Destroy,
+		messages: map[MessageType]string{
+			Success: "Destroyed PXF on %d out of %d hosts\n",
+			Status:  "Destroying PXF on master and %d other hosts...\n",
+			Error:   "Could not destroy PXF on %d out of %d hosts\n",
```
denalex: seems we have word "failed" in all other error messages, should we say "Failed to destroy PXF on ..."

server/pxf-service/src/scripts/pxf-service
```diff
@@ -405,6 +428,9 @@ validate_user
 validate_system
 
 case $pxf_script_command in
+    'destroy')
+        doDestroy
+        ;;
```
denalex: move it down to the end, please

server/pxf-service/src/scripts/pxf-service
```diff
@@ -317,6 +317,29 @@ doHelp() {
     exit 0
 }
 
+# doDestroy handles the destroy command
+function doDestroy()
+{
+    if doStatus >/dev/null; then
+        echo "PXF is running. Will stop..."
+        doStop || return 1
```
denalex: should we rather fail if PXF is running as to not accidentally kill running queries ?

cli/go/src/pxf-cli/cmd/cluster.go
```diff
@@ -94,10 +94,25 @@ var (
 			exitWithReturnCode(err)
 		},
 	}
+
```
denalex: should we have an extra warning "All custom PXF settings will be lost, are you sure you want to proceed ?" before proceeding and cleaning stuff ?

server/pxf-service/src/scripts/pxf-service
```diff
@@ -317,6 +317,29 @@ doHelp() {
...
+    fi
+
```
denalex: do we also need a warning here ? we can accept an argument "-force" which will avoid asking for confirmation that cluster tool will send.

cli/go/src/pxf-cli/cmd/cluster.go
```diff
@@ -94,10 +94,25 @@ var (
...
+		Run: func(cmd *cobra.Command, args []string) {
+			command := &pxf.DestroyCommand
+			clusterData, err := doSetup(command)
+			if err == nil {
+				err = clusterRun(command, clusterData)
+			}
+
+			exitWithReturnCode(err)
+		},
+	}
```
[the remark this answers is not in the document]
oliverralbertini: Yes, that's a nice idea. We haven't always done the same thing (there was a time when we batched `sync` operation) but now we are.

cli/go/src/pxf-cli/cmd/cluster.go
oliverralbertini: That should be do-able.

server/pxf-service/src/scripts/pxf-service
raymondyin: Good question. The story says "make sure pxf is not running" when running destroy command. 

It would be nice to implement a "yes/No" prompt for users to decide whether to shutdown PXF or not when executing "destroy" against cluster with running PXF.

cli/go/src/pxf-cli/cmd/cluster_test.go
```diff
@@ -140,6 +150,12 @@ var _ = Describe("GenerateOutput()", func() {
 			Expect(testStdout).Should(gbytes.Say("PXF is not running on 1 out of 3 hosts"))
 			Expect(testStderr).Should(gbytes.Say("sdw2 ==> an error happened on sdw2"))
 		})
+
+		It("reports the number of hosts that failed to destroy", func() {
+			_ = cmd.GenerateOutput(&pxf.DestroyCommand, clusterData)
+			Expect(testStdout).Should(gbytes.Say("Could not destroy PXF on 1 out of 3 hosts"))
+			Expect(testStderr).Should(gbytes.Say("sdw2 ==> an error happened on sdw2"))
```
frankgh: should we have more descriptive error messages? can we get more detail from the reported error?

cli/go/src/pxf-cli/pxf/pxf.go
```diff
@@ -81,6 +81,7 @@ const (
 	Stop     = "stop"
 	Sync     = "sync"
 	Statuses = "status"
+	Destroy  = "destroy"
```
frankgh: have we settled on the `destroy` word? I read 
> Destroying PXF on master and 2 other hosts.
and got a little scared. Maybe we can come up with a different word? Does everyone like destroying?

server/pxf-service/src/scripts/pxf-service
frankgh: I agree with @denalex . Let's abort the command if PXF is running. To run pxf cluster destroy make sure pxf is not running.

server/pxf-service/src/scripts/pxf-service
frankgh: +1

server/pxf-service/src/scripts/pxf-service
```diff
@@ -317,6 +317,29 @@ doHelp() {
...
+    echo "Cleaning ${PXF_HOME}/conf/pxf-private.classpath..."
+    rm -f "${PXF_HOME}/conf/pxf-private.classpath"
+    echo "Cleaning ${PXF_CONF}..."
+    rm -rf "$PXF_CONF" # do we really want to do this?
```
frankgh: don't remove the $PXF_CONF directory! My vote is to never delete it. Personally, I have put a lot of effort on my PXF_CONF directory that I would get really mad if this command destroyed it. Maybe warn that PXF_CONF will not get destroyed

server/pxf-service/src/scripts/pxf-service
frankgh: +1

lavjain:
**Destroy PXF** sounds like an evil operation. It could be more appropriately names as either _Reset PXF to factory settings_ :-) or _cleanup PXF confs_.

cli/go/src/pxf-cli/cmd/cluster_test.go
oliverralbertini: Yes, whatever is in `stderr` will get reported to the user, that's what the test is verifying.

Also, we make sure no to print more than 3 lines of the error message, in case it happens to be really long. We decided to do that in case someone is using `pxf cluster *` commands on large clusters.

oliverralbertini:
I've renamed the `destroy` command to `reset`.

server/pxf-service/src/scripts/pxf-service
oliverralbertini: Now it just echos `"Ignoring ${PXF_CONF}..."`

lisakowen:
quick question, @oliverralbertini - is this new command available only at the cluster level (pxf cluster reset), or could the admin run it on a single segment host (pxf reset)?

oliverralbertini:
> quick question, @oliverralbertini <https://github.com/oliverralbertini> -
> is this new command available only at the cluster level (pxf cluster
> reset), or could the admin run it on a single segment host (pxf reset)?
>
Will be available at both the local and cluster levels.

cli/go/src/pxf-cli/cmd/cluster.go
```diff
@@ -20,89 +20,41 @@ type ClusterData struct {
...
-				err = clusterRun(command, clusterData)
-			}
-
-			exitWithReturnCode(err)
-		},
-	}
+var (
+	clusterCmd = createCobraCommand("cluster", "Perform <command> on each segment host in the cluster", nil)
+	initCmd    = createCobraCommand("init", "Initialize the PXF server instances on master, standby master, and the segment hosts", &pxf.InitCommand)
+	startCmd   = createCobraCommand("start", "Start the PXF server instances on the segment hosts", &pxf.StartCommand)
```
frankgh: maybe use all?
```suggestion
	startCmd   = createCobraCommand("start", "Start the PXF server instances on all segment hosts", &pxf.StartCommand)
```

cli/go/src/pxf-cli/cmd/cluster.go
```diff
@@ -20,89 +20,41 @@ type ClusterData struct {
...
+	stopCmd    = createCobraCommand("stop", "Stop the PXF server instances on the segment hosts", &pxf.StopCommand)
```
frankgh:
```suggestion
	stopCmd    = createCobraCommand("stop", "Stop the PXF server instances on all segment hosts", &pxf.StopCommand)
```

cli/go/src/pxf-cli/cmd/cluster.go
```diff
@@ -20,89 +20,41 @@ type ClusterData struct {
...
+	statusCmd  = createCobraCommand("status", "Get status of PXF servers on the segment hosts", &pxf.StatusCommand)
```
frankgh:
```suggestion
	statusCmd  = createCobraCommand("status", "Get status of PXF servers on all segment hosts", &pxf.StatusCommand)
```

cli/go/src/pxf-cli/cmd/cluster.go
```diff
@@ -20,89 +20,41 @@ type ClusterData struct {
...
+	syncCmd    = createCobraCommand("sync", "Sync PXF configs from master to standby master and the segment hosts", &pxf.SyncCommand)
```
frankgh:
```suggestion
	syncCmd    = createCobraCommand("sync", "Synchronize PXF configs from master to standby master and all segment hosts", &pxf.SyncCommand)
```

cli/go/src/pxf-cli/cmd/cluster.go
```diff
@@ -20,89 +20,41 @@ type ClusterData struct {
...
+	resetCmd   = createCobraCommand("reset", "Reset PXF (undo initialization) on the segment hosts", &pxf.ResetCommand)
```
frankgh:
```suggestion
	resetCmd   = createCobraCommand("reset", "Reset PXF (undo initialization) on all segment hosts", &pxf.ResetCommand)
```

server/pxf-service/src/scripts/pxf-service
```diff
@@ -306,6 +306,7 @@ doHelp() {
 	  restart             restart the local PXF server instance (not supported for cluster)
 	  status              show the status of the local PXF server instance (not supported for cluster)
 	  version             show the version of PXF server
+	  reset               undo the local PXF initialization
 	  cluster <command>   perform <command> on all the hosts in the cluster; try ${bold}pxf cluster help$normal
 
 	  sync <hostname>     sync \$PXF_CONF/{conf,lib,servers} directories onto <hostname>
```
frankgh:
```suggestion
	  sync <hostname>     synchronize \$PXF_CONF/{conf,lib,servers} directories onto <hostname>
```

server/pxf-service/src/scripts/pxf-service
```diff
@@ -317,6 +318,28 @@ doHelp() {
...
+# doReset handles the reset command
+function doReset()
+{
+    if doStatus >/dev/null; then
+        echo "PXF is running. Please stop PXF before running 'pxf reset'"
+        return 1
+    fi
+
+    echo "Cleaning ${PXF_HOME}/conf/pxf-private.classpath..."
+    rm -f "${PXF_HOME}/conf/pxf-private.classpath"
```
frankgh: I'm wondering how errors get handled, (i.e fails to rm directory, does the error get propagated all the way to master if someone runs pxf cluster reset and a single segment host fails to rm a directory/file?)

oliverralbertini: The `-f` flag gets us past errors that are due to the file not being present.

But if the file is owned by root, this can be problematic, unless it's in a directory owned by `gpadmin`. I think we're OK here since all the files we are removing should be under `$GPHOME/pxf` which is owned by `gpadmin`.

server/pxf-service/src/scripts/pxf-service
oliverralbertini: added a warning to `pxf cluster reset` and `pxf reset`.

cli/go/src/pxf-cli/pxf/pxf.go
```diff
@@ -1,9 +1,12 @@
 package pxf
 
 import (
+	"bufio"
```
frankgh: will this bring new OSL dependencies?

oliverralbertini: I don't think so, this is just part of the Go Programming Language. Here's the [doc](https://golang.org/pkg/bufio/).

server/pxf-service/src/scripts/pxf-service
```diff
@@ -317,6 +318,42 @@ doHelp() {
...
+}
+
+# doReset handles the reset command
+function doReset()
+{
+    local force=$2
+    local prompt='Ensure your local PXF instance is stopped before continuing. '
+    prompt+='This is a destructive action. Press y to continue:'
+    if [[ $force != -f && $force != --force ]] && ! promptUser "$prompt"; then
+        echo 'User did not confirm reset action'
```
frankgh: maybe just say `pxf reset cancelled`?

cli/go/src/pxf-cli/pxf/pxf.go
oliverralbertini: Also using the [io](https://golang.org/pkg/io/) and [strings](https://golang.org/pkg/strings/) packages.

PR #190, opened 2019-06-20
fix flakey build issue for connection pool test

raymondyin:
the issue seems disappeared after the change, tried to run on pipeline for 10 times and results are all green.


frankgh:
This seems to be still flaky: https://hush-house.pivotal.io/builds/407535

frankgh:
@raymondyin Please take a look at https://hush-house.pivotal.io/builds/407535. There seems to be flakyness in the test still.

raymondyin:
The flaky error from https://hush-house.pivotal.io/builds/407535 is no longer there for new builds. Also, the error appeared to be caused by failure of a different unit test, which no longer fails.

PR #191, opened 2019-06-25
Support config option to specify the server configuration directory

frankgh:
This is motivated by the following:
Access to Hadoop clusters is nuanced, because with a single set of configurations
users are able to access HDFS, Hive, HBase and other services.

Suppose an enterprise user has a Hortonworks hadoop installation that includes
HDFS, Hive, and HBase. We would configure one server per technology we access,
for example:

     CREATE SERVER hdfs_hdp
          FOREIGN DATA WRAPPER hdfs_pxf_fdw
          OPTIONS ( config 'hdp_1' );

     CREATE SERVER hive_hdp
          FOREIGN DATA WRAPPER hive_pxf_fdw
          OPTIONS ( config 'hdp_1' );

     CREATE SERVER hbase_hdp
          FOREIGN DATA WRAPPER hbase_pxf_fdw
          OPTIONS ( config 'hdp_1' );

To reduce the amount of configuration required for each server, we introduce a
new option `config`. This new option provides the name of the server directory where
the configuration files reside. In the example above, configuration files live
in the `$PXF_CONF/servers/hdp_1` directory, and all three servers share the same
configuration directory.

The `config` option is intended for Hadoop based configurations, but we do
not intend to restrict usage by other connectors.

### DIFF
--- a/server/pxf-api/src/main/java/org/greenplum/pxf/api/model/BasePlugin.java
+++ b/server/pxf-api/src/main/java/org/greenplum/pxf/api/model/BasePlugin.java
@@ -19,6 +19,7 @@ package org.greenplum.pxf.api.model;
  * under the License.
  */
 
+import org.apache.commons.lang.StringUtils;
 import org.apache.hadoop.conf.Configuration;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -51,8 +52,13 @@ public class BasePlugin implements Plugin {
     @Override
     public void initialize(RequestContext requestContext) {
         this.context = requestContext;
+
+        String serverName = !StringUtils.isBlank(context.getConfig()) ?
+                context.getConfig() :
+                context.getServerName();
+
         this.configuration = configurationFactory.
-                initConfiguration(context.getServerName(), context.getUser(), context.getAdditionalConfigProps());
+                initConfiguration(serverName, context.getUser(), context.getAdditionalConfigProps());
         this.initialized = true;
     }
 
--- a/server/pxf-api/src/main/java/org/greenplum/pxf/api/model/BasePlugin.java
+++ b/server/pxf-api/src/main/java/org/greenplum/pxf/api/model/BasePlugin.java
@@ -52,13 +52,8 @@ public class BasePlugin implements Plugin {
     @Override
     public void initialize(RequestContext requestContext) {
         this.context = requestContext;
-
-        String serverName = !StringUtils.isBlank(context.getConfig()) ?
-                context.getConfig() :
-                context.getServerName();
-
         this.configuration = configurationFactory.
-                initConfiguration(serverName, context.getUser(), context.getAdditionalConfigProps());
+                initConfiguration(context.getConfig(), context.getServerName(), context.getUser(), context.getAdditionalConfigProps());
         this.initialized = true;
     }
 
--- a/server/pxf-api/src/main/java/org/greenplum/pxf/api/model/RequestContext.java
+++ b/server/pxf-api/src/main/java/org/greenplum/pxf/api/model/RequestContext.java
@@ -609,7 +609,7 @@ public class RequestContext {
     public void setServerName(String serverName) {
         if (StringUtils.isNotBlank(serverName)) {
 
-            if (!Utilities.isValidDirectoryName(serverName)) {
+            if (!Utilities.isValidRestrictedDirectoryName(serverName)) {
                 throw new IllegalArgumentException(String.format("Invalid server name '%s'", serverName));
             }
 
@@ -737,10 +737,6 @@ public class RequestContext {
         // accessor and resolver are user properties, might be missing if profile is not set
         ensureNotNull("ACCESSOR", accessor);
         ensureNotNull("RESOLVER", resolver);
-
-        if (StringUtils.isNotBlank(config) && !Utilities.isValidDirectoryName(config)) {
-            fail("invalid CONFIG directory name '%s'", config);
-        }
     }
 
     private void ensureNotNull(String property, Object value) {
--- a/server/pxf-api/src/main/java/org/greenplum/pxf/api/model/RequestContext.java
+++ b/server/pxf-api/src/main/java/org/greenplum/pxf/api/model/RequestContext.java
@@ -43,6 +43,7 @@ public class RequestContext {
     // ----- NAMED PROPERTIES -----
     private String accessor;
     private EnumAggregationType aggType;
+    private String config;
     private int dataFragment = -1; /* should be deprecated */
     private String dataSource;
     private String fragmenter;
@@ -346,6 +347,27 @@ public class RequestContext {
      */
     public ColumnDescriptor getColumn(int index) {
         return tupleDescription.get(index);
+    }
+
+    /**
+     * Returns the name of the server configuration for this request.
+     *
+     * @return (optional) the name of the server configuration for this request
+     */
+    public String getConfig() {
+        return config;
+    }
+
+    /**
+     * Sets the name of the server configuration for this request.
+     *
+     * @param config the directory name for the configuration
+     */
+    public void setConfig(String config) {
+        if (StringUtils.isNotBlank(config) && !Utilities.isValidDirectoryName(config)) {
+            fail("invalid CONFIG directory name '%s'", config);
+        }
+        this.config = config;
     }
 
     /**
@@ -715,6 +737,10 @@ public class RequestContext {
         // accessor and resolver are user properties, might be missing if profile is not set
         ensureNotNull("ACCESSOR", accessor);
         ensureNotNull("RESOLVER", resolver);
+
+        if (StringUtils.isNotBlank(config) && !Utilities.isValidDirectoryName(config)) {
+            fail("invalid CONFIG directory name '%s'", config);
+        }
     }
 
     private void ensureNotNull(String property, Object value) {
--- a/server/pxf-api/src/test/java/org/greenplum/pxf/api/model/BaseConfigurationFactoryTest.java
+++ b/server/pxf-api/src/test/java/org/greenplum/pxf/api/model/BaseConfigurationFactoryTest.java
@@ -54,12 +54,27 @@ public class BaseConfigurationFactoryTest {
         factory = new BaseConfigurationFactory(mockServersDirectory);
         when(mockServersDirectory.listFiles(any(FileFilter.class))).thenReturn(new File[]{new File("a"), new File("b")});
 
-        factory.initConfiguration("dummy", "dummy", null);
+        factory.initConfiguration("dummy", "dummy", "dummy", null);
     }
 
     @Test
     public void testConfigurationsLoadedFromMultipleFilesForDefaultServer() {
-        Configuration configuration = factory.initConfiguration("default", "dummy", null);
+        Configuration configuration = factory.initConfiguration("default", "default", "dummy", null);
+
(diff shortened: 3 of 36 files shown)


server/pxf-service/src/main/java/org/greenplum/pxf/service/HttpRequestParser.java
```diff
@@ -81,6 +81,10 @@ public RequestContext parseRequest(HttpHeaders request) {
         context.setAccessor(params.removeUserProperty("ACCESSOR"));
         context.setAggType(EnumAggregationType.getAggregationType(params.removeOptionalProperty("AGG-TYPE")));
 
+        // An optional CONFIG value specifies the name of the server
+        // configuration
+        context.setConfig(params.removeUserProperty("CONFIG"));
```
denalex: I would do it here -- if the CONFIG option is provided, set context.setServerName to that value, otherwise set it to the value of SERVER option

server/pxf-api/src/main/java/org/greenplum/pxf/api/model/BasePlugin.java
```diff
@@ -51,8 +52,13 @@ public BasePlugin() {
     @Override
     public void initialize(RequestContext requestContext) {
         this.context = requestContext;
+
+        String serverName = !StringUtils.isBlank(context.getConfig()) ?
+                context.getConfig() :
+                context.getServerName();
```
denalex: this logic belongs in RequestParser, IMO

denalex: on the other hand, somewhere in the code, where we cache or report errors we mention server name, so if we were to preserve serverName as the value of SERVER option or server name from FDW, then we need to change more code to make directories work off CONFIG value and default config value to serverName value if it is not provided explicitly.

server/pxf-api/src/main/java/org/greenplum/pxf/api/model/RequestContext.java
```diff
@@ -329,6 +335,27 @@ public ColumnDescriptor getColumn(int index) {
...
+    }
+
+    /**
+     * Sets the name of the server configuration for this request.
+     *
+     * @param config the directory name for the configuration
+     */
+    public void setConfig(String config) {
+        if (StringUtils.isNotBlank(config) && !Utilities.isValidDirectoryName(config)) {
+            fail("invalid CONFIG directory name '%s'", config);
```
denalex: does it check directory exists and readable ? If yes, maybe we can say `"No valid directory found for option CONFIG = %s"`

server/pxf-api/src/test/java/org/greenplum/pxf/api/model/RequestContextTest.java
```diff
@@ -184,4 +184,26 @@ public void testReturnsUnmodifiableOptionsMap() {
...
+        context.setConfig("foobar");
+        assertEquals("foobar", context.getConfig());
+    }
+
+    @Test
+    public void testFailsWhenConfigOptionIsAnInvalidDirectoryName() {
+        thrown.expect(IllegalArgumentException.class);
+        thrown.expectMessage("invalid CONFIG directory name '../../malicious'");
+
+        context.setConfig("../../malicious");
```
denalex: should we be concerned, in general, by users arbitrarily pointing their server config to some other pre-configured server to which they don't have permissions on FDW level ? That way they can circumvent any access controls on GPDB side, which is not acceptable. The issue already exists in the current implementation with SERVER option, btw, I just have not yet thought about it in this way. the only protection in this case is user impersonation.

server/pxf-service/src/main/java/org/greenplum/pxf/service/HttpRequestParser.java
frankgh: makes sense

server/pxf-api/src/test/java/org/greenplum/pxf/api/model/RequestContextTest.java
frankgh: We also run the server through validDirectoryName to check for these characters in the server name https://github.com/greenplum-db/pxf/blob/master/server/pxf-api/src/main/java/org/greenplum/pxf/api/model/RequestContext.java#L553
We can rethink about this validation for FDW

server/pxf-api/src/main/java/org/greenplum/pxf/api/model/RequestContext.java
frankgh: it doesn't check if directory exists or if readable, it's just the same validation as here: https://github.com/greenplum-db/pxf/blob/master/server/pxf-api/src/main/java/org/greenplum/pxf/api/model/RequestContext.java#L553

server/pxf-api/src/test/java/org/greenplum/pxf/api/model/RequestContextTest.java
shivzone: I would assume `CREATE SERVER` is something which adminds use while users can have access to table level DDL. Definitely something worth discussing/documenting

frankgh: good point, I think we can defer this decision to later. I will add a TODO to remind us if we want to allow relative paths for the `config` parameter/option

server/pxf-api/src/main/java/org/greenplum/pxf/api/model/RequestContext.java
denalex: So, would we need to remove / enhance that logic from setServerName() since the server name can now be a non-valid directory name if the proper CONFIG value is provided ?

server/pxf-api/src/main/java/org/greenplum/pxf/api/model/RequestContext.java
```diff
@@ -678,6 +705,10 @@ public void validate() {
         // accessor and resolver are user properties, might be missing if profile is not set
         ensureNotNull("ACCESSOR", accessor);
         ensureNotNull("RESOLVER", resolver);
+
+        if (StringUtils.isNotBlank(config) && !Utilities.isValidDirectoryName(config)) {
+            fail("invalid CONFIG directory name '%s'", config);
+        }
```
denalex: why is it here also in addition to the same logic in setConfig() ?

frankgh: I meant to move this one, instead of copying it. I will remove it.

server/pxf-api/src/test/java/org/greenplum/pxf/api/model/RequestContextTest.java
frankgh: While writing the TODO I realized that we should not really be concerned about what users set in the config path. However, my concern is that in the external table framework, we could have a malicious entity using DDL to access a config that he is not allowed to access, for example:
`LOCATION ('pxf://unauthorized_server?PROFILE=hdfs:text&SERVER=foo&CONFIG=/etc/hadoop/conf')`

So for now, is it okay to leave the restriction until we figure out what we'd like to do in the case above? @denalex @shivzone

server/pxf-api/src/main/java/org/greenplum/pxf/api/model/RequestContext.java
frankgh: good question. I would prefer to leave the server name set to some alphanumeric string. What are your thoughts on that @denalex ?

denalex: yes, ok to leave server name validation. Should we implement TODO listed here before committing the PR ?

frankgh: yes, I implemented it.

server/pxf-api/src/test/java/org/greenplum/pxf/api/model/RequestContextTest.java
```diff
@@ -184,4 +184,25 @@ public void testReturnsUnmodifiableOptionsMap() {
...
+
+    @Test
+    public void testConfigOptionIsSetWhenProvided() {
+        context.setConfig("foobar");
+        assertEquals("foobar", context.getConfig());
+    }
+
+    @Test
+    public void testSucceedsWhenConfigOptionIsARelativeDirectoryName() {
+        context.setConfig("../../relative");
```
raymondyin: Do we need any assertions for this and the following two test cases?

server/pxf-api/src/test/java/org/greenplum/pxf/api/model/BaseConfigurationFactoryTest.java
```diff
@@ -54,12 +54,27 @@ public void testInitConfigurationFailsWhenMultipleDirectoriesWithSameName() {
...
+        factory.initConfiguration("dummy", "dummy", "dummy", null);
     }
 
     @Test
     public void testConfigurationsLoadedFromMultipleFilesForDefaultServer() {
-        Configuration configuration = factory.initConfiguration("default", "dummy", null);
+        Configuration configuration = factory.initConfiguration("default", "default", "dummy", null);
+
+        assertEquals("blue", configuration.get("test.blue"));
+        assertEquals("red", configuration.get("test.red"));
```
raymondyin: Should "test.blue" and "test.red" also return null because the file name does not end in -site.xml either?

server/pxf-api/src/test/java/org/greenplum/pxf/api/utilities/UtilitiesTest.java
```diff
@@ -245,23 +240,35 @@ public void testToCsvTextEscapesQuoteChar() {
...
+        assertFalse(Utilities.isValidRestrictedDirectoryName("abc ac"));
+        assertFalse(Utilities.isValidRestrictedDirectoryName("abc;ac"));
+        assertFalse(Utilities.isValidRestrictedDirectoryName("\\"));
+        assertFalse(Utilities.isValidRestrictedDirectoryName("a,b"));
+    }
+
+    @Test
+    public void validRestrictedDirectoryName() {
+        assertTrue(Utilities.isValidRestrictedDirectoryName("pxf"));
+        assertTrue(Utilities.isValidRestrictedDirectoryName("\uD83D\uDE0A"));
```
raymondyin: For "\uD83D\uDE0A", aren't these emojis? Should we allow them for directory name?

server/pxf-api/src/test/java/org/greenplum/pxf/api/model/RequestContextTest.java
frankgh: yes we do, thanks!

server/pxf-api/src/test/java/org/greenplum/pxf/api/utilities/UtilitiesTest.java
frankgh: we allow whatever the underlying OS allows, so yes, emojis are valid directory names :)

server/pxf-api/src/test/java/org/greenplum/pxf/api/model/BaseConfigurationFactoryTest.java
frankgh: we have resource files, and we only load resources that end in `-site.xml`. For this test, we are skipping the property `test.green` which is defined in a resource file that does not end in `-site.xml`

PR #192, opened 2019-06-25
JSON: fix type mismatch bug

oliverralbertini:
Previously, when a user created an external JSON table containing
records such as `{ "key": "(" }`, where the value was cast to a number
such as SMALLINT in the database, we sent along zero as the number.  If the
value were unquoted: `{ "key": ( }`, we sent an empty row.

This commit corrects that behavior by detecting if the value is actually
a numeric type (and also Boolean) and if it isn't, sends it to the
database as type TEXT.  The database can then raise a type mismatch
error, which can be suppressed for unclean data using the REJECT LIMIT
option.


automation/src/test/java/org/greenplum/pxf/automation/features/json/JsonTest.java
```diff
@@ -11,198 +11,271 @@
...
+				"type_bigint bigint",
+				"type_smallint smallint",
+				"type_float real",
+				"type_double float8",
+				"type_string1 text",
+				"type_string2 varchar",
+				"type_string3 bpchar",
+				"type_char char",
+				"type_boolean bool",
+				// "type_bytes bytea",
```
frankgh: any reason this is commented out?

automation/src/test/java/org/greenplum/pxf/automation/features/json/JsonTest.java
```diff
@@ -11,198 +11,271 @@
...
```
frankgh: same here

automation/tincrepo/main/pxf/features/hdfs/readable/json/mismatched_types_record_with_reject_limit/expected/query01.ans
```diff
@@ -0,0 +1,9 @@
+-- @description query01 for PXF HDFS Readable Json supported primitive types test cases
+
+SELECT * from jsontest_mismatched_types_record_with_reject_limit ORDER BY type_int;
+psql:/Users/oalbertini/workspace/pxf/automation/tincrepo/main/pxf/features/hdfs/readable/json/mismatched_types_record_with_reject_limit/output/query01.sql:7: NOTICE:  Found 6 data formatting errors (6 or more input rows). Rejected related input data.
```
frankgh: you can remove the psql part safely
```suggestion
NOTICE:  Found 6 data formatting errors (6 or more input rows). Rejected related input data.
```

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonResolver.java
```diff
@@ -80,6 +80,7 @@ public void initialize(RequestContext requestContext) {
 		JsonNode root = decodeLineToJsonNode(jsonRecordAsText);
 
 		if (root == null) {
+			// TODO: error out here as well
```
frankgh: do we still want to error out when decoding to a JSON node fails?

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonResolver.java
```diff
@@ -203,45 +197,66 @@ private void addFieldFromJsonNode(DataType type, JsonNode val) throws IOExceptio
...
+					break;
+				default:
+					throw new IOException("Unsupported type " + type);
 			}
 		}
 
 		oneFieldList.add(oneField);
 	}
 
+	private static void validateNumber(JsonNode val, OneField oneField) {
```
frankgh: let's add javadocs to this method and describe what it does

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonResolver.java
```diff
@@ -203,45 +197,66 @@ private void addFieldFromJsonNode(DataType type, JsonNode val) throws IOExceptio
...
+		if (val.asLong() == 0 && val.asLong(1) == 1) {
+			oneField.val = val.asText();
+			oneField.type = DataType.TEXT.getOID();
+		}
+	}
+
+	private static void validateBoolean(String value, OneField oneField) {
```
frankgh: same here

automation/src/test/java/org/greenplum/pxf/automation/features/json/JsonTest.java
oliverralbertini: vestige of the supported types test above

automation/tincrepo/main/pxf/features/hdfs/readable/json/mismatched_types_record_with_reject_limit/expected/query01.ans
oliverralbertini: oh yeah. thanks

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonResolver.java
oliverralbertini: that's a great question. maybe sending an empty row is OK. We should discuss that.

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonResolver.java
raymondyin: Yea I agree. The trick to validate the val is very smart but probably not intuitive by nature :D

denalex: why not pass a boolean value here instead of string ?

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonResolver.java
```diff
@@ -203,45 +197,66 @@ private void addFieldFromJsonNode(DataType type, JsonNode val) throws IOExceptio
...
-				break;
-			case BPCHAR:
-			case TEXT:
-			case VARCHAR:
-				oneField.val = val.asText();
-				break;
-			default:
-				throw new IOException("Unsupported type " + type);
+				case BIGINT:
+					oneField.val = val.asLong();
```
denalex: wouldn't this still return 0 if the actual character was non-numeric ?

server/pxf-service/src/main/java/org/greenplum/pxf/service/bridge/ReadBridge.java
```diff
@@ -126,8 +126,6 @@ public Writable getNext() throws Exception {
                 LOG.debug(ex.toString() + ": " + row_info);
             }
             output = outputBuilder.getErrorOutput(ex);
-        } catch (Exception ex) {
-            throw ex;
```
denalex: this is a big change, why was it removed ?

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonResolver.java
```diff
@@ -203,45 +197,66 @@ private void addFieldFromJsonNode(DataType type, JsonNode val) throws IOExceptio
...
+				default:
+					throw new IOException("Unsupported type " + type);
 			}
 		}
 
 		oneFieldList.add(oneField);
 	}
 
+	private static void validateNumber(JsonNode val, OneField oneField) {
+		if (val.asLong() == 0 && val.asLong(1) == 1) {
```
raymondyin: It'll be great to add comment for this if condition. The trick to validate the val is very smart but probably not intuitive by nature .

server/pxf-service/src/main/java/org/greenplum/pxf/service/bridge/ReadBridge.java
frankgh: I think I removed this one when we were debugging. Do we need to catch and rethrow here?

frankgh: if so, a comment would be useful

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonResolver.java
oliverralbertini: I made a javaDoc which explains it, but basically the `asBoolean()` method from `TextNode` (Jackson) doesn't do the right thing in the version we have, but it has been fixed by now: https://github.com/FasterXML/jackson-databind/blob/a6667bcd1f0114f5b004d5203f4139b1e6104a1c/src/main/java/com/fasterxml/jackson/databind/node/TextNode.java#L102

oliverralbertini: Here's the issue that led to that change, 5 years ago! https://github.com/FasterXML/jackson-databind/issues/496

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonResolver.java
oliverralbertini: Yes, but the `validateNumber()` method should take care of the case when it is not a number.

automation/src/test/java/org/greenplum/pxf/automation/features/json/JsonTest.java
oliverralbertini: getting rid of it

automation/src/test/java/org/greenplum/pxf/automation/features/json/JsonTest.java
```diff
@@ -11,198 +11,291 @@
...
+		exTable.setSegmentRejectLimit(2);
+
+		gpdb.createTableAndVerify(exTable);
+
+		// Verify results
+		runTincTest("pxf.features.hdfs.readable.json.malformed_record_with_reject_limit.runTest");
+	}
+
+	/**
+	 * Test JSON file with all supported types. Some of the records
```
frankgh: is this test still valid? are we still sending type mismatched entries?

automation/src/test/java/org/greenplum/pxf/automation/features/json/JsonTest.java
```diff
@@ -11,198 +11,291 @@
...
+		runTincTest("pxf.features.hdfs.readable.json.malformed_record.runTest");
+	}
+
+	/**
+	 * Test JSON file with pretty print format with reject limit configured. One of the records
+	 * is malformed. The query is allowed and a table is created.
+	 *
+	 * @throws Exception if test fails to run
+	 */
+	@Test(groups = {"features", "gpdb"})
```
frankgh: should we test this in the cloud as well?
```suggestion
	@Test(groups = {"features", "gpdb", "hcfs"})
```

automation/src/test/java/org/greenplum/pxf/automation/features/json/JsonTest.java
```diff
@@ -11,198 +11,291 @@
...
+	public void malFormatedRecordWithRejectLimit() throws Exception {
+
+		exTable.setName("jsontest_malformed_record_with_reject_limit");
+		exTable.setProfile("Json");
```
frankgh: if we want to test on the clouds
```suggestion
		exTable.setProfile(ProtocolUtils.getProtocol().value() + ":json");
```

automation/src/test/java/org/greenplum/pxf/automation/features/json/JsonTest.java
```diff
@@ -11,198 +11,291 @@
...
+	 * Test JSON file with all supported types. Some of the records
+	 * have type mismatches (e.g. an integer entered as '(').
+	 * In that case, the line will be sent to GPDB as TEXT, and we
+	 * expect GPDB to raise a type error. This table has reject limit
+	 * set high enough to get some data back.
+	 *
+	 * @throws Exception if test fails to run
+	 */
+	@Test(groups = {"features", "gpdb"})
+	public void mismatchedTypesRecordWithRejectLimit() throws Exception {
```
frankgh: here too? do we have mismatched types still?

automation/src/test/java/org/greenplum/pxf/automation/features/json/JsonTest.java
```diff
@@ -11,198 +11,291 @@
...
+		gpdb.createTableAndVerify(exTable);
+
+		// Verify results
+		runTincTest("pxf.features.hdfs.readable.json.mismatched_types.runTest");
+	}
+
+	/**
+	 * Test JSON file with all supported types. Some of the records
+	 * have type mismatches (e.g. an integer entered as '(').
+	 * In that case, the line will be sent to GPDB as TEXT, and we
```
frankgh: are we sending the line as text, or as en error?
```suggestion
	 * In that case, the line will be sent to GPDB as TEXT, and we
```

oliverralbertini: We forgot to update comments

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonResolver.java
oliverralbertini: We are sending an error to GPDB just for the row now (`BadRecordException`).

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonResolver.java
oliverralbertini: Now passing a `JsonNode` type.

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonResolver.java
```diff
@@ -189,59 +183,95 @@ private void addFieldFromJsonArray(DataType type, JsonNode node, int index) thro
...
+				case BIGINT:
+					validateNumber(type, val);
+					oneField.val = val.asLong();
+					break;
+				case BOOLEAN:
+					validateBoolean(val);
+					oneField.val = val.asBoolean();
+					break;
+				case BYTEA:
+					validateNumber(type, val);
```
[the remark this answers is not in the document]
oliverralbertini: Don't know how that go there!

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonResolver.java
```diff
@@ -189,59 +183,95 @@ private void addFieldFromJsonArray(DataType type, JsonNode node, int index) thro
...
+				case REAL:
+					validateNumber(type, val);
+					oneField.val = (float) val.asDouble();
+					break;
+				case INTEGER:
+					validateNumber(type, val);
+					oneField.val = val.asInt();
+					break;
+				case SMALLINT:
+					validateNumber(type, val);
```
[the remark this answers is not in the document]
oliverralbertini: yes

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonResolver.java
```diff
@@ -59,7 +59,7 @@
 	public void initialize(RequestContext requestContext) {
 		super.initialize(requestContext);
 		oneFieldList = new ArrayList<>();
-		mapper = new ObjectMapper(new JsonFactory());
+		mapper = new ObjectMapper();
```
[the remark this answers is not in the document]
oliverralbertini: yes they are already there. We should probably make an effort to purge out all the `codehaus` Jackson stuff.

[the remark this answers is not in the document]
frankgh: let's add a chore in the backlog to purge codehaus dependencies

automation/src/test/java/org/greenplum/pxf/automation/features/json/JsonTest.java
frankgh: we can defer this to a different story

server/pxf-service/src/main/java/org/greenplum/pxf/service/bridge/ReadBridge.java
oliverralbertini: Should we add this back in now?

automation/src/test/java/org/greenplum/pxf/automation/features/json/JsonTest.java
oliverralbertini: Yeah this test still matters because even though we aren't sending TEXT, we still deal with mismatched types, e.g. "(" masquerading as a numeric type.

automation/src/test/java/org/greenplum/pxf/automation/features/json/JsonTest.java
oliverralbertini: did it, thanks!

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/JsonResolverTest.java
```diff
@@ -0,0 +1,212 @@
...
+        List<OneField> fields = assertRow(row, 10);
+        assertField(fields, 0, 100000001, DataType.INTEGER);
+        assertField(fields, 1, 10101010101L, DataType.BIGINT);
+        assertField(fields, 2, (short) 13, DataType.SMALLINT);
+        assertField(fields, 3, (float) 1.1, DataType.REAL);
+        assertField(fields, 4, 1.1, DataType.FLOAT8);
+        assertField(fields, 5, "testing all supported types in JsonResolver 1", DataType.TEXT);
+        assertField(fields, 6, "testing all supported types in JsonResolver 2", DataType.VARCHAR);
+        assertField(fields, 7, "testing all supported types in JsonResolver 3", DataType.BPCHAR);
+        assertField(fields, 8, "z", DataType.BPCHAR);
```
oliverralbertini: We tried passing this in as a `char` but it fails the test. Not sure if it matters that we get back a string from the resolver.

automation/src/test/java/org/greenplum/pxf/automation/features/json/JsonTest.java
```diff
@@ -21,15 +22,29 @@
...
+            "type_bigint bigint",
+            "type_smallint smallint",
+            "type_float real",
+            "type_double float8",
+            "type_string1 text",
+            "type_string2 varchar",
+            "type_string3 bpchar",
+            "type_char char",
+            "type_boolean bool",
+            /* "type_bytes bytea",*/};
```
frankgh: let's remove it if we are not going to use it
```suggestion
            };
```

server/pxf-service/src/main/java/org/greenplum/pxf/service/bridge/ReadBridge.java
frankgh: no

PR #198, opened 2019-07-08
Support serializing a list of OneFields to CSV

frankgh:
When you create an external table that uses GPDBWritable serialization
you have to define it with CUSTOM format, for example

CREATE EXTERNAL TABLE read_pxf_parquet(location text, month text, number_of_orders int, total_sales double precision)
    LOCATION
    ('pxf://data/pxf_examples/pxf_parquet?PROFILE=s3:parquet?SERVER=s3')

For S3 Select support, we read data in CSV format, but would like to
continue supporting Parquet files as CSV, when S3-SELECT mode is AUTO.

CREATE EXTERNAL TABLE read_pxf_parquet(location text, month text, number_of_orders int, total_sales double precision)
    LOCATION
    ('pxf://data/pxf_examples/pxf_parquet?PROFILE=s3:parquet&SERVER=s3&S3-SELECT=AUTO')
    FORMAT 'CSV';

This commit allows GPDBWritable data to be serialized into CSV, when the
wire format is TEXT.

Pending:
- Support GPDBWritable errors
- Tests


frankgh:
@denalex we can support any profile using GPDBWritable to CSV with this approach. Let me know what your thoughts are on this.

frankgh:
this code is intended for FDW. All profiles should take advantage of this. For example, reading Parquet from HDFS.

frankgh:
@lavjain I should clarify that this will help S3 Select as well

server/pxf-service/src/main/java/org/greenplum/pxf/service/BridgeOutputBuilder.java
```diff
@@ -261,26 +262,33 @@ boolean isTypeInSchema(int recType, int schemaType) {
...
-                output = new BufferWritable((byte[]) val);
-                outputList.add(output); // TODO break output into lines
+                String textRec = (String) val;
+                output = new Text(textRec + "\n");
+                outputList.add(output);
             }
-        } else { // from QuotedLineBreakAccessor
-            String textRec = (String) val;
+        } else {
+            String textRec = new CSVSerializer().fieldListToCSVString(recFields);
```
denalex: that's pretty smart to intercept List or fields at this level and serialize to CSV. BridgeOutputBuilder is the last untouched core component and needs some refactoring. Until then this is a good idea, IMO. The `if` statement is a bit clunky, though with lines 286-288 and 291-293 almost identical and conditions being checked twice. Too late for me now to propose a better structure, though :)

server/pxf-api/src/main/java/org/greenplum/pxf/api/utilities/CSVSerializer.java
```diff
@@ -0,0 +1,29 @@
...
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+public class CSVSerializer {
+    public static final char QUOTE = '"';
+    public static final String VALUE_OF_NULL = "";
+
+    public String fieldListToCSVString(List<OneField> fields) {
+        return fields.stream()
+                .map(s -> {
```
denalex: can we rename `s` to `field` so the code below reads easier ?

server/pxf-api/src/main/java/org/greenplum/pxf/api/utilities/CSVSerializer.java
```diff
@@ -0,0 +1,29 @@
...
+                    if (s.val == null)
+                        return VALUE_OF_NULL;
+                    else if (s.type == DataType.BYTEA.getOID())
+                        return "\\x" + Hex.encodeHexString((byte[]) s.val);
+                    else if (s.type == DataType.NUMERIC.getOID() || !DataType.isTextForm(s.type))
+                        return Objects.toString(s.val, null);
+                    else
+                        return Utilities.toCsvText((String) s.val, QUOTE, true, true);
```
denalex: this would escape internal quotes, I presume ?

server/pxf-service/src/main/java/org/greenplum/pxf/service/BridgeOutputBuilder.java
```diff
@@ -261,26 +262,33 @@ boolean isTypeInSchema(int recType, int schemaType) {
...
             } else {
-                output = new BufferWritable((byte[]) val);
-                outputList.add(output); // TODO break output into lines
+                String textRec = (String) val;
+                output = new Text(textRec + "\n");
+                outputList.add(output);
             }
-        } else { // from QuotedLineBreakAccessor
-            String textRec = (String) val;
+        } else {
```
denalex: CSVSerializer is function only, no need to create a new object for every row. Either use a dependency injection or inline the method into this class. Also CSV is very customizable by GPDB user and here we are quite opinionated what quotes, null and escape chars are. In the future we might need to take DDL values into account.

server/pxf-api/src/main/java/org/greenplum/pxf/api/utilities/CSVSerializer.java
```diff
@@ -0,0 +1,29 @@
...
+
+import org.apache.commons.codec.binary.Hex;
+import org.greenplum.pxf.api.OneField;
+import org.greenplum.pxf.api.io.DataType;
+
+import java.util.List;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+public class CSVSerializer {
```
denalex: Ideally we probably need to define some kind of `Serializer` interface with `GPDBWritableSerializer` and `CSVSerializer` as implementors, but at this point I would just kill this class and inline the function in the builder (see other comments below) until we refactor this module altogether.

server/pxf-api/src/main/java/org/greenplum/pxf/api/utilities/CSVSerializer.java
frankgh: correct

server/pxf-service/src/main/java/org/greenplum/pxf/service/BridgeOutputBuilder.java
frankgh: with FDW, we will need to look at the delimiter, quote, escape char etc. Maybe create a story to address it, and make sure we add TODOs. Should we even consider sending that information from the C-side to Java in the external table framework so that FDW and ext table are consistent?

server/pxf-api/src/main/java/org/greenplum/pxf/api/utilities/CSVSerializer.java
frankgh: makes sense. WIll inline it for now.

PR #203, opened 2019-07-12
Support reading and writing of timestamp with time zone for Parquet

raymondyin:
implemented support for timestamp with time zone with unit tests and
tested manually for integration


server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
```diff
@@ -124,8 +126,17 @@ private void fillGroup(int index, OneField field, Group group, Type type) throws
                 }
                 group.add(index, Binary.fromReusedByteArray(bytes));
                 break;
-            case INT96:
-                group.add(index, ParquetTypeConverter.getBinaryFromTimestamp((String) field.val));
+            case INT96:  // SQL standard timestamp string value with or without time zone literals: https://www.postgresql.org/docs/9.1/datatype-datetime.html
+                String timestamp = (String) field.val;
+
+                Pattern p = Pattern.compile("[\\+-]\\d{2}(:\\d{2})?$");
```
raymondyin: @frankgh points out that Pattern.compile need to be replaced as each time doing a regex compile for a field is expensive and unnecessary.

raymondyin: refactored the variable to static class variable

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
```diff
@@ -124,8 +128,14 @@ private void fillGroup(int index, OneField field, Group group, Type type) throws
                 }
                 group.add(index, Binary.fromReusedByteArray(bytes));
                 break;
-            case INT96:
-                group.add(index, ParquetTypeConverter.getBinaryFromTimestamp((String) field.val));
+            case INT96:  // SQL standard timestamp string value with or without time zone literals: https://www.postgresql.org/docs/9.1/datatype-datetime.html
+                String timestamp = (String) field.val;
+                if (pattern.matcher(timestamp).find()) {
+                    // timestamp with timezone format
+                    group.add(index, ParquetTypeConverter.getBinaryFromTimestampWithTimeZone(timestamp));
```
frankgh: let's add a note saying that when converting to INT96 we will lose timezone information while preserving the correct value

frankgh:
@denalex @oliverralbertini can you review this PR when you get a chance. This was merged a bit prematurely. Thanks

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
```diff
@@ -47,6 +49,8 @@
     private SimpleGroupFactory groupFactory;
     private ObjectMapper mapper = new ObjectMapper();
 
+    private static final Pattern pattern = Pattern.compile("[\\+-]\\d{2}(:\\d{2})?$");
```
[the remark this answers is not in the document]
oliverralbertini: Also, I am wondering why we need `\` in the `[\\+-]`.

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/ParquetTypeConverterTest.java
```diff
@@ -57,4 +60,54 @@ public void testBinaryWithNanos() {
...
+        // Conversion roundtrip for test input (timestamp)
+        String timestamp = "2019-07-10 21:54:53.523485-04";
+        Binary binary = ParquetTypeConverter.getBinaryFromTimestampWithTimeZone(timestamp);
+        String convertedTimestamp = ParquetTypeConverter.bytesToTimestamp(binary.getBytes());
+
+        assertEquals(expectedTimestampInSystemTimeZone, convertedTimestamp);
+    }
+
+    @Test
+    public void testTimestampWithTimezoneWithMicrosecondsStringConversionRoundTrip1() {
```
oliverralbertini: This could probably go in the previous test. It's OK to have more than one assert in there.

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/ParquetTypeConverterTest.java
```diff
@@ -57,4 +60,54 @@ public void testBinaryWithNanos() {
...
+        String timestamp = "2019-07-11 07:39:47.354795+12:45";
+        Binary binary = ParquetTypeConverter.getBinaryFromTimestampWithTimeZone(timestamp);
+        String convertedTimestamp = ParquetTypeConverter.bytesToTimestamp(binary.getBytes());
+
+        assertEquals(expectedTimestampInSystemTimeZone, convertedTimestamp);
+    }
+
+    // Helper function
+    private String convertUTCToCurrentSystemTimeZone(String expectedUTC, DateTimeFormatter formatter) {
+        // convert expectedUTC string to ZonedDateTime d
```
oliverralbertini: maybe the `d` in the comment was meant to be `zdt`. Same comment on line 107

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/ParquetTypeConverterTest.java
```diff
@@ -57,4 +60,54 @@ public void testBinaryWithNanos() {
...
+        LocalDateTime date = LocalDateTime.parse(expectedUTC, formatter);
+        ZonedDateTime zdt = ZonedDateTime.of(date, ZoneOffset.UTC);
+        // convert d to Current Zone ID
+        ZonedDateTime systemZdt = zdt.withZoneSameInstant(ZoneId.systemDefault());
+        // convert date to string representation
+        return systemZdt.format(GreenplumDateTime.DATETIME_FORMATTER);
```
oliverralbertini: I'm not super clear on this, but why aren't we using `formatter` here?

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/ParquetTypeConverterTest.java
```diff
@@ -57,4 +60,54 @@ public void testBinaryWithNanos() {
...
+        Binary binary = ParquetTypeConverter.getBinaryFromTimestampWithTimeZone(timestamp);
+        String convertedTimestamp = ParquetTypeConverter.bytesToTimestamp(binary.getBytes());
+
+        assertEquals(expectedTimestampInSystemTimeZone, convertedTimestamp);
+    }
+
+    @Test
+    public void testTimestampWithTimezoneWithMicrosecondsStringConversionRoundTrip1() {
+        String expectedTimestampInUTC = "2019-07-10 18:54:47.354795";
+        String expectedTimestampInSystemTimeZone = convertUTCToCurrentSystemTimeZone(expectedTimestampInUTC, GreenplumDateTime.DATETIME_FORMATTER);
```
oliverralbertini: can we have comments on these lines indicating what these values would be in Pacific time? It would make it easier to follow what's happening here.

server/pxf-api/src/main/java/org/greenplum/pxf/api/GreenplumDateTime.java
```diff
@@ -4,11 +4,30 @@
...
+/**
+ * Provides formatters for Greenplum DateTime
+ */
 public class GreenplumDateTime {
 
     public static final String DATETIME_FORMATTER_BASE_PATTERN = "yyyy-MM-dd HH:mm:ss";
+
+    /**
+     * Supports date times with the format yyyy-MM-dd HH:mm:ss and
+     * optional milliseconds
```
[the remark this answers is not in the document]
frankgh: correct

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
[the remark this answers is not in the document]
raymondyin: > Also, I am wondering why we need `\` in the `[\\+-]`.

Because "+" in regex means to match one or more previous character. But in this case we want to match both "+" and "-" sign, so we need one "\" to escape in regex, and another "\" to escape the first "\" in Java string.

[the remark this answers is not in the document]
oliverralbertini: But inside of `[]` is just a group of characters that you are matching. So if you want to match `+` or `-`, it's just `[+-]`, no?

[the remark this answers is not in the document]
raymondyin: Thanks for the suggestions.

server/pxf-jdbc/src/test/java/org/greenplum/pxf/plugins/jdbc/utils/ConnectionManagerTest.java
```diff
@@ -190,7 +190,7 @@ public void testPoolExpirationWithActiveConnections() throws SQLException {
         manager.cleanCache();
 
         // wait for at least 3 iteration of sleeping
```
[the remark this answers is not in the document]
raymondyin: @denalex I'll need Alex's help on this question.

server/pxf-api/src/main/java/org/greenplum/pxf/api/GreenplumDateTime.java
```diff
@@ -4,11 +4,30 @@
 import java.time.format.DateTimeFormatterBuilder;
 import java.time.temporal.ChronoField;
 
+/**
+ * Provides formatters for Greenplum DateTime
+ */
 public class GreenplumDateTime {
 
     public static final String DATETIME_FORMATTER_BASE_PATTERN = "yyyy-MM-dd HH:mm:ss";
```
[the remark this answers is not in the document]
raymondyin: I'm not quite sure, but for this class, seems to me that it is more for utility (as the variables below are used multiple places externally, I think it would be good to have it remain public for other places that requires to build datetime formatters

server/pxf-api/src/main/java/org/greenplum/pxf/api/GreenplumDateTime.java
```diff
@@ -4,11 +4,30 @@
...
+
+    /**
+     * Supports date times with timezone and optional milliseconds
+     */
+    public static final DateTimeFormatter DATETIME_WITH_TIMEZONE_FORMATTER =
+            new DateTimeFormatterBuilder()
+                    .parseCaseInsensitive()
+                    .append(DATETIME_FORMATTER)
+                    // Make the mm optional since Greenplum will only send HH if mm == 00
+                    .appendOptional(new DateTimeFormatterBuilder().appendOffset("+HH:mm", "Z").toFormatter())
```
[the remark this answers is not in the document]
raymondyin: Done

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/ParquetTypeConverterTest.java
raymondyin: The method we use from ZonedDateTime class requires DateTimeFormatter object to give string output.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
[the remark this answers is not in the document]
raymondyin: You're right that's a good point. Tried on regex101.com, and [] indeed works without escape symbol to match "+" literally.

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/ParquetTypeConverterTest.java
oliverralbertini: what I mean is why not use the parameter `formatter` that the user passes to this method?

raymondyin: Good point. Changed. (sorry for some reason I kept looking at GreenplumDateTime when reading your comments)

server/pxf-jdbc/src/test/java/org/greenplum/pxf/plugins/jdbc/utils/ConnectionManagerTest.java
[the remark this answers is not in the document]
raymondyin: Seems to me that we want to get through at least 3 iterations. By increasing the time, it is expected to reach 3 iterations at least. 

The following code expects that:
```
verify(mockMBean, times(3)).getActiveConnections();
        verify(mockDataSource, times(1)).close(); // verify datasource is closed when evicted	        verify(mockDataSource, times(1)).close(); // verify datasource is closed when evicted
```

frankgh:
@raymondyin maybe the new commit should be in its own branch and you can open a PR for the new code

server/pxf-service/src/main/java/org/greenplum/pxf/service/BridgeOutputBuilder.java
```diff
@@ -257,9 +257,15 @@ void fillGPDBWritable(List<OneField> recFields) throws BadRecordException {
      * @return whether data type is string type
      */
     boolean isStringType(DataType type) {
-        return Arrays.asList(DataType.VARCHAR, DataType.BPCHAR, DataType.TEXT,
-                DataType.NUMERIC, DataType.TIMESTAMP, DataType.DATE).contains(
-                type);
+        return Arrays.asList(
```
raymondyin: @lavjain Lav mentioned in the other PR that 
>Arrays.asList should be a static variable that is initialized only once.

PR #204, opened 2019-07-15
CLI: update error message for `pxf init`

oliverralbertini:
If the instance already exists, we should tell users to use `pxf reset`,
not to manually delete `pxf-service/`.

### DIFF
--- a/server/pxf-service/src/scripts/pxf-service
+++ b/server/pxf-service/src/scripts/pxf-service
@@ -336,7 +336,7 @@ function doReset()
     fi
 
     if doStatus >/dev/null; then
-        echo "PXF is running. Please stop PXF before running 'pxf reset'"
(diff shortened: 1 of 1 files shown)


raymondyin:
Not sure if it's okay that I just hit "approve" button - it looks fine to me.

oliverralbertini:
> what happens if you run `pxf cluster init` and then run `pxf cluster init` again. What will the error message be?

Then the error message will be something like: 

```sh
$ pxf cluster init
Initializing PXF on master and 1 other hosts...
ERROR: PXF failed to initialize on 2 out of 2 hosts
mdw ==> Instance already exists. Use 'pxf reset' before attempting to re-initialize PXF
sdw1 ==> Instance already exists. Use 'pxf reset' before attempting to re-initialize PXF
```

PR #205, opened 2019-07-15
Add suggestions from S3 Select PR feedback

frankgh:
- Improve code readability
- Fix comments
- Fix test names


frankgh:
@oliverralbertini we will be touching the code suggested here https://github.com/greenplum-db/pxf/commit/81972f7e9f3a1cce9be89421e68612bc90d8018b, so we would prefer to defer the changes to a new PR

PR #206, opened 2019-07-16
Add S3 Select support for Parquet using S3-SELECT=AUTO

frankgh:
Fix CSV serialization when the input has special characters (NEWLINE,
DELIMITER, QUOTE).


server/pxf-api/src/main/java/org/greenplum/pxf/api/GreenplumCSV.java
```diff
@@ -0,0 +1,20 @@
+package org.greenplum.pxf.api;
+
+/**
+ * Greenplum CSV Default
+ */
+public class GreenplumCSV {
```
denalex: if just for constants, this can be an interface

server/pxf-api/src/main/java/org/greenplum/pxf/api/utilities/Utilities.java
```diff
@@ -56,7 +60,7 @@
     public static String toCsvText(String s,
                                    boolean prependQuoteChar,
                                    boolean appendQuoteChar) {
-        return toCsvText(s, CSV_QUOTE, prependQuoteChar, appendQuoteChar, false);
+        return toCsvText(s, QUOTE, ESCAPE, NEWLINE, DELIMITER, prependQuoteChar, appendQuoteChar, false);
```
denalex: might as well pass in a POJO that encapsulates the details for a given CSV definition, with `GreenpluDefaultCSV` being and singleton POJO for defaults. But also need to limit object creation, so for a given request a single POJO can be created (and stored in the RequestContext) so that each row transformation utilizes the same object.

server/pxf-api/src/main/java/org/greenplum/pxf/api/utilities/Utilities.java
```diff
@@ -73,24 +77,47 @@ public static String toCsvText(String s,
      */
     public static String toCsvText(String s,
```
denalex: I have hard time believing this problem has not been solved already a thousand times. Is there any standard library that does CSV conversion ?

server/pxf-service/src/main/java/org/greenplum/pxf/service/BridgeOutputBuilder.java
```diff
@@ -444,8 +446,15 @@ else if (field.type == DataType.TIMESTAMP.getOID())
...
+                                QUOTE,
+                                QUOTE,
+                                NEWLINE,
+                                DELIMITER,
+                                true,
+                                true,
+                                true);
                 })
-                .collect(Collectors.joining(DELIMITER, "", NEWLINE));
+                .collect(Collectors.joining(DELIMITER + "", "", NEWLINE));
```
denalex: what's the point of adding an empty string to the delimiter ?

server/pxf-api/src/main/java/org/greenplum/pxf/api/utilities/Utilities.java
```diff
@@ -73,24 +77,47 @@ public static String toCsvText(String s,
...
                                    char quoteChar,
+                                   char escapeChar,
```
lavjain: toCSVText function should belong to GreenplumCSV class

lavjain:
The following function (line 261) must be changed to use a static list.

```
boolean isStringType(DataType type)
```

server/pxf-api/src/main/java/org/greenplum/pxf/api/utilities/Utilities.java
lavjain: QUOTE, ESCAPE, NEWLINE, DELIMITER, etc should be properties of GreenplumCSV object

server/pxf-service/src/main/java/org/greenplum/pxf/service/BridgeOutputBuilder.java
frankgh: Let's use String.valueOf(DELIMITER) or define a DELIMITER_STRING
```suggestion
                .collect(Collectors.joining(String.valueOf(DELIMITER), "", NEWLINE));
```

server/pxf-s3/src/main/java/org/greenplum/pxf/plugins/s3/S3ProtocolHandler.java
```diff
@@ -25,14 +25,9 @@
             new SupportMatrixEntry("TEXT", OutputFormat.TEXT, S3Mode.ON),
             new SupportMatrixEntry("CSV", OutputFormat.TEXT, S3Mode.ON),
             new SupportMatrixEntry("JSON", OutputFormat.TEXT, S3Mode.ON),
-            new SupportMatrixEntry("PARQUET", OutputFormat.TEXT, S3Mode.AUTO),
-            new SupportMatrixEntry("TEXT", OutputFormat.TEXT, S3Mode.AUTO),
-            new SupportMatrixEntry("CSV", OutputFormat.TEXT, S3Mode.AUTO),
-            new SupportMatrixEntry("JSON", OutputFormat.TEXT, S3Mode.AUTO),
+            new SupportMatrixEntry("PARQUET", OutputFormat.TEXT, S3Mode.OFF),
```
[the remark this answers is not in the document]
shivzone: Other question, does compression affect s3 select support ?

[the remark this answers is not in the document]
raymondyin: @shivzone not necessarily. We did not support Parquet using TEXT or CSV format, only FORMAT='custom' was supported. 

For your second question, compression should affect s3-select support. We'll write a follow up story for compression when S3-SELECT=AUTO is used

server/pxf-api/src/main/java/org/greenplum/pxf/api/model/GreenplumCSV.java
```diff
@@ -0,0 +1,225 @@
...
+            if (delimiter != null && curr == delimiter) specialChars++;
+            if (newline != null && newline.length() > 0) {
+
+                j = 0;
+
+                // let's say we have input asd\r\nacd
+                // and newline \r\n then we need to
+                // increase the specialChars count by 1
+
+                while (i < length && j < newline.length()
```
oliverralbertini: this would seem to warrant tests for edge cases like newlines at the beginning/end/middle of a string?

server/pxf-api/src/test/java/org/greenplum/pxf/api/model/GreenplumCSVTest.java
```diff
@@ -0,0 +1,243 @@
...
+
+        assertEquals(quoteChar + expected, gpCSV.toCsvText(input, true, false, false));
+        assertEquals(expected, gpCSV.toCsvText(input, false, false, false));
+        assertEquals(expected + quoteChar, gpCSV.toCsvText(input, false, true, false));
+        assertEquals(quoteChar + expected + quoteChar,
+                gpCSV.toCsvText(input, true, true, false));
+    }
+
+    @Test
+    public void testWithValueOfNullValidDefaultInput() {
```
oliverralbertini: It's confusing to name all the tests with `*Input()`, can we use a different word to reflect that these are CSV options? Especially since the text itself is often named `input`.

server/pxf-s3/src/main/java/org/greenplum/pxf/plugins/s3/S3ProtocolHandler.java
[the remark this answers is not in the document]
oliverralbertini: I agree we need the `OFF` mode in the matrix, glad we got rid of the `AUTO` one.

server/pxf-s3/src/main/java/org/greenplum/pxf/plugins/s3/S3ProtocolHandler.java
```diff
@@ -87,19 +81,19 @@ private boolean useS3Select(RequestContext context) {
 
         switch (selectMode) {
             case ON:
-                return formatSupported(outputFormat, format, S3Mode.ON, compressionType, true);
```
oliverralbertini: I'm unclear on why we got rid of `compressionType` here.

server/pxf-s3/src/main/java/org/greenplum/pxf/plugins/s3/S3SelectFilterParser.java
```diff
@@ -142,7 +148,7 @@ private String mapValue(Object val, DataType dataType) {
             case TEXT:
             case VARCHAR:
             case BPCHAR:
-                return Utilities.toCsvText(val.toString(), '\'', true, true, false);
+                return SINGLE_QUOTE_ESCAPE.toCsvText(val.toString(), true, true);
```
oliverralbertini: I know this is just following the previous pattern, but it seems like an abuse of the `GreenplumCSV` class to use it when parsing the filter to prepare a string that represents a SQL query *and* when reading data from S3. I think it should still be used, but maybe we can come up with a better name since this is not strictly CSV. Maybe `GreenplumCsvText`?

server/pxf-s3/src/test/java/org/greenplum/pxf/plugins/s3/S3ProtocolHandlerTest.java
```diff
@@ -28,7 +28,7 @@
 
     private static String[] EXPECTED_RESOLVER_TEXT_ON = {STRING_PASS_RESOLVER, STRING_PASS_RESOLVER, STRING_PASS_RESOLVER, STRING_PASS_RESOLVER, NOT_SUPPORTED};
     private static String[] EXPECTED_RESOLVER_TEXT_AUTO_BENEFIT = {STRING_PASS_RESOLVER, STRING_PASS_RESOLVER, STRING_PASS_RESOLVER, STRING_PASS_RESOLVER, DEFAULT_RESOLVER};
-    private static String[] EXPECTED_RESOLVER_TEXT_AUTO_NO_BENEFIT = {STRING_PASS_RESOLVER, DEFAULT_RESOLVER, DEFAULT_RESOLVER, STRING_PASS_RESOLVER, DEFAULT_RESOLVER};
+    private static String[] EXPECTED_RESOLVER_TEXT_AUTO_NO_BENEFIT = {DEFAULT_RESOLVER, DEFAULT_RESOLVER, DEFAULT_RESOLVER, STRING_PASS_RESOLVER, DEFAULT_RESOLVER};
```
oliverralbertini: @raymondyin and I tried to break down why we want the defaults for Fragmenter, Accessor, Resolver, but couldn't come up with an explanation. Was this a mistake, or does this PR introduce changes that make this necessary?

server/pxf-s3/src/test/java/org/greenplum/pxf/plugins/s3/S3ProtocolHandlerTest.java
```diff
@@ -46,13 +46,18 @@
...
     private static String[] EXPECTED_ACCESSOR_TEXT_OFF = {DEFAULT_ACCESSOR, DEFAULT_ACCESSOR, DEFAULT_ACCESSOR, DEFAULT_ACCESSOR, DEFAULT_ACCESSOR};
 
     private static String[] EXPECTED_ACCESSOR_GPDB_WRITABLE_ON = {NOT_SUPPORTED, NOT_SUPPORTED, NOT_SUPPORTED, NOT_SUPPORTED, NOT_SUPPORTED};
     private static String[] EXPECTED_ACCESSOR_GPDB_WRITABLE_AUTO = {DEFAULT_ACCESSOR, NOT_SUPPORTED, NOT_SUPPORTED, DEFAULT_ACCESSOR, DEFAULT_ACCESSOR};
     private static String[] EXPECTED_ACCESSOR_GPDB_WRITABLE_OFF = {DEFAULT_ACCESSOR, DEFAULT_ACCESSOR, DEFAULT_ACCESSOR, DEFAULT_ACCESSOR, DEFAULT_ACCESSOR};
 
+    private static String[] EXPECTED_ACCESSOR_TEXT_AUTO_NO_BENEFIT_HAS_FORMAT_OPTIONS = {DEFAULT_ACCESSOR, S3_ACCESSOR, S3_ACCESSOR, S3_ACCESSOR, DEFAULT_ACCESSOR};
+    private static String[] EXPECTED_RESOLVER_TEXT_AUTO_NO_BENEFIT_HAS_FORMAT_OPTIONS = {DEFAULT_RESOLVER, STRING_PASS_RESOLVER, STRING_PASS_RESOLVER, STRING_PASS_RESOLVER, DEFAULT_RESOLVER};
+    private static String[] EXPECTED_FRAGMENTER_TEXT_AUTO_NO_BENEFIT_HAS_FORMAT_OPTIONS = {DEFAULT_FRAGMENTER, FILE_FRAGMENTER, FILE_FRAGMENTER, FILE_FRAGMENTER, DEFAULT_FRAGMENTER};
```
oliverralbertini: This test is super hard to understand!

oliverralbertini: Why are we using `S3_ACCESSOR` when there is no benefit to do so when there are format options present?

server/pxf-s3/src/main/java/org/greenplum/pxf/plugins/s3/S3ProtocolHandler.java
frankgh: There's a story in the backlog to add it back. It should be simple as in we need a map where the key is the format and the value is a set of supported formats. Then we would need a check similar to this

https://github.com/greenplum-db/pxf/blob/6fa8208c9215c2a733a0498c91a1bf03077c86e5/server/pxf-s3/src/main/java/org/greenplum/pxf/plugins/s3/S3ProtocolHandler.java#L73-L80

frankgh: maybe it should be part of this story actually

server/pxf-s3/src/main/java/org/greenplum/pxf/plugins/s3/S3SelectFilterParser.java
frankgh: it's ugly, I agree. I feel StringUtils.replace should be fine here

server/pxf-s3/src/test/java/org/greenplum/pxf/plugins/s3/S3ProtocolHandlerTest.java
frankgh: it's because the format options only affect the csv/text format, and not the parquet format

frankgh: the tests are hard to understand. I agree

server/pxf-api/src/test/java/org/greenplum/pxf/api/model/GreenplumCSVTest.java
raymondyin: Test method names changed.

server/pxf-s3/src/main/java/org/greenplum/pxf/plugins/s3/S3ProtocolHandler.java
raymondyin: Added compression type logic

server/pxf-s3/src/main/java/org/greenplum/pxf/plugins/s3/S3SelectFilterParser.java
raymondyin: Updated parsing logic for this case.

server/pxf-s3/src/test/java/org/greenplum/pxf/plugins/s3/S3ProtocolHandlerTest.java
raymondyin: Francisco explained that when there's no benefit, we use default resolver for "parquet", "text", "csv", "avro", and string pass resolver for "json".

server/pxf-s3/src/test/java/org/greenplum/pxf/plugins/s3/S3ProtocolHandlerTest.java
```diff
@@ -380,6 +380,42 @@ public void testSelectAutoUnsupportedFormat() {
...
+        context.addOption(S3SelectAccessor.COMPRESSION_TYPE, "foo");
+        // EXPECTED_RESOLVER_GPDB_WRITABLE_ON is used as its values are desirable as expected value
+        verifyResolvers(context, EXPECTED_RESOLVER_GPDB_WRITABLE_ON);
+    }
+
+    @Test
+    public void testWithSelectOffAndWithUnSupportedCompressionType() {
+        context.addOption("S3-SELECT", "off");
+        context.addOption(S3SelectAccessor.COMPRESSION_TYPE, "foo");
+        // EXPECTED_RESOLVER_GPDB_WRITABLE_ON is used as its values are desirable as expected value
```
oliverralbertini: you mean `EXPECTED_RESOLVER_GPDB_WRITABLE_OFF`?

raymondyin: Good catch! Correct.

Thank you Oliver.

PR #207, opened 2019-07-18
Enable support for PXF server to run with Java-11

denalex:
Changes here allow PXF server to be run with Java 11. 

While it now seems to be fully functional, there are still 2 minor issues:

1. during startup, Tomcat logs errors:

```
Jul 17, 2019 7:10:13 AM org.apache.catalina.startup.ContextConfig processAnnotationsJar
SEVERE: Unable to process Jar entry [module-info.class] from Jar [jar:file:/usr/local/greenplum-db-devel/pxf/lib/shared/istack
-commons-runtime-3.0.8.jar!/] for annotations
org.apache.tomcat.util.bcel.classfile.ClassFormatException: Invalid byte tag in constant pool: 19
    at org.apache.tomcat.util.bcel.classfile.Constant.readConstant(Constant.java:97)
```

2. during shutdown, Tomcat logs errors:

```
Jul 18, 2019 7:34:36 AM org.apache.catalina.loader.WebappClassLoader clearReferencesResourceBundles
SEVERE: Failed to clear ResourceBundle references for web application [/manager]
java.lang.NoSuchFieldException: loaderRef
    at java.base/java.lang.Class.getDeclaredField(Class.java:2412)
    at org.apache.catalina.loader.WebappClassLoader.clearReferencesResourceBundles(WebappClassLoader.java:2798)
```

There are also new dependencies (and new HBase version) we'll be shipping, so need to go through the OSL process.

Also, there is a new job in the pipeline, so all pipelines will need to be re-generated and re-flown to enable the new job.


concourse/scripts/pxf_common.bash
```diff
@@ -313,6 +319,7 @@ function start_pxf_server() {
 
 	echo 'Starting PXF service'
 	su gpadmin -c "./bin/pxf start"
+	ps -aef | grep tomcat
```
frankgh: what's the purpose of this line? just for debugging?

server/gradle.properties
```diff
@@ -20,7 +20,7 @@ license=ASL 2.0
 hadoopVersion=2.9.2
 htraceVersion=4.0.1-incubating
 hiveVersion=1.2.2
-hbaseVersion=1.1.2
+hbaseVersion=1.3.2
```
frankgh: why not use the latest (1.4.10) ? Is there anything specific we need about this version?

server/pxf-service/src/templates/tomcat/conf/server.xml
```diff
@@ -21,8 +21,6 @@ under the License.
 
 <Server port="${base.shutdown.port}"
         shutdown="SHUTDOWN">
-    <Listener className="org.apache.catalina.core.JasperListener"/>
```
frankgh: are these listeners no longer needed?

server/pxf-service/src/templates/tomcat/bin/catalina.sh
```diff
@@ -0,0 +1,568 @@
+#!/bin/sh
```
frankgh: is this a modified version of the default catalina.sh? if so what is the change?

concourse/scripts/pxf_common.bash
denalex: yes, just to verify in the job output that PXF is running with specific JDK version. Used it for debugging but decided to leave it in so the info is available for all runs

server/gradle.properties
denalex: This is the first version that has the issue with JDK 11 fixed. Jumping to 2.x breaks a lot of things, so this is the closest we get to the original version.

server/pxf-service/src/templates/tomcat/conf/server.xml
denalex: Jasper is to do with JSPs of which we have none, and LeakPrevention was logging nasty error on non-Oracle JVM failing to launch a thread, effectively becoming useless, so I got rid of them. It is useful when webapps are re-deployed, but we always shutdown Tomcat, so it serves no real purpose.

server/gradle.properties
frankgh: I would go with the latest 1.x version

server/pxf-service/src/templates/tomcat/bin/catalina.sh
denalex: Yes, it is modified from the standard one that Tomcat is shipping with. The change was removing "endorsed" -D property (no longer supported with Java 11) and adding LOGGING_CONFIG to the stop command, since JDK 11 ships default `logging.properties` in a different location and we should be using the ones we ship anyways.

denalex:
I am leaving on my vacation, @frankgh -- feel free to merge it in or leave till I come back. There are new OSL modules and versions to process here and all pipelines will need to be re-generated and re-flown for the new job to show up.

concourse/pipelines/templates/pxf-tpl.yml
```diff
@@ -39,6 +40,7 @@ groups:
       - compile_gpdb_centos6
 {% endif %}
       - test_pxf_cdh_centos6
+      - test_pxf_hdp_centos7_java11
```
oliverralbertini: We should label the other jobs with `_java8`.

server/pxf-service/src/templates/tomcat/bin/catalina.sh
oliverralbertini: Instead of vendoring our version of `catalina.sh`, can we do some `sed`, etc. to achieve the desired change?

concourse/pipelines/templates/pxf-tpl.yml
```diff
@@ -524,6 +532,46 @@ jobs:
...
+  - aggregate:
+    - get: gpdb_src
+      passed:
+      - compile_pxf
+{% if compile_gpdb %}
+      - compile_gpdb_centos6
+{% endif %}
+    - get: bin_gpdb
+      resource: bin_gpdb_centos6
+{% if compile_gpdb %}
```
oliverralbertini: Why don't we have a `compile_gpdb` option for `*_centos7_java11`?

PR #210, opened 2019-07-20
Concourse: create separate 5x/6x release pipelines

oliverralbertini:
This is so that we can have separate 5x/6x state-checker jobs.  Also,
it's more convenient and less confusing to have separate generated
pipeline files.


oliverralbertini:
@frankgh we talked about doing this a while back: here's the relevant PR for `state-checker`: https://github.com/pivotal/state-checker/pull/1

PR #211, opened 2019-07-23
Preserve error when re-throwing IOException

oliverralbertini:
When the read bridge encounters an error, it catches the exception and
rethrows as an IOException. However, the IOException is not preserving
the causing exception which is not letting us debug errors during bridge
read.

### DIFF
--- a/server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/BridgeResource.java
+++ b/server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/BridgeResource.java
@@ -146,7 +146,6 @@ public class BridgeResource extends BaseResource {
                     // Occurs whenever client (GPDB) decides the end the connection
                     LOG.error("Remote connection closed by GPDB", e);
                 } catch (Exception e) {
-                    LOG.error("Exception thrown when streaming", e);
                     throw new IOException(e.getMessage(), e);
                 } finally {
                     LOG.debug("Stopped streaming fragment {} of resource {}, {} records.", fragment, dataDir, recordCount);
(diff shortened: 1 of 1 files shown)


server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/BridgeResource.java
```diff
@@ -147,7 +147,7 @@ public void write(final OutputStream out) throws IOException,
                     LOG.error("Remote connection closed by GPDB", e);
                 } catch (Exception e) {
                     LOG.error("Exception thrown when streaming", e);
-                    throw new IOException(e.getMessage());
+                    throw new IOException(e.getMessage(), e);
```
shivzone: If we are rethrowing the exception, do we really need to log the exception here as well ?

frankgh: most likely not, we saw a double log entry for the same error. I think we should remove this log entry @oliverralbertini . Just make sure the error gets logged after removing it

oliverralbertini:
Thanks for the approval, @lavjain .

server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/BridgeResource.java
frankgh: this is being logged again at `org.apache.catalina.core.StandardWrapperValve:230`. However the log entry above goes to pxf-service.log and the other log entry goes to `localhost.<yyyy-MM-dd>.log

PR #212, opened 2019-07-31
Use format options for S3 Select

frankgh:
Now that greenplum passes the format options from the external table definition, we no longer need to specify these options in the DDL. Before the DDL for a table accessing S3 would look like this:

```
CREATE EXTERNAL TABLE foo (a int, b text)
LOCATION ('pxf://BUCKET/path?PROFILE=s3:csv&SERVER=s3&S3-SELECT=AUTO&DELIMITER=|')
FORMAT 'CSV';
```

Now, we can specify the delimiter as part of the format options and this will get propagated to PXF. A table definition would now look like this:

```
CREATE EXTERNAL TABLE foo (a int, b text)
LOCATION ('pxf://BUCKET/path?PROFILE=s3:csv&SERVER=s3&S3-SELECT=AUTO')
FORMAT 'CSV' (DELIMITER '|');
```

The only parameter we need to define in the query string is header information, like this:

```
CREATE EXTERNAL TABLE foo (a int, b text)
LOCATION ('pxf://BUCKET/path?PROFILE=s3:csv&SERVER=s3&S3-SELECT=AUTO&FILE_HEADER=[NONE|IGNORE|USE]')
FORMAT 'CSV' (DELIMITER '|');
```


automation/src/test/java/org/greenplum/pxf/automation/features/cloud/S3SelectTest.java
```diff
@@ -145,6 +153,8 @@ private void runTestScenario(String name, String server, String format, String s
         exTable.setProfile("s3:" + format);
         exTable.setServer(serverParam);
 
+        if (delimiter != null)
+            exTable.setDelimiter("|");
```
oliverralbertini: Why do we want to hardcode the delimiter to `|`, should we use `delimiter`?

server/pxf-api/src/main/java/org/greenplum/pxf/api/model/RequestContext.java
```diff
@@ -836,6 +834,6 @@ public String inferFormatName() {
         if (!StringUtils.isBlank(profile) && profile.contains(":")) {
             return profile.split(":")[1];
         }
-        return null;
```
oliverralbertini: I wonder why we were ever returning `null` here.

server/pxf-s3/src/main/java/org/greenplum/pxf/plugins/s3/S3ProtocolHandler.java
```diff
@@ -111,7 +111,7 @@ private boolean useS3Select(RequestContext context) {
                 // if supported for ON and not beneficial -> if supported for OFF -> use OFF, else use ON
                 // if not supported for ON -> if supported for OFF -> use OFF, else ERROR out
                 if (formatSupported(outputFormat, format, S3Mode.ON, false)) {
-                    if (willBenefitFromSelect(context) || hasFormatOptions(format, context)) {
+                    if (willBenefitFromSelect(context)) {
```
oliverralbertini: I'm glad `hasFormatOptions()` is going away.

server/pxf-s3/src/main/java/org/greenplum/pxf/plugins/s3/S3SelectAccessor.java
```diff
@@ -213,7 +212,7 @@ private OutputSerialization getOutputSerialization(RequestContext context) {
     InputSerialization getInputSerialization(RequestContext context) {
         InputSerialization inputSerialization = new InputSerialization();
 
-        String format = context.getFormat();
+        String format = context.inferFormatName();
```
oliverralbertini: My understanding is that we need to infer the format name in case of something like
```SQL
create external table s3_parquet_table (foo int)
  LOCATION ( 'pxf://...?PROFILE:s3:parquet' )
  FORMAT 'CSV';
```
the format on S3 should not be considered as 'CSV'.

server/pxf-s3/src/main/java/org/greenplum/pxf/plugins/s3/S3SelectAccessor.java
```diff
@@ -264,36 +263,29 @@ JSONInput getJSONInput(RequestContext context) {
...
+            LOG.debug("With CSV field delimiter '{}'", csv.getDelimiter());
+            csvInput.setFieldDelimiter(csv.getDelimiter());
         }
 
-        String quoteEscapeCharacter = context.getOption(QUOTE_ESCAPE_CHARACTER);
-        if (quoteEscapeCharacter != null) {
-            LOG.debug("With CSV quote escape character '{}'", quoteEscapeCharacter);
-            csvInput.setQuoteEscapeCharacter(quoteEscapeCharacter);
+        if (csv.getNewline() != null) {
+            LOG.debug("With CSV record delimiter '{}'", csv.getNewline());
```
oliverralbertini: This debug statement is wrong.

server/pxf-s3/src/test/java/org/greenplum/pxf/plugins/s3/S3SelectAccessorTest.java
```diff
@@ -27,11 +27,11 @@ public void testGetInputSerializationDefaults() {
         assertNotNull(inputSerialization.getCsv());
         assertNull(inputSerialization.getCsv().getAllowQuotedRecordDelimiter());
         assertNull(inputSerialization.getCsv().getComments());
-        assertNull(inputSerialization.getCsv().getFieldDelimiter());
         assertNull(inputSerialization.getCsv().getFileHeaderInfo());
-        assertNull(inputSerialization.getCsv().getQuoteCharacter());
-        assertNull(inputSerialization.getCsv().getQuoteEscapeCharacter());
-        assertNull(inputSerialization.getCsv().getRecordDelimiter());
+        assertEquals(',', inputSerialization.getCsv().getFieldDelimiter().charValue());
```
oliverralbertini: So are these going to come from the default request context now?

server/pxf-service/src/main/java/org/greenplum/pxf/service/BridgeOutputBuilder.java
```diff
@@ -123,9 +123,9 @@ public Writable getErrorOutput(Exception ex) throws Exception {
             // We create a row with an extra column containing the error information
             LOG.error(ex.getMessage(), ex);
             return new Text(
-                    StringUtils.repeat(",", context.getTupleDescription().size()) +
+                    StringUtils.repeat(String.valueOf(greenplumCSV.getDelimiter()), context.getTupleDescription().size()) +
                             greenplumCSV.toCsvField(ex.getMessage(), true, true, true) +
-                            greenplumCSV.getDelimiter());
+                            greenplumCSV.getNewline());
```
oliverralbertini: Why do we suddenly need a newline?

automation/src/main/java/org/greenplum/pxf/automation/structures/tables/utils/TableFactory.java
```diff
@@ -76,7 +76,6 @@ public static ReadableExternalTable getPxfHiveRcReadableTable(String tableName,
             exTable.setResolver("org.greenplum.pxf.plugins.hive.HiveColumnarSerdeResolver");
         }
         exTable.setDelimiter("E'\\x01'");
-        exTable.setUserParameters(new String[] { "delimiter=\\x01" });
```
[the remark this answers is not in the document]
frankgh: yeah, the changes in GPDB will impact hive profiles that specify the delimiter in the LOCATION

automation/src/main/java/org/greenplum/pxf/automation/testplugin/ColumnProjectionVerifyAccessor.java
```diff
@@ -23,7 +23,7 @@ public boolean openForRead() throws Exception {
 
         // TODO whitelist the option
         userData = new String(context.getFragmentUserData());
-        userDelimiter = context.getOption("DELIMITER");
+        userDelimiter = String.valueOf(context.getGreenplumCSV().getDelimiter());
```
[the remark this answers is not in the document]
frankgh: A char cannot be null as it is a primitive type. You could have a char with `\0` for null, but it would still be a char. String.valueOf is a way to get a String from a char. Possibly the most optimized way to get it also. Internally it creates a char[] with the char that you passed in the method

automation/src/test/java/org/greenplum/pxf/automation/features/cloud/S3SelectTest.java
frankgh: thanks! i completely missed this one. will fix it

server/pxf-s3/src/main/java/org/greenplum/pxf/plugins/s3/S3ProtocolHandler.java
frankgh: it might have to come back at some point :)

server/pxf-s3/src/main/java/org/greenplum/pxf/plugins/s3/S3SelectAccessor.java
frankgh: correct. I will add a comment clarifying this

server/pxf-s3/src/main/java/org/greenplum/pxf/plugins/s3/S3SelectAccessor.java
frankgh: the debug statement is correct. However, record delimiter is a little misleading. I also get confused with it all the time. record delimiter is the same as newline in greenplum/postgres (not to be confused with field delimiter)

frankgh: will change it to say "With CSV newline"

server/pxf-s3/src/test/java/org/greenplum/pxf/plugins/s3/S3SelectAccessorTest.java
frankgh: correct, we define defaults in GreenplumCSV

server/pxf-service/src/main/java/org/greenplum/pxf/service/BridgeOutputBuilder.java
frankgh: that was a bug :). We need a new line at the end of the string.

server/pxf-api/src/main/java/org/greenplum/pxf/api/model/RequestContext.java
frankgh: hmm, because we would only infer it from the profile name. Maybe it should return null, as the format is not inferred.

server/pxf-s3/src/main/java/org/greenplum/pxf/plugins/s3/S3SelectAccessor.java
oliverralbertini: thanks, that makes sense.

automation/src/test/java/org/greenplum/pxf/automation/features/cloud/S3SelectTest.java
```diff
@@ -71,69 +71,77 @@ protected void afterClass() throws Exception {
...
+            String server,
+            String format,
+            String s3Path,
+            String srcPath,
+            String filename,
+            String delimiter,
+            String[] userParameters)
             throws Exception {
         String tableName = "s3select_" + name;
         String serverParam = (server == null) ? null : "server=" + server;
```
raymondyin: If server is "", and serverParam becomes "server=", is this okay?

raymondyin: If server is "", and serverParam becomes "server=", is this okay?

server/pxf-s3/src/main/java/org/greenplum/pxf/plugins/s3/S3ProtocolHandler.java
```diff
@@ -111,7 +111,7 @@ private boolean useS3Select(RequestContext context) {
                 // if supported for ON and not beneficial -> if supported for OFF -> use OFF, else use ON
```
raymondyin: What do the "->"s mean in these two comment lines?

server/pxf-s3/src/main/java/org/greenplum/pxf/plugins/s3/S3SelectAccessor.java
```diff
@@ -194,11 +188,16 @@ SelectObjectContentRequest generateBaseCSVRequest(RequestContext context) {
      * @return a {@link OutputSerialization} object
      */
     private OutputSerialization getOutputSerialization(RequestContext context) {
+
+        GreenplumCSV csv = context.getGreenplumCSV();
```
raymondyin: Would it also be beneficial to have csvOutput initialized as static variable to avoid constant instantiation?

server/pxf-s3/src/main/java/org/greenplum/pxf/plugins/s3/S3SelectAccessor.java
```diff
@@ -194,11 +188,16 @@ SelectObjectContentRequest generateBaseCSVRequest(RequestContext context) {
...
+
         OutputSerialization outputSerialization = new OutputSerialization();
         CSVOutput csvOutput = new CSVOutput();
-        // TODO: FDW: In FDW world, these should be passed as part of the request
-        csvOutput.setQuoteCharacter("\"");
-        csvOutput.setQuoteEscapeCharacter("\"");
+        csvOutput.setFieldDelimiter(csv.getDelimiter());
```
raymondyin: Would it also be beneficial to have csvOutput initialized as static variable to avoid constant instantiation?

automation/src/test/java/org/greenplum/pxf/automation/features/cloud/S3SelectTest.java
frankgh: that is totally fine, the server will be default at the server side

server/pxf-s3/src/main/java/org/greenplum/pxf/plugins/s3/S3ProtocolHandler.java
frankgh: I think it just means nested ifs

server/pxf-s3/src/main/java/org/greenplum/pxf/plugins/s3/S3SelectAccessor.java
frankgh: well, the CSVOutput object depends on the parameters the user passes. So a static `final` would not be useful in this case

server/pxf-s3/src/main/java/org/greenplum/pxf/plugins/s3/S3ProtocolHandler.java
frankgh: or maybe it means then

oliverralbertini:
yeah it means then

PR #213, opened 2019-08-01
Rename S3-SELECT option -> S3_SELECT

frankgh:
Following recommendations from docs team, we will use underscore for
consistency. Even though PXF still mixes between dash and underscore, we
should make an effort to use underscores only. This will align us
better with FDW as well.

### DIFF
--- a/server/pxf-s3/src/test/java/org/greenplum/pxf/plugins/s3/S3ProtocolHandlerTest.java
+++ b/server/pxf-s3/src/test/java/org/greenplum/pxf/plugins/s3/S3ProtocolHandlerTest.java
@@ -82,7 +82,7 @@ public class S3ProtocolHandlerTest {
 
     @Test
     public void testTextWithSelectOnResolver() {
-        context.addOption("S3-SELECT", "on");
+        context.addOption("S3_SELECT", "on");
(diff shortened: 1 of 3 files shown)


server/pxf-s3/src/test/java/org/greenplum/pxf/plugins/s3/S3ProtocolHandlerTest.java
```diff
@@ -82,7 +82,7 @@ public void before() {
 
     @Test
     public void testTextWithSelectOnResolver() {
-        context.addOption("S3-SELECT", "on");
+        context.addOption("S3_SELECT", "on");
```
raymondyin: Just wondering if it's okay to have the strings like "on", "auto" or "off" to be lower cases.
The cases of string value for s3_select don't matter as we do "StringUtils.upperCase()" in verification logics.

frankgh: yeah, it doesn't matter what whether you use on or ON or oN or On. In fact, I would argue that this is a good test since I'd expect customers to enter any of the variations.

PR #214, opened 2019-08-05
Server: Upgrade com.fasterxml.jackson.core:jackson-databind jar to 2.9.9.2

frankgh:
The jar addresses 2 CVE:
CVE-2019-14379: SubTypeValidator.java in FasterXML jackson-databind
before 2.9.9.2 mishandles default typing when ehcache is used, leading
to remote code execution.

CVE-2019-14439: A Polymorphic Typing issue was discovered in FasterXML
jackson-databind 2.x before 2.9.9.2. This occurs when Default Typing is
enabled (either globally or for a specific property) for an externally
exposed JSON endpoint and the service has the logback jar in the
classpath.


shivzone:
Please update this in pxf-hdfs as well

frankgh:
@shivzone thanks ! done

PR #215, opened 2019-08-09
Docker pipeline: use Dockerfile that triggered build

oliverralbertini:
Previously we were `get`ing the `pxf_src` and Dockerfile resources
separately, and using the Dockerfile from `pxf_src`. This allowed
situations where the Dockerfile being used in the task was not the one
that actually triggered the build.

With the approach used in this commit, the pipeline task always uses the
Dockerfile that triggers it. The downside is that if you want to
re-trigger the docker image manually (to re-cache stuff on the image)
you will either need to temporarily change the pipeline to use the
Dockerfile from `pxf_src`, or push some change to the Dockerfile.

### DIFF
--- a/concourse/pipelines/docker-images.yml
+++ b/concourse/pipelines/docker-images.yml
@@ -176,7 +176,7 @@ resources:
       username: {{docker-username}}
       password: {{docker-password}}
 
-  - name: dockerfile-gpdb-dev-centos6-hdp-secure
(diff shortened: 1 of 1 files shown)


oliverralbertini:
Here's an example of this commit [in action](https://ud.ci.gpdb.pivotal.io/teams/main/pipelines/gpdb_pxf_docker-images/jobs/docker-gpdb-pxf-dev-centos6/builds/94). The build is red because of a flake, but this shows that the resources can be wired up this way.

PR #216, opened 2019-08-09
Run multinode tests on dataproc (with and without Kerberos)

oliverralbertini:
This PR defines a new "security" test group and  enables use of GCP dataproc
Hadoop clusters with and without Kerberos.

Because Terraform and thus CCP don't support Kerberized dataproc, we
have to use our own scripts to create/destroy the Hadoop clusters.
Timing of the destruction of the cluster is controlled by the parameter
in pxf-multinode-params.yml `dataproc-destroy-timeout`.

There is a new KERBEROS (true/false) parameter for installing dataproc,
PXF and testing PXF multinode.


oliverralbertini:
Here's the [testing pipeline](https://ud.ci.gpdb.pivotal.io/teams/main/pipelines/dev:pxf_fguerrero-kerberized-dataproc_5X_STABLE).

concourse/tasks/test_pxf_multinode.yml
```diff
@@ -6,10 +6,16 @@ inputs:
   - name: cluster_env_files
   - name: bin_gpdb
   - name: pxf_tarball
+  - name: dataproc_env_files
+    optional: true
 params:
+  ACCESS_KEY_ID:
   GROUP: feature
+  HIVE_VERSION: 1
```
oliverralbertini: @frankgh Should we be adjusting this parameter?

concourse/pipelines/templates/pxf-tpl.yml
```diff
@@ -808,6 +864,113 @@ jobs:
...
+      - task: intialize_greenplum
+        file: ccp_src/ci/tasks/gpinitsystem.yml
+    - task: Generate Hadoop Cluster
+      file: pxf_src/concourse/tasks/install_dataproc.yml
+      image: ccp-7
+      params:
+        GOOGLE_CREDENTIALS: {{google-service-account-key}}
+        GOOGLE_PROJECT_ID: {{google-project-id}}
+        GOOGLE_ZONE: {{google-zone}}
+        IMAGE_VERSION: "1.3"
```
frankgh: this needs to be parameterized

concourse/tasks/test_pxf_multinode.yml
frankgh: hive_version 1 is the default

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
```diff
@@ -344,7 +344,8 @@ public Connection getConnection() throws SQLException {
             if (e instanceof SQLException) {
                 throw (SQLException) e;
             } else {
-                throw new SQLException(e.getMessage(), e);
+                String msg = e.getMessage();
+                throw new SQLException(msg == null ? e.getCause().getMessage() : msg, e);
```
frankgh: we need to make sure the cause is not null also

concourse/pipelines/templates/pxf-tpl.yml
```diff
@@ -1102,6 +1265,7 @@ jobs:
     - test_pxf_hdp_secure_with_impersonation
     - test_pxf_no_impersonation
     - test_pxf_hdp_multinode_gpdb
+    - test_pxf_multinode_gpdb
```
denalex: should we call this `test_pxf_dataproc_multinode_gpdb` or do we plan to get rid of CCP-based multinode tests, such as `test_pxf_hdp_multinode_gpdb` ?

concourse/scripts/install_hadoop.bash
```diff
@@ -49,7 +49,7 @@ function _main() {
     cp -R cluster_env_files/.ssh/* /root/.ssh
 
     hadoop_ip=$( < cluster_env_files/etc_hostfile grep "edw0" | awk '{print $1}')
-    install_hadoop_single_cluster ${hadoop_ip}
+    install_hadoop_single_cluster "$hadoop_ip"
```
denalex: just wonder, what's the need for all quotes here ?

concourse/scripts/install_pxf.bash
```diff
@@ -73,133 +100,70 @@ function install_java() {
...
+		gpgcheck=1
+		gpgkey=http://public-repo-1.hortonworks.com/HDP/centos7/2.x/updates/2.6.5.0/RPM-GPG-KEY/RPM-GPG-KEY-Jenkins
+		enabled=1
+		priority=1
+	EOF
 	yum install -y -d 1 hadoop-client
 	echo "export HADOOP_VERSION=\${HADOOP_VER}" | sudo tee -a ~gpadmin/.bash_profile
 	echo "export HADOOP_HOME=/usr/hdp/\${HADOOP_VER}" | sudo tee -a ~gpadmin/.bash_profile
 	echo "export HADOOP_HOME=/usr/hdp/\${HADOOP_VER}" | sudo tee -a ~centos/.bash_profile
 }
```
denalex: why is below one no longer needed ?

concourse/scripts/install_hadoop.bash
oliverralbertini: To prevent globbing and word-splitting. IP addresses not likely to have separate words or `*` character, but it's a good pattern to follow.

oliverralbertini: https://github.com/koalaman/shellcheck/wiki/Sc2086

concourse/scripts/install_pxf.bash
oliverralbertini: We get our `*-site.xml` files off of the dataproc cluster.

oliverralbertini: @frankgh is this correct?

concourse/pipelines/templates/pxf-tpl.yml
frankgh: We should keep the `test_pxf_hdp_multinode_gpdb` tests for now. Eventually, we could replace it with dataproc. For now, I suggest we rename it to `test_pxf_secure_multinode_gpdb`

concourse/scripts/install_pxf.bash
frankgh: correct, we just follow the procedure to either copy *-site.xml files from dataproc or copy the templates into the server folder, depending on the configuration

oliverralbertini:
Diff between multinode test and secure multinode test:
```
○ → diff /tmp/{a,b}
3d2
< CloudAccessTest.testCloudAccessWithHdfsFailsWhenServerNoCredsInvalidConfigFileExists...OK
5,6d3
< CloudAccessTest.testCloudAccessWithHdfsOkWhenServerCredsInvalidConfigFileExists...OK
< CloudAccessTest.testCloudAccessWithHdfsOkWhenServerNoCredsValidConfigFileExists...OK
81c78
< HiveOrcTest.storeAsOrcAllTypes...OK
---
> HiveOrcTest.storeAsOrcAllTypesHive1AndHive2...OK
162d158
< MultiServerTest.testTwoServers...OK
170,178d165
< S3SelectTest.testBzip2CsvWithHeadersUsingHeaderInfo...OK
< S3SelectTest.testGzipCsvWithHeadersUsingHeaderInfo...OK
< S3SelectTest.testGzipParquet...OK
< S3SelectTest.testParquet...OK
< S3SelectTest.testPlainCsvWithHeaders...OK
< S3SelectTest.testPlainCsvWithHeadersUsingHeaderInfo...OK
< S3SelectTest.testPlainCsvWithNoHeaders...OK
< S3SelectTest.testSnappyParquet...OK
```

PR #217, opened 2019-08-11
CLI: reset on standby master and don't allow cluster init without PXF_CONF set

oliverralbertini:

### DIFF
--- a/cli/go/src/pxf-cli/cmd/cluster.go
+++ b/cli/go/src/pxf-cli/cmd/cluster.go
@@ -124,7 +124,7 @@ func doSetup(command *pxf.Command) (*ClusterData, error) {
 	}
 	segConfigs := cluster.MustGetSegmentConfiguration(connection)
 	clusterData := &ClusterData{Cluster: cluster.NewCluster(segConfigs), connection: connection}
-	if command.Name() == pxf.Sync || command.Name() == pxf.Init {
+	if command.Name() == pxf.Sync || command.Name() == pxf.Init || command.Name() == pxf.Reset {
 		err = clusterData.appendMasterStandby()
 		if err != nil {
 			return nil, err
--- a/server/pxf-service/src/scripts/pxf-env-default.sh
+++ b/server/pxf-service/src/scripts/pxf-env-default.sh
@@ -21,22 +21,10 @@ PARENT_SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd )"
 #############################################################################
 # The PXF_CONF property is updated by pxf init script, do not edit manually #
 #############################################################################
-export PXF_CONF=${PXF_CONF:=NOT_INITIALIZED}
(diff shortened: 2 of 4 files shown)


server/pxf-service/src/scripts/pxf-env-default.sh
```diff
@@ -21,18 +21,20 @@ PARENT_SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd )"
...
 # The PXF_CONF property is updated by pxf init script, do not edit manually #
 #############################################################################
-export PXF_CONF=${PXF_CONF:=NOT_INITIALIZED}
+export PXF_CONF=${PXF_CONF:-NOT_INITIALIZED}
 #############################################################################
 
 # load user environment first, warn if the script is missing when not in init mode
-if [[ $pxf_script_command != init && $pxf_cluster_command != init ]]; then
+if [[ $pxf_cluster_command == init ]]; then
+	[[ $PXF_CONF == NOT_INITIALIZED ]] && export PXF_CONF=''
```
frankgh: What are we doing here?

cli/go/src/pxf-cli/cmd/cluster.go
```diff
@@ -124,7 +124,7 @@ func doSetup(command *pxf.Command) (*ClusterData, error) {
 	}
 	segConfigs := cluster.MustGetSegmentConfiguration(connection)
 	clusterData := &ClusterData{Cluster: cluster.NewCluster(segConfigs), connection: connection}
-	if command.Name() == pxf.Sync || command.Name() == pxf.Init {
+	if command.Name() == pxf.Sync || command.Name() == pxf.Init || command.Name() == pxf.Reset {
```
frankgh: Should we have a function ShouldCommandRunOnStandby() instead here?

server/pxf-service/src/scripts/pxf-env-default.sh
oliverralbertini: I have simplified that logic, please take another look when you get a chance.

cli/go/src/pxf-cli/cmd/cluster.go
oliverralbertini: This is going to be short-lived, when [this PR](https://github.com/greenplum-db/gp-common-go-libs/pull/44/commits/e125161076c3bcbc2e9778ee8b6dad823dfbda81) gets merged on `gp-common-go-libs`, we should have a more elegant solution.

frankgh: what happens when that PR gets merged? do we fix our dependencies to a specific version? My understanding is that the PR in gp-common-go-libs has breaking changes, so we need to watch out when this PR gets merged

oliverralbertini: When we are ready to update, we should run `dep ensure -update` and that will update the `Gopkg.lock` file. We can do that with the PR that eventually will incorporate those changes from the gp-common-go-libs.

PR #220, opened 2019-08-21
Cleanup dataproc cluster when setup fails

oliverralbertini:
We have to destroy the dataproc cluster when setup_pxf fails, and we
should use a timeout in case we need to debug.


oliverralbertini:
@edespino noticed that we had some dataproc clusters hanging around, and it turns out they came from a pipeline where the `setup_pxf` job had failed. We already clean up when it's test failures.

oliverralbertini:
Should I re-generate 5X,6X release pipelines as part of this PR, @frankgh ?

frankgh:
@oliverralbertini yeah, let's regenerate the pipelines

PR #221, opened 2019-08-21
Consume the correct GPDB resource in pxf_release_6x

frankgh:
Previously, we were consuming GPDB release candidates from S3. Recently,
releng moved away from publishing artifacts to S3 and are only using
Google Storage. This commit fixes the Release Candidate artifact resource
and points to the correct Google Cloud Storage.


concourse/pipelines/templates/pxf-tpl.yml
```diff
@@ -258,27 +258,73 @@ resources:
...
     access_key_id: {{bucket-access-key-id}}
     region_name: {{aws-region}}
     secret_access_key: {{bucket-secret-access-key}}
-{% if compile_gpdb %}
     bucket: gpdb5-concourse-builds-dev
     versioned_file: {{bin_gpdb_centos_versioned_file}}
 {% elif pipeline_type == "release" %}
-    bucket: {{gpdb-stable-builds-bucket-name}}
 {% if gpdb_branch == '5X_STABLE' %}
+- name: bin_gpdb_centos6
```
oliverralbertini: can we move the common code out before the if?

oliverralbertini: I would say the same below...

frankgh: I found it very confusing. I can change it back, but I am afraid it might hurt understandability.

oliverralbertini:
No, that's OK.

concourse/pipelines/templates/pxf-tpl.yml
```diff
@@ -258,27 +258,73 @@ resources:
...
+    json_key: ((concourse-gcs-resources-service-account-key))
+    regexp: server/published/gpdb6/server-rc-(.*)-rhel6_x86_64.tar.gz
 {% endif %}
 {% elif pipeline_type == "pxf" %}
+- name: bin_gpdb_centos6
+  type: s3
+  source:
+    access_key_id: {{bucket-access-key-id}}
+    region_name: {{aws-region}}
+    secret_access_key: {{bucket-secret-access-key}}
```
oliverralbertini: Does this mean that we want to use gpdb5 for pxf pipelines?

concourse/pipelines/templates/pxf-tpl.yml
```diff
@@ -258,27 +258,73 @@ resources:
...
+{% endif %}
+{% elif pipeline_type == "pxf" %}
+- name: bin_gpdb_centos7
+  type: s3
+  source:
+    access_key_id: {{bucket-access-key-id}}
+    region_name: {{aws-region}}
+    secret_access_key: {{bucket-secret-access-key}}
+    bucket: {{icw_green_bucket}}
+    versioned_file: bin_gpdb_centos7/gpdb_branch_((gpdb-branch))/icw_green/bin_gpdb.tar.gz
```
oliverralbertini: Won't this be a problem if we have the master or 6X branch of GPDB?

concourse/pipelines/templates/pxf-tpl.yml
frankgh: hmm, this seems wrong. I will fix it

concourse/pipelines/templates/pxf-tpl.yml
frankgh: yeah. I think ICW green still goes to S3 though. I will double check

concourse/pipelines/templates/pxf-tpl.yml
frankgh: ok, after giving another look at this, it seems that ICW_green is always pushed to S3 .
For 6X, it gets pushed to 
bin_gpdb_centos6/gpdb_branch_6X_STABLE/icw_green/bin_gpdb.tar.gz
For master, it gets pushed to
bin_gpdb_centos6/gpdb_branch_master/icw_green/bin_gpdb.tar.gz

so the condition is correct

concourse/pipelines/templates/pxf-tpl.yml
frankgh: this is fine too

concourse/pipelines/5x-release_pipeline-generated.yml
```diff
@@ -351,7 +360,7 @@ jobs:
       passed:
       - compile_pxf
     - get: bin_gpdb
-      resource: bin_gpdb_centos6
+      resource: bin_gpdb_centos7
```
denalex: here my intention was to test on Centos7 the artifact that was build on Centos6, as this is how we ship it, no ? This change will use GPDB artifact build on centos7, which is generally correct, but is not what the customers will do, unless I am mistaken.

frankgh: we ship multiple versions of the artifact, one for centos 6 and a different one for centos 7. The centos 7 artifact is compiled and tested against 7. Maybe your statement is true for 5x?

PR #222, opened 2019-08-23
Fix cloud access tests

oliverralbertini:


oliverralbertini:
Here's a [green build](https://ud.ci.gpdb.pivotal.io/teams/main/pipelines/cli-demo/jobs/test_pxf_secure_multinode_gpdb/builds/27).

PR #225, opened 2019-08-28
pipeline: Consume GPDB 7 binary for CentOS 7

oliverralbertini:
For 5X and 6X, whether we are making regular PXF or a release pipeline,
we use release candidates. We don't consume GPDB 7 (master) artifacts in
the release pipelines, but now we are picking that up from the correct
place for CentOS 7.


oliverralbertini:
Thanks for looking at it, I think this will have to be re-worked.

This PR does change what artifacts are consumed for release/non-release pipelines (which I had initially been confusing with dev/prod).

oliverralbertini:
Closing because the desired changes are already addressed in 2bfec36914e45a94574e0b30b77ae1fdaefe0ba1.

PR #227, opened 2019-09-02
The newest version of PSQL introduces a + in multiline text

frankgh:
To make the output consistent accross all versions of GPDB we add \pset
format unaligned


frankgh:
@denalex @oliverralbertini merging this to make the pipelines less red

PR #228, opened 2019-09-03
Make JDBC profile not fail when MAPR JAR files override default Hadoop ones

denalex:


concourse/pipelines/templates/pxf-tpl.yml
```diff
@@ -607,15 +606,16 @@ jobs:
     file: pxf_src/concourse/tasks/test_pxf.yml
     image: gpdb-pxf-dev-centos6-mapr-image
     params:
-      GROUP: hcfs
+      ACCESS_KEY_ID: {{tf-machine-access-key-id}}
```
frankgh: do we want the `ACCESS_KEY_ID` and `SECRET_ACCESS_KEY` if we are not running any test accessing S3

denalex: tried to run cloud tests, but they fail due to `Configuration` APIs, so will remove the creds.

denalex:
@oliverralbertini -- that pipeline is red for a different reason due to flaky HCFS test, which I will try to look into, but seems unrelated to this PR.

concourse/pipelines/templates/pxf-tpl.yml
frankgh: mapr has it's own aws jars, we might be able to use them:
`/opt/mapr/hadoop/hadoop-2.7.0/share/hadoop/common/lib/aws-java-sdk-1.7.15.jar`
`/opt/mapr/hadoop/hadoop-2.7.0/share/hadoop/common/lib/hadoop-aws-2.7.0-mapr-1707.jar`

automation/src/test/java/org/greenplum/pxf/automation/features/hdfs/HdfsReadableTextTest.java
```diff
@@ -304,9 +304,10 @@ public void emptyTextFile() throws Exception {
         // define and create external table
         exTable.setProfile(ProtocolUtils.getProtocol().value() + ":text");
         exTable.setDelimiter(",");
+        exTable.setPath(exTable.getPath() + "_empty");
         gpdb.createTableAndVerify(exTable);
         // write empty data to HDFS
-        hdfs.writeTableToFile(hdfsFilePath, new Table("emptyTable", null), ",");
+        hdfs.writeTableToFile(hdfsFilePath + "_empty", new Table("emptyTable", null), ",");
```
oliverralbertini: Is this the change that fixed the flakiness? How does it work?

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
```diff
@@ -527,5 +528,29 @@ private void assertMandatoryParameter(String value, String paramName, String opt
...
+     * @return mapping of configuration properties with prefix stripped
+     */
+    private Map<String, String> getPropsWithPrefix(Configuration configuration, String confPrefix) {
+        Map<String, String> configMap = new HashMap<>();
+        Iterator<Map.Entry<String, String>> it = configuration.iterator();
+        while (it.hasNext()) {
+            String propertyName = it.next().getKey();
+                if (propertyName.startsWith(confPrefix)) {
+                    // do not use value from the iterator as it might not come with variable substitution
+                    String value = configuration.get(propertyName);
```
oliverralbertini: Is this what makes it different from Hadoop's `Configuration#getPropsWithPrefix()`?

automation/src/test/java/org/greenplum/pxf/automation/features/hdfs/HdfsReadableTextTest.java
denalex: the empty file was not being overwritten in time by the next test, so now it is a separate file used by this test only and not impacting anything else.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
denalex: no, they use the same, but since MAPR's version of Configuration is older and does not include this method, we had to pull it here.

PR #229, opened 2019-09-05
Allow users to set hadoop login user with HADOOP_USER_NAME

frankgh:
Allow users to set the HADOOP_USER_NAME environment variable to define
the hadoop login user. By default, hadoop will use the OS user name.
By setting the HADOOP_USER_NAME property in pxf-conf.sh, a different
user can be specified.

When PXF is configured with multiple hadoop environments, this
environment variable will apply to all the non-kerberized environments.
For example, let's say we have 3 servers default (kerberized), hdp1, and
hdp2, and the HADOOP_USER_NAME is set to gpuser. The login user for hdp1
and hdp2 will be gpuser.

### DIFF
--- a/server/pxf-service/src/scripts/pxf-env-default.sh
+++ b/server/pxf-service/src/scripts/pxf-env-default.sh
@@ -73,3 +73,6 @@ export PXF_OOM_KILL=${PXF_OOM_KILL:-true}
 
 # Dump heap on OutOfMemoryError, set to dump path to enable
 # export PXF_OOM_DUMP_PATH=
+
+# Login user for hadoop, defaults to the OS user that started PXF process
+export HADOOP_USER_NAME=${HADOOP_USER_NAME:-$(whoami)}
--- a/server/pxf-service/src/templates/user/conf/pxf-env.sh
+++ b/server/pxf-service/src/templates/user/conf/pxf-env.sh
@@ -39,3 +39,6 @@ PXF_CONF="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd )"
 
 # Dump heap on OutOfMemoryError, set to dump path to enable
 # export PXF_OOM_DUMP_PATH=/tmp/pxf_heap_dump
+
+# Login user for hadoop, defaults to the OS user that started PXF process
+# export HADOOP_USER_NAME=$(whoami)


server/pxf-service/src/scripts/pxf-env-default.sh
```diff
@@ -73,3 +73,6 @@ export PXF_OOM_KILL=${PXF_OOM_KILL:-true}
 
 # Dump heap on OutOfMemoryError, set to dump path to enable
 # export PXF_OOM_DUMP_PATH=
+
+# Login user for hadoop, defaults to the logged in user in the OS
```
denalex:
```suggestion
# Login user for hadoop, defaults to the OS user that started PXF process
```

server/pxf-service/src/scripts/pxf-env-default.sh
```diff
@@ -73,3 +73,6 @@ export PXF_OOM_KILL=${PXF_OOM_KILL:-true}
...
+# export HADOOP_USER_NAME=
```
denalex: we advise people not to touch this file, so what is the point of this line ? Maybe have it like this:
```suggestion
export HADOOP_USER_NAME=${HADOOP_USER_NAME:-$(whoami)}
```

server/pxf-service/src/templates/user/conf/pxf-env.sh
```diff
@@ -39,3 +39,6 @@ PXF_CONF="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd )"
 
 # Dump heap on OutOfMemoryError, set to dump path to enable
 # export PXF_OOM_DUMP_PATH=/tmp/pxf_heap_dump
+
+# Login user for hadoop, defaults to the logged in user in the OS
+# export HADOOP_USER_NAME=
```
denalex: same here, the design idea was to still work with defaults if the line is uncommented, so maybe
```suggestion
# export HADOOP_USER_NAME=$(whoami)
```

server/pxf-service/src/scripts/pxf-env-default.sh
frankgh: I think it's a good idea to keep these two files in sync, to keep track of the environment variables we use

server/pxf-service/src/scripts/pxf-env-default.sh
```diff
@@ -73,3 +73,6 @@ export PXF_OOM_KILL=${PXF_OOM_KILL:-true}
 
 # Dump heap on OutOfMemoryError, set to dump path to enable
 # export PXF_OOM_DUMP_PATH=
+
+# Login user for hadoop, defaults to the OS user that started PXF process
+export HADOOP_USER_NAME=${HADOOP_USER_NAME:-$(whoami)}
```
oliverralbertini: I prefer to either use the value of `USER` or use the `id` command. From the (BSD) man page for `whoami`:

```
DESCRIPTION
     The whoami utility has been obsoleted by the id(1) utility, and is equivalent to ``id -un''.  The command ``id -p'' is suggested for normal interactive use.

     The whoami utility displays your effective user ID as a name.
```

So let's use `$(id -u -n)` or `${USER}`.

denalex: yes, ${USER} seems simpler, thanks @oliverralbertini

frankgh:
waiting on merging/closing until we define if we want to support user-per-server setup. Will add details in tracker.

frankgh:
Closing in favor of supporting the ability to configure a per server hadoop login user

PR #231, opened 2019-09-05
Create pipelines for producing HDP3-based docker images

denalex:


concourse/pipelines/docker-images.yml
```diff
@@ -23,6 +23,7 @@ groups:
   - docker-gpdb-pxf-dev-centos7-mapr-server
   - docker-gpdb-pxf-dev-centos7-cdh-server
   - docker-gpdb-pxf-dev-centos7-hdp2-server
+  - docker-gpdb-pxf-dev-centos7-hdp3-server
```
oliverralbertini: Should we also add to the centos group?

denalex: there's no `centos` group, only `centos6` and `centos7` and the job has been already added to `centos7` one (line 49)

concourse/pipelines/docker-images.yml
```diff
@@ -593,7 +622,26 @@ jobs:
     - task: package_singlecluster
       file: pxf_src/singlecluster/package_singlecluster_hdp.yml
       image: gpdb-dev-centos6-image
-    - put: singlecluster-HDP
+    - put: singlecluster-HDP2
+      params:
+        file: artifacts/singlecluster-HDP.tar.gz
```
frankgh: will this conflict with the file for HDP3? the file names are the same, not sure if that is solved by the singlecluster-HDP2 resource?

denalex: yes, I hope it will be solved by the resource `versioned_file: singlecluster/singlecluster-HDP2.tar.gz` (need to be tested) and since the build job always creates `artifacts/singlecluster-HDP.tar.gz` I think during `put` the file will be renamed when copied to S3

PR #234, opened 2019-09-07
Port smoke test group from Tinc to pg_regress

oliverralbertini:
Uses sed to search and replace some templates for PXF_CONF and HDFS_CMD
(path to `hdfs` CLI). Generated files have the same name as the
sql/*.sql and expected/*.out files, but with a leading underscore.

Currently relies on copying the PXF_CONF/servers/default ->
PXF_CONF/servers/hdfs directory, because default is a protected word in
Postgres.


regression/substitute.bash
```diff
@@ -0,0 +1,8 @@
+for f in sql/*.sql expected/*.out; do
+	basename=${f##*/}
```
denalex: wodoo magic, please comment for less advanced bash users :-)

regression/sql/HdfsSmokeTest.sql
```diff
@@ -0,0 +1,42 @@
...
+		dub DOUBLE PRECISION,
+		longNum BIGINT,
+		bool BOOLEAN
+	) SERVER hdfs
+	OPTIONS (resource '/tmp/pxf_automation_data', format 'csv');
+SELECT * FROM hdfs_smoke_small_fdw ORDER BY name;
+SELECT name, num FROM hdfs_smoke_small_fdw WHERE num > 50 ORDER BY name;
+
+-- clean up HDFS
+\!%HDFS_CMD% dfs -rm -r /tmp/pxf_automation_data
```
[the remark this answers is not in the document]
oliverralbertini: It seems to execute everything, even when there are failures.

regression/Makefile
```diff
@@ -0,0 +1,30 @@
+ifndef HDFS_CMD
```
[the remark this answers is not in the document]
oliverralbertini: I think that is a cleaner solution.

regression/expected/HdfsSmokeTest.out
```diff
@@ -0,0 +1,355 @@
...
+		case when (i%2) = 0 then 'true' else 'false' end
+		from generate_series(1, 100) s(i)
+	) to '/tmp/pxf_automation_data/data.txt' WITH (FORMAT 'csv');
+\!%HDFS_CMD% dfs -mkdir -p /tmp/pxf_automation_data
+\!%HDFS_CMD% dfs -copyFromLocal /tmp/pxf_automation_data/data.txt /tmp/pxf_automation_data
+-- External Table test
+CREATE EXTENSION pxf;
+CREATE EXTERNAL TABLE pxf_smoke_small_data
+	(name TEXT, num INTEGER, dub DOUBLE PRECISION, longNum BIGINT, bool BOOLEAN)
+	LOCATION('pxf://tmp/pxf_automation_data/data.txt?PROFILE=hdfs:text') FORMAT 'CSV' (DELIMITER ',');
```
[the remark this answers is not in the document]
frankgh: totally agree, and when issues surface, we don't need to go to a different places to dig out the issue

regression/expected/HdfsSmokeTest.out
```diff
@@ -0,0 +1,355 @@
...
+ row_98  |  98
+ row_99  |  99
+(50 rows)
+
+-- FDW test
+\!cp -r %PXF_CONF%/servers/{default,hdfs}
+CREATE EXTENSION pxf_fdw;
+CREATE SERVER hdfs
+	FOREIGN DATA WRAPPER hdfs_pxf_fdw;
+CREATE USER MAPPING FOR CURRENT_USER SERVER hdfs;
```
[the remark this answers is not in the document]
frankgh: another thing to think about is how to run different `GROUP`s. Is this supported by the pg_regress framework?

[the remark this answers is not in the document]
oliverralbertini: yeah there is a flag called `--schedule` that can point to a file containing a list of test cases. See this commit: e4590c8bb7d90d7bb250f325de01b00da36e74e8

regression/sql/HdfsSmokeTest.sql
```diff
@@ -0,0 +1,41 @@
+-- data prep
+\!mkdir -p /tmp/pxf_automation_data
```
frankgh: maybe let's try to keep a similar directory structure as in the automation framework. I suggest `./regression/sql/smoke/HdfsSmokeTest.sql` ?

regression/expected/HdfsSmokeTest.out
```diff
@@ -0,0 +1,354 @@
...
+COPY (
+	SELECT format('row_%s',i::varchar(255)),
+		i,
+		i+0.0001,
+		i*100000000000,
+		CASE WHEN (i%2) = 0 THEN 'true' ELSE 'false' END
+		from generate_series(1, 100) s(i)
+	) TO '/tmp/pxf_automation_data/data.txt'
+	WITH (FORMAT 'csv');
+\!%HDFS_CMD% dfs -mkdir -p /tmp/pxf_automation_data
```
frankgh: maybe remove the directory before the test starts? can we have unique directories for different tests? This would let us consider running the regression framework in parallel, and it would let us speed up our feedback cycle.

frankgh: In our existing tests we do that: `Hdfs  -> Remove Directory tmp/pxf_automation_data`

regression/expected/HdfsSmokeTest.out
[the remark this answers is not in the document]
frankgh: will we be able to have a single test in multiple groups (schedules)? Can we have a smoke, gpdb, hive, s3, etc schedules?

regression/sql/HdfsSmokeTest.sql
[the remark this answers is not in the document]
frankgh: I have seen this behaviour, so it looks like it will be executed regardless. Maybe we should also consider how to integrate PXF_TEST_DEBUG into this, where we don't cleanup data after the test runs for debugging purposes.

regression/Makefile
```diff
@@ -0,0 +1,32 @@
...
+PGXS := $(shell $(PG_CONFIG) --pgxs)
+include $(PGXS)
+else
+subdir = contrib/pxf_fdw
+top_builddir = ../..
+include $(top_builddir)/src/Makefile.global
+include $(top_srcdir)/contrib/contrib-global.mk
+endif
+
+smoke : REGRESS_OPTS += --schedule=smoke_schedule
```
frankgh: I can see how we can run the smoke schedule now, nice!  Can a test be part of multiple schedules?

regression/expected/HdfsSmokeTest.out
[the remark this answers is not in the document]
oliverralbertini: yes multiple schedules and tests in more than one schedule is all possible

regression/expected/HdfsSmokeTest.out
oliverralbertini: OK

regression/Makefile
oliverralbertini: Yes

denalex: The way it starts to look we'll be married to Postgres / Greenplum to even be able to run the integration tests. Not sure what to think about it, we need those, of course, at runtime for the tests to run,  but what if PXF were to become an independent engine ...

frankgh: what would a more generic alternative would look like? Or is our existing automation framework flexible enough for that?

oliverralbertini: What's so bad about being married?

denalex: or, nothing, as long as you're married to the right thing :-D

denalex: but seriously, we'll not be able to run automation tests without having gpdb (or postgres) installed on the same node and having their `include $(top_srcdir)/contrib/contrib-global.mk` to be in place because pg_regress is integrated into postgres/greenplum and not a standlone framework. We are going from the need to have `gpdb` SOMEWHERE during RUNTIME to having `gpdb` LOCALLY during BUILD time which is too much coupling, IMO.

regression/substitute.bash
oliverralbertini: commented. It's just removing prefix from var `foo`: `${foo#prefix}`. With two `#` it removes the longest possible match of prefix. Replacing `#` with `%`, its the same, but for suffixes.

regression/Makefile
frankgh: isn't this true already with our existing automation framework?

regression/expected/FDW_HBaseSmokeTest.out
```diff
@@ -0,0 +1,178 @@
...
+CREATE USER MAPPING FOR CURRENT_USER
+	SERVER h_base_smoke_test_server;
+CREATE FOREIGN TABLE h_base_smoke_test_foreign_table  (
+		name text,
+		num int,
+		dub double precision,
+		longnum bigint,
+		bool boolean
+	)
+	SERVER h_base_smoke_test_server OPTIONS (resource 'hbase_table', format 'csv');
```
frankgh: hbase has no format :)

regression/hbase/drop_small_data.rb
```diff
@@ -0,0 +1,6 @@
+disable 'hbase_table'
```
frankgh: maybe have unique names for the table to run parallel tests

shivzone:
@oliverralbertini nice effort ! Would be very useful for context and future reference to list out the directory structure and the general strategy/approach on how and where would the component specific DDLs exist

oliverralbertini:
I think this is a good time to think about merging this work. Currently all *SmokeTest are ported to run in pg_regress using external and foreign tables. Here's a [pipeline](https://ud.ci.gpdb.pivotal.io/teams/main/pipelines/pg_regress/jobs/test_pxf/builds/14).

concourse/pipelines/pg_regress_pipeline.yml
```diff
@@ -0,0 +1,169 @@
...
+    params:
+      CONFIGURE_FLAGS: '--enable-cassert --enable-debug-extensions'
+      TARGET_OS: centos
+      TARGET_OS_VERSION: 7
+      BLD_TARGETS: "clients loaders"
+    timeout: 30m
+  - put: bin_gpdb
+    params: { file: gpdb_artifacts/bin_gpdb.tar.gz }
+
+- name: test_pxf
```
frankgh: maybe this should be our `test_pxf_no_impersonation` test, which only runs smoke with no impersonation.

regression/Makefile
```diff
@@ -0,0 +1,50 @@
...
+include $(top_srcdir)/contrib/contrib-global.mk
+endif
+
+ifdef PXF_TEST_DEBUG
+SUBSTITUTE_COMMAND += --debug
+REGRESS_OPTS += --debug
+endif
+
+setup_test_dir :
+	mkdir -p /tmp/pxf_automation_data
```
frankgh: one last thing we need to take into account is when we run the tests using PROTOCOL=[s3|gs|adl|...] . This PR is coming along, but I think we still haven't solved this:
- https://github.com/greenplum-db/pxf/blob/master/automation/src/test/resources/sut/default.xml#L18
- https://github.com/greenplum-db/pxf/blob/master/automation/src/test/resources/sut/default.xml#L24
- https://github.com/greenplum-db/pxf/blob/master/automation/src/test/resources/sut/default.xml#L38
- https://github.com/greenplum-db/pxf/blob/master/automation/src/test/resources/sut/default.xml#L45

I think a way to do is always have an `UUID` generated for every test, including for hdfs (i.e. `tmp/pxf_automation_data/__UUID__` ).
We also need to provide the `workingDirectory` as an environment variable or something along the lines of the hdfs command.

Hopefully this makes sense

regression/init_file
```diff
@@ -0,0 +1,17 @@
...
+m/^SLF4J: Class path contains multiple SLF4J bindings\.$/
+m/^SLF4J: Found binding in \[jar:file:.*\/hbase\/lib\/slf4j-log4j12-\d{1,3}\.\d{1,3}\.\d{1,3}\.jar\!\/org\/slf4j\/impl\/StaticLoggerBinder\.class\]$/
+m/^SLF4J: Found binding in \[jar:file:.*\/hadoop\/share\/hadoop\/common\/lib\/slf4j-log4j12-\d{1,3}\.\d{1,3}\.\d{1,3}\.jar\!\/org\/slf4j\/impl\/StaticLoggerBinder\.class\]$/
+m/^SLF4J: See http:\/\/www\.slf4j\.org\/codes\.html#multiple_bindings for an explanation\.$/
+m/^SLF4J: Actual binding is of type \[org\.slf4j\.impl\.Log4jLoggerFactory\]$/
+m/^INFO:  extractPxfAttributes: unsupported node tag 406, unable to extract attribute from qualifier$/
+-- end_matchignore
+
+-- start_matchsubs
+m/0 row\(s\) in \d{1}\.\d{4} seconds/
```
frankgh: does this only match for `0 row[s]`?

regression/Makefile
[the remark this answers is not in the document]
frankgh: can we rename `HDFS_CMD` to `HCFS_CMD`?

frankgh:
@oliverralbertini also, I think @shivzone brings a good point. Maybe let's create a README.md file with the folder layout.

regression/init_file
oliverralbertini: yeah that's something that gets output by the beeline client, I believe.

frankgh: should we change it to `\d+ row[s]...` in case the beeline client output changes?

regression/Makefile
oliverralbertini: Yeah, I hadn't considered that we should run these tests against all the cloud providers. I guess the normal way with Tinc is to export PROTOCOL=s3, etc. to get those tests to run.

I agree if we are going to use a random UUID, then should also do this for HDFS. This avoids issues with tests running concurrently (even the same test in different pipelines, etc.).

As for giving the workingDirectory back to the user/test-runner, would it suffice just to print it to the screen?

regression/init_file
oliverralbertini: done

regression/Makefile
oliverralbertini: It's now being tested against S3, and I will try it with the other clouds next (ADL, GS, WASBS).

https://ud.ci.gpdb.pivotal.io/teams/main/pipelines/pg_regress/jobs/test_pxf_s3/builds/3

Should I make a new schedule for HCFS? It's kind of awkward in CI right now because when we set PROTOCOL, we end up not turning on HBase/Hive in `test_pxf.bash`. So I am setting HCFS_PROTOCOL instead, that way the entire smoke_schedule passes. But maybe we should have a hcfs_smoke_schedule with just the tests HdfsSmokeTest, WritableSmokeTest, and MultiBlockDataSmokeTest.

denalex: maybe we can have s3_schedule, gs_schedule, etc ? That way it will be easier to specify what test you want to run, instead of messing up with PROTOCOL variables.

frankgh: @denalex that makes sense, but it'll be a little hard to maintain because if you add a new HCFS test, you need to add it to all schedules. Up to you Oliver

oliverralbertini:
@shivzone @frankgh @denalex @vraghavan78 Please take a look at the new README for regression. Also here's a pipeline where the regression suite is running against multiple clouds:
https://ud.ci.gpdb.pivotal.io/teams/main/pipelines/pg_regress

regression/expected/FDW_HBaseSmokeTest.out
```diff
@@ -0,0 +1,178 @@
+-- Create Hbase tables hbase_table and pxflookup
+\!{{HBASE_CMD}} shell {{FULL_PATH_TO_HBASE_GEN_SCRIPT}} >/dev/null
+-- FDW test
+CREATE SERVER h_base_smoke_test_server
+	FOREIGN DATA WRAPPER hbase_pxf_fdw
+	OPTIONS (config 'default');
+CREATE USER MAPPING FOR CURRENT_USER
```
frankgh: nice, I haven't used this one, but it makes sense

regression/hbase/drop_small_data.rb
oliverralbertini: Table names are now unique.

concourse/pipelines/pg_regress_pipeline.yml
oliverralbertini: Yeah that would be great.

regression/Makefile
oliverralbertini: Resolving because tests are being run against all the clouds plus Minio. The only thing that's missing is running against remote Greenplum/HDFS. The test code can handle this (verified on local workstation) but we don't have the scripts in CI for it yet. I think it should be separate work.

concourse/pipelines/pg_regress_pipeline.yml
```diff
@@ -0,0 +1,350 @@
...
+        TARGET_OS_VERSION: 7
+        HADOOP_CLIENT: HDP
+        GROUP: hcfs_schedule,fdw_hcfs_schedule
+        PG_REGRESS: true
+        PROTOCOL: s3
+        ACCESS_KEY_ID: {{tf-machine-access-key-id}}
+        SECRET_ACCESS_KEY: {{tf-machine-secret-access-key}}
+      run:
+        path: pxf_src/concourse/scripts/test_pxf.bash
+    timeout: 2h
```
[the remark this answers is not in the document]
oliverralbertini: that's just a remnant from copy-paste.

concourse/scripts/test_pxf.bash
```diff
@@ -17,12 +18,48 @@ else
...
+		export PGPORT=15432
+		export HCFS_CMD=${GPHD_ROOT}/bin/hdfs
+		export HCFS_PROTOCOL=${PROTOCOL}
+		export HBASE_CMD=${GPHD_ROOT}/bin/hbase
+		export BEELINE_CMD=${GPHD_ROOT}/hive/bin/beeline
+		export HCFS_BUCKET=${HCFS_BUCKET}
+		# hive-specific vars
+		# export HIVE_IS_REMOTE= HIVE_HOST= HIVE_PRINCIPAL=
+
+		time make -C ${PWD}/pxf_src/regression ${GROUP//,/ }
```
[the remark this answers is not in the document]
oliverralbertini: replace all commas with a space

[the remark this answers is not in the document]
oliverralbertini: adding a comment in the script for posterity :)

regression/README.md
```diff
@@ -0,0 +1,144 @@
...
+This is different from our automation frame work (`pxf/automation`) which fires up a Java process that uses, e.g. HDFS APIs to prepare the data, then performs SQL queries for the rest.
+
+Running the tests
+===================================
+
+## Pre-requisites
+
+You need a running instance of Greenplum and PXF, along with a local installation of Greenplum (to be able to use the `pg_regress` framework).
+The variables `PGHOST` and `PGPORT` must be pointing at the Greenplum master node, and Greenplum environment scripts like `${GPHOME}/greenplum_path.sh` and `gpdb/gpAux/gpdemo/gpdemo-env.sh` should be sourced.
+`pg_config` must be on your path.
```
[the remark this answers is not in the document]
oliverralbertini: `${GPHOME}/bin`, although I just realized I've been using `/usr/local/bin/pg_config` this whole time that I've worked on pg_regress. (That's the one from homebrew).

PR #235, opened 2019-09-07
added test_pxf_hdp3_centos7 job

denalex:


automation/src/test/java/org/greenplum/pxf/automation/features/hive/HiveTest.java
```diff
@@ -114,6 +114,7 @@ private void prepareManyPartitionedData() throws Exception {
         hive.createTableAndVerify(hiveManyPartitionsTable);
         hive.runQuery("SET hive.exec.dynamic.partition = true");
         hive.runQuery("SET hive.exec.dynamic.partition.mode = nonstrict");
+        hive.runQuery("SET hive.stats.autogather = false");
```
oliverralbertini: this is the secret sauce?

concourse/scripts/pxf_common.bash
```diff
@@ -242,6 +242,47 @@ EOF
...
+		<name>datanucleus.schema.autoCreateAll</name>
+		<value>True</value>
+	</property>
+	<property>
+		<name>metastore.metastore.event.db.notification.api.auth</name>
+		<value>false</value>
+	</property>
+EOF
+	sed -i -e '/<configuration>/r patch.xml' ${GPHD_ROOT}/hive/conf/hive-site.xml
+	rm patch.xml
```
oliverralbertini: Nit: If you're going to overwrite `patch.xml` below you don't need to `rm` it.

concourse/scripts/pxf_common.bash
```diff
@@ -242,6 +242,47 @@ EOF
...
+		<name>tez.use.cluster.hadoop-libs</name>
+		<value>true</value>
+	</property>
+EOF
+	sed -i -e '/<configuration supports_final="true">/r patch.xml' ${GPHD_ROOT}/tez/conf/tez-site.xml
+	rm patch.xml
+
+	# update properties in yarn-site.xml
+	sed -i -e 's|HADOOP_CONF|HADOOP_CONF_DIR|g' ${GPHD_ROOT}/hadoop/etc/hadoop/yarn-site.xml
+	sed -i -e 's|HADOOP_ROOT|HADOOP_HOME|g' ${GPHD_ROOT}/hadoop/etc/hadoop/yarn-site.xml
```
oliverralbertini: can be one call to sed:

```bash
sed -ie 's|HADOOP_CONF|HADOOP_CONF_DIR|g' -e 's|HADOOP_ROOT|HADOOP_HOME|g' ${GPHD_ROOT}/hadoop/etc/hadoop/yarn-site.xml
```
You can throw in line breaks too if it's too long:
```bash
sed -ie 's|HADOOP_CONF|HADOOP_CONF_DIR|g' \
    -e 's|HADOOP_ROOT|HADOOP_HOME|g' ${GPHD_ROOT}/hadoop/etc/hadoop/yarn-site.xml
```

concourse/scripts/pxf_common.bash
```diff
@@ -242,6 +242,47 @@ EOF
...
+	<property>
+		<name>datanucleus.schema.autoCreateAll</name>
+		<value>True</value>
+	</property>
+	<property>
+		<name>metastore.metastore.event.db.notification.api.auth</name>
+		<value>false</value>
+	</property>
+EOF
+	sed -i -e '/<configuration>/r patch.xml' ${GPHD_ROOT}/hive/conf/hive-site.xml
```
oliverralbertini: I prefer these to be one call to sed when possible:

```bash
sed -ie '/<configuration>/r patch.xml' -e 's/>mr/>tez/g' "${GPHD_ROOT}/hive/conf/hive-site.xml"
```

Note that GPHD_ROOT is quoted.

concourse/scripts/pxf_common.bash
```diff
@@ -242,6 +242,47 @@ EOF
...
+	</property>
+	<property>
+		<name>datanucleus.schema.autoCreateAll</name>
+		<value>True</value>
+	</property>
+	<property>
+		<name>metastore.metastore.event.db.notification.api.auth</name>
+		<value>false</value>
+	</property>
+EOF
```
oliverralbertini: You can tab-indent this EOF as well. You can also further indent the contents of the heredoc one level.

concourse/scripts/test_pxf.bash
```diff
@@ -82,6 +82,9 @@ function setup_hadoop() {
 	if [[ -n ${GROUP} ]]; then
 		export SLAVES=1
 		setup_impersonation ${hdfsrepo}
+		if [[ $(grep 'hadoop-3' ${hdfsrepo}/versions.txt) ]]; then
```
oliverralbertini: You probably want:
```bash
if grep hadoop-3 "${hdfsrepo}/versions.txt"; then
```

If the grep finds a match it returns 0 (success), and it goes inside the `if` block. Use command substitution, e.g. `$(command)` only when you need the _output_, like:

```bash
if [[ $(grep foo /tmp/bar) == foo ]]; then
```

Otherwise just rely on the return code.

concourse/scripts/pxf_common.bash
```diff
@@ -242,6 +242,47 @@ EOF
...
 }
 
+function adjust_for_hadoop3() {
+	local GPHD_ROOT=${1}
+
+	# remove deprecated conf from hive-env.sh
+	sed -i -e 's/-hiveconf hive.log.dir=$LOGS_ROOT //g' ${GPHD_ROOT}/hive/conf/hive-env.sh
+
+	# add properties to hive-site.xml
+    sed -i -e 's/>mr/>tez/g' ${GPHD_ROOT}/hive/conf/hive-site.xml
```
oliverralbertini: This should be a tab so it lines up with the rest of the code in this func, but see below's edit (eliminates the need for this line).

concourse/scripts/pxf_common.bash
```diff
@@ -242,6 +242,47 @@ EOF
...
+	sed -i -e '/<configuration>/r patch.xml' ${GPHD_ROOT}/hive/conf/hive-site.xml
+	rm patch.xml
+
+	# add properties to tez-site.xml
+	cat > patch.xml <<-EOF
+	<property>
+		<name>tez.use.cluster.hadoop-libs</name>
+		<value>true</value>
+	</property>
+EOF
```
oliverralbertini: Same as above, the EOF can be tabbed to line up with `cat` call.

concourse/scripts/test_pxf.bash
oliverralbertini: More info here: https://github.com/koalaman/shellcheck/wiki/SC2143. As mentioned there, you can use the `grep` option `-q` for quiet output.

concourse/scripts/test_pxf.bash
```diff
@@ -82,6 +82,9 @@ function setup_hadoop() {
...
+		    adjust_for_hadoop3 ${hdfsrepo}
+		fi
```
oliverralbertini: `hdfsrepo` should be double-quoted everywhere it appears.

automation/src/test/java/org/greenplum/pxf/automation/features/hive/HiveTest.java
frankgh: it is :)

denalex: one of them, gathering stats after dynamically creating a partition required a reducer step which was failing for some reason, so we got rid of the need to run that reducer.

concourse/scripts/pxf_common.bash
denalex: Sounds good, do you prefer them quoted in the snippet above as well ?

concourse/scripts/pxf_common.bash
denalex: yea, but tab-indenting EOF screws up the view in IntelliJ for some reason

concourse/scripts/test_pxf.bash
denalex: thanks, will do, even though leaving code after if "totally naked" doesn't feel comfortable :)

denalex:
@oliverralbertini -- I addressed the comments, thanks, please take a look

PR #237, opened 2019-09-11
Hive metastore client patch

frankgh:
This PR brings HiveMetaStoreClient from the Hive 2.3.6 branch and fixes the issue in JDK 11, in a separate commit. This issue has been fixed in Hive version 2.3.7 and is scheduled to be released in November. I suggest we keep this file until Hive 2.3.7 is released and we can safely remove it then.


frankgh:
Closing after talking to @denalex

PR #247, opened 2019-10-08
Add support for multiple kerberos Hadoop and Hive servers

frankgh:
This PR adds support in PXF for multiple kerberized hadoop / hive clusters. This PR also makes sure that non-kerberized clusters can work alongside keberized clusters in PXF.

- Remove the restriction to configure the kerberized cluster on the default server
- Add support for multiple hadoop clusters
- Add support for multiple keberized hadoop clusters
- Add support for multiple kerberized hive clusters
- Allow custom user identity per Hadoop server for non-secure clusters
- Enable setting user impersonation per server
- Test one keytab per PXF server (before we were testing headless keytabs)


lavjain:
Wow, this is a massive commit :-)

Kudos to PXF team for the great work!

PR #248, opened 2019-10-08
Enable Avro write

oliverralbertini:


oliverralbertini:
This is still preliminary, but need some feedback from @frankgh and @denalex.

Schema is written into file, and can be auto-generated based on schema in GPDB table.

The following works:

```sql
DROP EXTERNAL TABLE avro_write_primitive;
CREATE WRITABLE EXTERNAL TABLE avro_write_primitive (id int, foo text, f real)
	LOCATION ('pxf:///tmp/pxf_automation_data/avro/write_test?PROFILE=hdfs:avro')
	FORMAT 'CUSTOM' ( formatter='pxfwritable_export');
\!/Users/oalbertini/workspace/singlecluster-HDP/bin/hdfs dfs -rm -r -f /tmp/pxf_automation_data/avro/write_test
INSERT INTO avro_write_primitive VALUES ( 1, 'foobar', 0.123456789 );

DROP EXTERNAL TABLE avro_read_primitive;
CREATE EXTERNAL TABLE avro_read_primitive (LIKE avro_write_primitive)
	LOCATION ('pxf:///tmp/pxf_automation_data/avro/write_test?PROFILE=hdfs:avro')
	FORMAT 'CUSTOM' ( formatter='pxfwritable_import');

SELECT * FROM avro_read_primitive;
```
giving the following output:
```
DROP EXTERNAL TABLE
CREATE EXTERNAL TABLE
19/10/08 11:55:32 WARN util.NativeCodeLoader: Unable to load native-hadoop library for your platform... using builtin-java classes where applicable
Deleted /tmp/pxf_automation_data/avro/write_test
INSERT 0 1
DROP EXTERNAL TABLE
CREATE EXTERNAL TABLE
 id |  foo   |    f     
----+--------+----------
  1 | foobar | 0.123457
(1 row)
```

Rough edges: 

1) Looking up the schema will fail on the write path if the file you are writing to doesn't exist. This happens in `AvroFileAccessor#initialize()` which I tried to mitigate by checking the data fragment is -1 (is that the case for write path only?).
2) On the other hand in `AvroResolver#initialize()` we have a null schema during write, and we have to be careful about trying to call `getFields()` on that object.
3) If the file already exists, we fail at this command in `AvroFileAccessor#openForWrite()`:
```java
FSDataOutputStream avroOut = fs.create(file, false);
```
Maybe there is an append flag that should be added?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/AvroFileAccessor.java
```diff
@@ -62,7 +77,12 @@ public AvroFileAccessor() {
     @Override
     public void initialize(RequestContext requestContext) {
         super.initialize(requestContext);
+        rowsWritten = totalRowsWritten = 0;
 
+        // If we are about to write to HDFS, we don't need to locate schema
+        if (context.getDataFragment() == -1) {
+            return;
+        }
```
denalex: nit pick - it might be used to write to S3, probably, not necessarily HDFS, so this condition is common for all write cases, correct ?

But now, as this will support both read / write, maybe move respective blocks to `openForRead` / `openForWrite` methods and not try to determine if this is read / write case by analyzing `dataFragment`

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/AvroFileAccessor.java
```diff
@@ -122,7 +142,21 @@ public OneRow readNextObject() throws IOException {
...
+        String schemaFile = context.getOption("SCHEMA");
+        Schema schema = (schemaFile != null) ? getAvroSchema(jobConf, schemaFile) :
+                generateAvroSchema(context.getTupleDescription());
+        context.setMetadata(schema);
+        // make writer
+        // how do we write the schema file?
+        DatumWriter<GenericRecord> datumWriter = new GenericDatumWriter<GenericRecord>(schema);
+        writer = new DataFileWriter<>(datumWriter);
+        Path file = new Path(context.getDataSource());
+        FileSystem fs = file.getFileSystem(jobConf);
```
denalex: don't we need to do `getHcfsType().getUriForWrite()` to resolve the full URI ? Also, don't we need to append segmentId / transactionId to the path ?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/AvroFileAccessor.java
```diff
@@ -122,7 +142,21 @@ public OneRow readNextObject() throws IOException {
...
+        FSDataOutputStream avroOut = fs.create(file, false);
+        writer.create(schema, avroOut);
```
denalex: can this operation fail ? If yes, should we attempt to close `writer` and / or `FSDataOutputStream` in the `finally` block ?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/AvroFileAccessor.java
```diff
@@ -144,7 +180,11 @@ public boolean writeNextObject(OneRow onerow) throws Exception {
      */
     @Override
     public void closeForWrite() throws Exception {
-        throw new UnsupportedOperationException();
+        if (writer != null) {
+            writer.close();
+            totalRowsWritten += rowsWritten;
+        }
+        LOG.debug("Wrote a TOTAL of {} rows", totalRowsWritten);
```
denalex: need more context here, if possible -- server / segment / path / etc

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/AvroFileAccessor.java
```diff
@@ -144,7 +180,11 @@ public boolean writeNextObject(OneRow onerow) throws Exception {
      */
     @Override
     public void closeForWrite() throws Exception {
-        throw new UnsupportedOperationException();
+        if (writer != null) {
+            writer.close();
+            totalRowsWritten += rowsWritten;
```
denalex: why to keep 2 counters and not just use `rowsWritten` ? Accessor object will be used only for one transaction and `totalRowsWritten` is not static, so will be re-inited for every new transaction

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/AvroFileAccessor.java
```diff
@@ -62,7 +77,12 @@ public AvroFileAccessor() {
     @Override
     public void initialize(RequestContext requestContext) {
         super.initialize(requestContext);
+        rowsWritten = totalRowsWritten = 0;
```
denalex: is there a need to do this, they will already be inited with 0 ?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/AvroFileAccessor.java
```diff
@@ -167,4 +207,88 @@ Schema getAvroSchema(Configuration conf, String dataSource)
         dummyFileReader.close();
         return schema;
     }
+
+    private Schema generateAvroSchema(List<ColumnDescriptor> tupleDescription) throws IOException {
+        String colName;
+        int colType;
+
+        Schema schema = Schema.createRecord("tableName", "", COMMON_NAMESPACE, false);
+        List<Schema.Field> fields = new ArrayList<Schema.Field>();
```
denalex:
```suggestion
        List<Schema.Field> fields = new ArrayList<>();
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/AvroFileAccessor.java
```diff
@@ -167,4 +207,88 @@ Schema getAvroSchema(Configuration conf, String dataSource)
...
+            // columnDelimiter = delim == null ? ',' : delim.charAt(0);
+            fields.add(new Schema.Field(colName, getFieldSchema(DataType.get(colType), 0, 1), "", null));
+        }
+
+        schema.setFields(fields);
+
+        return schema;
+    }
+
+    private Schema getFieldSchema(DataType type, int notNull, int dim) throws IOException {
```
denalex:
```suggestion
    private Schema getFieldSchema(DataType type, boolean notNull, int dim) throws IOException {
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/AvroFileAccessor.java
```diff
@@ -167,4 +207,88 @@ Schema getAvroSchema(Configuration conf, String dataSource)
...
+        List<Schema> unionList = new ArrayList<Schema>();
```
denalex:
```suggestion
        List<Schema> unionList = new ArrayList<>();
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/AvroResolver.java
```diff
@@ -91,7 +97,8 @@ public void initialize(RequestContext requestContext) {
         }
 
         reader = new GenericDatumReader<>(schema);
-        fields = schema.getFields();
+        // schema may be null when we are writing data to HDFS
+        if (schema != null) fields = schema.getFields();
```
denalex:
```suggestion
        if (schema != null) {
            fields = schema.getFields();
        }
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/AvroResolver.java
```diff
@@ -141,7 +148,11 @@ public void initialize(RequestContext requestContext) {
      */
     @Override
     public OneRow setFields(List<OneField> record) throws Exception {
-        throw new UnsupportedOperationException();
+        GenericRecord genericRecord = new GenericData.Record((Schema) context.getMetadata());
```
denalex: probably would be better to make `schema` an object field and assign upon initialization ?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/AvroResolver.java
```diff
@@ -141,7 +148,11 @@ public void initialize(RequestContext requestContext) {
...
+        for (int i = 0; i < record.size(); i++) {
```
denalex:
```suggestion
        for (OneField field : record) {
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/AvroFileAccessor.java
oliverralbertini: In case of splits, which hasn't been worked out yet.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/AvroFileAccessor.java
oliverralbertini: yes, that was added in a later commit

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/AvroFileAccessor.java
oliverralbertini: I have addressed this by caching the type of request that we are processing inside `RequestContext`.

If we are writing, then we skip out of this `AvroFileAccessor#initialize()` early.

Would be nice to set up schema in `openForRead()`, but can't because the schema/metadata is needed in `AvroResolver#initialize()`, and all the `initialize()` methods are called up front, before `openForRead()`/`openForWrite()`.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/AvroResolver.java
oliverralbertini: Doing this in both the resolver and accessor.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/AvroResolver.java
oliverralbertini: I wanted to do this but I need the counter in the loop.

oliverralbertini:
> It looks good in general. I added some comments. Do we need to migrate integration tests?

Yes, going to add some integration tests soon.

oliverralbertini:
>I have thought about this and I believe a good place to add this information is at the entry point as part of the RequestContext. (i.e. https://github.com/greenplum-db/pxf/blob/master/server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/BridgeResource.java#L100)

@frankgh I ended up putting a new field on RequestContext that tells us what the request type is. It is added inside `HttpRequestParser#parseRequest()`. I think it is more complete this way since the request can be fragmenter, or read/write bridge.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/AvroFileAccessor.java
oliverralbertini: I believe we are covered in `AvroFileAccessor#closeForWrite()` when `writer.close()` is called. `writer` is wrapping the `FSDataOutputStream`, `avroOut`, so it should close both.

I think we need to close `datumWriter`, though.

oliverralbertini: but then again, a DatumWriter has no close() method.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/AvroFileAccessor.java
frankgh: a new class will be created per split, no? I'm not really sure what a split is

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/AvroFileAccessor.java
```diff
@@ -134,7 +167,9 @@ public boolean openForWrite() throws Exception {
      */
     @Override
     public boolean writeNextObject(OneRow onerow) throws Exception {
-        throw new UnsupportedOperationException();
+        writer.append((GenericRecord) onerow.getData());
```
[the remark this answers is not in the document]
oliverralbertini: Actually this is the reason why there are two counters. It's not about splits.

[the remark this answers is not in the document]
frankgh: ah, that would make sense. In that case you'd need to implement splitting.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HcfsType.java
```diff
@@ -10,13 +10,21 @@
...
 import static org.apache.hadoop.fs.FileSystem.FS_DEFAULT_NAME_KEY;
 
 public enum HcfsType {
-    ADL,
+    ADL {
+        @Override
+        public String getProfileScheme() {
+            return "adl://";
+        }
+    },
```
oliverralbertini: @denalex @frankgh 

I realized that doing `hcfsType.getDataUri()` doesn't help in the case of finding the schema for Avro. That will always give the dataSource file, but I have added a `HcfsType#getProfileScheme()` to be able to correctly build the path to the schema file on HCFS, it defaults to `file://`, see below.

automation/src/test/java/org/greenplum/pxf/automation/features/avro/HdfsWritableAvroTest.java
```diff
@@ -0,0 +1,319 @@
...
+            hdfs.copyToLocal(srcPath, filePath + ".avro");
+            hdfs.writeJsonFileFromAvro("file://" + filePath + ".avro", filePath + ".json");
+            filesToDelete.add(new File(filePath + ".json"));
+        }
+        Process p = Runtime
+                .getRuntime()
+                .exec(new String[]{
+                        "bash",
+                        "-c",
+                        "diff <(cat " + publicStage + "file_*.json | sort -d) " + resourcePath + compareFile
```
frankgh: 👏  I think this is a great approach to check whether what we wrote on the remote server makes sense. Very interesting, are you comparing the resulting file with some source file? can we also make sure that we are able to query the data we just wrote.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/AvroUtilities.java
```diff
@@ -0,0 +1,287 @@
...
+
+    private Schema readSchemaFromUserProvidedJson(RequestContext context, Configuration configuration, String schemaName, HcfsType hcfsType) throws IOException, URISyntaxException {
+
+        InputStream schemaStream = null;
+        // try searching local disk (full path name and relative path in classpath) first.
+        // by searching locally first we can avoid an extra call to external (fs.exists())
+        try {
+            File file = fileSearcher.searchForFile(schemaName);
+            if (file == null) {
+                Path path = new Path(schemaName);
```
frankgh: You need to use HcfsType to get the URI string. I realize that there is no method in HcfsType, but it needs to be added there.
```suggestion
                String uri = hcfsType.getDataUri(configuration, schemaName);
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/AvroUtilities.java
```diff
@@ -0,0 +1,287 @@
...
+            }
+            return (new Schema.Parser()).parse(schemaStream);
+        } finally {
+            if (schemaStream != null) {
+                schemaStream.close();
+            }
+        }
+    }
+
+    static Path schemaPathFromContext(RequestContext context, String schemaName) {
```
frankgh: this logic is already part of HcfsType.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/AvroUtilities.java
```diff
@@ -0,0 +1,287 @@
...
+     * @throws IOException if I/O error occurred while accessing Avro schema file
+     */
+    private Schema readSchemaFromUserProvidedAvroBinary(RequestContext context, Configuration configuration, String avroFile) throws IOException {
+        DatumReader<GenericRecord> datumReader = new GenericDatumReader<>();
+        DataFileReader<GenericRecord> fileReader = null;
+
+        try {
+            File file = fileSearcher.searchForFile(avroFile);
+            if (file == null) {
+                final Path path = schemaPathFromContext(context, avroFile);
```
frankgh: use HcfsType here

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/AvroUtilities.java
```diff
@@ -0,0 +1,287 @@
...
+import java.io.InputStream;
+import java.io.UnsupportedEncodingException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.net.URL;
+import java.net.URLDecoder;
+import java.util.ArrayList;
+import java.util.List;
+
+public final class AvroUtilities {
```
frankgh: HcfsType is being passed to the class, but it's never been used.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/AvroUtilities.java
```diff
@@ -0,0 +1,287 @@
...
+            case BOOLARRAY:
+            case TEXTARRAY:
+            case REALARRAY:
+            case DOUBLEARRAY:
+            case ENUM:
+            default:
+                unionList.add(Schema.create(Schema.Type.STRING));
+                // Below is code for adding _actual_ array types to the schema
+
+                // if (!isArrayType(type.getOID())) {
```
frankgh: let's remove any commented out code if we don't intend to use it, or is this code needed?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/AvroUtilities.java
```diff
@@ -0,0 +1,287 @@
...
+            case INTEGER:
+                unionList.add(Schema.create(Schema.Type.INT));
+                break;
+            case REAL:
+                unionList.add(Schema.create(Schema.Type.FLOAT));
+                break;
+            case FLOAT8:
+                unionList.add(Schema.create(Schema.Type.DOUBLE));
+                break;
+            case VARCHAR:
```
frankgh: why all the cases, if default is doing string too?
```suggestion
            default:
                unionList.add(Schema.create(Schema.Type.STRING));
```

server/pxf-service/src/main/java/org/greenplum/pxf/service/HttpRequestParser.java
```diff
@@ -1,5 +1,6 @@
 package org.greenplum.pxf.service;
 
+import com.sun.jersey.spi.container.ContainerRequest;
```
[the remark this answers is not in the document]
oliverralbertini: yes, you're right.

server/pxf-service/src/test/java/org/greenplum/pxf/service/HttpRequestParserTest.java
```diff
@@ -541,7 +543,27 @@ public void testInvalidHandlerCausesException() {
...
+    @Test
+    public void testWritePath() {
+        when(mockRequestHeaders.getPath()).thenReturn("v199/Writable/stream");
+        RequestContext context = parser.parseRequest(mockRequestHeaders, RequestType.WRITE_BRIDGE);
+        assertEquals(RequestType.WRITE_BRIDGE, context.getRequestType());
+    }
+
+    @Test
+    public void testFragmenterPath() {
+        when(mockRequestHeaders.getPath()).thenReturn("v1999/Fragmenter/getFragments");
```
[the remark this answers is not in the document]
oliverralbertini: correct.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/AvroUtilities.java
oliverralbertini: there's `HcfsType#getDataUri()` but that returns the data source, and we need to construct the correct Path to the schema file, which is not necessarily in the same place. This method is trying to combine correct scheme + bucket and the path to user-provided schema.

Maybe it should live in HcfsType, though?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/AvroUtilities.java
oliverralbertini: Yes, good catch. I was trying to use it, but then realized the issue with schema path != datasource path. See comment above.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/AvroUtilities.java
oliverralbertini: It will be needed in future work, but yeah I agree we can get rid of it.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/AvroUtilities.java
oliverralbertini: Actually then we would need to pass in context (we need to get the bucket name from `context.getDataSource()`).

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/AvroUtilities.java
oliverralbertini: Yeah I reduced it down. Thanks.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/AvroFileAccessor.java
```diff
@@ -167,4 +206,88 @@ Schema getAvroSchema(Configuration conf, String dataSource)
         dummyFileReader.close();
         return schema;
     }
+
+    private Schema generateAvroSchema(List<ColumnDescriptor> tupleDescription) throws IOException {
+        String colName;
```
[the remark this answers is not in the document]
oliverralbertini: This is now in its own class, called `AvroUtilities`. Maybe it should be called `AvroSchemaGenerator` instead?

automation/tincrepo/main/pxf/features/hdfs/writable/avro/complex_generate_schema/expected/query01.ans
```diff
@@ -0,0 +1,111 @@
...
+       92 | (f,184)     | sad            | {92,920,9200}    | {92.0,9200.0,920000.0}    | {"item 910","item 920","item 930"}
+       93 | (t,186)     | happy          | {93,930,9300}    | {93.0,9300.0,930000.0}    | {"item 920","item 930","item 940"}
+       94 | (f,188)     | sad            | {94,940,9400}    | {94.0,9400.0,940000.0}    | {"item 930","item 940","item 950"}
+       95 | (t,190)     | happy          | {95,950,9500}    | {95.0,9500.0,950000.0}    | {"item 940","item 950","item 960"}
+       96 | (f,192)     | sad            | {96,960,9600}    | {96.0,9600.0,960000.0}    | {"item 950","item 960","item 970"}
+       97 | (t,194)     | happy          | {97,970,9700}    | {97.0,9700.0,970000.0}    | {"item 960","item 970","item 980"}
+       98 | (f,196)     | sad            | {98,980,9800}    | {98.0,9800.0,980000.0}    | {"item 970","item 980","item 990"}
+       99 | (t,198)     | happy          | {99,990,9900}    | {99.0,9900.0,990000.0}    | {"item 980","item 990","item 1000"}
+      100 | (f,200)     | sad            | {100,1000,10000} | {100.0,10000.0,1000000.0} | {"item 990","item 1000","item 1010"}
```
denalex: seems like being odd means being "happy " :)

server/pxf-api/src/main/java/org/greenplum/pxf/api/io/DataType.java
```diff
@@ -58,6 +58,9 @@
     INT8ARRAY(1016),
     BOOLARRAY(1000),
     TEXTARRAY(1009),
+    REALARRAY(1021),
+    DOUBLEARRAY(1022),
+    ENUM(58377),
```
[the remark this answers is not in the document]
denalex: Will this cause any issues for any other connectors that do not support these types ?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/AvroFileAccessor.java
```diff
@@ -55,37 +60,43 @@ public AvroFileAccessor() {
...
-        // 1. Accessing the avro file through the "unsplittable" API just to get the schema.
-        //    The splittable API (AvroInputFormat) which is the one we will be using to fetch
-        //    the records, does not support getting the avro schema yet.
-        Schema schema;
-        try {
-            schema = getAvroSchema(configuration, context.getDataSource());
-        } catch (IOException e) {
-            throw new RuntimeException("Failed to obtain Avro schema for " + context.getDataSource(), e);
-        }
+        schema = AvroUtilities.getInstance().obtainSchema(context, configuration);
```
denalex: we should be doing `AvroUtilities.getInstance()` in the constructor and creating another constructor with this parameter so that it can be mocked easily by the unit tests

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/AvroFileAccessor.java
```diff
@@ -122,9 +133,21 @@ public OneRow readNextObject() throws IOException {
      */
     @Override
     public boolean openForWrite() throws Exception {
-        throw new UnsupportedOperationException();
+        // make writer
+        DatumWriter<GenericRecord> datumWriter = new GenericDatumWriter<>(schema);
+        writer = new DataFileWriter<>(datumWriter);
```
denalex:
```suggestion
        writer = new DataFileWriter<>(new GenericDatumWriter<GenericRecord>(schema));
```
to make it a 1-liner since you're not using `datumWriter` anywhere else.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/AvroFileAccessor.java
```diff
@@ -144,27 +169,14 @@ public boolean writeNextObject(OneRow onerow) throws Exception {
...
-                inStream, dummyReader);
-        Schema schema = dummyFileReader.getSchema();
-        dummyFileReader.close();
-        return schema;
+        if (writer != null) {
+            writer.close();
+            totalRowsWritten += rowsWritten;
+        }
+        LOG.debug("Segment {}: writer closed, wrote a TOTAL of {} rows to {} on server {}",
+                context.getSegmentId(),
```
denalex: might be useful to provide more context (like txid and user) to cross correlate. I think we have this SesionID we use for cache keys, not sure if we expose it in the RequestContext. If not, we probably should and standardize our logging on this.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/AvroResolver.java
```diff
@@ -76,21 +76,11 @@
...
-            } else {
-                try (InputStream externalSchema = openExternalSchema()) {
-                    schema = (new Schema.Parser()).parse(externalSchema);
-                }
-            }
-        } catch (Exception e) {
-            throw new RuntimeException("Failed to initialize AvroResolver: " + e.getMessage(), e);
-        }
+        hcfsType = HcfsType.getHcfsType(configuration, context);
+        Schema schema = AvroUtilities.getInstance().obtainSchema(context, configuration);
```
denalex: same comment about singleton pattern

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/AvroResolver.java
denalex: didn't we want to pull the schema from context once upon initialization ?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/AvroResolver.java
```diff
@@ -137,22 +127,18 @@ public void initialize(RequestContext requestContext) {
...
-     */
-    boolean isAvroFile() {
-        return context.getAccessor().toLowerCase().contains("avro");
+    public OneRow setFields(List<OneField> record) {
+        GenericRecord genericRecord = new GenericData.Record((Schema) context.getMetadata());
+        int cnt = 0;
+        for (OneField field : record) {
+            // Avro does not seem to understand regular byte arrays
+            if (field.val instanceof byte[])
+                field.val = ByteBuffer.wrap((byte[]) field.val);
```
denalex: we prefer even 1-liners inside if() to be in { } to avoid errors and improve readability.
```suggestion
                {
                    field.val = ByteBuffer.wrap((byte[]) field.val);
                }
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/AvroUtilities.java
```diff
@@ -0,0 +1,252 @@
...
+        File searchForFile(String filename);
+    }
+
+    // default constructor
+    private AvroUtilities() {
+        fileSearcher = (file) -> {
+            try {
+                return searchForFile(file);
+            } catch (UnsupportedEncodingException e) {
+                e.printStackTrace();
```
denalex: this should be logged, not printed

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/AvroUtilities.java
```diff
@@ -0,0 +1,252 @@
...
+        Schema schema = (Schema) context.getMetadata();
+
+        if (schema != null) {
+            return schema;
+        }
+        try {
+            schemaPath = context.getDataSource();
+            schema = readOrGenerateAvroSchema(context, configuration);
+        } catch (Exception e) {
+            throw new RuntimeException("Failed to obtain Avro schema from '" + schemaPath + "'", e);
```
denalex:
```suggestion
            throw new RuntimeException(String.format("Failed to obtain Avro schema from '%s'", schemaPath), e);
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/AvroUtilities.java
```diff
@@ -0,0 +1,252 @@
...
+        // user-provided schema trumps everything
+        String userProvidedSchemaFile = context.getOption("SCHEMA");
+        if (userProvidedSchemaFile != null) {
+            schemaPath = userProvidedSchemaFile;
+            if (userProvidedSchemaFile.matches("^.*\\.avsc$")) {
+                return readSchemaFromUserProvidedJson(context, configuration, userProvidedSchemaFile);
+            }
+            return readSchemaFromUserProvidedAvroBinary(context, configuration, userProvidedSchemaFile);
+        }
+
```
denalex: in a pure OO design, you would have a `AvroSchemaFileReader` interface, `JsonAvroSchemaFileReader` and `BinaryAvroSchemaFileReader` implementations and a factory that would contain the above `if` statement and give you an instance of one of the implementations to use, so here you would not care which one it actually is.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/AvroUtilities.java
```diff
@@ -0,0 +1,252 @@
...
+        if (context.getProfileScheme() == null) {
+            return new Path(schemaName);
+        }
+        // grab the bucket name from dataSource
+        // read/write are inconsistent about whether context.getDataSource()
+        // includes the profile scheme or not
+        String bucket = context
+                .getDataSource()
+                .replaceFirst("^(.*://)?/*([^:/]+)/.*$", "$2");
+        return new Path(context.getProtocol() + "://" + bucket + schemaName);
```
denalex: I am not sure what we are doing here. Do we assume path to schema does not include bucket name ? If yes, why not ?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/AvroUtilities.java
```diff
@@ -0,0 +1,252 @@
...
+    private static Schema generateSchema(List<ColumnDescriptor> tupleDescription) throws IOException {
+        String colName;
+        int colType;
+
+        Schema schema = Schema.createRecord("tableName", "", COMMON_NAMESPACE, false);
+        List<Schema.Field> fields = new ArrayList<>();
+
+        for (ColumnDescriptor cd : tupleDescription) {
+            colName = cd.columnName();
+            colType = cd.columnTypeCode();
```
denalex: if you inline these 2 vars in the line below, you will not have to declare them and the method becomes shorter

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/AvroFileAccessorTest.java
```diff
@@ -0,0 +1,48 @@
...
+
+import static org.junit.Assert.assertNotNull;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class AvroFileAccessorTest {
+    AvroFileAccessor accessor;
+    RequestContext context;
+    String avroDirectory;
+    @Mock
```
denalex: hmm. I thought for this to work you need `@RunWith(MockitoJUnitRunner.class)` at the class level.

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/AvroFileAccessorTest.java
```diff
@@ -0,0 +1,48 @@
...
+    ConfigurationFactory mockConfigurationFactory;
+
+    @Before
+    public void setup() {
+        accessor = new AvroFileAccessor();
+        context = new RequestContext();
+        Configuration configuration = new Configuration();
+        mockConfigurationFactory = mock(ConfigurationFactory.class);
```
denalex: right, this will not be needed if you use annotation above

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/AvroFileAccessorTest.java
```diff
@@ -0,0 +1,48 @@
...
+        context.setTransactionId("testID");
+        context.setProfileScheme("localfile");
+        context.setRequestType(RequestContext.RequestType.READ_BRIDGE);
+        context.setDataSource(this.getClass().getClassLoader().getResource("avro/").getPath() + "test.avro");
+    }
+
+    @Test
+    public void testInitialize() {
+        accessor.initialize(context);
+        assertNotNull(context.getMetadata());
```
denalex: you probably want to assert more than just not null, like that is actually schema object with expected fields there, no ?

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/utilities/AvroTypeConverter.java
```diff
@@ -0,0 +1,188 @@
...
+
+import org.apache.avro.Schema;
+import org.apache.avro.generic.GenericRecord;
+import org.greenplum.pxf.api.io.DataType;
+import org.greenplum.pxf.api.utilities.ColumnDescriptor;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+public enum AvroTypeConverter {
```
denalex: this is in test package, why do we need such a complex logic here ?

server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/BridgeResource.java
```diff
@@ -97,7 +97,7 @@ public BridgeResource() {
     public Response read(@Context final ServletContext servletContext,
                          @Context HttpHeaders headers) throws Exception {
 
-        RequestContext context = parseRequest(headers);
+        RequestContext context = parseRequest(headers, RequestContext.RequestType.READ_BRIDGE);
```
denalex: the preferred OO-design way would be to define type at the `BaseResource` level and have subclasses provide it as a parameter to `super()` constructor, so that the previous API will stay the same, but the implementation in the superclass will add a second parameter.

server/pxf-service/src/test/java/org/greenplum/pxf/service/HttpRequestParserTest.java
```diff
@@ -59,7 +60,7 @@
     private MultivaluedMap<String, String> parameters;
     private HttpRequestParser parser;
     @Mock
-    private HttpHeaders mockRequestHeaders;
+    private ContainerRequest mockRequestHeaders;
```
denalex: not sure why you needed this

automation/src/test/java/org/greenplum/pxf/automation/features/avro/HdfsReadableAvroTest.java
```diff
@@ -424,22 +537,16 @@ public void avroCodecs() throws Exception {
...
-            } else {
-                Object[] data = FileFormatsUtils.prepareData(
-                        new CustomAvroPreparer(schemaName), 50, dataTable);
-                hdfs.writeAvroFile(fileName, schemaName, codec,
-                        (IAvroSchema[]) data);
-            }
+            Object[] data = FileFormatsUtils.prepareData(
+                    new CustomAvroPreparer(schemaName), 50, dataTable);
+            hdfs.writeAvroFile(fileName, schemaName, codec,
+                    (IAvroSchema[]) data);
```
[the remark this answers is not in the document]
oliverralbertini: I guess it's hard to say since we don't actually test on centos5

automation/tincrepo/main/pxf/features/hdfs/readable/avro/errors/no_schema_file/expected/query01.ans
```diff
@@ -12,13 +12,20 @@
 -- m/(ERROR|WARNING):.*javax.servlet.ServletException: java.lang.Exception:.*/
 -- s/javax.servlet.ServletException: //
 --
--- m/  Description  .*/
--- s/Description  .*/Description  DESCRIPTION/
+-- m/ +(d|D)escription  .*/
+-- s/ +(d|D)escription  .*/ Description DESCRIPTION/
+--
+-- m/ +Failed to obtain Avro schema from 'i_do_not_exist'/
+-- s/ +Failed to obtain Avro schema from 'i_do_not_exist'/ Failed to obtain Avro schema from 'i_do_not_exist'/
```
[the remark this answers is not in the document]
oliverralbertini: No, there is more whitespace in the search string that gets squashed down to a single space in the replacement string.

server/pxf-service/src/test/java/org/greenplum/pxf/service/HttpRequestParserTest.java
oliverralbertini: because only `ContainerRequest` has `getPath()`, not `HttpHeaders`.

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/utilities/AvroTypeConverter.java
oliverralbertini: I actually think it should become production code at some point, but it would require an additional refactor.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/AvroUtilities.java
oliverralbertini: path to schema is given in the URL like this:

```sql
CREATE WRITABLE EXTERNAL TABLE avro_write_with_schema (id INT, foo TEXT[], ints TEXT[])
	LOCATION ('pxf:///tmp/pxf_automation_data/avro/write_array_test_with_schema?PROFILE=hdfs:avro&SCHEMA=generated.avsc')
```
for java classpath and
```sql
CREATE WRITABLE EXTERNAL TABLE avro_write_arrays_with_schema (id INT, foo TEXT[], ints TEXT[])
	LOCATION ('pxf://bucket/tmp/pxf_automation_data/avro/write_array_test_with_schema?PROFILE=hdfs:avro&SCHEMA=/tmp/foobar/generated.avsc')
```
for schema in filesystem or HDFS.

We assume that the data is in the same bucket as the data path.

automation/src/main/java/org/greenplum/pxf/automation/components/hdfs/Hdfs.java
```diff
@@ -350,6 +353,21 @@ public void writeAvroFileFromJson(String pathToFile, String schemaName,
         ReportUtils.stopLevel(report);
     }
 
+    public void writeJsonFileFromAvro(String pathToFile, String pathToJson)
+            throws Exception {
+        Tool tool = new DataFileReadTool();
+        List<String> args = new ArrayList<>();
+        args.add(pathToFile);
+
+        FileOutputStream out = new FileOutputStream(new File(pathToJson));
```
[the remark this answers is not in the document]
oliverralbertini: I added a try-with-resources block like this:

```java
try (PrintStream printStream = new PrintStream(new FileOutputStream(new File(pathToJson)))) {
            tool.run(null, printStream, System.err, args);
        }
```

server/pxf-api/src/main/java/org/greenplum/pxf/api/io/DataType.java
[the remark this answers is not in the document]
oliverralbertini: > what is this type in the db? I coudn't find it in the greenplum source code. Maybe this OID is not a constant, and it varies depending on the greenplum installation?

The OID gets created whenever the user creates the ENUM, so it is wrong to have an enum type in here at all. We should just rely on the default case in which we send text to the external source.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/AvroFileAccessor.java
```diff
@@ -122,9 +133,21 @@ public OneRow readNextObject() throws IOException {
      */
     @Override
     public boolean openForWrite() throws Exception {
-        throw new UnsupportedOperationException();
+        // make writer
+        DatumWriter<GenericRecord> datumWriter = new GenericDatumWriter<>(schema);
+        writer = new DataFileWriter<>(datumWriter);
+        Path file = new Path(hcfsType.getUriForWrite(configuration, context, true) + ".avro");
+        FileSystem fs = file.getFileSystem(jobConf);
+        FSDataOutputStream avroOut = fs.create(file, false);
```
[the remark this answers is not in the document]
oliverralbertini: Using try-with-resources here too:
```java
try (FSDataOutputStream avroOut = fs.create(file, false)) {
            writer.create(schema, avroOut);
        } catch (IOException e) {
            throw e;
        }
        return true;
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/AvroResolver.java
```diff
@@ -233,46 +218,43 @@ int populateRecord(List<OneField> record, Object fieldValue,
...
-                ret = addOneFieldToRecord(record, DataType.INTEGER, value);
+                ret = addOneFieldToRecord(record, DataType.INTEGER, fieldValue);
                 break;
             case DOUBLE:
-                ret = addOneFieldToRecord(record, DataType.FLOAT8, value);
+                ret = addOneFieldToRecord(record, DataType.FLOAT8, fieldValue);
                 break;
             case STRING:
-                value = (fieldValue != null) ? String.format("%s", fieldValue)
+                fieldValue = (fieldValue != null) ? String.format("%s", fieldValue)
```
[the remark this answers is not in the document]
oliverralbertini: I will just use `toString()`

automation/tincrepo/main/pxf/features/hdfs/writable/avro/null_values/runTest.py
```diff
@@ -0,0 +1,12 @@
+from mpp.models import SQLTestCase
+from mpp.models import SQLConcurrencyTestCase
+
+class PxfHdfsAvroWritableNoSchema(SQLConcurrencyTestCase):
```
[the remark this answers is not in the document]
oliverralbertini: I changed the name to match what the test is doing.

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/utilities/AvroTypeConverter.java
```diff
@@ -0,0 +1,188 @@
...
+    BOOLEAN {
+        @Override
+        public DataType getDataType() {
+            return DataType.BOOLEAN;
+        }
+
+        @Override
+        public Object getValue(GenericRecord genericRecord, int columnIndex, Schema.Type type) {
+            return null;
```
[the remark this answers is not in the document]
oliverralbertini: good catch, it used to do something before.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/AvroFileAccessor.java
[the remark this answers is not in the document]
oliverralbertini: That didn't work, so I took your suggestion.

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/AvroFileAccessorTest.java
oliverralbertini: yes, I am now verifying the Schema that we generate.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/AvroUtilities.java
oliverralbertini: it's now assumed that the user will provide the bucket.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/AvroFileAccessor.java
oliverralbertini: I have added that now.

automation/src/test/java/org/greenplum/pxf/automation/features/avro/HdfsWritableAvroTest.java
```diff
@@ -0,0 +1,319 @@
...
+                });
+        p.waitFor();
+        assert p.exitValue() == 0;
```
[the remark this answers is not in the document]
oliverralbertini: switched to reading the files line by line into `JsonNode` objects and `assertEquals()`. This way we can avoid shelling out, avoid the Java assert, and also the order of the objects within each Json no longer matters.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/AvroUtilities.java
oliverralbertini: done.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/BinaryAvroSchemaFileReader.java
```diff
@@ -0,0 +1,46 @@
...
+                fileReader = new DataFileReader<>(inStream, datumReader);
+            } else {
+                fileReader = new DataFileReader<>(file, datumReader);
+            }
+        } finally {
+            if (fileReader != null) {
+                fileReader.close();
+            }
+        }
+        return fileReader.getSchema();
```
denalex: find it a bit unusual that you'd first close the reader and then get the schema, especially without eating anything from the reader, but maybe that's how Avro reader behaves. Also, not null check is either redundant in line 40 or needed here as well, since if the reader is null it will NPE here.

oliverralbertini: good catch, it was meant to happen inside the try block.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/JsonAvroSchemaFileReader.java
```diff
@@ -0,0 +1,37 @@
+package org.greenplum.pxf.plugins.hdfs.utilities;
```
frankgh: I think we should move all these avro classes to the `org.greenplum.pxf.plugins.hdfs.avro` package

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/avro/AvroSchemaFileReaderFactoryTest.java
```diff
@@ -0,0 +1,26 @@
...
+public class AvroSchemaFileReaderFactoryTest {
+
+    private AvroSchemaFileReaderFactory avroSchemaFileReaderFactory;
+
+    @Before
+    public void setup() {
+        avroSchemaFileReaderFactory = new AvroSchemaFileReaderFactory();
+    }
+
+    @Test
```
oliverralbertini: @frankgh @denalex I forgot to add a test for the `AvroSchemaFileReaderFactory`, so here it is.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/JsonAvroSchemaFileReader.java
oliverralbertini: Done.

PR #250, opened 2019-10-10
Use custom user identity per Hadoop server (non-secure)

frankgh:
Allow users to set the `pxf.service.user.name` property to define
the hadoop login user on a per server basis. By default, hadoop
will use the OS user name.

PXF provides a template `$PXF_CONF/templates/pxf-site.xml`.
To configure the `pxf.service.user.name` property follow these
steps:

1. Copy the pxf-site.xml template to your non-secure hadoop
server configuration

        cp $PXF_CONF/templates/pxf-site.xml $PXF_CONF/servers/non-secure-hadoop-server

2. Edit $PXF_CONF/servers/non-secure-hadoop-server-configuration/pxf-site.xml
by replacing the value for the `pxf.service.user.name` property
to your hadoop user.

### DIFF
--- a/concourse/scripts/pxf_common.bash
+++ b/concourse/scripts/pxf_common.bash
@@ -4,6 +4,7 @@ GPHOME=/usr/local/greenplum-db-devel
 PXF_HOME=${GPHOME}/pxf
 MDD_VALUE=/data/gpdata/master/gpseg-1
 PXF_COMMON_SRC_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
+PROXY_USER=${PROXY_USER:-pxfuser}
 
 # on purpose do not call this PXF_CONF so that it is not set during pxf operations
 PXF_CONF_DIR=~gpadmin/pxf
@@ -319,6 +320,9 @@ function init_and_configure_pxf_server() {
 	if [[ ! ${IMPERSONATION} == true ]]; then
 		echo 'Impersonation is disabled, updating pxf-env.sh property'
 		su gpadmin -c "echo 'export PXF_USER_IMPERSONATION=false' >> ${PXF_CONF_DIR}/conf/pxf-env.sh"
+	else
+		cp ${PXF_CONF_DIR}/templates/pxf-site.xml ${PXF_CONF_DIR}/servers/default/pxf-site.xml
+		sed -i -e "s|\${user.name}|${PROXY_USER}|g" ${PXF_CONF_DIR}/servers/default/pxf-site.xml
 	fi
 
 	# update runtime JDK value based on CI parameter
--- a/concourse/scripts/pxf_common.bash
+++ b/concourse/scripts/pxf_common.bash
@@ -5,6 +5,7 @@ PXF_HOME=${GPHOME}/pxf
 MDD_VALUE=/data/gpdata/master/gpseg-1
 PXF_COMMON_SRC_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
 PROXY_USER=${PROXY_USER:-pxfuser}
+PROTOCOL=${PROTOCOL:-}
 
 # on purpose do not call this PXF_CONF so that it is not set during pxf operations
 PXF_CONF_DIR=~gpadmin/pxf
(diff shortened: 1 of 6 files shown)


concourse/scripts/pxf_common.bash
```diff
@@ -4,6 +4,8 @@ GPHOME=/usr/local/greenplum-db-devel
 PXF_HOME=${GPHOME}/pxf
 MDD_VALUE=/data/gpdata/master/gpseg-1
 PXF_COMMON_SRC_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
+PROXY_USER=${PROXY_USER:-pxfuser}
```
denalex: are you sure we want to default to `pxfuser` and not `gpadmin` ? If people turn off impersonation in PXF, and if PXF is run as `gpadmin`, then they would need to have `gpadmin` as the proxy user in hadoop, which I think should be default for our tests

concourse/scripts/pxf_common.bash
```diff
@@ -200,11 +202,11 @@ function setup_impersonation() {
 		echo 'Impersonation is enabled, adding support for gpadmin proxy user'
 		cat > proxy-config.xml <<-EOF
 			<property>
-			  <name>hadoop.proxyuser.gpadmin.hosts</name>
+			  <name>hadoop.proxyuser.pxfuser.hosts</name>
```
denalex: should this be using the variable ?
```suggestion
			  <name>hadoop.proxyuser.${PROXY_USER}.hosts</name>
```

concourse/scripts/pxf_common.bash
```diff
@@ -200,11 +202,11 @@ function setup_impersonation() {
...
 			  <value>*</value>
 			</property>
 			<property>
-			  <name>hadoop.proxyuser.gpadmin.groups</name>
+			  <name>hadoop.proxyuser.pxfuser.groups</name>
```
denalex: same here

server/pxf-service/src/templates/user/templates/pxf-site.xml
```diff
@@ -12,7 +12,7 @@
     </property>
     <property>
         <name>pxf.service.user.name</name>
-        <value>${user}</value>
+        <value>${user.name}</value>
```
denalex: right ! how was it working till this change, though ?

frankgh: we haven't used it for non-kerberos until now. I had to fix it for this work

concourse/scripts/pxf_common.bash
frankgh: hmm, how is our job without impersonation working then?
https://ud.ci.gpdb.pivotal.io/teams/main/pipelines/dev:pxf_fguerrero-custom_username_per_server_5X_STABLE/jobs/test_pxf_no_impersonation/builds/5

frankgh: Ok, I looked at this. If impersonation is disabled, then we don't need to configure a proxy user.

shivzone:
What is the motivation behind having a separate pxf user per server ?

frankgh:
@shivzone the proxy user configured in hadoop might not necessarily be called `gpadmin`. We had   requests for allowing configuring a different proxy user other than gpadmin ([for example](https://groups.google.com/a/greenplum.org/d/msg/gpdb-users/u7QdDEt18QE/54RlOFHjAgAJ)).

With multi hadoop support, we support any proxy user configured in Hadoop for each different server. It's possible that each Hadoop cluster will have a different proxy user configured for PXF.

shivzone:
I initially though of proxy user for pxf as one specific proxy for the entire pxf service. But i see your point. If there is a new hadoop server with an already configured proxyuser say `proxy1` it would be useful to allow them to use that for the given server

concourse/scripts/pxf_common.bash
frankgh: done

concourse/scripts/pxf_common.bash
frankgh: done

PR #251, opened 2019-10-14
Fix regression in *:text:multi profiles when using wildcards

frankgh:
The HdfsFileFragmenter did not support wildcards and it caused
regressions in *:text:multi profiles because the HdfsFileFragmenter was
used for this profiles in favor of HdfsDataFragmenter.

Also, when using S3 Select, wildcards were not supported in the location
path of the create external table definition.

### DIFF
--- a/server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HdfsFileFragmenter.java
+++ b/server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HdfsFileFragmenter.java
@@ -1,12 +1,12 @@
 package org.greenplum.pxf.plugins.hdfs;
 
+import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.FileSystem;
-import org.apache.hadoop.fs.LocatedFileStatus;
 import org.apache.hadoop.fs.Path;
-import org.apache.hadoop.fs.RemoteIterator;
 import org.greenplum.pxf.api.model.Fragment;
 import org.greenplum.pxf.plugins.hdfs.utilities.HdfsUtilities;
 
+import java.io.IOException;
 import java.net.URI;
 import java.util.List;
 
@@ -18,6 +18,18 @@ import java.util.List;
  */
 public class HdfsFileFragmenter extends HdfsDataFragmenter {
 
+    // The hostname is no longer used, hardcoding it to localhost
+    private static final String[] HOSTS = {"localhost"};
+    private static byte[] DUMMY_METADATA;
+
+    static {
+        try {
+            DUMMY_METADATA = HdfsUtilities.prepareFragmentMetadata(0, 0, HOSTS);
+        } catch (IOException ignored) {
+            // Should not fail
+        }
+    }
+
     /**
      * Gets the fragments for a data source URI that can appear as a file name,
      * a directory name or a wildcard. Returns the data fragments in JSON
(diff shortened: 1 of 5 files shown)


frankgh:
@oliverralbertini @denalex green pipelines:
https://ud.ci.gpdb.pivotal.io/teams/main/pipelines/dev:pxf_fguerrero-regression-wildcard-hcfs_5X_STABLE
https://ud.ci.gpdb.pivotal.io/teams/main/pipelines/dev:pxf_fguerrero-regression-wildcard-hcfs_6X_STABLE

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HdfsFileFragmenter.java
```diff
@@ -27,23 +39,36 @@
...
         FileSystem fs = FileSystem.get(URI.create(fileName), configuration);
-        RemoteIterator<LocatedFileStatus> fileStatusListIterator =
-                fs.listFiles(path, false);
-
-        while (fileStatusListIterator.hasNext()) {
-            LocatedFileStatus fileStatus = fileStatusListIterator.next();
-            String sourceName = fileStatus.getPath().toUri().toString();
-            Fragment fragment = new Fragment(sourceName, hosts, dummyMetadata);
-            fragments.add(fragment);
+        FileStatus[] fileStatusList = fs.globStatus(path);
```
denalex: this will theoretically allow for more than just `*` patterns, but I am not sure how URL-friendly are the other pattern characters: https://hadoop.apache.org/docs/current/api/org/apache/hadoop/fs/FileSystem.html#globStatus-org.apache.hadoop.fs.Path-

Also, will it work if the pattern is in the middle of the path ? You have a lot of tests 👍 maybe you already test that scenario ?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HdfsFileFragmenter.java
```diff
@@ -27,23 +39,36 @@
...
+     *
+     * @param fs   the filesystem
+     * @param path the path to list
+     * @throws IOException for any IO error
+     */
+    private void listFragments(FileSystem fs, Path path) throws IOException {
+        FileStatus[] fileStatusList = fs.listStatus(path);
+        for (FileStatus fileStatus : fileStatusList) {
+            if (fileStatus.isDirectory()) {
+                listFragments(fs, fileStatus.getPath());
```
denalex: this might be a problem if there are circular links (might not be a problem in HDFS / S3, but might be on a file system [that we do not actively support]). I'd build a protection by fixing max depth to smth (say 100)  ?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HdfsFileFragmenter.java
frankgh: yeah, I tried the following

```
# CREATE EXTERNAL TABLE file_as_row_multi_files
(text_blob text)
LOCATION (E'pxf://tmp/pxf_automation_data/file_as_row/m*ti/*line?PROFILE=hdfs:text:multi&FILE_AS_ROW=true')
FORMAT 'CSV';
```
and it works!
```
select * from file_as_row_multi_files ;
                          text_blob
--------------------------------------------------------------
 there are two lines in this file                            +

 ""Lorem ipsum dolor sit amet, consectetur adipiscing elit,  +
```

I will add a test in the automation framework

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HdfsFileFragmenter.java
frankgh: we could skip symlinks. `if (fileStatus.isDirectory() && !fileStatus.isSymlink())`
will that be sufficient?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HdfsFileFragmenter.java
frankgh: added a test with a wildcard in the path as well as the name

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HdfsFileFragmenter.java
frankgh: added a max depth of 32

PR #252, opened 2019-10-14
Allow configurable PGPORT

frankgh:
In GPDB, a PORT_BASE [was introduced](https://github.com/greenplum-db/gpdb/commit/704d2bdcab3b3ba413abb1aaebbb365e487ca79c)
for demo cluster. In GPDB 7 base port is 7000.

This commit allows pipelines to define a base port for the GPDB cluster and
default to 15432.

### DIFF
--- a/concourse/deploy
+++ b/concourse/deploy
@@ -89,7 +89,7 @@ def how_to_use_generated_pipeline_message():
             cmd += '    -v gcs-bucket-resources-prod=pivotal-gpdb-concourse-resources-prod\\\n'
     if ARGS.compile_gpdb:
         cmd += '    -v configure_flags=--enable-cassert -v gpaddon-git-branch=master \\\n'
-    cmd += '    -v gpdb-branch=' + ARGS.actual_gpdb_branch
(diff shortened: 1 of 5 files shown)


concourse/deploy
```diff
@@ -184,6 +185,7 @@ if __name__ == "__main__":
         ARGS.ICW_GREEN_BUCKET = 'gpdb5-assert-concourse-builds'
     elif ARGS.gpdb_branch == 'master':
         ARGS.ICW_GREEN_BUCKET = 'gpdb5-assert-concourse-builds'
+        ARGS.PGPORT = '7000'
```
denalex: this is fine, just wonder if it is still correct to set `ARGS.ICW_GREEN_BUCKET = 'gpdb5-assert-concourse-builds` for all non-5x branches/master

oliverralbertini: yeah that's confusing

frankgh:
@oliverralbertini yeah, I have green pipelines for master (problematic branch) and 5X_STABLE (still using 15432)

concourse/deploy
frankgh: yeah, it's the right bucket, but naming should be different ¯\_(ツ)_/¯

PR #253, opened 2019-10-22
Improve performance of HdfsFileFragmenter

frankgh:
While trying the current HdfsFileFragmenter code on a directory with
40K+ files, fs.globStatus(...) was taking over 20 minutes before having
to cancel the query. The new code takes a few seconds to load the
metadata, and allows the query to run to completion.

### DIFF
--- a/server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/HdfsUtilities.java
+++ b/server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/HdfsUtilities.java
@@ -102,7 +102,7 @@ public class HdfsUtilities {
     public static byte[] prepareFragmentMetadata(long start, long length, String[] locations)
             throws IOException {
 
-        ByteArrayOutputStream byteArrayStream = Utilities.writeBaseFragmentInfo(start, length, locations);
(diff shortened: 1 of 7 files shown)


frankgh:
@denalex addressed your feedback

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/HdfsUtilities.java
```diff
@@ -88,23 +88,24 @@ public static boolean isThreadSafe(Configuration config, String dataDir, String
...
+     * hosts) using {@link ObjectOutputStream}.
+     *
+     * @param start     the file split start
+     * @param length    the file split length
+     * @param locations the data node locations for this split
+     * @return byte serialization of the file split
+     * @throws IOException if I/O errors occur while writing to the underlying
+     *                     stream
+     */
     public static byte[] prepareFragmentMetadata(long start, long length, String[] locations)
```
denalex: I would've move this whole function to Utilities and return byte[] since that what Fragment static init is doing anyways. Also an option to define `public static byte[] getEmptyFragmentMetadata()` in Utilities itself to drive static init out of Fragment object which is supposed to be like POJO.

PR #257, opened 2019-10-29
Add Support for Kerberized Hive 3

frankgh:
- Test against a static kerberized ambari server running HDFS 3.1.1 and Hive 3.1.0
- Add groups: multiClusterSecurity, proxySecurity for finer grain test grouping
- Remove `test_pxf_hdp2_secure_no_impersonation` and `test_pxf_hdp2_secure_with_impersonation` jobs from the pipeline as they were running a very basic test with Kerberos. These jobs are now replaced with `test_pxf_dataproc_multinode_with_impersonation` and `test_pxf_dataproc_multinode_no_impersonation`
- Add __UUID__ to the workingDirectory for hdfs

### DIFF
--- a/concourse/pipelines/templates/pxf-tpl.yml
+++ b/concourse/pipelines/templates/pxf-tpl.yml
@@ -15,7 +15,7 @@ groups:
       - test_pxf_hdp2_ubuntu18
 {% endif %}
       - test_pxf_hdp2_multinode_gpdb
-      - test_pxf_secure_multinode_gpdb
+      - test_pxf_dataproc_multinode_with_impersonation
       - test_pxf_s3_no_impersonation
       - test_pxf_s3_with_impersonation
       - test_pxf_no_impersonation
@@ -25,6 +25,7 @@ groups:
       - test_pxf_hdp3_centos7
 {% if pipeline_type == "release" %}
       - test_pxf_hdp3_secure_with_impersonation
+      - test_pxf_dataproc_multinode_no_impersonation
       - test_pxf_hdp2_centos6
       - test_pxf_adl
       - test_pxf_gs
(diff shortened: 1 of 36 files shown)


concourse/pipelines/templates/pxf-tpl.yml
```diff
@@ -49,11 +49,12 @@ groups:
       - test_pxf_hdp2_ubuntu18
 {% endif %}
       - test_pxf_hdp2_multinode_gpdb
-      - test_pxf_secure_multinode_gpdb
+      - test_pxf_dataproc_multinode_with_impersonation
```
denalex: does "dataproc" here means it is also and always "secure" ?

concourse/pipelines/templates/pxf-tpl.yml
```diff
@@ -840,64 +836,27 @@ jobs:
       passed:
       - compile_pxf
       trigger: true
-    - get: gpdb-pxf-dev-centos6-hdp2-secure
-  - task: test_pxf_secure
-    file: pxf_src/concourse/tasks/test_pxf_secure.yml
-    image: gpdb-pxf-dev-centos6-hdp2-secure
-    privileged: true
-    attempts: 2
+    - get: gpdb-pxf-dev-centos6-hdp2-server
```
denalex: do we really need server-based docker to run the tests ? if yes, we also have hdp3-based server docker image

frankgh: we need the server based image for gpdb and singlecluster jars. I switched it to `gpdb-pxf-dev-centos7-hdp3-server`

concourse/pipelines/templates/pxf-tpl.yml
frankgh: added secure for this jobs

PR #260, opened 2019-11-06
A new UGI is created every time in some cases

frankgh:
When security is disabled, a new UserGroupInformation is created every
time when the pxf.service.kerberos.principal or
pxf.service.kerberos.keytab or
hadoop.kerberos.min.seconds.before.relogin properties are set in the
configuration. This can occur when the pxf-site.xml template is copied
into the server configuration of the non-secured server. Even though the
values in these properties are dummy values that take no effect, but
they were still being used by the LoginSession for non-secure servers.

### DIFF
--- a/automation/pom.xml
+++ b/automation/pom.xml
@@ -229,7 +229,7 @@
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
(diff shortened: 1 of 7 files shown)


denalex:
Looks good, let me update this branch with the changes needed for the security case and the test cases, then we can merge it at once.

PR #261, opened 2019-11-06
Impersonation per server

frankgh:
Add tests for impersonation per server

### DIFF
--- a/concourse/scripts/pxf_common.bash
+++ b/concourse/scripts/pxf_common.bash
@@ -331,6 +331,19 @@ function init_and_configure_pxf_server() {
 		fi
 	fi
 
+	if [[ ${IMPERSONATION} == true ]]; then
+		cp -r ${PXF_CONF_DIR}/servers/default ${PXF_CONF_DIR}/servers/default-no-impersonation
+
+		if [[ ! -f ${PXF_CONF_DIR}/servers/default-no-impersonation/pxf-site.xml ]]; then
+			cp ${PXF_CONF_DIR}/templates/pxf-site.xml ${PXF_CONF_DIR}/servers/default-no-impersonation/pxf-site.xml
+		fi
+
+		sed -i \
+			-e "/<name>pxf.service.user.impersonation<\/name>/ {n;s|<value>.*</value>|<value>false</value>|g;}" \
+			-e "/<name>pxf.service.user.name<\/name>/ {n;s|<value>.*</value>|<value>foobar</value>|g;}" \
+			${PXF_CONF_DIR}/servers/default-no-impersonation/pxf-site.xml
+	fi
+
 	# update runtime JDK value based on CI parameter
 	RUN_JDK_VERSION=${RUN_JDK_VERSION:-8}
 	if [[ $RUN_JDK_VERSION == 11 ]]; then
--- a/concourse/scripts/pxf_common.bash
+++ b/concourse/scripts/pxf_common.bash
@@ -331,19 +331,6 @@ function init_and_configure_pxf_server() {
 		fi
 	fi
 
-	if [[ ${IMPERSONATION} == true ]]; then
-		cp -r ${PXF_CONF_DIR}/servers/default ${PXF_CONF_DIR}/servers/default-no-impersonation
-
-		if [[ ! -f ${PXF_CONF_DIR}/servers/default-no-impersonation/pxf-site.xml ]]; then
-			cp ${PXF_CONF_DIR}/templates/pxf-site.xml ${PXF_CONF_DIR}/servers/default-no-impersonation/pxf-site.xml
-		fi
-
-		sed -i \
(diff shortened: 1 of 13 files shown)


frankgh:
This PR needs to be merged after merging https://github.com/greenplum-db/pxf/pull/260

server/pxf-api/src/main/java/org/greenplum/pxf/api/security/SecureLogin.java
```diff
@@ -209,12 +209,13 @@ private LoginSession getServerLoginSession(final String serverName, final String
...
-        long kerberosMinMillisBeforeRelogin = Utilities.isSecurityEnabled(configuration) ?
+        boolean securityEnabled = Utilities.isSecurityEnabled(configuration);
+        long kerberosMinMillisBeforeRelogin = securityEnabled ?
                 PxfUserGroupInformation.getKerberosMinMillisBeforeRelogin(serverName, configuration) : 0L;
         LoginSession expectedLoginSession = new LoginSession(
                 configDirectory,
-                SecureLogin.getServicePrincipal(serverName, configuration),
-                SecureLogin.getServiceKeytab(serverName, configuration),
+                securityEnabled ? SecureLogin.getServicePrincipal(serverName, configuration) : null,
+                securityEnabled ? SecureLogin.getServiceKeytab(serverName, configuration) : null,
```
denalex: this has changed, you need to rebase off master

concourse/scripts/pxf_common.bash
```diff
@@ -534,6 +534,19 @@ function configure_pxf_default_server() {
 			cp /etc/hbase/conf/*-site.xml "${PXF_CONF_DIR}/servers/default"
 		fi
 	fi
+
+	if [[ ${IMPERSONATION} == true ]]; then
+		cp -r ${PXF_CONF_DIR}/servers/default ${PXF_CONF_DIR}/servers/default-no-impersonation
+
+		if [[ ! -f ${PXF_CONF_DIR}/servers/default-no-impersonation/pxf-site.xml ]]; then
```
oliverralbertini: What's the situation where this file would already be there?

If there isn't any, I suggest replacing this `if` and the `sed` call below with:
```bash
sed \
			-e "/<name>pxf.service.user.impersonation<\/name>/ {n;s|<value>.*</value>|<value>false</value>|g;}" \
			-e "/<name>pxf.service.user.name<\/name>/ {n;s|<value>.*</value>|<value>foobar</value>|g;}" \
			"${PXF_CONF_DIR}/templates/pxf-site.xml" > "${PXF_CONF_DIR}/servers/default-no-impersonation/pxf-site.xml"
```

concourse/scripts/test_pxf_multinode.bash
```diff
@@ -111,6 +111,161 @@ function setup_pxf_on_cluster() {
...
+		if [[ ${IMPERSONATION} == true ]]; then
+			cp -r ${PXF_CONF_DIR}/servers/default ${PXF_CONF_DIR}/servers/default-no-impersonation
+
+			if [[ ! -f ${PXF_CONF_DIR}/servers/default-no-impersonation/pxf-site.xml ]]; then
+				cp ${PXF_CONF_DIR}/templates/pxf-site.xml ${PXF_CONF_DIR}/servers/default-no-impersonation/pxf-site.xml
+			fi
+			sed -i \
+			-e '/<name>pxf.service.user.impersonation<\/name>/ {n;s|<value>.*</value>|<value>false</value>|g;}' \
+			-e '/<name>pxf.service.user.name<\/name>/ {n;s|<value>.*</value>|<value>foobar</value>|g;}' \
+			${PXF_CONF_DIR}/servers/default-no-impersonation/pxf-site.xml
```
oliverralbertini: This looks like the same code that's in `pxf_common.bash`. Might be worth refactoring to a function that lives there.

server/pxf-api/src/main/java/org/greenplum/pxf/api/security/SecureLogin.java
frankgh: yes!

concourse/scripts/pxf_common.bash
frankgh: depends on the job we are running, `pxf-site.xml` may or may not exist. We need to be careful not to overwrite it if it already exists .

concourse/scripts/test_pxf_multinode.bash
frankgh: the difference is that this code runs in an ssh block, not sure if it's possible to call a function in an ssh block if the function is not available in that block

oliverralbertini: I would scp pxf_common.bash to the node and source pxf_common.bash in the ssh block. but feel free to just leave it as is, we can address this at a later time.

frankgh: I expect to see this code refactored at a later point in the future :) most likely you will take care of make it into proper bash

PR #263, opened 2019-11-18
log re-login message at INFO level

denalex:


denalex:
@oliverralbertini -- we log the principal name, which actual user makes the request is irrelevant in this case.

PR #267, opened 2019-11-23
Log the number of fragments retrieved just once

frankgh:
When the fragment cache is enabled, we are logging the number of
fragments returned by the fragmenter for every thread requesting the
fragment list. We observed in large clusters this was producing a large
amount of information being logged at the info level. With this
approach, only the thread performing the fragmentation call will log at
info level, while other threads should still log at debug level.

When the fragment cache is disabled, all threads will log this
information at info level.

### DIFF
--- a/server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/FragmenterResource.java
+++ b/server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/FragmenterResource.java
@@ -20,6 +20,7 @@ package org.greenplum.pxf.service.rest;
  */
 
 import com.google.common.util.concurrent.UncheckedExecutionException;
+import org.apache.log4j.Level;
 import org.greenplum.pxf.api.model.Fragment;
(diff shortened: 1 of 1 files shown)


server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/FragmenterResource.java
```diff
@@ -107,12 +111,14 @@ public Response getFragments(@Context final ServletContext servletContext,
         List<Fragment> fragments;
 
         if (Utilities.isFragmenterCacheEnabled()) {
+            final boolean[] didThreadProcessFragmentCall = {false};
```
denalex: wonder if you can move `boolean didProcess = false` to become class field and then I think it doesn't have to be final (only local vars need to be final ?) to get set in the closure, so you would not need to wrap it into array.

frankgh: done!

PR #268, opened 2019-11-26
fallback on using LineRecordReader when reading encrypted files

denalex:


frankgh:
Feel free to merge it after you restore the logic to use ChunkRecordReader

PR #270, opened 2019-11-27
Host information is not needed for the split information

frankgh:
This is a minor optimization, no need to get the hosts as they are most likely hardcoded and not needed for the split information


frankgh:
green pipeline https://ud.ci.gpdb.pivotal.io/teams/main/pipelines/dev:pxf_fguerrero-host-info-not-needed_5X_STABLE

PR #271, opened 2019-11-28
Prevent make from running with bad PROTOCOL

oliverralbertini:
Previously if you ran `make -C automation` with env var PROTOCOL set to
something not quite correct (S3 instead of s3) you got a mysterious
error from Amazon. Now the Makefile will prevent you from making that
mistake.

### DIFF
--- a/automation/Makefile
+++ b/automation/Makefile
@@ -27,6 +27,9 @@ else
 	MAVEN_TEST_OPTS+= -o
 endif
 
+# lowercase the protocol
+PROTOCOL := $(shell echo $(PROTOCOL) | tr A-Z a-z)
+
 ifeq "$(PROTOCOL)" "minio"
 	MINIO=true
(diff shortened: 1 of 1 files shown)


automation/Makefile
```diff
@@ -27,6 +27,12 @@ else
 	MAVEN_TEST_OPTS+= -o
 endif
 
+# lowercase the protocol
+PROTOCOL_LC := $(shell echo $(PROTOCOL) | tr A-Z a-z)
+ifneq "$(PROTOCOL_LC)" "$(PROTOCOL)"
+$(error Use lower case when defining PROTOCOL; $(PROTOCOL) should be $(PROTOCOL_LC))
```
denalex: why not force convert to lower case and use it instead of giving the error ?

oliverralbertini: 👍 this has been done.

PR #281, opened 2019-12-24
gitignore: ignore dev pipeline yml files

oliverralbertini:

### DIFF
--- a/concourse/pipelines/.gitignore
+++ b/concourse/pipelines/.gitignore
@@ -1,0 +1,1 @@
+pxf_pipeline-dev-*.yml


.gitignore
```diff
@@ -6,3 +6,4 @@
 server/*/out
 server/tmp
 server/pxf-cli/go/pkg
+concourse/pipelines/pxf_pipeline-dev*.yml
```
denalex: how are you getting these ? When I do deploy from concourse directory I get bunch of `.adenissov-*.yml` files generatd, but they are already hidden by this file: https://github.com/greenplum-db/pxf/blob/master/concourse/.gitignore

PR #283, opened 2019-12-27
Fix parquet write decimal

frankgh:
Writing decimals to parquet produces the incorrect parquet file, and it cannot be read back correctly again.


server/build.gradle
```diff
@@ -188,6 +188,7 @@ project('pxf-hdfs') {
         compile "org.apache.parquet:parquet-encoding:${parquetVersion}"
         compile "org.apache.parquet:parquet-generator:${parquetVersion}"
         compile "org.apache.parquet:parquet-hadoop:${parquetVersion}"
+        compile "org.apache.hive:hive-storage-api:2.4.0" // for parquet writing
```
denalex: new dependency that we ship ?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetFileAccessor.java
```diff
@@ -67,15 +68,23 @@
...
     private static final CompressionCodecName DEFAULT_COMPRESSION = CompressionCodecName.SNAPPY;
 
+    public static final int PRECISION_TO_BYTE_COUNT[] = new int[38];
+
+    static {
+        for (int prec = 1; prec <= 38; prec++) {
+            // Estimated number of bytes needed.
+            PRECISION_TO_BYTE_COUNT[prec - 1] = (int)
+                    Math.ceil((Math.log(Math.pow(10, prec) - 1) / Math.log(2) + 1) / 8);
+        }
```
denalex: was this copied from somewhere ?

server/build.gradle
frankgh: not really, this dependency already existed for hive

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetFileAccessor.java
frankgh: yes. I will add a comment

frankgh:
@denalex DECIMAL/NUMERIC types that defined a precision/scale were not working. ie

```
CREATE WRITABLE EXTERNAL TABLE foo (
a NUMERIC (5,2)
) .....
```

This PR fixes NUMERIC/DECIMAL types that have a precision/scale defined

PR #286, opened 2020-01-06
Parquet record filter

frankgh:
Enable pushing down of predicates to parquet. Note that filtering
happens after reading the file on PXF. This incurs in network traffic
by bringing the entire parquet file into the PXF JVM, but CPU and I/O
are reduced because Parquet skips reading groups based on
statistics and other factors.

**Pending in this PR**:
- ~~Pruning filters of INT96 and FIXED_LEN_BYTE_ARRAY types~~
- ~~Unit tests for column projection~~
- ~~Integration tests~~

**Open questions**:
- testing comparing strings with trailing white spaces


server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetRecordFilterBuilder.java
```diff
@@ -0,0 +1,273 @@
...
+
+    /**
+     * Returns the built record filter
+     *
+     * @return the built record filter
+     */
+    public FilterCompat.Filter getRecordFilter() {
+        FilterPredicate predicate = filterQueue.poll();
+        if (!filterQueue.isEmpty()) {
+            throw new IllegalStateException("Filter queue is not empty after visiting all nodes");
```
oliverralbertini: How do we know all the nodes have been visited?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetRecordFilterBuilder.java
```diff
@@ -0,0 +1,273 @@
...
+    /**
+     * Returns the FilterPredicate function that supports equals and not equals
+     * for the given operator
+     *
+     * @param operator the operator
+     * @param <T>      the type
+     * @param <C>      the column type
+     * @return the FilterPredicate function
+     */
+    private static <T extends Comparable<T>, C extends Operators.Column<T> & Operators.SupportsEqNotEq> BiFunction<C, T, FilterPredicate> getOperatorWithEqNotEqSupport(Operator operator) {
```
oliverralbertini: what?

server/pxf-hdfs/src/test/resources/parquet/generate_parquet_types.bash
```diff
@@ -0,0 +1,23 @@
...
+$HDFS_CMD dfs -rm -r -f "$HDFS_DIR"
+$HDFS_CMD dfs -mkdir -p "$HDFS_DIR"
+# Copy source CSV file to HDFS
+$HDFS_CMD dfs -copyFromLocal "$SRC_DIR/parquet_types.csv" "$HDFS_DIR"
+# Run the HQL file
+$HIVE_CMD -f "$SRC_DIR/generate_parquet_types.hql"
+
+rm -f "$SRC_DIR/parquet_types.parquet"
+# Copy file to the directory where this script resides
+$HDFS_CMD dfs -copyToLocal /hive/warehouse/parquet_types/000000_0 "$SRC_DIR/parquet_types.parquet"
```
oliverralbertini: `HDFS_CMD` and `HIVE_CMD` should all be quoted here (lines 14, 15, 17, 19, 23).

oliverralbertini: I think the point of this script is to re-create `parquet_types.parquet` using PXF. Should we do a diff somewhere?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetTypeConverter.java
```diff
@@ -68,7 +69,11 @@ public Object getValue(Group group, int columnIndex, int repeatIndex, Type type)
         @Override
         public DataType getDataType(Type type) {
             OriginalType originalType = type.getOriginalType();
-            if (originalType == OriginalType.INT_8 || originalType == OriginalType.INT_16) {
+            if (originalType == OriginalType.DATE) {
+                return DataType.DATE;
+            } else if (originalType == OriginalType.DECIMAL) {
+                return DataType.NUMERIC;
+            } else if (originalType == OriginalType.INT_8 || originalType == OriginalType.INT_16) {
```
oliverralbertini: We add support for date and decimal types?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetTypeConverter.java
```diff
@@ -94,12 +105,22 @@ public void addValueToJsonArray(Group group, int columnIndex, int repeatIndex, T
...
         }
 
         @Override
         public Object getValue(Group group, int columnIndex, int repeatIndex, Type type) {
-            return group.getLong(columnIndex, repeatIndex);
+            long value = group.getLong(columnIndex, repeatIndex);
+            OriginalType originalType = type.getOriginalType();
+            if (originalType == OriginalType.DECIMAL) {
+                int scale = type.asPrimitiveType().getDecimalMetadata().getScale();
+                return new BigDecimal(BigInteger.valueOf(value), scale);
```
oliverralbertini: maybe we should refactor this to a function that takes a long:
```java
private BigDecimal bigDecimalFromLong(long val)
```
and above (int32) you could cast the int to long.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/SupportedParquetPrimitiveTypePruner.java
```diff
@@ -0,0 +1,75 @@
...
+
+        return super.visit(node, level);
+    }
+
+    private PrimitiveType.PrimitiveTypeName getPrimitiveType(OperatorNode operatorNode) {
+        ColumnIndexOperandNode columnIndexOperand = operatorNode.getColumnIndexOperand();
+        ColumnDescriptor columnDescriptor = columnDescriptors.get(columnIndexOperand.index());
+        String filterColumnName = columnDescriptor.columnName();
+        Type type = fields.get(filterColumnName);
+        return type.asPrimitiveType().getPrimitiveTypeName();
```
oliverralbertini: this seems like a lot of work to get the type for a node. Is it possible that this could be cached on the node itself?

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetBaseTest.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    private Map<String, Type> getOriginalFieldsMap(MessageType originalSchema) {
+        Map<String, Type> originalFields = new HashMap<>(originalSchema.getFieldCount() * 2);
+
+        // We need to add the original name and lower cased name to
+        // the map to support mixed case where in GPDB the column name
+        // was created with quotes i.e "mIxEd CaSe". When quotes are not
+        // used to create a table in GPDB, the name of the column will
+        // always come in lower-case
```
oliverralbertini: interesting!

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetBaseTest.java
```diff
@@ -0,0 +1,83 @@
...
+        originalSchema.getFields().forEach(t -> {
+            String columnName = t.getName();
+            originalFields.put(columnName, t);
+            originalFields.put(columnName.toLowerCase(), t);
+        });
+
+        return originalFields;
+    }
```
oliverralbertini: no tests?

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetFilterPushDownTest.java
```diff
@@ -0,0 +1,836 @@
...
+        context.setFilterString("a5c16s4dtrueo0");
+        helper(expectedRows);
+
+        // a5 <> true
+        expectedRows = new int[]{1, 3, 5, 7, 9, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25};
+        context.setFilterString("a5c16s4dtrueo6");
+        helper(expectedRows);
+
+        // a5 == false
+        expectedRows = new int[]{1, 3, 5, 7, 9, 11, 12, 13, 14, 15, 16, 17, 18, 20, 21, 22, 23, 24, 25,};
```
oliverralbertini: oh my gosh an extra comma

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetFilterPushDownTest.java
```diff
@@ -0,0 +1,836 @@
...
+        context.setTupleDescription(columnDescriptors);
+
+        accessor.initialize(context);
+        resolver.initialize(context);
+    }
+
+    @Test
+    public void testNoFilter() throws Exception {
+        // all rows are expected
+        helper(COL1);
```
oliverralbertini: I like this way of testing. Can we make an alias for `COL1` called `ALL`? `helper()` could then be named `assertRowsReturned()`. This line would then read something like `assertRowsReturned(ALL)` which would be easier to understand than `helper(COL1)`.

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetRecordFilterBuilderTest.java
```diff
@@ -0,0 +1,180 @@
...
+    @Rule
+    public ExpectedException thrown = ExpectedException.none();
+
+    @Test
+    public void testUnsupportedOperationError() throws Exception {
+        thrown.expect(UnsupportedOperationException.class);
+        thrown.expectMessage("not supported IN");
+
+        // a16 in (11, 12)
+        helper("a16m1007s2d11s2d12o10");
```
oliverralbertini: In this case you might rename `helper` to `filterBuilderFromFilterString()`

concourse/scripts/pxf-perf-multi-node.bash
```diff
@@ -205,10 +205,20 @@ EOF
...
+    local run_id=${2}
+    # create text tables
     readable_external_table_text_query "${name}" "pxf://data-gpdb-ud-tpch/${SCALE}/lineitem_data/?PROFILE=gs:text&SERVER=gsbenchmark"
-    writable_external_table_text_query "${name}" "pxf://data-gpdb-ud-pxf-benchmark/output/${SCALE}/${UUID}/?PROFILE=gs:text&SERVER=gsbenchmark"
+    writable_external_table_text_query "${name}" "pxf://data-gpdb-ud-pxf-benchmark/output/${SCALE}/${UUID}-${run_id}/?PROFILE=gs:text&SERVER=gsbenchmark"
+}
+
+function create_gcs_parquet_tables() {
+    local name=${1}
+    local run_id=${2}
```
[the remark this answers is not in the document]
frankgh: we didn't run concurrency for GCS, but this allows us to run concurrency. The run_id is used to get a unique path for the execution of multiple queries

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetFileAccessor.java
```diff
@@ -86,53 +98,66 @@
...
-        } catch (Exception e) {
-            fileReader.close();
-            throw new IOException(e);
-        }
+        // Read the original schema from the parquet file
+        MessageType originalSchema = getSchema(file, fileSplit);
+        // Get a map of the column name to Types for the given schema
+        Map<String, Type> originalFieldsMap = getOriginalFieldsMap(originalSchema);
+        // Get the read schema in case of column projection
+        MessageType readSchema = buildReadSchema(originalFieldsMap, originalSchema);
```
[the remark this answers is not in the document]
frankgh: I'd prefer to keep it as `readSchema` because it's the schema we are reading. If there's no projection, it will be end up being all the columns of the Greenplum schema. I realize that the comment might not be accurate though. What do you think?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetFileAccessor.java
```diff
@@ -145,36 +170,16 @@ public boolean openForRead() throws IOException {
...
-        rowsInRowGroup = currentRowGroup.getRowCount();
-
-        LOG.debug("Reading {} rows (rowgroup {})", rowsInRowGroup, rowGroupsReadCount);
-        return true;
-    }
-
-    private int getOption(String optionName, int defaultValue) {
-        String optionStr = context.getOption(optionName);
-        return optionStr != null ? Integer.parseInt(optionStr) : defaultValue;
+        return null;
```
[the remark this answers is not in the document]
frankgh: definitely, the new Parquet APIs help!

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetFileAccessor.java
```diff
@@ -186,7 +191,19 @@ private int getOption(String optionName, int defaultValue) {
...
+            long average = totalReadTimeInNanos / totalRowsRead;
+            LOG.debug("{}-{}: Read TOTAL of {} rows from file {} on server {} in {} ms. Average speed: {} nanoseconds",
+                    context.getTransactionId(),
+                    context.getSegmentId(),
+                    totalRowsRead,
+                    context.getDataSource(),
+                    context.getServerName(),
+                    millis,
+                    average);
+        }
```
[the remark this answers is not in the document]
frankgh: yeah, I've been thinking about this. This definitely helps debug things when you have a large greenplum cluster with concurrent queries. We definitely need better instrumentation to trace the source of the error.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetFileAccessor.java
```diff
@@ -269,19 +287,94 @@ public void closeForWrite() throws IOException {
...
+                fileSplit.getStart(), fileSplit.getStart() + fileSplit.getLength());
+        ParquetReadOptions parquetReadOptions = HadoopReadOptions
+                .builder(configuration)
+                .withMetadataFilter(filter)
+                .build();
+        HadoopInputFile inputFile = HadoopInputFile.fromPath(parquetFile, configuration);
+        try (ParquetFileReader parquetFileReader =
+                     ParquetFileReader.open(inputFile, parquetReadOptions)) {
+            FileMetaData metadata = parquetFileReader.getFileMetaData();
+            if (LOG.isDebugEnabled()) {
```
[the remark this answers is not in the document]
frankgh: I assume that `getRecordCount()` and `getRowGroups()` might be expensive calls, so I prefer to avoid calling them at debug level, when debugging is not on

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetRecordFilterBuilder.java
frankgh: We rely on TreeTraverser to be correct. TreeTraverser has a set of unit tests that should give us some degree of guarantee that we'll visit all the nodes. In theory, there shouldn't be a situation where the queue is not empty after visiting all nodes

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetRecordFilterBuilder.java
frankgh: :) generics. This was a refactoring I did because the code was a bit messy when Alex and I reviewed it. Basically, this will give you the predicate for the given operator

server/pxf-hdfs/src/test/resources/parquet/generate_parquet_types.bash
frankgh: actually, the point of the script is to generate `parquet_types.parquet` without PXF. We use Hive directly to generate the file (take a look at the `.hql` files in the PR). The idea is that we assume that Hive is doing the right thing when generating the parquet file, so we don't rely on PXF.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetTypeConverter.java
frankgh: hmm, good point

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/SupportedParquetPrimitiveTypePruner.java
frankgh: Well, the node is agnostic to parquet types. These nodes are the result of the expression tree that we obtain from greenplum. We look up the types by name in a hashmap. I think hashmaps should be fast enough for this operation

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetBaseTest.java
frankgh: some of the tests in this namespace share the same schema for test `ParquetBaseTest`. To avoid code duplication, I created a base class where all the other classes that need the schema inherit.

frankgh:
thanks for the review @oliverralbertini , I will be addressing your comments

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetFilterPushDownTest.java
frankgh: how do you even see this!

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetFilterPushDownTest.java
frankgh: great suggestion

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetTypeConverter.java
frankgh: so date is serialized as an integer by hive, and this supports reading from parquet file generated by hive.

frankgh:
@denalex @oliverralbertini I'm merging this today if there are no objections

PR #287, opened 2020-01-08
Update GCS connector jar in automation/prod

oliverralbertini:
This updated version handles filenames with spaces inside correctly.
The error that the old connector gives looks like this:

java.net.URISyntaxException: Illegal character in path


oliverralbertini:
Let me know if and when it's OK to merge.

frankgh:
I think it should be okay to merge this now

PR #288, opened 2020-01-09
Minor code optimizations

frankgh:
While profiling the code, I noticed some hotspots in

DataType#isTextForm: Optimized by using an array, iterating over a small
array is significantly faster than using a HashSet.

GreenplumCSV#toCsvField: I noticed a high amount of calls to
String.length. Assign the length to a variable once and use it to
optimize it.

### DIFF
--- a/server/pxf-api/src/main/java/org/greenplum/pxf/api/io/DataType.java
+++ b/server/pxf-api/src/main/java/org/greenplum/pxf/api/io/DataType.java
@@ -64,7 +64,8 @@ public enum DataType {
     UNSUPPORTED_TYPE(-1);
 
     private static final Map<Integer, DataType> lookup = new HashMap<>();
-    private static final Set<Integer> notText = new HashSet<>();
+    private static final int[] NOT_TEXT = {BIGINT.OID, BOOLEAN.OID, BYTEA.OID,
+            FLOAT8.OID, INTEGER.OID, REAL.OID, SMALLINT.OID};
 
     static {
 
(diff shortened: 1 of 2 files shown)


server/pxf-api/src/main/java/org/greenplum/pxf/api/io/DataType.java
```diff
@@ -64,7 +64,8 @@
     UNSUPPORTED_TYPE(-1);
 
     private static final Map<Integer, DataType> lookup = new HashMap<>();
-    private static final Set<Integer> notText = new HashSet<>();
+    private static final int[] NOT_TEXT = {BIGINT.OID, BOOLEAN.OID, BYTEA.OID,
+            FLOAT8.OID, INTEGER.OID, REAL.OID, SMALLINT.OID};
```
denalex: wonder how BitSet would behave, the largest int value is 701 (FLOAT8), which will make BitSet-backing array about 88 bytes long, which is not a big deal and checking is a binary operation (actually check : isText = (oid > set.size()-1 || !set.get(oid))

Also, there are other OIDs (like NUMERIC(1700)) that are not lister here, but probably are not text also ?

server/pxf-api/src/main/java/org/greenplum/pxf/api/model/GreenplumCSV.java
```diff
@@ -174,28 +174,28 @@ public String toCsvField(String s,
         if (s == null) return null;
 
         final int length = s.length();
-        int i, j, quotes = 0, specialChars = 0, pos = 0, total = length;
+        int i, j, quotes = 0, specialChars = 0, pos = 0, total = length,
+                newLineLength = newline != null ? newline.length() : 0;
```
denalex: even better to move to constructor and to `withNewLine()` method so it is analyzed only once per instance.

server/pxf-api/src/main/java/org/greenplum/pxf/api/io/DataType.java
frankgh: Well, the array only uses 4 bytes/element * 7 elements = 28 bytes. I think speed-wise, the speed difference between BitSet#get vs iterating over a small array should be negligible. I'm not sure why NUMERIC is not included, will add it and see if something breaks

frankgh: Ok, it turns out that NUMERIC is treated as TEXT because GPDBWritable serializes it as a string:
https://github.com/greenplum-db/pxf/blob/master/server/pxf-service/src/main/java/org/greenplum/pxf/service/BridgeOutputBuilder.java#L409

PR #290, opened 2020-01-22
Log ClientAbortException at debug level

frankgh:
While running a large cluster and queries fail, ClientAbortExceptions
become intrusive in the error log. To better debug queries while running
on large clusters (10s of segments per segment host), log the stacktrace
at the debug level, and give hints how to surface the exception.

### DIFF
--- a/server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/BridgeResource.java
+++ b/server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/BridgeResource.java
@@ -142,7 +142,9 @@ public class BridgeResource extends BaseResource {
                     LOG.debug("Finished streaming fragment {} of resource {}, {} records.", fragment, dataDir, recordCount);
                 } catch (ClientAbortException e) {
                     // Occurs whenever client (GPDB) decides to end the connection
-                    LOG.error("Remote connection closed by GPDB", e);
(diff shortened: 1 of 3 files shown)


server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/BridgeResource.java
```diff
@@ -142,7 +142,9 @@ public void write(final OutputStream out) throws IOException,
                     LOG.debug("Finished streaming fragment {} of resource {}, {} records.", fragment, dataDir, recordCount);
                 } catch (ClientAbortException e) {
                     // Occurs whenever client (GPDB) decides to end the connection
-                    LOG.error("Remote connection closed by GPDB", e);
+                    LOG.error("Remote connection closed by GPDB (Enable debug for additional details)");
+                    // Stacktrace in debug
+                    LOG.debug("Details: Remote connection closed by GPDB", e);
```
denalex: good idea, but maybe log one message or another, depending on the log level. Having 2 log lines for DEBUG level will be messy when all threads are logging, it will be hard to cross-correlate

oliverralbertini: that's a good point, Alex.

frankgh: done

PR #293, opened 2020-01-28
Use Zonal DNS when accessing GCP servers

frankgh:
Our GCP dataproc instances are being deployed with `VmDnsSetting= ZonalOnly`, but our scripts are using `Global (project wide) DNS`.
https://cloud.google.com/compute/docs/internal-dns


frankgh:
Closing this PR as it seems that dataproc behavior went back to what we expect.

PR #294, opened 2020-01-29
Update pg regress pipeline

oliverralbertini:
The pg_regress pipeline had become a bit stale and these changes help
bring it back up to speed.

### DIFF
--- a/concourse/README.md
+++ b/concourse/README.md
@@ -163,7 +163,7 @@ fly -t ud set-pipeline -p pg_regress \
 	-l ~/workspace/gp-continuous-integration/secrets/gpdb6-integration-testing.dev.yml \
 	-l ~/workspace/gp-continuous-integration/secrets/ccp-integration-pipelne-secrets.yml \
 	-l ~/workspace/gp-continuous-integration/secrets/gpdb_common-ci-secrets.yml \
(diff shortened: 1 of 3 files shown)


oliverralbertini:
There is one change for the `compile_pxf` step to pass, we run the compile step twice. But this is not intended to be merged in.

concourse/scripts/compile_pxf.bash
```diff
@@ -9,7 +9,9 @@ PXF_ARTIFACTS_DIR=${PWD}/${OUTPUT_ARTIFACT_DIR}
 bash --login -c "
 	export PXF_HOME=${GPHOME}/pxf BUILD_NUMBER=${TARGET_OS}
 	make -C '${PWD}/pxf_src' test install
-"
+" || bash --login -c "
```
frankgh: what is this doing? Running tests twice?

oliverralbertini: It's been removed now. I have seen some green builds with only one attempt, but we should still keep an eye on this. It looks to be kind of flaky right now.

oliverralbertini:
Pipeline is looking good at the moment, so will merge. Thanks for the review, @frankgh.

PR #295, opened 2020-02-03
Use the first version of Guava that has a stable Cache API

frankgh:
Currently, we are using an old version of Guava that does not provide a
stable cache API and warnings are showing in IDE and during compilation.
The API became stable starting version 19.0


frankgh:
green pipelines:
https://ud.ci.gpdb.pivotal.io/teams/main/pipelines/dev:pxf_fguerrero-guava-stable-cache-api_5X_STABLE

https://ud.ci.gpdb.pivotal.io/teams/main/pipelines/dev:pxf_fguerrero-guava-stable-cache-api_6X_STABLE

PR #296, opened 2020-02-03
Update pxf_pr pipeline to not use Greenplum master

oliverralbertini:
We recently merged a PR to Greenplum master which removes the PXF
external table extension in favor of FDW. To continue testing we should
use 6X_STABLE in our pxf_pr pipelines.


oliverralbertini:
I have already flown the pipelines with this change on hush-house as well as UD's CI deployment.
Thanks for the help, @divyabhargov !

PR #297, opened 2020-02-07
CI: Do not trigger pxf build on every commit to gpdb master [ci_skip]

frankgh:
Currently, we trigger a PXF build for the pg_regress pipeline on every
commit to GPDB. Adding a trigger to execute on changes to gpcontrib/pxf_fdw.

We still trigger a build on GPDB for every commit, but it should be
limited to changes to FDW and Copy in the future.

### DIFF
--- a/concourse/README.md
+++ b/concourse/README.md
@@ -159,11 +159,11 @@ For example, you may want to work off of a development branch for PXF or Greenpl
 
 ```
 fly -t ud set-pipeline -p pg_regress \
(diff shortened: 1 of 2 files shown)


oliverralbertini:
Did you update the pipeline with this change?

frankgh:
yes! I did

oliverralbertini:
Here's the [pg_regress pipeline](https://ud.ci.gpdb.pivotal.io/teams/main/pipelines/pg_regress).

PR #298, opened 2020-02-10
Implement `pxf cluster restart`

oliverralbertini:

### DIFF
--- a/cli/go/src/pxf-cli/cmd/cluster.go
+++ b/cli/go/src/pxf-cli/cmd/cluster.go
@@ -38,13 +38,14 @@ func createCobraCommand(use string, short string, command *pxf.Command) *cobra.C
 }
 
 var (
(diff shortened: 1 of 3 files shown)


oliverralbertini:
I tried this out on `phoenix` cluster: 
image

denalex:
It says "Restarting on master and 5 other hosts" and then "Restarted on 5 hosts". So what about master ?

oliverralbertini:
> It says "Restarting on master and 5 other hosts" and then "Restarted on 5 hosts". So what about master ?

Great catch, yes it was an issue with wording, which I fixed on a force push, see [here](https://github.com/greenplum-db/pxf/compare/407f84791af976acec5d6e7d60146d2f27a594e7..e118bc67796534b01cc8b3a7de2214403e4936ad).

This mistake was from using the `pxf cluster reset` command as a template. BTW that cluster has 5 hosts including master, but master also has segments. Should not mention master but rather 5 hosts.

PR #299, opened 2020-02-10
Log org.apache.parquet at WARN level [ci_skip]

frankgh:
After implementing parquet predicate pushdown, an increase of INFO level
log entries are appearing when there is a predicate filter. This commit
defaults the level to WARN for org.apache.parquet in the
pxf-log4j.properties.

### DIFF
--- a/server/pxf-service/src/templates/user/conf/pxf-log4j.properties
+++ b/server/pxf-service/src/templates/user/conf/pxf-log4j.properties
@@ -27,7 +27,7 @@ log4j.logger.org.apache.hadoop=WARN
 log4j.logger.org.apache.zookeeper=WARN
 log4j.logger.hive.metastore=WARN
 
-log4j.logger.org.apache.parquet.hadoop=WARN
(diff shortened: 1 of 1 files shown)


server/pxf-service/src/templates/user/conf/pxf-log4j.properties
```diff
@@ -26,6 +26,7 @@ log4j.logger.com.google.cloud.hadoop.fs.gcs.GoogleHadoopFileSystemBase=WARN
 log4j.logger.org.apache.hadoop=WARN
 log4j.logger.org.apache.zookeeper=WARN
 log4j.logger.hive.metastore=WARN
+log4j.logger.org.apache.parquet=WARN
 
 log4j.logger.org.apache.parquet.hadoop=WARN
```
denalex: then remove this line

frankgh: done

PR #300, opened 2020-02-10
Parquet Filter: Pad the predicate value to the length of the char field

frankgh:
While querying parquet files generated by PXF, and in the presence of a
predicate for a char field type, the predicate does not match the value
when the predicate value has length < char field length. When these
predicates are present, the query does not return any results.

To illustrate the issue, let's consider the following example:

```
CREATE WRITABLE EXTERNAL TABLE foo_w (a char(3))
LOCATION ('pxf://tmp/foo/?PROFILE=hdfs:parquet&SERVER=local')

INSERT INTO foo_w VALUES ('s');
```

The commands above will generate a parquet file with a field of value
's  ' (with 2 blank characters at the end). When we query this data we
get the following:

```
CREATE EXTERNAL TABLE foo_r (a char(3))
LOCATION ('pxf://tmp/foo/?PROFILE=hdfs:parquet&SERVER=local')
FORMAT 'CUSTOM' (formatter='pxfwritable_import');

SELECT * FROM foo_r; -- Produces 1 row

SELECT * FROM foo_r WHERE a = 's'; -- Produces 0 rows

SELECT * FROM foo_r WHERE a = 's  '; -- Produces 1 row
```

To mantain consistency with Greenplum's behavior we need to pad the
predicate value for char fields with a defined length to match the
length of the field.

### DIFF
--- a/server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetRecordFilterBuilder.java
+++ b/server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetRecordFilterBuilder.java
@@ -1,5 +1,6 @@
 package org.greenplum.pxf.plugins.hdfs.parquet;
 
+import org.apache.commons.lang.StringUtils;
 import org.apache.parquet.filter2.compat.FilterCompat;
 import org.apache.parquet.filter2.predicate.FilterApi;
 import org.apache.parquet.filter2.predicate.FilterPredicate;
@@ -13,6 +14,7 @@ import org.greenplum.pxf.api.filter.OperandNode;
 import org.greenplum.pxf.api.filter.Operator;
 import org.greenplum.pxf.api.filter.OperatorNode;
 import org.greenplum.pxf.api.filter.TreeVisitor;
+import org.greenplum.pxf.api.io.DataType;
 import org.greenplum.pxf.api.utilities.ColumnDescriptor;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetRecordFilterBuilder.java
```diff
@@ -174,8 +176,22 @@ private void processSimpleColumnOperator(OperatorNode operatorNode) {
...
             case BINARY:
+                String value = null;
+
+                if (valueOperand != null) {
+                    value = valueOperand.toString();
+
+                    // Special case for char, we need to right pad the value to the width of the char
+                    if (columnDescriptor.getDataType() == DataType.BPCHAR &&
+                            columnDescriptor.columnTypeModifiers() != null &&
+                            columnDescriptor.columnTypeModifiers().length > 0) {
```
denalex: ArrayUtils.isNotEmpty(columnDescriptor.columnTypeModifiers()) -- should test for both null and empty

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetRecordFilterBuilder.java
```diff
@@ -174,8 +176,22 @@ private void processSimpleColumnOperator(OperatorNode operatorNode) {
...
+                        int width = columnDescriptor.columnTypeModifiers()[0];
+                        value = StringUtils.rightPad(value, width);
```
denalex: I wonder who does original right-padding on Parquet Write - writer or GPDB ?

frankgh: I think we should determine what's Hive's behavior, because we're most likely be reading lines from parquet files generated by hive.

frankgh:
Hive does not pad char(n) types. PXF should be consistent with that behavior and right trim strings when writing char(n) fields to parquet. Closing this PR and a new one should be opened addressing Parquet write

PR #302, opened 2020-02-12
Make pxf cluster sync delete files

oliverralbertini:
Previously pxf cluster sync was not deleting files from PXF_CONF that
had been deleted on master. By adding the `--delete` flag to the rsync
commands, we delete files from remote hosts.


oliverralbertini:
Confirmed that this works on `phoenix` cluster.

oliverralbertini:
@denalex The flag you recommended is working, to avoid deletion, you can use `--non-destructive`. I am open to suggestions on the name.

oliverralbertini:
```
pxf cluster sync --help
Sync PXF configs from master to standby master and all segment hosts

Usage: pxf cluster <command>
       pxf cluster {-h | --help}

Flags:
  -h, --help              help for sync
  -n, --non-destructive   Don't delete files on remote host
```

cli/go/src/pxf-cli/cmd/cluster.go
```diff
@@ -54,6 +56,7 @@ func init() {
 	clusterCmd.AddCommand(startCmd)
 	clusterCmd.AddCommand(stopCmd)
 	clusterCmd.AddCommand(statusCmd)
+	syncCmd.Flags().BoolVarP(&NonDestructiveSync, "non-destructive", "n", false, "Don't delete files on remote host")
```
frankgh: maybe be more explicit here:
```suggestion
	syncCmd.Flags().BoolVarP(&NonDestructiveSync, "non-destructive", "n", false, "Do not delete files on segment hosts that have been deleted on master host")
```

cli/go/src/pxf-cli/cmd/pxf.go
```diff
@@ -0,0 +1,205 @@
+package cmd
```
frankgh: I assume this is the file that was moved from `.../pxf/` to `cmd` correct?

oliverralbertini: yes

oliverralbertini:
I've incorporated @lisakowen 's feedback; `pxf cluster sync` only deletes files if user passes `-d` or `--delete`.

```
$ pxf cluster sync --help
Sync PXF configs from master to standby master and all segment hosts. Use --delete to delete extraneous remote files

Usage: pxf cluster <command>
       pxf cluster {-h | --help}

Flags:
  -d, --delete   Delete extraneous files on remote host
  -h, --help     help for sync

$ pxf cluster --help
Perform <command> on each segment host in the cluster

Usage: pxf cluster <command>
       pxf cluster {-h | --help}

List of Commands:
  init        Initialize the PXF server instances on master, standby master, and all segment hosts
  reset       Reset PXF (undo initialization) on all segment hosts
  restart     Restart the PXF server on all segment hosts
  start       Start the PXF server instances on all segment hosts
  status      Get status of PXF servers on all segment hosts
  stop        Stop the PXF server instances on all segment hosts
  sync        Sync PXF configs from master to standby master and all segment hosts. Use --delete to delete extraneous remote files

Flags:
  -h, --help   help for cluster

Use "pxf cluster [command] --help" for more information about a command.
```

lisakowen:
thanks, @oliverralbertini .  question:  we support `pxf sync <gphost>` from the master.  just want to verify that the new `{-d | --delete}` option is valid in the single-host-sync case.

oliverralbertini:
> thanks, @oliverralbertini . question: we support `pxf sync <gphost>` from the master. just want to verify that the new `{-d | --delete}` option is valid in the single-host-sync case.

that's a great point. For `pxf init` we have the user pass the flag `-y` after the command. We can do the same for `pxf sync`, so with a8f3eb6eb70c3244627a1a060b2915038980cd91 it's  working like this `pxf sync -d host` or `pxf sync --delete host`. Note that the flag cannot be after the hostname.

denalex:
`pxf sync --delete host` gives an impression that we are trying to take a given host out of PXF cluster, just saying ...

lisakowen:
good point, alex.  what about --delete-files instead?

cli/go/src/pxf-cli/cmd/cluster.go
```diff
@@ -111,12 +117,12 @@ func GenerateOutput(command *pxf.Command, clusterData *ClusterData) error {
 		}
 		response += fmt.Sprintf("%s ==> %s\n", host, errorMessage)
 	}
-	gplog.Info("ERROR: "+command.Messages(pxf.Error), numErrors, numHosts)
+	gplog.Info("ERROR: "+cmd.messages[err], numErrors, numHosts)
```
denalex: should ERROR be logged at ERROR level rather than INFO ?

cli/go/src/pxf-cli/cmd/cluster.go
```diff
@@ -146,28 +152,28 @@ func adaptContentIDToHostname(cluster *cluster.Cluster, f func(string) string) f
...
 }
 
-func clusterRun(command *pxf.Command, clusterData *ClusterData) error {
+func clusterRun(cmd *command, clusterData *ClusterData) error {
 	defer clusterData.connection.Close()
 
-	err := command.Warn(os.Stdin)
+	err := cmd.Warn(os.Stdin)
 	if err != nil {
 		gplog.Info(fmt.Sprintf("%s", err))
```
denalex: should this be Warn level ?

cli/go/src/pxf-cli/cmd/cluster.go
oliverralbertini: thanks for the feedback, I will address these in an upcoming PR.

cli/go/src/pxf-cli/cmd/cluster.go
oliverralbertini: see the upcoming PR to adjust these log levels.

PR #304, opened 2020-02-21
Harden PXF's Tomcat configuration

frankgh:
- Remove default applications
- Remove Tomcat version information
- Set logs folder to 300

### DIFF
--- a/server/pxf-service/src/scripts/pxf-service
+++ b/server/pxf-service/src/scripts/pxf-service
@@ -164,7 +164,7 @@ function waitForTomcat()
     # wait until tomcat is up:
     sleep 2
     echo 'Checking if tomcat is up and running...'
-    until $curl --silent --connect-timeout 1 -I "http://localhost:$PXF_PORT" | grep 'Coyote'; do
+    until $curl --silent --connect-timeout 1 -I "http://localhost:$PXF_PORT" | grep 'PXF Server'; do
         if (( ++attempts == max_attempts )); then
             echo 'ERROR: PXF is down - tomcat is not running'
             return 1
@@ -248,7 +248,6 @@ function generateUserConfigs()
     setup_conf_directory "${PXF_CONF}/conf" "${PXF_HOME}/templates/user/conf"
     setup_conf_directory "${PXF_CONF}/keytabs"
     setup_conf_directory "${PXF_CONF}/lib"
-    setup_conf_directory "${PXF_CONF}/logs"
     setup_conf_directory "${PXF_CONF}/servers/default"
     setup_conf_directory "${PXF_CONF}/templates" "${PXF_HOME}/templates/user/templates" 'override'
 }
--- a/server/pxf-service/src/templates/tomcat/conf/server.xml
+++ b/server/pxf-service/src/templates/tomcat/conf/server.xml
@@ -22,14 +22,6 @@ under the License.
 <Server port="${base.shutdown.port}"
         shutdown="SHUTDOWN">
     <Listener className="org.apache.catalina.mbeans.GlobalResourcesLifecycleListener"/>
-    <GlobalNamingResources>
-        <Resource auth="Container"
-                  description="User database that can be updated and saved"
-                  factory="org.apache.catalina.users.MemoryUserDatabaseFactory"
-                  name="UserDatabase"
-                  pathname="conf/tomcat-users.xml"
-                  type="org.apache.catalina.UserDatabase"/>
-    </GlobalNamingResources>
     <Service name="Catalina">
         <Executor maxThreads="${executor.threads.max}"
                   minSpareThreads="25"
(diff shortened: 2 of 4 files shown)


server/pxf-service/src/scripts/pxf-service
```diff
@@ -118,6 +118,20 @@ function fail()
...
+function tomcatHardening()
+{
+    # Remove default/unwanted Applications
+    rm -rf "$tomcat_root"/webapps/ROOT
+    rm -rf "$tomcat_root"/webapps/docs
+    rm -rf "$tomcat_root"/webapps/examples
+    rm -rf "$tomcat_root"/webapps/host-manager
+    rm -rf "$tomcat_root"/webapps/manager
+}
+
```
denalex: this is good, but I think we should do it during Tomcat packaging into the build, it's a bit silly trying to remove something after we install it, and might be tricky for RPM install as well.

server/pxf-service/src/scripts/pxf-service
```diff
@@ -214,7 +228,8 @@ function createLogsDir()
     if [[ ! -d $PXF_LOGDIR ]]; then
         echo "Creating PXF logs directory $PXF_LOGDIR ..."
         mkdir -p "$PXF_LOGDIR"
-        chmod 700 "$PXF_LOGDIR"
+        # only write/execute access
+        chmod 300 "$PXF_LOGDIR"
```
denalex: why so restrictive ? even owner will not be able to list files in this directory

server/pxf-service/src/templates/tomcat/conf/server.xml
```diff
@@ -52,9 +52,13 @@ under the License.
...
+                       showReport="true"
+                       showServerInfo="false"/>
             </Host>
         </Engine>
-        <Connector acceptCount="100"
+        <Connector server="PXF Server"
+                   acceptCount="100"
                    connectionTimeout="20000"
                    executor="tomcatThreadPool"
                    maxKeepAliveRequests="15"
```
denalex: `web.xml` also needs to be updated, it contains JSP servlet, welcome pages, etc

server/pxf-service/src/templates/tomcat/conf/server.xml
```diff
@@ -52,9 +52,13 @@ under the License.
                        pattern="%h %l %u %t &quot;%r&quot; %s %b"
```
denalex: we also do not use "in-memory user database" in lines 26-31

server/pxf-service/src/scripts/pxf-service
frankgh: this is the recommended for log files, only write/execute access for the catalina user

frankgh: In my local environment, it has become annoying to tail the pxf logs. Even though 300 is recommended, maybe it's not practical?

oliverralbertini: How is it annoying, what's the workaround?

server/pxf-service/src/templates/tomcat/conf/server.xml
```diff
@@ -52,9 +40,13 @@ under the License.
...
                        prefix="localhost_access_log."
                        suffix=".txt"/>
+                <Valve className="org.apache.catalina.valves.ErrorReportValve"
+                       showReport="true"
+                       showServerInfo="false"/>
             </Host>
         </Engine>
-        <Connector acceptCount="100"
+        <Connector server="PXF Server"
```
oliverralbertini: what does adding server field do for security?

server/pxf-service/src/scripts/pxf-service
frankgh: sudo tail, but it's not practical, I agree. Maybe we should leave it at 700, but recommend 300 in the documentation?

frankgh: after the directory exists, `pxf init` will not do anything, so maybe it's okay to ship with 300 and let the users decide to change it to 700?

server/pxf-service/src/templates/tomcat/conf/server.xml
frankgh: It displays `PXF Server` for the server value in the response headers
```
±  |harden-tomcat → origin {34} U:1 ?:20 ✗| → curl -i http://localhost:5888/pxf/v0
HTTP/1.1 500 Internal Server Error
Content-Type: text/plain
Transfer-Encoding: chunked
Date: Tue, 25 Feb 2020 13:15:23 GMT
Connection: close
Server: PXF Server
```

server/pxf-service/src/scripts/pxf-service
denalex: let's not over-restrict things -- they should be usable out-of-the-box and users still can harden it further if required.

PR #305, opened 2020-02-25
pxf cluster init: enforce JAVA_HOME is set

oliverralbertini:
Sometimes users want to use a specific version of Java with PXF. In the
current world, initializing PXF can become awkward. We would ask the
user to temporarily set `JAVA_HOME` in `~/.bashrc` to perform the init
(or `pxf-env-default.sh`, which users shouldn't touch). Then after init,
we would ask them to manually adjust `JAVA_HOME` in
`$PXF_CONF/conf/pxf-env.sh` for their custom PXF value and optionally
switch back in `~/.bashrc`.

This PR proposes a better scenario where we enforce that `JAVA_HOME` is
valid in `pxf cluster init` by passing from the command's environment to
the remote hosts. Command then looks like this:

`JAVA_HOME=/correct/java/home pxf cluster init`

The `pxf-service` script also edits `$PXF_CONF/conf/pxf-env.sh` to make sure that
PXF will always have the correct value of `JAVA_HOME`. Users can thus have
whatever they like in their startup scripts, etc. for `JAVA_HOME`.

Also `pxf-service` stops telling users to edit `pxf-env-default.sh`.

### DIFF
--- a/server/pxf-service/src/scripts/pxf-service
+++ b/server/pxf-service/src/scripts/pxf-service
@@ -103,11 +103,11 @@ function get_environment()
 
 }
 
-checkJavaHome()
+function checkJavaHome()
 {
     # validate JAVA_HOME
     if [[ ! -x ${JAVA_HOME}/bin/java ]]; then
-        fail "\$JAVA_HOME=$JAVA_HOME is invalid. Set \$JAVA_HOME in ~/.bashrc or $default_env_script on all Greenplum hosts."
+        fail "\$JAVA_HOME=$JAVA_HOME is invalid. Set \$JAVA_HOME in your environment before initializing PXF."
     fi
 }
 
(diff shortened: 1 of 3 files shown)


oliverralbertini:
Sorry about the extra commits, they will go away once [this PR](https://github.com/greenplum-db/pxf/pull/302) is merged.

server/pxf-service/src/scripts/pxf-service
```diff
@@ -107,7 +107,7 @@ checkJavaHome()
 {
     # validate JAVA_HOME
     if [[ ! -x ${JAVA_HOME}/bin/java ]]; then
-        fail "\$JAVA_HOME=$JAVA_HOME is invalid. Set \$JAVA_HOME in ~/.bashrc or $default_env_script on all Greenplum hosts."
+        fail "\$JAVA_HOME=$JAVA_HOME is invalid. Set \$JAVA_HOME in your environment before initializing PXF."
```
denalex: consistency: other functions start with keyword `function`

server/pxf-service/src/scripts/pxf-service
```diff
@@ -371,12 +371,19 @@ function doInit()
...
     generatePrivateClasspath || return 1
     generateUserConfigs || return 1
+    editPxfEnvSh || return 1
     createInstance || return 1
     deployWebapp || return 1
     createLogsDir || return 1
     createRunDir  || return 1
 }
 
+editPxfEnvSh()
```
denalex: consistency: other functions start with keyword `function`

server/pxf-service/src/scripts/pxf-service
```diff
@@ -371,12 +371,19 @@ function doInit()
...
+{
+	[[ -z ${JAVA_HOME} ]] && return 1
```
denalex: in the overall flow we call checkJavaHome (line 371) before calling this function, so we are guaranteed it is set to correct path and probably there's no need to check here again ?

oliverralbertini: yes, it's kind of overkill

Issue #307, opened 2020-02-27
ESCAPE 'OFF' is not processed correctly on PXF side


kapustor:
GP 6.4, PXF 5.10.1.
We can create a valid external table with ESCAPE 'OFF' clause:
```
CREATE EXTERNAL TABLE test (test text)
LOCATION ('pxf://folder?PROFILE=hdfs:text')
FORMAT 'TEXT' ( ESCAPE 'OFF');
```
The table is created successfully, but while selecting we got the following error:
```
ERROR:  remote component error (500) from '127.0.0.1:5888':  Type  Exception Report   Message  invalid ESCAPE character 'OFF'. Only single character is allowed for ESCAPE.   Description  The server encountered an unexpected condition that prevented it from fulfilling the request.   Exception   java.lang.IllegalArgumentException: invalid ESCAPE character 'OFF'. Only single character is allowed for ESCAPE. (libchurl.c:920)  (seg0 slice1 10.92.8.11:10000 pid=23082) (libchurl.c:920)
CONTEXT:  External table test, line 1 of file pxf://folder?PROFILE=hdfs:text
```
It seems that PXF should process ESCAPE 'OFF' correctly and ignore all escape symbols.

oliverralbertini:
Thanks for the report, @kapustor . Could you also add some example data that causes this error?

kapustor:
Hi @oliverralbertini, the error occurs earlier than PXF gets any data, even without connected HDFS. Just deploy any GP cluster, install and run PXF and create external table.

PR #308, opened 2020-02-27
Fix ESCAPE 'OFF' is not processed correctly on PXF side error

frankgh:
Currently, PXF is not handling ESCAPE 'OFF' and DELIMITER 'OFF' correctly.
These are valid keywords that can be used to create an external table.

If provided, PXF should not ESCAPE or add DELIMITERS when reading data.

Fixes https://github.com/greenplum-db/pxf/issues/307

Integrations tests to come

### DIFF
--- a/server/pxf-api/src/main/java/org/greenplum/pxf/api/model/GreenplumCSV.java
+++ b/server/pxf-api/src/main/java/org/greenplum/pxf/api/model/GreenplumCSV.java
@@ -18,7 +18,7 @@ public class GreenplumCSV {
 
     private String valueOfNull;
     private char quote;
-    private char escape;
+    private Character escape;
     private String newline;
     private Character delimiter;
 
@@ -43,7 +43,7 @@ public class GreenplumCSV {
         return quote;
     }
 
-    public char getEscape() {
+    public Character getEscape() {
         return escape;
     }
 
@@ -105,7 +105,9 @@ public class GreenplumCSV {
      * @return GreenplumCSV object for builder pattern
      */
     public GreenplumCSV withEscapeChar(String escapeString) {
-        if (StringUtils.isNotEmpty(escapeString)) {
+        if (StringUtils.equalsIgnoreCase("OFF", escapeString)) {
+            escape = null;
+        } else if (StringUtils.isNotEmpty(escapeString)) {
             validateSingleCharacter(escapeString, "ESCAPE");
             withEscapeChar(escapeString.charAt(0));
         }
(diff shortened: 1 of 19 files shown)


server/pxf-api/src/main/java/org/greenplum/pxf/api/model/GreenplumCSV.java
```diff
@@ -105,7 +105,9 @@ public GreenplumCSV withQuoteChar(char quoteChar) {
      * @return GreenplumCSV object for builder pattern
      */
     public GreenplumCSV withEscapeChar(String escapeString) {
-        if (StringUtils.isNotEmpty(escapeString)) {
+        if (StringUtils.equalsIgnoreCase("OFF", escapeString)) {
+            escape = null;
+        } else if (StringUtils.isNotEmpty(escapeString)) {
             validateSingleCharacter(escapeString, "ESCAPE");
```
denalex: can ESCAPE be a multi-char string ?

server/pxf-service/src/main/java/org/greenplum/pxf/service/HttpRequestParser.java
```diff
@@ -163,6 +163,11 @@ public RequestContext parseRequest(HttpHeaders request, RequestContext.RequestTy
         // parse CSV format information
         parseGreenplumCSV(params, context);
 
+        // Only single column tables support 'OFF' delimiter
+        if (context.getTupleDescription().size() != 1 && context.getGreenplumCSV().getDelimiter() == null) {
```
denalex: wouldn't this be a problem for non-CSV tables where there's no delimiter ?

frankgh: good point. It looks like it isn't as all the tests are passing, but I'd like to understand why the tests are passing. I will get back to you on this

server/pxf-api/src/main/java/org/greenplum/pxf/api/model/GreenplumCSV.java
frankgh: nope, Greenplum only support single character for the escape character

server/pxf-service/src/main/java/org/greenplum/pxf/service/HttpRequestParser.java
frankgh: It turns out that the default CSV/TEXT options are still passed from the C side even with 'CUSTOM' format type. I added an if guarding the checks.

PR #311, opened 2020-03-04
JDBC: Validate batchsize only on write

oliverralbertini:
This property is only needed for the write case.

### DIFF
--- a/server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
+++ b/server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
@@ -223,13 +223,15 @@ public class JdbcBasePlugin extends BasePlugin {
 
         // Optional parameters
         batchSizeIsSetByUser = configuration.get(JDBC_STATEMENT_BATCH_SIZE_PROPERTY_NAME) != null;
-        batchSize = configuration.getInt(JDBC_STATEMENT_BATCH_SIZE_PROPERTY_NAME, DEFAULT_BATCH_SIZE);
+        if (context.getRequestType() == RequestContext.RequestType.WRITE_BRIDGE) {
+            batchSize = configuration.getInt(JDBC_STATEMENT_BATCH_SIZE_PROPERTY_NAME, DEFAULT_BATCH_SIZE);
 
-        if (batchSize == 0) {
-            batchSize = 1; // if user set to 0, it is the same as batchSize of 1
-        } else if (batchSize < 0) {
-            throw new IllegalArgumentException(String.format(
-                    "Property %s has incorrect value %s : must be a non-negative integer", JDBC_STATEMENT_BATCH_SIZE_PROPERTY_NAME, batchSize));
+            if (batchSize == 0) {
+                batchSize = 1; // if user set to 0, it is the same as batchSize of 1
+            } else if (batchSize < 0) {
+                throw new IllegalArgumentException(String.format(
+                        "Property %s has incorrect value %s : must be a non-negative integer", JDBC_STATEMENT_BATCH_SIZE_PROPERTY_NAME, batchSize));
+            }
         }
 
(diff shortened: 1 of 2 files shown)


oliverralbertini:
This PR stops the validation process on the non-write paths of JDBC, that is fragmentation and read paths.

@lisakowen This would have an implication for [docs](https://gpdb.docs.pivotal.io/6-4/pxf/jdbc_pxf.html) under the `BATCH_SIZE` table entry.

@denalex you mentioned that we should throw an error when this is present outside of the write situation, but @frankgh mentioned that for FDW tables will be able to read _and_ write.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
```diff
@@ -1,582 +1,584 @@
...
-                String value = configuration.get(propertyName);
-                String keyName = propertyName.substring(confPrefix.length());
-                configMap.put(keyName, value);
-            }
-        }
-        return configMap;
-    }
-
-}
+package org.greenplum.pxf.plugins.jdbc;
```
frankgh: this file has formatting issues. Can you fix it before reviewing?

server/pxf-jdbc/src/test/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePluginTestInitialize.java
```diff
@@ -189,7 +193,35 @@ public void testBatchSize2() throws Exception {
...
+        prepareBaseConfigurationFactory(configuration);
+        JdbcBasePlugin plugin = new JdbcBasePlugin();
+        RequestContext context = makeContext();
+        context.setRequestType(RequestContext.RequestType.READ_BRIDGE);
+        plugin.initialize(context);
+
+        // should not error because we don't validate this on the READ path
+    }
+
+    @Test(expected = NumberFormatException.class)
```
denalex: while this is fine, using @Rule method would allow to check for the error message as well.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
oliverralbertini: yes, in the second commit I stripped carriage returns (^M) but the [first commit](https://github.com/greenplum-db/pxf/pull/311/commits/551c363873a62443d95b2680e49c49c15513f191) has the changes.

server/pxf-jdbc/src/test/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePluginTestInitialize.java
oliverralbertini: How'd I know you would say this?

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
frankgh: any functional reason we want to do that? I find it useful to be able to do a "blame/annotate" to look at a code in the change. However, when doing these changes, we make it more difficult to quickly determine changes in the source code

server/pxf-jdbc/src/test/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePluginTestInitialize.java
oliverralbertini: I switched all the tests to use `@Rule`

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
oliverralbertini: OK, switched back to carriage returns.

PR #313, opened 2020-03-04
Update jackson-databind

oliverralbertini:
https://github.com/advisories/GHSA-gww7-p5w4-wrfv

### DIFF
--- a/automation/pom.xml
+++ b/automation/pom.xml
@@ -229,7 +229,7 @@
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
-            <version>2.9.10.1</version>
+            <version>2.9.10.3</version>
         </dependency>
 
         <dependency>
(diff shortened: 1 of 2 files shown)


automation/pom.xml
```diff
@@ -223,19 +223,19 @@
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-core</artifactId>
-            <version>2.9.10</version>
+            <version>2.9.10.2</version>
```
denalex: does this exist, I do not see this version for the core package here: https://mvnrepository.com/artifact/com.fasterxml.jackson.core/jackson-core

if it does, then we should also change runtime version n build.gradle line 150

automation/pom.xml
```diff
@@ -223,19 +223,19 @@
...
             <artifactId>jackson-databind</artifactId>
-            <version>2.9.10.1</version>
+            <version>2.9.10.2</version>
         </dependency>
 
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-annotations</artifactId>
-            <version>2.9.10</version>
+            <version>2.9.10.2</version>
```
denalex: same here

PR #316, opened 2020-03-11
Address CVE-2020-1938

frankgh:
The AJP protocol is enabled by default in Tomcat versions < 7.0.100,
with the AJP connector listening in TCP port 8009 by default. This is
low impact on PXF because PXF should only be accessible from segments
collocated with the PXF Server. It is important to note that mitigation
is only required if an AJP port is accessible to untrusted users.

Tomcat 7.0.100 disables AJP protocol by default


server/pxf-service/src/main/java/org/greenplum/pxf/service/HttpRequestParser.java
```diff
@@ -391,24 +394,51 @@ private int parsePositiveIntOrError(String s, String propName) {
...
+        RequestMap(MultivaluedMap<String, String> requestHeaders) throws UnsupportedEncodingException {
             super(String.CASE_INSENSITIVE_ORDER);
+
+            boolean decodeHeaderValue = false;
             for (Map.Entry<String, List<String>> entry : requestHeaders.entrySet()) {
-                List<String> values = entry.getValue();
-                if (values == null) continue;
+                if (StringUtils.equalsIgnoreCase(ENCODED_HEADER_VALUES_NAME, entry.getKey())) {
+                    String value = getValue(entry.getValue());
+                    decodeHeaderValue = StringUtils.equals("true", value);
```
oliverralbertini: Shouldn't we ignore case on the value as well?

server/pxf-service/src/main/java/org/greenplum/pxf/service/HttpRequestParser.java
```diff
@@ -391,24 +394,51 @@ private int parsePositiveIntOrError(String s, String propName) {
         private static final String PROP_PREFIX = "X-GP-";
         private static final String USER_PROP_PREFIX = "X-GP-OPTIONS-";
         private static final String USER_PROP_PREFIX_LOWERCASE = "x-gp-options-";
+        private static final String ENCODED_HEADER_VALUES_NAME = PROP_PREFIX + "ENCODED-HEADER-VALUES";
 
-        RequestMap(MultivaluedMap<String, String> requestHeaders) {
+        RequestMap(MultivaluedMap<String, String> requestHeaders) throws UnsupportedEncodingException {
```
denalex: why throw this exception ? We can't recover from it, might as well throw RuntimeException and simplify method signatures

server/pxf-service/src/main/java/org/greenplum/pxf/service/HttpRequestParser.java
```diff
@@ -391,24 +394,51 @@ private int parsePositiveIntOrError(String s, String propName) {
...
+            for (Map.Entry<String, List<String>> entry : requestHeaders.entrySet()) {
+                String value = getValue(entry.getValue());
                 if (value == null) continue;
 
-                // Converting to value UTF-8 encoding
                 String key = entry.getKey();
-                value = new String(value.getBytes(StandardCharsets.ISO_8859_1), StandardCharsets.UTF_8);
+                if (decodeHeaderValue && StringUtils.startsWithIgnoreCase(key, PROP_PREFIX)) {
+                    value = URLDecoder.decode(value, StandardCharsets.UTF_8.name());
+                }
```
denalex: should we have unit tests that test that we trigger the logic on the header value, case insensitive and that decoding is correct ?

frankgh: yes! definitely

server/pxf-service/src/main/java/org/greenplum/pxf/service/HttpRequestParser.java
frankgh: yeah, good point

server/pxf-service/src/main/java/org/greenplum/pxf/service/HttpRequestParser.java
frankgh: @denalex  I just added unit tests

server/pxf-service/src/main/java/org/greenplum/pxf/service/HttpRequestParser.java
```diff
@@ -391,24 +393,55 @@ private int parsePositiveIntOrError(String s, String propName) {
...
+                if (StringUtils.equalsIgnoreCase(ENCODED_HEADER_VALUES_NAME, entry.getKey())) {
+                    String value = getValue(entry.getValue());
+                    decodeHeaderValue = StringUtils.equalsIgnoreCase("true", value);
+                    break;
+                }
+            }
 
-                String value = values.size() > 1 ? StringUtils.join(values, ",") : values.get(0);
+            for (Map.Entry<String, List<String>> entry : requestHeaders.entrySet()) {
+                String value = getValue(entry.getValue());
```
frankgh: @oliverralbertini the logic was moved to the `getValue` method

server/pxf-service/src/main/java/org/greenplum/pxf/service/HttpRequestParser.java
frankgh: it is `equalsIgnoreCase` now

PR #317, opened 2020-03-12
Add test demo profile

oliverralbertini:
Will help users make sure that PXF is working, even before an external
data source is configured.

### DIFF
--- a/automation/src/test/java/org/greenplum/pxf/automation/features/demo/ReadableDemoTest.java
+++ b/automation/src/test/java/org/greenplum/pxf/automation/features/demo/ReadableDemoTest.java
@@ -1,0 +1,24 @@
+package org.greenplum.pxf.automation.features.demo;
+
+import org.greenplum.pxf.automation.features.BaseFeature;
+import org.greenplum.pxf.automation.structures.tables.pxf.ReadableExternalTable;
+import org.testng.annotations.Test;
+
+public class ReadableDemoTest extends BaseFeature {
+    @Override
+    public void beforeMethod() throws Exception {
+        // default external table with common settings
+        exTable = new ReadableExternalTable("demo_table", null, "/tmp/dummy/path", "text");
+        exTable.setHost(pxfHost);
+        exTable.setPort(pxfPort);
+        exTable.setProfile("demo");
+        exTable.setDelimiter(",");
+        exTable.setFields(new String[]{"a text", "b text", "c text"});
+        gpdb.createTableAndVerify(exTable);
+    }
+
+    @Test(groups = {"features", "gpdb", "hcfs", "security"})
+    public void testDemo() throws Exception {
+        runTincTest("pxf.features.demo.readableDemo.runTest");
+    }
+}
--- a/server/pxf-service/src/main/resources/pxf-profiles-default.xml
+++ b/server/pxf-service/src/main/resources/pxf-profiles-default.xml
@@ -749,4 +749,13 @@ under the License.
         </plugins>
         <protocol>wasbs</protocol>
     </profile>
+    <profile>
+        <name>demo</name>
+        <description>Test the PXF demo</description>
+        <plugins>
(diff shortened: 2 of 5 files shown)


frankgh:
An integration test would be nice :)

server/pxf-service/src/main/resources/pxf-profiles-default.xml
```diff
@@ -749,4 +749,13 @@ under the License.
         </plugins>
         <protocol>wasbs</protocol>
     </profile>
+    <profile>
+        <name>test:demo</name>
```
frankgh: can we just call this profile `demo`? In FDW, we have a demo already defined

oliverralbertini:
> An integration test would be nice :)

OK, I've added one. It was easier than I expected because DemoFragmenter always returns 3 fragments regardless of how many segments there are.

automation/src/test/java/org/greenplum/pxf/automation/features/demo/ReadableDemoTest.java
```diff
@@ -0,0 +1,24 @@
+package org.greenplum.pxf.automation.features.demo;
+
```
frankgh: I think @denalex had mentioned that we need 1 fragment per segment, which would mean changing `DemoFragmenter` to read the segment count and produce 1 fragment per segment

oliverralbertini:
Here's a pipeline: https://ud.ci.gpdb.pivotal.io/teams/main/pipelines/dev:pxf_oalbertini-add-test-demo-profile_6X_STABLE

automation/src/test/java/org/greenplum/pxf/automation/features/demo/ReadableDemoTest.java
oliverralbertini: That's a good point, thanks @denalex and @frankgh .

denalex: My comment was similar, but specifically to the point that if on multi-node cluster people (or GP4K) want to make sure PXF runs and responds on all nodes, then having 1 fragment per segment might not be sufficient, as in the edge case a few of fragments might hash-function to the same segment id and some PXF JVMs might not get any requests. But choosing the number of fragments per segment to be 5 or 10 will likely reduce the risk. The operator will then have to make sure that all fragments got returned and all hosts have responded. In all honesty this better be done in a SQL function that does counting / grouping / validation against `gp_segment_configuration`. Maybe this PR is a bit pre-mature and we need to think about production-quality way to verify PXF is up/running from SQL command.

oliverralbertini: That's fair. I appreciate the need for a better verification. I thought about having the DemoFragmenter perform some JDBC call back to Greenplum to get the number of segments and do some multiple of that for number of fragments. But then we might require some JDBC server to be set up in `$PXF_CONF/servers`. Thoughts?

denalex: that'd be prone to config errors. this would require JDBC server setup, JDBC driver present and more critically `pg_hba.conf` entry to allow such a connection. In my opinion having a SQL function will the best: `select * from pxf_status()`, which can do a join between `pg_segment_configuration` and info obtained from PXF (it will have to return hostnames) to get smth like:
```
hostname | status
----------+-------
sdw0         | UP
sdw1          | UP
sdw2         | DOWN 
```

denalex: Also, best to talk to Gp4k guys to understand what they are after, maybe they just need 0 / 1 type of output.

oliverralbertini:
Closing because we need to figure out what kind of functionality users need out of a demo profile.

PR #318, opened 2020-03-19
pxf-public classpath is no longer being used

frankgh:
This is generating unwanted log entries in PXF's log. I have this fix in
a different branch, but it belong on its own branch.


frankgh:
The log message that we are effectively removing with this PR looks like this:

> WARNING: Failed to read classpath file: java.nio.file.NoSuchFileException: /usr/local/gpdb6/pxf/conf/pxf-public.classpath

PR #319, opened 2020-03-23
SecureLoginTest: Fix test pollution

oliverralbertini:
SecureLoginTest has been failing sporadically on centos7 in Concourse
during the compile_pxf job. It turns out that if one of these tests:

* org.apache.hadoop.security.LoginSessionTest
* org.apache.hadoop.security.PxfUserGroupInformationTest

runs before SecureLoginTest, we get an error when SecureLogin tries to
get the default realm when looking up the user. This has to do with
test pollution when newing up Hadoop `User`s, which end up
setting the default Kerberos realm for the JVM. This is the Hadoop code
that up does that[0].

This commit makes sure that

* if we change system properties, we change them back after the tests
* we have default realm set to something so that we don't get empty
  string

[0] https://github.com/apache/hadoop/blob/826afbeae31ca687bc2f8471dc841b66ed2c6704/hadoop-common-project/hadoop-auth/src/main/java/org/apache/hadoop/security/authentication/util/KerberosName.java#L86

### DIFF
--- a/server/pxf-api/src/test/java/org/apache/hadoop/security/PxfUserGroupInformationTest.java
+++ b/server/pxf-api/src/test/java/org/apache/hadoop/security/PxfUserGroupInformationTest.java
@@ -3,7 +3,9 @@ package org.apache.hadoop.security;
 import com.google.common.collect.Sets;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.security.authentication.util.KerberosUtil;
+import org.junit.AfterClass;
 import org.junit.Before;
+import org.junit.BeforeClass;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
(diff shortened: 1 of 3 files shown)


server/pxf-api/src/test/java/org/greenplum/pxf/api/security/SecureLoginTest.java
```diff
@@ -34,10 +36,45 @@
...
+    @AfterClass
+    public static void resetProperties() {
+        resetProperty(PROPERTY_KEY_USER_IMPERSONATION, userImpersonationEnabled);
+        resetProperty(PROPERTY_KEY_SERVICE_PRINCIPAL, kerberosPrincipal);
+        resetProperty(PROPERTY_KEY_SERVICE_KEYTAB, kerberosKeytab);
+        resetProperty(PROPERTY_KEY_KERBEROS_KDC, kdcDefault);
+        resetProperty(PROPERTY_KEY_KERBEROS_REALM, realmDefault);
+    }
+
+    private static void resetProperty(String key, String val) {
```
frankgh: minor: private methods should go after public/protected

server/pxf-api/src/test/java/org/apache/hadoop/security/PxfUserGroupInformationTest.java
```diff
@@ -51,6 +53,32 @@
...
+        realmDefault = System.setProperty(PROPERTY_KEY_KERBEROS_REALM, "DEFAULT_REALM");
+    }
+
+    @AfterClass
+    public static void resetProperties() {
+        resetProperty(PROPERTY_KEY_KERBEROS_KDC, kdcDefault);
+        resetProperty(PROPERTY_KEY_KERBEROS_REALM, realmDefault);
+    }
+
+    private static void resetProperty(String key, String val) {
```
frankgh: minor: private methods should go after public/protected

frankgh:
It'd be nice to see how this behaves on the failing pipeline.

oliverralbertini:
﻿﻿﻿﻿﻿I pointed the [`pg_regress`](https://ud.ci.gpdb.pivotal.io/teams/main/pipelines/pg_regress) pipeline to use this branch instead of master, and the last 5 builds [231](https://ud.ci.gpdb.pivotal.io/teams/main/pipelines/pg_regress/jobs/compile_pxf/builds/231), [232](https://ud.ci.gpdb.pivotal.io/teams/main/pipelines/pg_regress/jobs/compile_pxf/builds/232), [233](https://ud.ci.gpdb.pivotal.io/teams/main/pipelines/pg_regress/jobs/compile_pxf/builds/233), [234](https://ud.ci.gpdb.pivotal.io/teams/main/pipelines/pg_regress/jobs/compile_pxf/builds/234), and [235](https://ud.ci.gpdb.pivotal.io/teams/main/pipelines/pg_regress/jobs/compile_pxf/builds/235) have gone green.

PR #320, opened 2020-03-24
docker-images: build when deps are updated

oliverralbertini:


oliverralbertini:
We are re-building docker images only when `concourse/docker/pxf-dev-base/Dockerfile` is modified. This commit expands that to also include Makefiles, important gradle, Maven, and Gopkg files. This is in hopes of triggering image building whenever there is a change in dependencies or build process.

PR #322, opened 2020-03-25
FDW to use spring boot

frankgh:
Integrate pxf_fdw with the spring boot code:

- allows users to set pxf_host, pxf_port, pxf_protocol
  - using gucs
  - using alter commands on FDW or Servers
  - support legacy environment variable (backwards compatibility with pxf ext table)
- fixes issue with column projection on attrs from predicate
- add pxf version function pxf_fdw_version

```
regression=# select pxf_fdw_version();
  pxf_fdw_version
-------------------
 1.0.0-DEVELOPMENT
(1 row)

Time: 5.221 ms
```

- add demo profile
- cleanup curl resources when query is cancelled
- support column_name option for foreign columns

```
CREATE FOREIGN TABLE foo (
  a TEXT OPTIONS ( column_name 'name_on_remote_system' )
) SERVER bar
OPTIONS ( resource '/bar/foo' );
```

- Remove unnecessary headers
- Add TLS support (untested)


frankgh:
Precedence for pxf_port, pxf_host and pxf_protocol:
1. OPTIONS
2. GUC
3. Environment variables

frankgh:
closing since the commits are moved to the `pxf-refactor` branch

PR #323, opened 2020-03-29
Create a copy of the server project on the boot-server directory

frankgh:
This is a copy of the server/ directory at SHA
5f2c436550e40d9e9888dafd14a8f10d4d9a3c6c that will be the base of the
boot project. This will help us maintain git history and it will make it
easier to review changes to the boot-server project.


frankgh:
This needs 4 commits to be able to preserver history, used the approach mentioned here: https://stackoverflow.com/a/44036771

denalex:
Git trick seems a bit complex. What's the intended structure of the project after all is done ? I thought `boot-server` will be on the same level as `pxf-service`, not at the root of the whole thing.

frankgh:
Closing this. We probably won't use this approach

PR #324, opened 2020-04-01
Parquet: Right trim char types during insert

frankgh:
Currently, there is an issue when inserting parquet data with char types
when they are padded. For example,

    CREATE WRITABLE EXTERNAL TABLE foo (id int, code char(5))
    LOCATION ('pxf://bucket/path?PROFILE=s3:parquet&SERVER=s3')

    INSERT INTO foo VALUE (1, '95126'), (2, '9'), (3, '9');

This will cause postgres to right pad tuples 2 and 3 with whitespace
after the char '9', which will be sent to PXF Server as '9    ', and
we'll store it as such in a parquet file. However, this is inconsistent
with Hive's behavior and it is also causing issues with predicate
pushdown. Queries like the following will produce no tuples:

    SELECT * FROM foo WHERE code = '9';

With this patch, we right-trim values of type char coming from Greenplum
before storing them to parquet. This brings us on-par with the way Hive
stores char types and it fixes predicate pushdown.


automation/tincrepo/main/pxf/features/parquet/padded_char_pushdown/expected/query01.ans
```diff
@@ -0,0 +1,54 @@
...
+-- @description query01 for pushing down predicates of type char that were
+-- padded during write. Hive right trims values for type char, for example
+-- when a value 'a  ' is inserted into a Hive table stored as parquet, Hive will
+-- only store 'a' in the parquet file.
+-- Display on for output consistency between GPDB 5 and 6
+\x on
+Expanded display is on.
+\pset format unaligned
+Output format is unaligned.
+SELECT s1, s2, n1, d1, dc1, f, bg, b, tn, sml, vc1, c1, bin FROM parquet_view WHERE c1 = 'a' OR c1 = e'b\t' OR c1 = e'c\n' ORDER BY s1;
```
denalex, line 12: would be nice to have a query that also uses WHERE c1 = 'a  ' to make sure if a user specifies a padded value we can handle that as well.

server/pxf-api/src/test/java/org/greenplum/pxf/api/utilities/UtilitiesTest.java
```diff
@@ -168,6 +168,20 @@ public boolean isThreadSafe() {
         }
     }
 
+    @Test
+    public void testRightTrimWhitespace() {
+        assertNull(Utilities.rightTrimWhiteSpace(null));
```
denalex, line 173:
```
assertEquals("", Utilities.rightTrimWhiteSpace(""));
```

server/pxf-api/src/main/java/org/greenplum/pxf/api/utilities/Utilities.java
```diff
@@ -332,4 +330,31 @@ public static boolean isSecurityEnabled(Configuration configuration) {
...
+     */
+    public static String rightTrimWhiteSpace(String s) {
+        if (s == null) return null;
+        int length = s.length();
+        while (length > 0 && s.charAt(length - 1) == ' ') length--;
+
+        if (length == s.length()) return s;
+        if (length == 0) return "";
+        return s.substring(0, length);
+    }
```
oliverralbertini, line 359: I would be more in favor of this:
```java
public static String rightTrimWhiteSpace(String s) {
        if (s == null) return null;
        return s.replaceAll(" +$", "");
    }
```

oliverralbertini: Do we need tests in UtilitiesTest.java?

frankgh: I'm not a huge fan of regex. We have tests in UtilitiesTest.java as part of the PR.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
```diff
@@ -90,7 +91,19 @@ public OneRow setFields(List<OneField> record) throws IOException {
...
+            OneField field = record.get(i);
+            ColumnDescriptor columnDescriptor = context.getTupleDescription().get(i);
+
+            /*
+             * We need to right trim the incoming value from Greenplum. This is
+             * consistent with the behaviour in Hive, where char fields are right
+             * trimmed during write. Note that String and varchar Hive types are
+             * not right trimmed. Hive does not trim tabs or newlines
+             */
+            if (columnDescriptor.getDataType() == DataType.BPCHAR && field.val instanceof String) {
```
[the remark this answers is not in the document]
frankgh: because field.val is of Object type

automation/tincrepo/main/pxf/features/parquet/padded_char_pushdown/expected/query01.ans
frankgh: good point, but I'm pretty sure it won't be handled. Should we trim the predicate? Or should we add an OR for these predicates on the server side ( c1 = 'a ' OR c1 = 'a' )

PR #325, opened 2020-04-01
pgregress: fix HCFS cleanup script template

oliverralbertini:
It was using HCFS_LOCATION, which is no longer used in the substitution
script, in favor of TEST_LOCATION.


oliverralbertini:
The scripts that get generated from this template are meant to be run after testing with PXF_TEST_DEBUG=1. After you don't need the data on HCFS (after a debugging session) you can run the generated script to delete the data.

PR #330, opened 2020-04-02
pxf-cli: set GOPATH using direnv

oliverralbertini:
.envrc will be sourced if user has direnv.


oliverralbertini:
> This is only for developers, right ? The file will not be shipped ?

That's correct.

PR #331, opened 2020-04-02
pxf-cli: Update gp-common-go-libs to latest

oliverralbertini:

### DIFF
--- a/cli/go/src/pxf-cli/Gopkg.toml
+++ b/cli/go/src/pxf-cli/Gopkg.toml
@@ -31,7 +31,7 @@
 
 [[constraint]]
   name = "github.com/greenplum-db/gp-common-go-libs"
-  revision = "1a4969cbabb6e2f16bb2db542deccbee9965d33c"
+  branch = "master"
 
 [prune]
   go-tests = true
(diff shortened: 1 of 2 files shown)


cli/go/src/pxf-cli/Gopkg.toml
```diff
@@ -31,7 +31,7 @@
 
 [[constraint]]
   name = "github.com/greenplum-db/gp-common-go-libs"
-  revision = "1a4969cbabb6e2f16bb2db542deccbee9965d33c"
+  branch = "master"
```
frankgh, line 34: if there are new commits to master, will it pick up the latest version? will that be an issue with dependencies and licensing?

oliverralbertini: It should only pick up new commits if someone runs `dep ensure -update` then commits the change to this repo, which is not in our normal workflow. For building, testing we run `dep ensure` without the update flag.

oliverralbertini: BTW, here's an explanation: https://golang.github.io/dep/docs/Gopkg.toml.html#branch

Issue #328, opened 2020-04-02
PXF: Error while reading timestamp column from Hive parquet file


frankgh:
https://groups.google.com/a/greenplum.org/forum/?utm_medium=email&utm_source=footer#!msg/gpdb-users/iYNLT7_FSeM/c3XdFKesBAAJ

From the open source channel:

> Hello,
> 
> getting error while trying to read a Hive table (parquet format)containing timestamp column (created via Spark) - PXF v5.11.2 .
> Seems like joda-time dependency is missing.
> 
> Best,
> 
> Ales

```
Apr 01, 2020 4:19:46 PM org.apache.catalina.core.StandardWrapperValve invoke
SEVERE: Servlet.service() for servlet [PXF REST Service] in context with path [/pxf] threw exception [javax.servlet.ServletException: Servlet execution threw an exception] with root cause
java.lang.NoClassDefFoundError: jodd/datetime/JDateTime
        at org.apache.hadoop.hive.ql.io.parquet.timestamp.NanoTimeUtils.getTimestamp(NanoTimeUtils.java:93)
        at org.apache.hadoop.hive.ql.io.parquet.convert.ETypeConverter$9$1.convert(ETypeConverter.java:205)
        at org.apache.hadoop.hive.ql.io.parquet.convert.ETypeConverter$9$1.convert(ETypeConverter.java:196)
        at org.apache.hadoop.hive.ql.io.parquet.convert.ETypeConverter$BinaryConverter.setDictionary(ETypeConverter.java:283)
        at org.apache.parquet.column.impl.ColumnReaderImpl.<init>(ColumnReaderImpl.java:341)
        at org.apache.parquet.column.impl.ColumnReadStoreImpl.newMemColumnReader(ColumnReadStoreImpl.java:80)
        at org.apache.parquet.column.impl.ColumnReadStoreImpl.getColumnReader(ColumnReadStoreImpl.java:75)
        at org.apache.parquet.io.RecordReaderImplementation.<init>(RecordReaderImplementation.java:271)
        at org.apache.parquet.io.MessageColumnIO$1.visit(MessageColumnIO.java:147)
        at org.apache.parquet.io.MessageColumnIO$1.visit(MessageColumnIO.java:109)
        at org.apache.parquet.filter2.compat.FilterCompat$NoOpFilter.accept(FilterCompat.java:165)
        at org.apache.parquet.io.MessageColumnIO.getRecordReader(MessageColumnIO.java:109)
[...]

PR #332, opened 2020-04-03
Hive: Add missing transitive dependency when reading parquet files

frankgh:
When reading Hive tables that are stored as parquet files, and the
tables have timestamp types (INT96 primitive type), then PXF is unable
to read the table because PXF is missing a transitive dependency
(org.jodd:jodd-core). This commit adds the missing transitive
dependency, and adds tests to make sure this issue is not encountered in
the future.


frankgh:
Fixes https://github.com/greenplum-db/pxf/issues/328

automation/tincrepo/main/pxf/features/hive/parquet_timestamp/sql/query01.sql
```diff
@@ -0,0 +1,3 @@
+-- @description query01 for PXF Hive parquet timestamp
+
+SELECT * FROM pxf_hive_parquet_timestamp;
```
denalex, line 3: Don't you need ORDER BY to guarantee order of records ?

oliverralbertini: I would also feel better if there was an ORDER BY.

frankgh: A single file is created because a single insert statement is performed. But I will add the order by clause for peace of mind
https://github.com/greenplum-db/pxf/pull/332/files#diff-8d1b90518e58a14500ec6a5046cce97dR355

PR #334, opened 2020-04-06
pgregress: fix external table tests

oliverralbertini:
* on Ubuntu, we need the uuid-runtime package installed
* make the WritableSmokeTest more robust by checking the entire output
  of the `hdfs -cat` command


oliverralbertini:
> ```
> cat /dev/urandom | base64 | tr -cd '[:alnum:]._-' | head -c 16
> ```

Looks like on Linux, we can do this `cat /proc/sys/kernel/random/uuid`,
but for local envs with MacOS we can continue using `uuidgen`. I will
probably add an if statement that checks `uname -s` command for Linux.

oliverralbertini:
[Here's](https://serverfault.com/a/529319) a reference.

concourse/docker/pxf-dev-base/Dockerfile
```diff
@@ -42,16 +42,19 @@ RUN ["/bin/bash", "-c", "export BUILD_PARAMS=-Dorg.gradle.daemon=false && \
     chown -R gpadmin:gpadmin ~gpadmin && \
     rm -rf /tmp/pxf_src && \
     if grep 'CentOS release 6' /etc/centos-release; then \
-      yum install -y krb5-workstation python-paramiko && yum clean all; \
+      yum install -y krb5-workstation python-paramiko; \
     else \
       pip install psi paramiko --no-cache-dir; \
     fi && \
     if grep 'VERSION=.*18.04.*LTS.*Bionic Beaver.*$' /etc/os-release; then \
+      apt-get update && apt-get install -y zip libaprutil1 && rm -rf /var/lib/apt/lists/* && \
```
frankgh, line 50: what are these new dependencies for?

oliverralbertini:
> what are these new dependencies for?

@denalex and I decided to switch to an rpm/apt based installation, and
these dependencies are needed to do that.

PR #336, opened 2020-04-12
Add User Option to allow invalid input paths

frankgh:
While running performance tests, some external tables did not have any
data, and when writting data to the partition, they did not create the
expected path while reading. To illustrate this better, this is a sample
code of what happened:

    CREATE TABLE foo (id int, some_date date)
    DISTRIBUTED BY (id)
    PARTITION BY RANGE (some_date)
    (start('1992-01-01') INCLUSIVE end ('1998-12-31') INCLUSIVE every
    (30), default partition others);

The stamement above will create 87 partitions. In our experiment, some
of the partitions did not have any data (there were no records for some
date ranges).

We then replace the internal partition with an external partition, as
illustrated below:

    CREATE EXTERNAL TABLE replace_partition_87 (id int, some_date date)
    LOCATION (
    'pxf://bucket/non-existent-path/?PROFILE=s3:parquet&SERVER=s3'
    ) FORMAT 'CUSTOM' (formatter = 'pxfwritable_import');

    ALTER TABLE foo
    EXCHANGE PARTITION FOR (RANK(87)) WITH TABLE replace_partition_87
    WITHOUT VALIDATION;

Because some of the external tables end up with no data, queries fail
when reading the original partitioned table `foo`. This commit
introduces a new user property `IGNORE_INVALID_INPUT`, which when set to
true, it will ignore the error and return an empty list of fragments.
The new table definition would look like this:

    CREATE EXTERNAL TABLE replace_partition_87(id int, some_date date)
    LOCATION (
    'pxf://bucket/non-existent-path/?PROFLE=s3:parquet&IGNORE_INVALID_INPUT=true'
    ) FORMAT 'CUSTOM' (formatter = 'pxfwritable_import');

This commit does not change the default behaviour of erroring out when
the partition is missing, but rather gives users an option to not error
out when the input path is missing.

### DIFF
--- a/server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HdfsDataFragmenter.java
+++ b/server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HdfsDataFragmenter.java
@@ -20,9 +20,11 @@ package org.greenplum.pxf.plugins.hdfs;
  */
 
 
+import org.apache.commons.lang.StringUtils;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.mapred.FileSplit;
 import org.apache.hadoop.mapred.InputSplit;
+import org.apache.hadoop.mapred.InvalidInputException;
 import org.apache.hadoop.mapred.JobConf;
 import org.greenplum.pxf.api.model.BaseFragmenter;
 import org.greenplum.pxf.api.model.Fragment;
@@ -64,7 +66,16 @@ public class HdfsDataFragmenter extends BaseFragmenter {
     @Override
     public List<Fragment> getFragments() throws Exception {
         Path path = new Path(hcfsType.getDataUri(jobConf, context));
-        List<InputSplit> splits = getSplits(path);
+        List<InputSplit> splits;
+        try {
+            splits = getSplits(path);
+        } catch (InvalidInputException e) {
+            if (StringUtils.equalsIgnoreCase("true", context.getOption("IGNORE_INVALID_INPUT"))) {
+                LOG.debug("Ignoring InvalidInputException", e);
+                return fragments;
+            }
+            throw e;
+        }
 
         LOG.debug("Total number of fragments = {}", splits.size());
         for (InputSplit split : splits) {
--- a/server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HdfsDataFragmenter.java
+++ b/server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HdfsDataFragmenter.java
@@ -46,6 +46,8 @@ import java.util.List;
  */
 public class HdfsDataFragmenter extends BaseFragmenter {
 
+    protected static final String IGNORE_MISSING_PATH_OPTION = "IGNORE_MISSING_PATH";
+
     protected JobConf jobConf;
     protected HcfsType hcfsType;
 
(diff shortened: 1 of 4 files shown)


server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HdfsDataFragmenter.java
```diff
@@ -64,7 +66,16 @@ public void initialize(RequestContext context) {
     @Override
     public List<Fragment> getFragments() throws Exception {
         Path path = new Path(hcfsType.getDataUri(jobConf, context));
-        List<InputSplit> splits = getSplits(path);
+        List<InputSplit> splits;
+        try {
+            splits = getSplits(path);
+        } catch (InvalidInputException e) {
+            if (StringUtils.equalsIgnoreCase("true", context.getOption("IGNORE_INVALID_INPUT"))) {
```
denalex, line 73: maybe declare "IGNORE_INVALID_INPUT" as a constant on RequestContext ?

denalex: can we think of a better name ? "Invalid input" is not smth exposed to end-users defining path in LOCATION Url, for them it is either "path" or "resource", in fact we call it in our docs "path to data" `LOCATION('pxf://<path-to-data>?PROFILE=<profile_name>`, so maybe at least `IGNORE_INVALID_PATH`. It is also not that the path is invalid (like in violation of syntax rules) but more as non-existent, but `IGNORE_NON_EXISTENT_PATH` is too long and prone to typing errors. I'd prefer 1-2 simple words max.

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/HdfsFileFragmenterTest.java
```diff
@@ -74,4 +74,37 @@ public void testFragmenterWilcardPath() throws Exception {
...
+        context.setDataSource("/tmp/non-existent-path-on-disk/*.csv");
+
+        Fragmenter fragmenter = new HdfsFileFragmenter();
+        fragmenter.initialize(context);
+
+        List<Fragment> fragmentList = fragmenter.getFragments();
+        assertNotNull(fragmentList);
+        assertEquals(0, fragmentList.size());
+    }
 }
```
denalex, line 110: what about other profiles ? If we introduce such an option, users might expect it to work similarly for other profiles (Hive, JDBC, etc). So maybe have other profiles proactively throw an error if they encounter this setting ? Or will this be too restrictive, since they will essentially ignore the option and error out if they determine actually missing path.

frankgh: I thought about that, but I think this concept of `PATH` does not apply to other profiles. In essence, this feature allows supporting partitioned tables. Not in the Hive partitioned tables sense, but rather Greenplum partitioned tables backed by external tables accessing Hadoop Compatible Filesystems.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HdfsDataFragmenter.java
frankgh: maybe something like `IGNORE_MISSING_PATH` could also work

frankgh: Using `IGNORE_MISSING_PATH`

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/HdfsFileFragmenterTest.java
oliverralbertini: Is it fair to say that the only profiles that don't use Hdfs{Data,File}Fragmenter are Hive and JDBC?

frankgh: we also have Hbase and S3 Select

oliverralbertini:
> we also have Hbase and S3 Select

thanks!

PR #342, opened 2020-04-21
changed Makefiles to build tarball with all components

denalex:


Makefile
```diff
@@ -1,18 +1,27 @@
...
-    endif
+    PXF_HOME = "$(GPHOME)/pxf"
 endif
 
 SHELL := /bin/bash
 
+export GPHOME
 export PXF_HOME
 
+PXF_VERSION := $(shell grep version= server/gradle.properties | cut -d= -f2)
```
frankgh, line 14: Should we take this opportunity to bump the version of our PXF server to `5.12.0-SNAPSHOT` here https://github.com/greenplum-db/pxf/blob/master/server/gradle.properties? Maybe it's an opportunity to start using snapshots moving forward. This might break CI code that assumes the version number is major, minor, rev

package/install_component
```diff
@@ -0,0 +1,43 @@
...
+    fi
+
+    if [[ ! -f ${GPHOME}/greenplum_path.sh ]]; then
+        echo "Error: environment variable GPHOME must be set to a valid Greenplum installation"
+        exit 1
+    fi
+}
+
+function remove_old_component() {
+    echo "Removing old PXF ..."
```
frankgh, line 19: minor: this does not necessarily have to be an old version of PXF
```suggestion
    echo "Removing existing PXF ..."
```

package/install_component
```diff
@@ -0,0 +1,43 @@
...
+
+}
+
+function install_new_component() {
+    echo "Installing new PXF version ${PXF_VERSION} ..."
```
frankgh, line 24: minor:
```suggestion
    echo "Installing PXF version ${PXF_VERSION} ..."
```

package/install_component
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env bash
+
+CWDIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
+PXF_VERSION=$(<"${CWDIR}/pxf/version")
+
+function check_gphome() {
+    if [[ -z "${GPHOME}" ]]; then
```
frankgh, line 7: I think we should allow a $PXF_HOME environment variable to be defined, and if not. defined then PXF_HOME=GPHOME/pxf. We install the PXF Server + CLI components in PXF_HOME, and the extension inside GPHOME

Makefile
oliverralbertini: I agree, this might be the right time to start versioning all our tarballs on commit, and reserve the non SHA versions for releases.

oliverralbertini: perhaps we can have some logic to check if we are on a git tag, otherwise we tack on a SHA

package/install_component
denalex: thanks, will update, but this generally would require way more logic to check for existing PXF install, make sure it is stopped / killed, and removing bits, which will be a separate story.

package/install_component
```diff
@@ -0,0 +1,84 @@
...
+    fi
+}
+
+function remove_old_component() {
+    if [[ ! -z "${PXF_HOME}" && -d ${PXF_HOME} ]]; then
+        if [[ -f ${PXF_HOME}/version ]]; then
+            OLD_PXF_VERSION="version $(<"${PXF_HOME}/version") "
+        fi
+
+        echo "Removing existing PXF ${OLD_PXF_VERSION}from ${PXF_HOME} ..."
```
frankgh, line 48:
```suggestion
        echo "Removing existing PXF ${OLD_PXF_VERSION} from ${PXF_HOME} ..."
```

server/Makefile
```diff
@@ -17,30 +17,34 @@
...
+default: stage-notest
 
-BUILD_PARAMS+= -DdeployPath="$(PXF_HOME)"
 TOMCAT_VERSION="7.0.100"
 TOMCAT_DIR="tomcat/build"
 TOMCAT_TAR="apache-tomcat-${TOMCAT_VERSION}.tar.gz"
 TOMCAT_URL="http://archive.apache.org/dist/tomcat/tomcat-7/v${TOMCAT_VERSION}/bin/${TOMCAT_TAR}"
 
+PXF_ROOT_DIR := $(shell cd .. && pwd)
+PXF_VERSION ?= $(shell $(PXF_ROOT_DIR)/version)
```
oliverralbertini, line 30: same comment as for pxf-cli/Makefile

server/pxf-jdbc/src/test/java/org/greenplum/pxf/plugins/jdbc/utils/ConnectionManagerTest.java
```diff
@@ -76,6 +77,7 @@ public void testGetConnectionPoolDisabled() throws SQLException {
     }
 
     @Test
+    @Ignore
```
oliverralbertini, line 80: We going to keep this?

denalex: I'll try uncommenting this, will still be failing on local Docker I guess.

concourse/scripts/test.bash
```diff
@@ -0,0 +1,250 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
denalex: yes, we will at some later point when the old pipeline is obsolete

cli/go/src/pxf-cli/Makefile
```diff
@@ -1,28 +1,30 @@
...
 
 TARGET := $(shell echo $${PWD\#\#*/})
 PXF_ROOT_DIR := $(shell cd ../../../.. && pwd)
-GO_DEP_CACHE := ${PXF_ROOT_DIR}/cli/go/pkg/dep
 PXF_SERVER_DIR := ${PXF_ROOT_DIR}/server
-PXF_STAGING_DIR := ${PXF_SERVER_DIR}/build/stage/bin
-PXF_SERVICE_SCRIPTS := ${PXF_SERVER_DIR}/pxf-service/src/scripts
-PXF_VERSION := $(shell grep version= ${PXF_SERVER_DIR}/gradle.properties | cut -d= -f2)
+PXF_CLI_DIR := ${PXF_ROOT_DIR}/cli
+PXF_VERSION ?= $(shell $(PXF_ROOT_DIR)/version)
```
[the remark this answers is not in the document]
denalex: good point on missing `cat`, wonder how it works now then ? Including in `common.mk` will be a problem, though, as the path to version file is different from different Makefiles (go Makefile is buried too deep from the top)

concourse/scripts/build.bash
```diff
@@ -26,39 +23,17 @@ function install_gpdb() {
...
-    pushd dist > /dev/null
-    tar cvzf ${package_name}.tar.gz ${package_name}
-    popd > /dev/null
-
     # verify contents
-    ls -al dist
-    tar -tvzf dist/${package_name}.tar.gz
+    ls -al pxf_src/build/dist
+    tar -tvzf pxf_src/build/dist/pxf-*.tar.gz
+    cp pxf_src/build/dist/pxf-*.tar.gz dist
```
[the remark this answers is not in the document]
denalex: I don't think it belongs in Makefile, this just list the content of the tarball and copies it into CI job's output location.

concourse/scripts/install_hadoop.bash
```diff
@@ -12,8 +12,14 @@ REMOTE_GPHD_ROOT=~centos/singlecluster
...
 	local hadoop_ip=${1}
 
-	tar -xzf pxf_tarball/pxf.tar.gz -C /tmp
-	cp /tmp/pxf/lib/pxf-hbase-*.jar "${LOCAL_GPHD_ROOT}/hbase/lib"
+	tar -xzf pxf_tarball/pxf*.tar.gz -C /tmp
+	if [[ -d /tmp/pxf/lib ]]; then
+	    # regular tarball based install (old)
+	    cp /tmp/pxf/lib/pxf-hbase-*.jar "${LOCAL_GPHD_ROOT}/hbase/lib"
+	else
+	    # componet tarball based install (new)
```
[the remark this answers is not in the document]
denalex: thanks for noticing the typo. As for PXF_COMPONENT, maybe , this was done before it and here we can actually look at what we got, in other cases the code was expanding the tarball right into GPHOME/pxf, here we're expanding into /tmp. Maybe we will revisit once we change tarball into RPM, will leave this as is for now.

dev/install_gpdb_package.bash
```diff
@@ -0,0 +1,60 @@
...
+CWDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+# TODO: check if gpadmin-limits.conf already exists and bail out if it does
+>/etc/security/limits.d/gpadmin-limits.conf cat <<-EOF
+gpadmin soft core unlimited
+gpadmin soft nproc 131072
+gpadmin soft nofile 65536
+EOF
+
+>>/home/gpadmin/.bash_profile cat <<EOF
```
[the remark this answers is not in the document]
denalex: will take a look, dev docker image story is not yet complete

cli/go/src/pxf-cli/Makefile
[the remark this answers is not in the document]
oliverralbertini: I think it's grabbing the version from somewhere else (I notice the ?= assignment which allows previous values to live on).

PR #343, opened 2020-04-21
Upgrade hive 2.3.7 and support Java 11 for Hive Profiles

frankgh:
When supporting hive 3, 2 and 1, supporting Hive on Java 11 broke. We documented this as a limitation

https://gpdb.docs.pivotal.io/6-2/relnotes/gpdb-62-release-notes.html

> 168689202 PXF PXF fails to run any query on Java 11 that specifies a Hive* profile due to this Hive known issue: ClassCastException when initializing HiveMetaStoreClient on JDK10 or newer.
> Workaround: Run PXF on Java 8 or use the PXF JDBC Connector to access Hive.

We pushed a commit https://github.com/greenplum-db/pxf/commit/041cbc7cf2aaf93ec9fc0978c7123d092d7566c7 to disable testing Hive on Java 11.

There are two commits:
1. Revert commit https://github.com/greenplum-db/pxf/commit/041cbc7cf2aaf93ec9fc0978c7123d092d7566c7: Restores tests for Hive on the Java11 job
2.  Upgrade Hive to version 2.3.7


oliverralbertini:
> not sure why the changes to "pushdown" test group, but otherwise looks good.

Is it that the pushdown group was created just to get around this limitation? that's the impression that I got also for cloud-access-and-hdfs group. Now that we have the new jar, we can just run the regular big groups. Is that right, @frankgh ?

frankgh:
yes, that's correct, we still wanted to run some tests on the job, so we created the group to be able to run those tests

PR #345, opened 2020-04-22
Convert the pxf-api project to a boot project

frankgh:
- Provide a top level build.gradle with only pxf-api enabled
- Provide a build.gradle for pxf-api
- Migrate unit tests to JUnit 5
- Make all unit tests pass


server/pxf-api/src/main/java/org/greenplum/pxf/api/security/SecureLogin.java
```diff
@@ -68,6 +68,11 @@
 
     private static final SecureLogin instance = new SecureLogin();
 
+    /**
+     * Package-private for testing
+     */
+    static PxfUserGroupInformation pxfUserGroupInformation = PxfUserGroupInformation.getInstance();
```
denalex, line 74: can maybe this and PxfUserGroupInformation be Spring services, Spring will manage their "singleton-ness" and have them autowired ?

server/pxf-api/src/test/java/org/apache/hadoop/security/PxfUserGroupInformationTest.java
```diff
@@ -2,97 +2,104 @@
...
         // prepare objects
         nowMs = System.currentTimeMillis();
         configuration = new Configuration();
         user = new User("user");
         serverName = "server";
+        pxfUserGroupInformation = PxfUserGroupInformation.getInstance();
 
         // prepare common mocks
-        mockTGT = PowerMockito.mock(KerberosTicket.class); // has final methods, needs PowerMock to mock it
+        mockTGT = mock(KerberosTicket.class);
```
denalex, line 85: it can now mock classes with final methods without PowerMock? cool

server/pxf-api/src/test/java/org/apache/hadoop/security/PxfUserGroupInformationTest.java
```diff
@@ -144,128 +151,106 @@ public void testLoginFromKeytabMinMillisFromDefault() throws Exception {
...
+        assertEquals(" loginUserFromKeyTab must be done first", e.getMessage());
     }
 
     /* ---------- Test below follow full login path via a few alternatives ---------- */
 
     @Test
     public void testReloginFromKeytabNoValidTGT() throws Exception {
 
-        assertEquals(1, subject.getPrivateCredentials().size()); // subject has 1 ticket
+        assertEquals(2, subjectWithKerberosKeyTab.getPrivateCredentials().size()); // subject has 2 tickets
```
denalex, line 239: not sure why things changed here

server/pxf-api/src/test/java/org/apache/hadoop/security/PxfUserGroupInformationTest.java
```diff
@@ -274,7 +259,7 @@ public void testReloginFromKeytabNoValidTGT() throws Exception {
         /* subject's non-TGT ticket has been removed, in reality another one would be created by login process,
          * but we are not mocking it here.
          */
-        assertTrue(subject.getPrivateCredentials().isEmpty());
+        assertEquals(1, subject.getPrivateCredentials().size());
```
denalex, line 262: and here ...

server/build.gradle
```diff
@@ -17,345 +17,97 @@
  * under the License.
  */
 
+buildscript {
+    repositories {
+        jcenter()
```
denalex, line 22: that's already declared below (line 45), do we need in both places ?

server/gradlew
```diff
@@ -172,111 +154,30 @@ if $cygwin ; then
...
-
-  eval splitJvmOpts $DEFAULT_JVM_OPTS $JAVA_OPTS $GRADLE_OPTS
-  JVM_OPTS[${#JVM_OPTS[*]}]="-Dorg.gradle.appname=$APP_BASE_NAME"
-  exec "$JAVACMD" "${JVM_OPTS[@]}" -cp $(classpath) org.gradle.launcher.GradleMain "$@"
-}
+# Collect all arguments for the java command, following the shell quoting and substitution rules
+eval set -- $DEFAULT_JVM_OPTS $JAVA_OPTS $GRADLE_OPTS "\"-Dorg.gradle.appname=$APP_BASE_NAME\"" -classpath "\"$CLASSPATH\"" org.gradle.wrapper.GradleWrapperMain "$APP_ARGS"
 
-main "$@"
+exec "$JAVACMD" "$@"
```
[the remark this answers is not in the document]
frankgh: correct

server/pxf-api/build.gradle
```diff
@@ -0,0 +1,47 @@
...
+
+    /*******************************
+     * Implementation Dependencies
+     *******************************/
+
+    implementation("org.springframework.boot:spring-boot-starter-log4j2")
+    implementation("commons-codec:commons-codec")
+    implementation("com.fasterxml.jackson.core:jackson-core")
+    implementation("com.fasterxml.jackson.core:jackson-databind")
+
```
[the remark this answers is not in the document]
frankgh: Here's some info about API vs implementation: https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_separation This is also a good read: https://stackoverflow.com/questions/44413952/gradle-implementation-vs-api-configuration

Here's some info about annotationProcessor: https://docs.gradle.org/current/userguide/java_plugin.html#making_an_annotation_processor_incremental

server/pxf-api/src/main/java/org/greenplum/pxf/api/security/SecureLogin.java
frankgh: yes, but we're deferring for later for now

server/pxf-api/src/test/java/org/apache/hadoop/security/PxfUserGroupInformationTest.java
frankgh: mockito-inline supports it: https://github.com/greenplum-db/pxf/blob/5f3680ba48faef81bde209a78c7040f4b8860e57/server/pxf-api/build.gradle#L39

server/pxf-api/src/test/java/org/apache/hadoop/security/PxfUserGroupInformationTest.java
frankgh: We can no longer mock the static `KerberosUtil.hasKerberosKeyTab(subject)` method. So we need to make KerberosUtil.hasKerberosKeyTab return the expected value. For that, when we want a subject with a Kerberos Keytab, we create a subject like this:
https://github.com/greenplum-db/pxf/blob/5f3680ba48faef81bde209a78c7040f4b8860e57/server/pxf-api/src/test/java/org/apache/hadoop/security/PxfUserGroupInformationTest.java#L92
which has two items in the private credentials as opposed to 1 before

server/pxf-api/src/test/java/org/apache/hadoop/security/PxfUserGroupInformationTest.java
frankgh: see my comment above

server/build.gradle
frankgh: i believe so, I will confirm.

PR #346, opened 2020-04-22
Remove Tomcat templates

frankgh:
- Remove CustomWebapploader
- Remove tomcat download from Makefile
- Remove `pxf-service/src/main/webapp/WEB-INF/web.xml`
- Remove `pxf-service/src/main/webapp/META-INF/context.xml`


server/pxf-service/src/scripts/pxf-service
```diff
@@ -137,7 +137,7 @@ function createInstance()
 
     # copy configuration files into instance
     cp "$tomcat_templates"/bin/{kill-pxf,setenv,catalina}.sh "${instance}/bin"
```
denalex, line 139: are we still gonna use `catalina.sh` ?

denalex: there are also references to tomcat elsewhere in this script (line 131)

frankgh: I think we were planning to reuse `catalina.sh` but we need to modify. There are some stories that address all scripts

PR #358, opened 2020-05-15
Fix Glob Patterns for Hadoop-Compatible FileSystems

frankgh:
- Added unit tests
- Added integration tests that test the following:

| Pattern        | Description                                                                                                                                                    |
|----------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------|
| ?              | Matches any single character                                                                                                                                   |
| *              | Matches zero or more characters                                                                                                                                |
| [abc]          | Matches a single character from character set {a,b,c}.                                                                                                         |
| [a-b]          | Matches a single character from the character range {a…b}.                                                                                                     |
| [^a]           | Matches a single character that is not from character set or range {a}. Note that the “^”character must occur immediately to the right of the opening bracket. |
| \c             | Removes (escapes) any special meaning of character c.                                                                                                          |
| {ab,cd}        | Matches a string from the string set {ab, cd}.                                                                                                                 |
| {ab,c{de, fh}} | Matches a string from the string set {ab, cde, cfh}.                                                                                                           |


server/pxf-api/src/test/java/org/greenplum/pxf/api/utilities/UtilitiesTest.java
```diff
@@ -413,4 +413,48 @@ public void testSecurityIsEnabledWithKerberosAuthentication() {
...
+        assertNull(Utilities.getHost("file:///foo/bar.txt"));
+        assertEquals("0.0.0.0", Utilities.getHost("hdfs://0.0.0.0:8020"));
+        assertEquals("abc", Utilities.getHost("hdfs://abc:8020/foo/bar.txt"));
+        assertEquals("0.0.0.0", Utilities.getHost("hdfs://0.0.0.0:8020/tmp/issues/172848577/[a-b].csv"));
+
+        assertEquals("0.0.0.0", Utilities.getHost("hdfs://0.0.0.0#anchor"));
+        assertEquals("0.0.0.0", Utilities.getHost("hdfs://0.0.0.0/p"));
+        assertEquals("0.0.0.0", Utilities.getHost("hdfs://0.0.0.0?PROFILE=foo"));
+        assertEquals("www.example.com", Utilities.getHost("www.example.com"));
+        assertEquals("10.0.0.15", Utilities.getHost("10.0.0.15"));
```
denalex, line 458: nice test, would be good to also see smith starting with a slash `/foo/bar` and variations

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HcfsType.java
```diff
@@ -235,15 +233,12 @@ protected String getDataUriForPrefix(Configuration configuration, String dataSou
...
-        if (Utilities.isSecurityEnabled(configuration))
+        if (Utilities.isSecurityEnabled(configuration) || StringUtils.isBlank(uri))
             return;
 
-        // find the "host" that TokenCache will check against the exclusion list, for cloud file systems (like S3)
-        // it might actually be a bucket in the full resource path
-        String host = URI.create(StringUtils.replace(uri, " ", "%20")).getHost();
-        // String host = URI.create(uri).getHost();
-        LOG.debug("Disabling token renewal for host {} for path {}", host, uri);
+        String host = Utilities.getHost(uri);
```
denalex, line 239: I would like to preserve the comment on why we are finding the hostname in the first place

server/pxf-api/src/test/java/org/greenplum/pxf/api/utilities/UtilitiesTest.java
frankgh: I just added one like that! Oliver also pointed it out.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HcfsType.java
frankgh: good catch. Didn't mean to remove it

PR #360, opened 2020-05-16
Add additional information regarding the pxf.service.user.name property

frankgh:
Add additional information regarding the pxf.service.user.name property
in pxf-site.xml. This property has been causing confusion in users and
it seems they are modifying the value of this property even in
environments where they should not. In this commit, I am moving the
property to the bottom of the file, and I add a comment in hopes that
users do read this comment, because they usually don't read the
<description> tag

### DIFF
--- a/server/pxf-service/src/templates/user/templates/pxf-site.xml
+++ b/server/pxf-service/src/templates/user/templates/pxf-site.xml
@@ -11,13 +11,21 @@
         <description>Kerberos path to keytab file owned by pxf service with permissions 0400</description>
     </property>
     <property>
-        <name>pxf.service.user.name</name>
-        <value>${user.name}</value>
-        <description>Login user for remote system, defaults to the OS user that started PXF process</description>
-    </property>
-    <property>
         <name>pxf.service.user.impersonation</name>
         <value>${pxf.service.user.impersonation.enabled}</value>
         <description>End-user identity impersonation, set to true to enable, false to disable</description>
+    </property>
+    <property>
+        <name>pxf.service.user.name</name>
+        <value>${user.name}</value>
+        <description>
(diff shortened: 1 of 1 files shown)


server/pxf-service/src/templates/user/templates/pxf-site.xml
```diff
@@ -10,14 +10,20 @@
...
         <name>pxf.service.user.impersonation</name>
         <value>${pxf.service.user.impersonation.enabled}</value>
         <description>End-user identity impersonation, set to true to enable, false to disable</description>
     </property>
+
+    <!--
+    The pxf.service.user.name property below is used to specify the login user,
+    typically when connecting to a non-Kerberized hadoop cluster, using a
+    different user other than the user that started the PXF process.
+    -->
```
denalex, line 23: How about smth like this (to explain why it is needed, what default means and why to change it):
"The pxf.service.user.name property below is used to specify the login identity when connecting to a remote system, typically an unsecured Hadoop cluster. By default it is set to the user that started the PXF process. If PXF impersonation feature is used, this is the identity that needs to be granted Hadoop proxy user privileges. Change it ONLY if you would like to use another identity to login to an unsecured Hadoop cluster.

denalex: actually we can just add it to description attribute below, no reason for XML comment.

frankgh: mostly because people don't read the description. @denalex should we keep it as both an XML comment and the description tag?

denalex: I don't know, I believe as long as the description is there and it's informative enough we should be ok, nothing will cause people to read things they are not willing to read :) You decide.

PR #361, opened 2020-05-19
added packaging README

denalex:


package/README.md
```diff
@@ -0,0 +1,55 @@
+PXF Packaging
+============
+
+PXF consists of 3 groups of artifacts, each developed using a different underlying technology:
+
+* Greenplum extension -- written in C, represented by a `pxf.so` library and configuration files
+* PXF Server -- written in Java, represented by a `pxf.war` file, Tomcat server, dependent JAR files, templates and scripts
+* Script Cluster Plugin -- written in Go, represented by `pxf-cli` executable
+
+When packaged and shipped, PXF is represented by an RPM package on Centos platform and a DEB package on Ubuntu platform,
```
lisakowen, line 10: seems like the info on this page is geared both towards the consumer of the packages and a developer who would be building the package on their own.  i might say "The PXF packaging framework can build and generate an RPM package for CentOS platforms and a DEB package for the Ubuntu platform".  or "You can build an RPM ..."

package/README.md
```diff
@@ -0,0 +1,55 @@
...
+respectively. The package is different for every major Greenplum version that PXF has been compiled with.
```
lisakowen, line 11: how about something like "PXF compiles against and generates a different package for every major Greenplum version."

package/README.md
```diff
@@ -0,0 +1,55 @@
...
+`pxf-gpX.spec` files in this directory. The following key design decisions were made:
+
+* the name of the RPM package is `pxf-gpX`, where X is the major Greenplum version (e.g. `pxf-gp5`, `pxf-gp6`)
+* the install a newer RPM package for the same Greenplum major release, a user will have to upgrade the PXF RPM
+* the RPM installs PXF server into `/usr/local/pxf-gpX` directory (e.g. `/usr/local/pxf-gp6`)
+* the RPM is relocatable, a user can specify --prefix option when installing the RPM to install the server into another directory
+* the PXF greenplum extension is initially installed by RPM alongside the PXF server and is not initially active
+* the PXF greenplum extension is copied into Greenplum install location during `pxf init` command issued by a user after the install
+* the PXF RPM version number follows 3-number semantic versioning and must be provided during the RPM build process
+* the PXF RPM release number is usually specified as `1`
```
lisakowen, line 28: just curious, when would PXF release a package with `-2`?

package/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+## RPM installation process
+To install PXF from an RPM:
+1. Build or download PXF RPM for the corresponding major version of Greenplum. The following example will assume
+   that PXF version `1.2.3` will be installed to work with with Greenplum 5.
+2. Decide which OS user will own the PXF installation. If PXF is installed alongside Greenplum, the user that owns the PXF
+installation should either be the same as the one owning the Greenplum installation or have write privilleges to the
+Greenplum installation directory. This is necessary to be able to register the PXF Greenplum extension with Greenplum.
+3. If a previous PXF version has been installed, stop pxf server. 
+4. As a superuser, run `rpm -Uvh pxf-gp5-1.2.3-1.el7.x86_64.rpm` to install the RPM into `/usr/local/pxf-gp5`
```
lisakowen, line 48: so, the user wouldn't run `yum` to install the RPM?

will the user still need to install Java separately?

package/README.md
```diff
@@ -0,0 +1,55 @@
...
+5. As a superuser, run `chown gpadmin:gpadmin /usr/local/pxf-gp5` to change ownership of PXF installation to the user `gpadmin`.
+Specify a different user other than `gpadmin`, if desired.
+
+
+After these steps, the PXF product will be installed and is ready to be configured. If there was a previous installation of
+PXF for the same major Greenplum version, the files and the runtime directories from the older version will be removed.
```
lisakowen, line 54: does this mean that installing a PXF RPM removes the $GPHOME/pxf directory?

package/README.md
denalex: thanks, I will change to "The PXF build system can create an RPM package ..."

package/README.md
denalex: probably never, release number is the RPM thing where each software version might have multiple releases (builds) that are available. Our dev team issues 1 artifact per version only and is not planning to use multiple release numbers. However, other vendors might decide otherwise.

package/README.md
denalex: they can run `yum` as well, let me add it, thanks for catching this. User will still have to install Java, correct, maybe I should mention this as well, however, I don't want to turn this into a detailed installation guide, since this is developer's documentation primarily

package/README.md
denalex: no, it does not remove it. But this guide assumes developer's focus going forward. Moving from non-RPM to RPM-based install belongs in migration guide that is not the focus here. "older versions" here are assumed to be installed by RPM as well.

package/README.md
```diff
@@ -0,0 +1,55 @@
...
+For example, `pxf-gp5-1.2.3-1.el7.x86_64.rpm` represents an RPM package of PXF version 1.2.3 intended to work with
+Greenplum 5 on Centos / Redhat 7 operating systems.  
+
+
+## RPM specification
+On Centos platforms PXF product is packaged as an RPM. The specification on how to build the RPM is provided by the
+`pxf-gpX.spec` files in this directory. The following key design decisions were made:
+
+* the name of the RPM package is `pxf-gpX`, where X is the major Greenplum version (e.g. `pxf-gp5`, `pxf-gp6`)
+* the install a newer RPM package for the same Greenplum major release, a user will have to upgrade the PXF RPM
```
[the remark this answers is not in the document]
denalex: yes, `package` is a new dir, I think it is justifiable, as this is a project-level package build specifications and helper scripts

PR #365, opened 2020-06-03
Migrate pxf-json to Spring Boot

frankgh:
- Update top level build.gradle to include pxf-json
- Add new build.gradle for pxf-json
- all existing unit tests should be ported to junit 5
- make test on java 11
- automation tests passing for JSON


server/pxf-json/build.gradle
```diff
@@ -0,0 +1,31 @@
...
+     * API Dependencies
+     *******************************/
+
+    /*******************************
+     * Implementation Dependencies
+     *******************************/
+
+    implementation(project(':pxf-api'))
+    implementation(project(':pxf-hdfs'))
+    implementation("org.springframework.boot:spring-boot-starter-log4j2")
```
denalex, line 18: wonder why we need to depend on this and not only on slf4j-api ?

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonResolver.java
```diff
@@ -51,15 +54,14 @@
     private ObjectMapper mapper;
 
     @Override
-    public void initialize(RequestContext requestContext) {
-        super.initialize(requestContext);
+    public void initialize() {
         oneFieldList = new ArrayList<>();
         mapper = new ObjectMapper();
```
denalex, line 59: can we reuse a single object Mapper if we do not customize its behavior ?

server/pxf-json/build.gradle
frankgh: There's a good explanation here: https://docs.spring.io/spring-boot/docs/2.0.x/reference/html/howto-logging.html but basically this provides an abstraction for logging in spring boot.

PR #366, opened 2020-06-03
Migrate pxf-s3 to Spring Boot

frankgh:
- Update top level build.gradle to include pxf-s3
- Add new build.gradle for pxf-s3
- all existing unit tests should be ported to junit 5


server/pxf-s3/src/main/java/org/greenplum/pxf/plugins/s3/S3SelectAccessor.java
```diff
@@ -129,9 +139,16 @@ public void closeForRead() throws Exception {
...
+                error = defaultIfNull(error, e);
             }
         }
 
+        if (error != null) {
+            // If there's an error during closing of streams, report the first
+            // encountered error
+            throw error;
+        }
+
```
denalex, line 151: why do we want to do this now ? by this point we already read all data and just failed to close the stream, this is usually ignored as the stream itself will be GC-ed in a short while and the caller does not get any impact from this error ?

frankgh: The error is swallowed by the `BridgeResponse` take a look at: https://github.com/greenplum-db/pxf/blob/spring-boot/server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/BridgeResponse.java#L87 but I still think we should report errors generated during the stream closing, and they should be handled by the caller. The result is the same for the end user

PR #367, opened 2020-06-04
pxf-dev-base: add LANG to all images

oliverralbertini:
This prevents failures on unit tests like
BridgeOutputBuilder#convertTextDataToLines() where the text contains
special characters.


oliverralbertini:
The reason we are seeing this error is we have this change on the `image-loading` branch:
image

oliverralbertini:
> do we need it in the ubuntu image?
It's already in there, actually...

PR #369, opened 2020-06-04
README: point status badges to new pipelines

oliverralbertini:


oliverralbertini:
image
 ^ This is how it would look. You can check it out at https://github.com/greenplum-db/pxf/tree/fix-badge

oliverralbertini:
> Maybe have Master FDW be 3rd in the order as it is not as important as the other 2

Aww shoot just merged this. My bad.. We will do that later.

PR #371, opened 2020-06-05
Publish PXF releases to all RelEng buckets

oliverralbertini:


concourse/pipelines/templates/build_pipeline-tpl.yml
```diff
@@ -278,6 +278,40 @@ resources:
     branch: ((ccp-git-branch))
     private_key: ((ccp-git-key))
     uri: ((ccp-git-remote))
+
+- name: pxf_gp5_rhel6_release
```
denalex, line 282: I would add a suffix `_releng` to the end of these resource names so it is clear we are uploading them to the releng realm and not our own release buckets

frankgh: or `_prod` will also work I guess

PR #383, opened 2020-06-16
Avro: support writing SMALLINT to Avro

oliverralbertini:
The Postgres/GPDB type SMALLINT has no corresponding Avro type[0].

In this commit we make SMALLINT types being written to Avro external
sources appear as Avro integer type. For reading Avro data, the
Greenplum external table should always use INT or BIGINT.

[0] https://avro.apache.org/docs/current/spec.html#schema_primitive


automation/src/test/java/org/greenplum/pxf/automation/features/avro/HdfsWritableAvroTest.java
```diff
@@ -24,8 +24,21 @@
 
     private ReadableExternalTable readableExternalTable;
     private ArrayList<File> filesToDelete;
-    private final String[] avroPrimitiveTableCols = new String[]{
+    private final String[] avroPrimitiveWritableTableCols = new String[]{
```
frankgh, line 27: these should be static final String arrays
```suggestion
    private static final String[] AVRO_PRIMITIVE_WRITABLE_TABLE_COLS = new String[]{
```

automation/src/test/java/org/greenplum/pxf/automation/features/avro/HdfsWritableAvroTest.java
```diff
@@ -24,8 +24,21 @@
...
+            "type_long bigint",
+            "type_float real",
+            "type_double float8",
+            "type_string text",
+            "type_bytes bytea",
+            "type_boolean bool"
+    };
+    // values that were written from a smallint column (see above type_smallint)
+    // must be read back into integer columns
+    private final String[] avroPrimitiveReadableTableCols = new String[]{
```
frankgh, line 39:
```suggestion
    private static final String[] AVRO_PRIMITIVE_READABLE_TABLE_COLS = new String[]{
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/AvroResolver.java
```diff
@@ -145,6 +144,10 @@ public OneRow setFields(List<OneField> record) {
             if (field.val instanceof byte[]) {
                 field.val = ByteBuffer.wrap((byte[]) field.val);
             }
+            // Avro doesn't have a short, just an int type
```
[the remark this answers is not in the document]
oliverralbertini: Couldn't figure out how to do a switch with non-constant types (`DataType.BYTEA.getOID()`, etc.) but this does seem better now

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/AvroResolver.java
```diff
@@ -141,9 +140,12 @@ public OneRow setFields(List<OneField> record) {
...
         int cnt = 0;
         for (OneField field : record) {
-            // Avro does not seem to understand regular byte arrays
-            if (field.val instanceof byte[]) {
+            if (field.type == DataType.BYTEA.getOID()) {
+                // Avro does not seem to understand regular byte arrays
                 field.val = ByteBuffer.wrap((byte[]) field.val);
+            } else if (field.type == DataType.SMALLINT.getOID()) {
+                // Avro doesn't have a short, just an int type
+                field.val = (int) (short) field.val;
```
denalex, line 148: wonder if you really need double cast, would just `(int) field.val` work ?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/AvroResolver.java
```diff
@@ -84,7 +83,7 @@ public AvroResolver() {
     public void initialize(RequestContext requestContext) {
         super.initialize(requestContext);
 
-        hcfsType = HcfsType.getHcfsType(configuration, context);
+        HcfsType hcfsType = HcfsType.getHcfsType(configuration, context);
```
denalex, line 86: wonder why this is no longer a member field ?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/AvroResolver.java
oliverralbertini: With just the single cast, the editor warns of redundant cast. An alternative is this if you prefer:

```java
field.val = ((Short) field.val).intValue();
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/AvroResolver.java
oliverralbertini: Was only needed for `initialize()`.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/AvroResolver.java
frankgh: I think it's fine to keep it as is

oliverralbertini: To expand on this a bit, when just using `(Integer) field.val` or `(int) field.val` I see this error when the writable table has `SMALLINT` type:
```java
type  Exception report   message   java.lang.Short cannot be cast to java.lang.Integer
```
When casting between primitive types the above should work fine, but with reference types (I _think_) you need to cast twice in order to affect the underlying data (cast of a reference just changes the reference, not the data itself).

PR #386, opened 2020-06-17
docs - updates for rpm install, upgrade, related edits and file delet…

lisakowen:
…ions

in this PR:
- add a new pxf upgrade topic for upgrading a pxf installation with a pxf rpm.  (this installation may be an embedded pxf installation.  should also handle upgrade of pxf installed via rpm.)
- open source book update to reference new pxf upgrade topic
- update the "about pxf directories" topic with new file
- remove greenplum upgrade and migration topics - they will stay in greenplum docs
- add `pxf [cluster] register` command info
- add a few notes that commands assume `$PXF_INSTALL_DIR/bin` has been added to path; could then remove `$GPHOME/pxf/bin` prefix from MANY `pxf` command examples
- a few other misc edits

doc review site link:  http://docs-lisa-pxf-docset-rpm.cfapps.io/pxf/wip/release/index.html


docs/content/init_pxf.html.md.erb
```diff
@@ -47,7 +49,11 @@ Refer to [PXF User Configuration Directories](about_pxf_dir.html#usercfg) for a
 
 The PXF server runs on Java 8 or 11. You identify the PXF `$JAVA_HOME` and `$PXF_CONF` settings at PXF initialization time.
 
-Initializing PXF creates the PXF Java web application, and generates PXF internal configuration files, setting default properties specific to your configuration.
+Initializing PXF:
+
+- Copies PXF extension files from `$PXF_INSTALL_DIR/gpextable` to the Greenplum Database install directory `$GPHOME`.
```
frankgh, line 54: copying of the extension files depends on whether `$GPHOME/bin` is in the `$PATH`. If `$GPHOME/bin` is not in the path, a warning will displayed but the operation will succeed. I'm not sure how to put this in the documentation not to making too confusing

frankgh: maybe eventually we'll have instructions to install PXF outside of a greenplum cluster, and that would be easier?

docs/content/init_pxf.html.md.erb
```diff
@@ -86,7 +92,7 @@ Perform the following procedure to initialize PXF on each segment host in your G
 4. Run the `pxf cluster init` command to initialize the PXF service on the master, standby master, and on each segment host. For example, the following command specifies `/usr/local/greenplum-pxf` as the PXF user configuration directory for initialization:
 
     ``` shell
-    gpadmin@gpmaster$ PXF_CONF=/usr/local/greenplum-pxf $GPHOME/pxf/bin/pxf cluster init
+    gpadmin@gpmaster$ PXF_CONF=/usr/local/greenplum-pxf pxf cluster init
```
frankgh, line 95: Ensure `$GPHOME/bin` is in the `$PATH` for the PXF extension to be installed.

docs/content/init_pxf.html.md.erb
```diff
@@ -96,11 +102,11 @@ Perform the following procedure to initialize PXF on each segment host in your G
...
+When you reset PXF, PXF prompts you to confirm the operation. If you confirm, PXF removes the following runtime files and directories:
 
-- `$PXF_HOME/conf/pxf-private.classpath`
-- `$PXF_HOME/pxf-service`
-- `$PXF_HOME/run`
+- `$PXF_INSTALL_DIR/conf/pxf-private.classpath`
+- `$PXF_INSTALL_DIR/pxf-service`
+- `$PXF_INSTALL_DIR/run`
 
 PXF does not remove the `$PXF_CONF` directory during a reset operation.
```
frankgh, line 111:
```suggestion
PXF does not remove the `$PXF_CONF` directory during a reset operation nor does PXF removes the PXF extension files from the Greenplum Database installation .
```

docs/content/upgrade_pxf_rpm.html.md.erb
```diff
@@ -1,36 +1,40 @@
...
+title: Upgrading PXF
 ---
 
-If you are using PXF in your current Greenplum Database 6.x installation, you must upgrade the PXF service when you upgrade to a newer version of Greenplum Database 6.x.
+If you have installed the PXF `.rpm` and have initialized, configured, and are using PXF in your current Greenplum Database 5.21.2+ or 6.x installation, you must perform some upgrade actions when you install a new version of PXF.
 
-The PXF upgrade procedure describes how to upgrade PXF in your Greenplum Database installation. This procedure uses *PXF.from* to refer to your currently-installed PXF version and *PXF.to* to refer to the PXF version installed when you upgrade to the new version of Greenplum Database.
-
-The PXF upgrade procedure has two parts. You perform one procedure before, and one procedure after, you upgrade to a new version of Greenplum Database:
+The PXF upgrade procedure has two parts. You perform one procedure before, and one procedure after, you install a new version to upgrade PXF:
```
frankgh, line 7: can you downgrade your PXF version with an older version of the RPM?

docs/content/init_pxf.html.md.erb
lisakowen: $GPHOME/bin, or $GPHOME - i thought that just $GPHOME needs to be set.  

when would the user want to init without copying over the extension files?  if there is a real use case here we should call it out.

docs/content/upgrade_pxf_rpm.html.md.erb
lisakowen: there is only one version of the pxf rpm right now, so not possible to downgrade to an older "pxf installed via rpm".  will need to add these docs when we release the next version of the rpm.

that does bring up a good point, though.  can the user downgrade to an embedded pxf?  for sure not if they remove $GPHOME/pxf after installing the rpm.  even if they keep this directory around, the C library and extension files would not be recoverable from the current installation, right?

docs/content/init_pxf.html.md.erb
frankgh: I think our goal is to eventually allow installation of PXF on any host. In that case, you would only want to install the server portion on the PXF host, and the extension portion on the greenplum host

frankgh: I'm pretty sure that `$GPHOME/bin` needs to be on the path. Can @oliverralbertini or @denalex confirm?

docs/content/upgrade_pxf_rpm.html.md.erb
frankgh: true, it doesn't make sense for us to have a downgrade section for this release. To downgrade to the PXF version installed from Greenplum they would need to 
1. backup their extension files
2. backup their PXF server if they removed it previously
3. RPM uninstall
4. Restore files from 1  & 2

lisakowen: ok, in the pre-upgrade (step 1), lets have the user copy $GPHOME/pxf OR $PXF_INSTALL_DIR (depending on where pxf.current is located) and the PXF extension files somewhere.  they won't likely use them this release, but they will be available to downgrade to embedded pxf if they want to after first install of rpm.  what do you think about this?

but wouldn't downgrading pxf via rpm just involve removing current pxf rpm, install new pxf rpm, init, ...?  i.e. why would the user need to save anything?  maybe i am missing something?

related:  what do you think is the best way to determine if pxf.current was installed via rpm or is embedded?  we can't necessarily go by existence of $GHOME/pxf.  and the user can install the pxf rpm in a location other than the default (/usr/local/pxf-gpX), so i can't just check for the existence of that directory.  i have been instructing the  user to run "pxf version", but that is kind of nebulous, too - assumes they added their pxf bin/ directory to PATH.

docs/content/upgrade_pxf_rpm.html.md.erb
```diff
@@ -1,36 +1,40 @@
...
 
-3. Upgrade to the new version of Greenplum Database and then continue your PXF upgrade with [Step 2: Upgrading PXF](#pxfup).
+3. Install the new version of PXF, identify and note the new PXF version number, and then continue your PXF upgrade with [Step 2: Completing the PXF Upgrade](#pxfup).
 
 
-## <a id="pxfup"></a>Step 2: Upgrading PXF
+## <a id="pxfup"></a>Step 2: Completing the PXF Upgrade
 
-After you upgrade to the new version of Greenplum Database, perform the following procedure to upgrade and configure the *PXF.to* software:
+After you upgrade to the new version of PXF, perform the following procedure:
```
dyozie, line 37: This sentence also suggests to me that the pre-upgrade steps has already upgraded PXF.

docs/content/upgrade_pxf_rpm.html.md.erb
frankgh: > but wouldn't downgrading pxf via rpm just involve removing current pxf rpm, install new pxf rpm, init, ...? i.e. why would the user need to save anything? maybe i am missing something?

In my previous comment, I didn't mean for users to downgrade using an RPM, but rather what they would need to do to be able to restore their Greenplum + PXF installation after they have installed an RPM

> related: what do you think is the best way to determine if pxf.current was installed via rpm or is embedded? we can't necessarily go by existence of $GHOME/pxf. and the user can install the pxf rpm in a location other than the default (/usr/local/pxf-gpX), so i can't just check for the existence of that directory. i have been instructing the user to run "pxf version", but that is kind of nebulous, too - assumes they added their pxf bin/ directory to PATH.

I guess two ways: 1. the `pxf version` approach which as you mention would require `pxf` to be on the `$PATH`. 2. Run some RPM query `rpm -qi pxf-gp6` to determine if PXF was installed by RPM. This can bring some problems too, because even though you can have a PXF RPM installed, you can still be running the PXF version that was installed through a Greenplum installation.

docs/content/init_pxf.html.md.erb
oliverralbertini: It's just `GPHOME`, see [here](https://github.com/greenplum-db/pxf/blob/aed2b6968e44639a9e085e2bbef4e3e2a73bb13d/server/pxf-service/src/scripts/pxf#L369-L381) for details.

docs/content/init_pxf.html.md.erb
frankgh: apologies, `$GPHOME/bin` is required for building, not installing the extension

docs/content/init_pxf.html.md.erb
frankgh: Apologies, I confused building the extension with installing the extension.

docs/content/about_pxf_dir.html.md.erb
```diff
@@ -2,24 +2,27 @@
 title: About the PXF Installation and Configuration Directories
 ---
 
-PXF is installed on your master and segment nodes when you install Greenplum Database. 
+When you install the PXF `.rpm` package, PXF is installed to `/usr/local/pxf-gp<greenplum-major-version>`, or to a directory of your choosing. This documentation uses `$PXF_INSTALL_DIR` to refer to this directory.
```
oliverralbertini, line 5: I've been thinking of this directory as `PXF_HOME` still. I guess that's never been a part of the docs? Maybe we should align the docs with what we have in our [README](https://github.com/greenplum-db/pxf#how-to-install-pxf).

docs/content/cfginitstart_pxf.html.md.erb
```diff
@@ -27,9 +27,11 @@ PXF provides two management commands:
 - `pxf cluster` - manage all PXF service instances in the Greenplum Database cluster
 - `pxf` - manage the PXF service instance on a specific Greenplum Database host
 
-The [`pxf cluster`](ref/pxf-cluster.html) command supports `init`, `start`, `restart`, `status`, `stop`, and `sync` subcommands. When you run a `pxf cluster` subcommand on the Greenplum Database master host, you perform the operation on all segment hosts in the Greenplum Database cluster. PXF also runs the `init` and `sync` commands on the standby master host.
+The [`pxf cluster`](ref/pxf-cluster.html) command supports `init`, `reset`, `start`, `restart`, `status`, `stop`, `register`, and `sync` subcommands. When you run a `pxf cluster` subcommand on the Greenplum Database master host, you perform the operation on all segment hosts in the Greenplum Database cluster. PXF also runs the `init` and `sync` commands on the standby master host.
```
oliverralbertini, line 30: `register` also runs on the standby master.

docs/content/cfginitstart_pxf.html.md.erb
```diff
@@ -78,10 +80,10 @@ Perform the following procedure to start PXF on each segment host in your Greenp
     $ ssh gpadmin@<gpmaster>
     ```
 
-3. Run the `pxf cluster start` command to start PXF on each segment host. For example:
+3. Run the `pxf cluster start` command to start PXF on each segment host:
 
     ```shell
-    gpadmin@gpmaster$ $GPHOME/pxf/bin/pxf cluster start
+    gpadmin@gpmaster$ pxf cluster start
```
oliverralbertini, line 86: I really like this way of specifying the commands.

docs/content/upgrade_pxf_rpm.html.md.erb
```diff
@@ -1,36 +1,40 @@
...
+2. Identify and note the version of PXF currently running in your Greenplum cluster. For example:
+
+    ``` shell
+    gpadmin@gpmaster$ pxf version
+    ```
+
 2. Stop PXF on each segment host as described in [Stopping PXF](cfginitstart_pxf.html#stop_pxf).
 
-3. Upgrade to the new version of Greenplum Database and then continue your PXF upgrade with [Step 2: Upgrading PXF](#pxfup).
+3. Install the new version of PXF, identify and note the new PXF version number, and then continue your PXF upgrade with [Step 2: Completing the PXF Upgrade](#pxfup).
```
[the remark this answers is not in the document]
lisakowen: true.  but there is at least one other entry point to the postupgrade section - the installing topic xrefs to the Step2: Completing ....

docs/content/about_pxf_dir.html.md.erb
lisakowen: can the user set PXF_HOME?  is this environment variable, that is set by PXF, available to the user?  if so, when does PXF set it?

docs/content/cfginitstart_pxf.html.md.erb
```diff
@@ -27,7 +27,7 @@ PXF provides two management commands:
 - `pxf cluster` - manage all PXF service instances in the Greenplum Database cluster
 - `pxf` - manage the PXF service instance on a specific Greenplum Database host
 
-The [`pxf cluster`](ref/pxf-cluster.html) command supports `init`, `reset`, `start`, `restart`, `status`, `stop`, `register`, and `sync` subcommands. When you run a `pxf cluster` subcommand on the Greenplum Database master host, you perform the operation on all segment hosts in the Greenplum Database cluster. PXF also runs the `init` and `sync` commands on the standby master host.
+The [`pxf cluster`](ref/pxf-cluster.html) command supports `init`, `reset`, `start`, `restart`, `status`, `stop`, and `sync` subcommands. When you run a `pxf cluster` subcommand on the Greenplum Database master host, you perform the operation on all segment hosts in the Greenplum Database cluster. PXF also runs the `init`, `registe`, and `sync` commands on the standby master host.
```
dyozie, line 30: Typo introduced here:  registe

lisakowen:
@oliverralbertini 

> I would like to discuss a bit further whether we should use PXF_INSTALL_DIR or just PXF_HOME as we call it in the README. Maybe it's best to align the two.

i think we decided to use PXF_HOME, but not have the user explicitly set the environment variable.
 
> We should also discuss further how we want to approach the ability to revert to an older PXF version that came with a Greenplum installation.

i added a step to the pre-upgrade to copy/backup the $GPHOME/pxf and the extension bits.  let me know if that is not sufficient.

PR #389, opened 2020-06-19
Avro: add compression

oliverralbertini:
With this commit we can support writing (and reading back) Avro data
with Snappy, Deflate compressions along with no compression (by
specifying COMPRESSION_CODEC=null). For XZ and BZIP2 we cannot currently
support as it requires adding the Apache Commons Compression jar[0].

For Deflate and in the future XZ you can supply a compression level
between 1-9 inclusive, where higher is better compression, but slower
performance.

[0] https://commons.apache.org/proper/commons-compress/


automation/src/main/java/org/greenplum/pxf/automation/components/hdfs/Hdfs.java
```diff
@@ -365,6 +366,17 @@ public void writeJsonFileFromAvro(String pathToFile, String pathToJson)
...
     }
 
+    public void writeAvroMetadata(String pathToFile, String pathToMetadata)
+            throws Exception {
+        Tool tool = new DataFileGetMetaTool();
+        List<String> args = new ArrayList<>();
+        args.add(pathToFile);
+
+        try (PrintStream printStream = new PrintStream(new FileOutputStream(new File(pathToMetadata)))) {
+            tool.run(null, printStream, System.err, args);
```
denalex, line 376: not sure what this does, but we don't want to write to stderr from the app

automation/src/main/java/org/greenplum/pxf/automation/components/hdfs/Hdfs.java
```diff
@@ -365,6 +366,17 @@ public void writeJsonFileFromAvro(String pathToFile, String pathToJson)
         }
     }
 
+    public void writeAvroMetadata(String pathToFile, String pathToMetadata)
+            throws Exception {
+        Tool tool = new DataFileGetMetaTool();
```
denalex, line 371: is Tool thread safe and stateless ? If yes, we should be using a single (static) instance of it

automation/src/main/java/org/greenplum/pxf/automation/components/hdfs/Hdfs.java
```diff
@@ -365,6 +366,17 @@ public void writeJsonFileFromAvro(String pathToFile, String pathToJson)
...
+        List<String> args = new ArrayList<>();
+        args.add(pathToFile);
```
denalex, line 373: `List<String> args = Arrays.asList(pathToFile)`

automation/src/test/java/org/greenplum/pxf/automation/features/avro/HdfsWritableAvroTest.java
```diff
@@ -60,7 +61,7 @@
     public void beforeClass() throws Exception {
         // path for storing data on HDFS (for processing by PXF)
         hdfsPath = hdfs.getWorkingDirectory() + "/writableAvro/";
-        String absolutePath = getClass().getClassLoader().getResource("data").getPath();
+        String absolutePath = Objects.requireNonNull(getClass().getClassLoader().getResource("data")).getPath();
```
denalex, line 64: what's the point, it will throw NPE in either case (before and after the change) ? This is mostly used for checking input params to the functions are not nulls.

automation/src/test/java/org/greenplum/pxf/automation/features/avro/HdfsWritableAvroTest.java
```diff
@@ -96,12 +97,43 @@ public void generateSchemaPrimitive() throws Exception {
...
+    @Test(groups = {"features", "gpdb", "hcfs", "security"})
+    public void generateSchemaPrimitive_withNoCompression() throws Exception {
+        gpdbTable = "writable_avro_primitive_no_compression";
+        fullTestPath = hdfsPath + "generate_schema_primitive_types_with_no_compression";
+        exTable = new WritableExternalTable(gpdbTable + "_writable", avroPrimitiveTableCols, fullTestPath, "custom");
+        exTable.setHost(pxfHost);
+        exTable.setPort(pxfPort);
+        exTable.setFormatter("pxfwritable_export");
+        exTable.setProfile(protocol.value() + ":avro");
+        exTable.setUserParameters(new String[]{"COMPRESSION_CODEC=null"});
```
denalex, line 115: should we use the value `uncompressed` (like in other profiles, I believe) instead of `null` ?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/AvroFileAccessor.java
```diff
@@ -138,6 +142,28 @@ public OneRow readNextObject() throws IOException {
     public boolean openForWrite() throws Exception {
         // make writer
         writer = new DataFileWriter<>(new GenericDatumWriter<>(schema));
+        String codec = context.getOption(COMPRESSION_CODEC_OPTION, "deflate");
+        int codecCompressionLevel = context.getOption(CODEC_COMPRESSION_LEVEL_OPTION, DEFAULT_CODEC_COMPRESSION_LEVEL);
+        switch (codec) {
+            case "deflate":
```
denalex, line 148: I suggest to make all these as constants

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/AvroFileAccessor.java
```diff
@@ -138,6 +142,28 @@ public OneRow readNextObject() throws IOException {
...
+                // writer.setCodec(CodecFactory.xzCodec(codecCompressionLevel));
+                // break;
+            case "bzip2":
+                throw new RuntimeException("Avro Compression codec bzip2 currently not supported");
+                // writer.setCodec(CodecFactory.bzip2Codec());
+                // break;
+            case "snappy":
+                writer.setCodec(CodecFactory.snappyCodec());
+                break;
+            default:
```
denalex, line 162: the default case should handle all unsupported ones and throw exception, while no-compression case should be one of the explicit choices above

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HcfsBaseAccessor.java
```diff
@@ -0,0 +1,53 @@
...
+     *
+     * @return if the request can be run in multi-threaded mode.
+     */
+    @Override
+    public boolean isThreadSafe() {
+        if (context.getProfile() == null) {
+            return true;
+        }
+        String compCodec = context.getOption("COMPRESSION_CODEC", null);
+        if (context.getProfile().matches("^.*:parquet$")) {
```
denalex, line 24: I don't like going back to profile and relying on its name, let's think if this can be done in a better way

frankgh:
@oliverralbertini also `COMPRESSION_CODEC=null` doesn't make sense. For parquet we specify `uncompressed`, I think we should stick to `uncompressed` to make consistent across profiles

automation/src/main/java/org/greenplum/pxf/automation/components/hdfs/Hdfs.java
oliverralbertini: Thanks for the review. This is used in automation test for Writable Avro tables, in which we write the contents of PXF-generated Avro data/metadata into files to make sure that what ends up in external is what we think (avoids potential garbage in/garbage out problem when testing with writable/readable tables).

The API for these Avro tools takes a stream for stdout and stderr, but mostly we care about stdout in this case which will go to a file.

automation/src/main/java/org/greenplum/pxf/automation/components/hdfs/Hdfs.java
oliverralbertini: Since it's not being used in the app, we can ignore that.

automation/src/test/java/org/greenplum/pxf/automation/features/avro/HdfsWritableAvroTest.java
oliverralbertini: yeah, this was suggested by the editor for some reason.

automation/src/main/java/org/greenplum/pxf/automation/components/hdfs/Hdfs.java
oliverralbertini: with editor help, ended up with
```java
List<String> args = Collections.singletonList(pathToFile);
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HcfsBaseAccessor.java
oliverralbertini: Simplest approach is just to make sure that `COMPRESSION_CODEC != "bzip2"`, this works regardless of the type of data (Avro, Parquet, or TEXT).

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HcfsBaseAccessor.java
```diff
@@ -0,0 +1,27 @@
...
+    @Override
+    public boolean isThreadSafe() {
+        String compCodec = context.getOption("COMPRESSION_CODEC");
+        if (compCodec == null) {
+            // check for file extensions indicating bzip2 (Parquet only)
+            Class<? extends CompressionCodec> codecClass = codecFactory.getCodecClassByPath(configuration, context.getDataSource());
+            return (codecClass == null || !BZip2Codec.class.isAssignableFrom(codecClass));
+        }
+        // bzip2 for Avro, and BZip2Codec for Parquet
+        return !compCodec.equalsIgnoreCase("bzip2") && !compCodec.equalsIgnoreCase("org.apache.hadoop.io.compress.BZip2Codec");
```
denalex, line 25: I am not sure this logic is equivalent (or a superset) or what was removed, but it might as well be. Actually, I prefer the way comparison was done before: `BZip2Codec.class.isAssignableFrom(..)` instead of comparing class names. There was also getting Hadoop mapping `compressionCodecName.getHadoopCompressionCodecClass()` which is no longer done here.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HcfsBaseAccessor.java
```diff
@@ -0,0 +1,27 @@
...
+     * Checks if requests should be handled in a single thread or not.
+     *
+     * @return if the request can be run in multi-threaded mode.
+     */
+    @Override
+    public boolean isThreadSafe() {
+        String compCodec = context.getOption("COMPRESSION_CODEC");
+        if (compCodec == null) {
+            // check for file extensions indicating bzip2 (Parquet only)
+            Class<? extends CompressionCodec> codecClass = codecFactory.getCodecClassByPath(configuration, context.getDataSource());
```
denalex, line 21: what if datasource points to directory full of bz2 files (and not an individual file), will this detect it ?

oliverralbertini: Good catch, no it will not. This isn't an issue with my change, however, this was always the case. On the other hand, I think that the Accessor stage in PXF should only see files, not directories. If this were happening during fragmentation of the external resource, it would be more concerning.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HcfsBaseAccessor.java
oliverralbertini: That brings us right back around to the Parquet vs Avro issue. How do we know which type of compression we are dealing with?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/CodecFactory.java
```diff
@@ -80,6 +81,24 @@ public CompressionCodec getCodec(String name, Configuration conf) {
...
+     * @return true only if it's thread safe
+     */
+    public boolean isCodecThreadSafe(String compCodec, String dataSource, Configuration configuration) {
+        if (compCodec == null) {
+            // check for file extensions indicating bzip2 (Text only)
+            // currently doesn't check for bzip2 in .avro files
+            Class<? extends CompressionCodec> codecClass = getCodecClassByPath(configuration, dataSource);
+            return (codecClass == null || !BZip2Codec.class.isAssignableFrom(codecClass));
+        }
+        return !compCodec.equalsIgnoreCase("bzip2") && !compCodec.equalsIgnoreCase(BZip2Codec.class.getName());
```
denalex, line 99: I was suggesting to collapse into 1 return statement, like this:

```suggestion
        Class<? extends CompressionCodec> codecClass = null;
        if (compCodec == null) {
            // check for file extensions indicating bzip2 (Text only)
            // currently doesn't check for bzip2 in .avro files
            codecClass = getCodecClassByPath(configuration, dataSource);
        }
        return !( "bzip2".equalsIgnoreCase(compCodec) ||
                       BZip2Codec.class.getName().equalsIgnoreCase(compCodec) ||
                       (codecClass != null && BZip2Codec.class.isAssignableFrom(codecClass))
                      );
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HcfsBaseAccessor.java
```diff
@@ -0,0 +1,20 @@
...
+    private static final CodecFactory codecFactory = CodecFactory.getInstance();
+
+    /**
+     * Checks if requests should be handled in a single thread or not.
+     *
+     * @return if the request can be run in multi-threaded mode.
+     */
+    @Override
+    public boolean isThreadSafe() {
+        return codecFactory.isCodecThreadSafe(context.getOption("COMPRESSION_CODEC"), context.getDataSource(), configuration);
```
denalex, line 18: should we have `context.getCompressionCodec()` method not to deal with strings here ? That method can also deal with short-name vs. fully qualified class (if needed).

oliverralbertini: This would be a great thing to tackle in #393 . That's where we start to allow short names in the text profiles.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/CodecFactory.java
oliverralbertini: This looks equivalent, I will give it a try.

PR #393, opened 2020-06-23
Allow short names for text compression codecs

oliverralbertini:
In Parquet and Avro we allow the user to specify COMPRESSION_CODEC by
short names like snappy, gzip, bzip2, deflate, etc. With this change, we
can allow the user to specify one of the three supported codecs for text
data using shortcuts: bzip2, default, gzip.

### DIFF
--- a/server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/CodecFactory.java
+++ b/server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/CodecFactory.java
@@ -10,10 +10,18 @@ import org.apache.parquet.hadoop.metadata.CompressionCodecName;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.HashMap;
+import java.util.Map;
+
 public class CodecFactory {
 
-    private static Logger LOG = LoggerFactory.getLogger(CodecFactory.class);
+    private static final Logger LOG = LoggerFactory.getLogger(CodecFactory.class);
     private static final CodecFactory codecFactoryInstance = new CodecFactory();
+    private static final Map<String, String> codecMap = new HashMap<String, String>() {{
+        put("bzip2", "org.apache.hadoop.io.compress.BZip2Codec");
+        put("default", "org.apache.hadoop.io.compress.DefaultCodec");
+        put("gzip", "org.apache.hadoop.io.compress.GzipCodec");
+    }};
 
     /**
      * Returns the {@link CompressionCodecName} for the given name, or default if name is null
@@ -44,6 +52,9 @@ public class CodecFactory {
      * @return generated CompressionCodec
      */
     public CompressionCodec getCodec(String name, Configuration conf) {
+        if (codecMap.containsKey(name)) {
+            name = codecMap.get(name);
+        }
         return ReflectionUtils.newInstance(getCodecClass(name, conf), conf);
     }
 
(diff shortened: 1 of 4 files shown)


oliverralbertini:
Please ignore all but the last commit, after merging #389 will clean this branch up.

frankgh:
can we still using the full classname ? is this a backwards compatible change?

oliverralbertini:
> can we still using the full classname ? is this a backwards compatible change?

Yes, it's backwards compatible.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/CodecFactory.java
```diff
@@ -10,10 +10,18 @@
...
+
 public class CodecFactory {
 
-    private static Logger LOG = LoggerFactory.getLogger(CodecFactory.class);
+    private static final Logger LOG = LoggerFactory.getLogger(CodecFactory.class);
     private static final CodecFactory codecFactoryInstance = new CodecFactory();
+    private static final Map<String, String> codecMap = new HashMap<String, String>() {{
+        put("bzip2", "org.apache.hadoop.io.compress.BZip2Codec");
+        put("default", "org.apache.hadoop.io.compress.DefaultCodec");
+        put("gzip", "org.apache.hadoop.io.compress.GzipCodec");
```
denalex, line 23: would we allow to support LZO (com.hadoop.compression.lzo.LzopCodec) in case customers are willing to download the JAR themselves (it is GPL licensed):  https://www.oreilly.com/library/view/hadoop-the-definitive/9780596521974/ch04.html

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/CodecFactory.java
```diff
@@ -10,10 +10,18 @@
...
 
+import java.util.HashMap;
+import java.util.Map;
+
 public class CodecFactory {
 
-    private static Logger LOG = LoggerFactory.getLogger(CodecFactory.class);
+    private static final Logger LOG = LoggerFactory.getLogger(CodecFactory.class);
     private static final CodecFactory codecFactoryInstance = new CodecFactory();
+    private static final Map<String, String> codecMap = new HashMap<String, String>() {{
```
denalex, line 20: should we use standard lookup instead of reinventing it: https://hadoop.apache.org/docs/r2.6.1/api/org/apache/hadoop/io/compress/CompressionCodecFactory.html#getCodecByName(java.lang.String)

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/CodecFactory.java
denalex: also `snappy` ? I think if we use the Hadoop library (see comment for line 20), it will auto discover and register codecs available on the classpath

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/CodecFactory.java
```diff
@@ -44,6 +52,9 @@ public CompressionCodecName getCodec(String name, CompressionCodecName defaultCo
      * @return generated CompressionCodec
      */
     public CompressionCodec getCodec(String name, Configuration conf) {
+        if (codecMap.containsKey(name)) {
+            name = codecMap.get(name);
+        }
```
denalex, line 57: this whole method will not be needed if using https://hadoop.apache.org/docs/r2.6.1/api/org/apache/hadoop/io/compress/CompressionCodecFactory.html#getCodecByName(java.lang.String)

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/CodecFactory.java
denalex: in fact, I struggle to see what this factory adds on top of existing Hadoop one ? Can we just use `CompressionCodecFactory` from Hadoop to get our codec by either short name or class name ?

ashuka24:
Closed in favor of https://github.com/greenplum-db/pxf/pull/535

PR #394, opened 2020-06-23
Fix queries with limit

frankgh:
There is an issue with async threads, where we are swallowing the
ClientAbortException when Greenplum closes the connection. This lack of
communication that an IOException has occurred is causing problems with
subsequent requests. Re-throwing the exception allows Spring MVC to
catch it and mark the connection as broken so it stops writing.

The short version of what happens is:
- client drops connection async thread is writing to
- Tomcat cleans this up
- Request/Response/Processor gets put back in pool
- New connections come in
- Processor above gets allocated to one of them
- Original async thread tries writing again, fails and puts the processor into an error state
- Tomcat tries to process the new connection
- Processor is in an error state so the connection is closed before the request is even read

Co-authored-by: Mark Thomas <thomasma@vmware.com>

### DIFF
--- a/server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/BridgeResponse.java
+++ b/server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/BridgeResponse.java
@@ -73,17 +73,24 @@ public class BridgeResponse implements StreamingResponseBody {
                 ++recordCount;
             }
             LOG.debug("Finished streaming fragment {} of resource {}, {} records.", fragment, dataDir, recordCount);
-        } catch (ClientAbortException e) {
-            // Occurs whenever client (GPDB) decides to end the connection
-            if (LOG.isDebugEnabled()) {
-                // Stacktrace in debug
-                LOG.debug("Remote connection closed by GPDB", e);
-            } else {
-                LOG.error("Remote connection closed by GPDB (Enable debug for stacktrace)");
-            }
-            throw e;
         } catch (Exception e) {
-            throw e instanceof IOException ? (IOException) e : new IOException(e.getMessage(), e);
+            IOException ioe;
+            if (e instanceof ClientAbortException) {
+                // Occurs whenever client (GPDB) decides to end the connection
+                if (LOG.isDebugEnabled()) {
+                    // Stacktrace in debug
+                    LOG.debug("Remote connection closed by GPDB", e);
+                } else {
+                    LOG.error("Remote connection closed by GPDB (Enable debug for stacktrace)");
+                }
+                ioe = (ClientAbortException) e;
+            } else if (e instanceof IOException) {
+                ioe = (IOException) e;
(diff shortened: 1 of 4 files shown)


server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/BridgeResponse.java
```diff
@@ -73,17 +73,24 @@ private Void writeToInternal(OutputStream out) throws IOException {
...
-                LOG.debug("Remote connection closed by GPDB", e);
+        } catch (Exception e) {
+            IOException ioe;
+            if (e instanceof ClientAbortException) {
+                // Occurs whenever client (GPDB) decides to end the connection
+                if (LOG.isDebugEnabled()) {
+                    // Stacktrace in debug
+                    LOG.debug("Remote connection closed by GPDB", e);
+                } else {
+                    LOG.error("Remote connection closed by GPDB (Enable debug for stacktrace)");
```
denalex, line 84: I don't know if we even should log this as error, since this is not really an error condition when GPDB routinely does this. maybe use WARN ?

server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/BridgeResponse.java
```diff
@@ -73,17 +73,24 @@ private Void writeToInternal(OutputStream out) throws IOException {
...
+                ioe = (IOException) e;
             } else {
-                LOG.error("Remote connection closed by GPDB (Enable debug for stacktrace)");
+                ioe = new IOException(e.getMessage(), e);
             }
-            throw e;
-        } catch (Exception e) {
-            throw e instanceof IOException ? (IOException) e : new IOException(e.getMessage(), e);
+            // Re-throw the exception so Spring MVC is aware that an IO error has occurred
+            throw ioe;
```
denalex, line 93: I kinda like the previous logic better, where we handle ClientAbortException only and the common block of wrapping any exception as IOException, unless it is already an IOException.

server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/WritableResource.java
```diff
@@ -165,6 +165,8 @@ public WritableResource(BridgeFactory bridgeFactory, SecurityService securitySer
                 } else {
                     LOG.error("Remote connection closed by GPDB (Enable debug for stacktrace)");
                 }
+                ex = cae;
+                throw ex;
```
denalex, line 169: why just not `throw cae` ?

server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/BridgeResponse.java
frankgh: true, it should be a warn

PR #396, opened 2020-06-24
Address Feedback fix spring boot TODOs

frankgh:
- install-dev -> install-server
- fix expected output on some files
- refactor trace/hint information on the c-side
- remove path from Writable/stream endpoint as it is no longer used
- refactor PxfErrorAttributes
- address spring_boot_todo
- expose only a few endpoints (health, info, shutdown)


external-table/src/libchurl.c
```diff
@@ -640,12 +640,12 @@ flush_internal_buffer(churl_context *context)
 		multi_perform(context);
 	}
 
+	check_response(context);
+
```
denalex, line 644: wonder why this was needed

server/pxf-service/src/main/java/org/greenplum/pxf/service/PxfConfiguration.java
```diff
@@ -9,30 +10,64 @@
...
+import org.springframework.web.servlet.config.annotation.WebMvcConfigurer;
 
 /**
  * Declares the registerSecurityServletFilter bean method to be processed by
  * the Spring container
  */
 @Configuration
 @EnableConfigurationProperties(PxfServerProperties.class)
-public class PxfConfiguration {
+public class PxfConfiguration implements WebMvcConfigurer {
```
denalex, line 27: the comment above lines 21-24 is no longer relevant (registerSecurityServletFilter)

server/pxf-service/src/main/java/org/greenplum/pxf/service/PxfConfiguration.java
```diff
@@ -9,30 +10,64 @@
...
+     * Configures and builds the {@link ThreadPoolTaskExecutor}
      *
-     * @return the {@link TaskExecutorBuilder} object
+     * @return the {@link ThreadPoolTaskExecutor}
      */
-    @Bean
-    public TaskExecutorBuilder taskExecutorBuilder(PxfServerProperties pxfServerProperties,
-                                                   ObjectProvider<TaskExecutorCustomizer> taskExecutorCustomizers,
-                                                   ObjectProvider<TaskDecorator> taskDecorator) {
+    @Lazy
```
denalex, line 64: I thought we did not want lazy init unless absolutely required

server/pxf-service/src/main/java/org/greenplum/pxf/service/PxfConfiguration.java
```diff
@@ -9,30 +10,64 @@
...
  */
 @Configuration
 @EnableConfigurationProperties(PxfServerProperties.class)
-public class PxfConfiguration {
+public class PxfConfiguration implements WebMvcConfigurer {
+
+    /**
+     * Bean name of PXF's {@link TaskExecutor}.
+     */
+    public static final String PXF_APPLICATION_TASK_EXECUTOR_BEAN_NAME = "pxfApplicationTaskExecutor";
```
denalex, line 32: can we fine-tune a name -- it is not executing any arbitrary pxf tasks, this is for async response streaming only, in other words we should not be submitting any arbitrary async work here, right ? Maybe `pxfResponseStreamTaskExecutor` or smith even simpler

server/pxf-service/src/main/java/org/greenplum/pxf/service/PxfConfiguration.java
```diff
@@ -9,30 +10,64 @@
...
     /**
-     * Configures the TaskExecutorBuilder to be used for async requests (i.e. Bridge
+     * Configures the TaskExecutor to be used for async requests (i.e. Bridge
      * Read).
+     */
+    @Override
+    public void configureAsyncSupport(AsyncSupportConfigurer configurer) {
+        Object taskExecutor = this.beanFactory
+                .getBean(PXF_APPLICATION_TASK_EXECUTOR_BEAN_NAME);
+        if (taskExecutor instanceof AsyncTaskExecutor) {
```
denalex, line 54: why `if` ? We expect our own bean here, so cast or let it fail the startup if somehow misconfigured

external-table/src/libchurl.c
frankgh: in one case the response was closed by the server, but we already had a response in the buffer, so we could actually take some action in that case. For example, check the response code

server/pxf-service/src/main/java/org/greenplum/pxf/service/PxfConfiguration.java
frankgh: Yeah, I'll address it

frankgh:
@denalex thanks for the review. I have addressed your concerns

PR #397, opened 2020-06-25
Fix the MapR job

frankgh:
When doing classloading, the LOADER_PATH is ending up with something
like this:

export LOADER_PATH=,file:/home/gpadmin/pxf/conf,....

Apparently, PropertiesLauncher is parsing the string above and doing a
split, and for them "" (the first element in the array as a result of a split)
means the user wants root of archive but not
current directory. This is making the root archive to be on the
classpath first, then the PXF specific classpath, and then again the
root archive for some reason. This is preventing us to classload the PXF
libraries before the application libraries. This commit checks whether
the PXF_LOADER_PATH environment variable is set, and if it's not, we
don't add the comma at the beginning.

- Fix a test for CloudAccess
- warn instead of error ignored exception

### DIFF
--- a/server/pxf-service/src/scripts/pxf-env-default.sh
+++ b/server/pxf-service/src/scripts/pxf-env-default.sh
@@ -63,8 +63,8 @@ export PXF_OOM_KILL=${PXF_OOM_KILL:-true}
 # export PXF_OOM_DUMP_PATH=
 
 # Additional locations to be class-loaded by the application
-if [ -n "$PXF_LOADER_PATH" ]; then
-  export LOADER_PATH="$PXF_LOADER_PATH,file:${PXF_CONF}/conf,file:${PXF_HOME}/conf,file:${PXF_CONF}/lib"
+if [[ -n "${PXF_LOADER_PATH}" ]]; then
+  export LOADER_PATH="${PXF_LOADER_PATH},file:${PXF_CONF}/conf,file:${PXF_HOME}/conf,file:${PXF_CONF}/lib"
 else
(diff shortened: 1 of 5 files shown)


server/pxf-service/src/scripts/pxf-env-default.sh
```diff
@@ -63,4 +63,8 @@ export PXF_OOM_KILL=${PXF_OOM_KILL:-true}
 # export PXF_OOM_DUMP_PATH=
 
 # Additional locations to be class-loaded by the application
-export LOADER_PATH="${PXF_LOADER_PATH},file:${PXF_CONF}/conf,file:${PXF_HOME}/conf,file:${PXF_CONF}/lib"
+if [ -n "$PXF_LOADER_PATH" ]; then
```
denalex, line 66:
```suggestion
if [[ -n "${PXF_LOADER_PATH}" ]]; then
```

server/pxf-service/src/scripts/pxf-env-default.sh
```diff
@@ -63,4 +63,8 @@ export PXF_OOM_KILL=${PXF_OOM_KILL:-true}
...
+  export LOADER_PATH="$PXF_LOADER_PATH,file:${PXF_CONF}/conf,file:${PXF_HOME}/conf,file:${PXF_CONF}/lib"
```
denalex, line 67:
```suggestion
  export LOADER_PATH="${PXF_LOADER_PATH},file:${PXF_CONF}/conf,file:${PXF_HOME}/conf,file:${PXF_CONF}/lib"
```

PR #398, opened 2020-06-30
Spring boot Performance

frankgh:
Settings used to make the performance pipeline pass

### DIFF
--- a/server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/BridgeResponse.java
+++ b/server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/BridgeResponse.java
@@ -77,9 +77,9 @@ public class BridgeResponse implements StreamingResponseBody {
             // Occurs whenever client (GPDB) decides to end the connection
             if (LOG.isDebugEnabled()) {
                 // Stacktrace in debug
-                LOG.debug("Remote connection closed by GPDB", e);
+                LOG.debug(String.format("Remote connection closed by GPDB (segment %s)", context.getSegmentId()), e);
             } else {
-                LOG.warn("Remote connection closed by GPDB (Enable debug for stacktrace)");
+                LOG.warn("Remote connection closed by GPDB (segment {}) (Enable debug for stacktrace)", context.getSegmentId());
             }
             // Re-throw the exception so Spring MVC is aware that an IO error has occurred
             throw e;
--- a/server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/BridgeResponse.java
+++ b/server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/BridgeResponse.java
@@ -77,7 +77,7 @@ public class BridgeResponse implements StreamingResponseBody {
             // Occurs whenever client (GPDB) decides to end the connection
             if (LOG.isDebugEnabled()) {
                 // Stacktrace in debug
-                LOG.debug(String.format("Remote connection closed by GPDB (segment %s)", context.getSegmentId()), e);
+                LOG.warn(String.format("Remote connection closed by GPDB (segment %s)", context.getSegmentId()), e);
             } else {
                 LOG.warn("Remote connection closed by GPDB (segment {}) (Enable debug for stacktrace)", context.getSegmentId());
             }
--- a/server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/WritableResource.java
+++ b/server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/WritableResource.java
@@ -160,7 +160,7 @@ public class WritableResource extends BaseResource {
                 // Occurs whenever client (GPDB) decides to end the connection
                 if (LOG.isDebugEnabled()) {
                     // Stacktrace in debug
-                    LOG.debug(String.format("Remote connection closed by GPDB (segment %s)", context.getSegmentId()), cae);
+                    LOG.warn(String.format("Remote connection closed by GPDB (segment %s)", context.getSegmentId()), cae);
                 } else {
                     LOG.warn("Remote connection closed by GPDB (segment {}) (Enable debug for stacktrace)", context.getSegmentId());
                 }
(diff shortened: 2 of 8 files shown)


server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/BridgeResponse.java
```diff
@@ -77,9 +77,9 @@ private Void writeToInternal(OutputStream out) throws IOException {
             // Occurs whenever client (GPDB) decides to end the connection
             if (LOG.isDebugEnabled()) {
                 // Stacktrace in debug
-                LOG.debug("Remote connection closed by GPDB", e);
+                LOG.debug(String.format("Remote connection closed by GPDB (segment %s)", context.getSegmentId()), e);
```
denalex, line 80: wonder, should we still log at WARN level here to be consistent. Extra context if debug is on is ok.

server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/WritableResource.java
```diff
@@ -160,9 +160,9 @@ public WritableResource(BridgeFactory bridgeFactory, SecurityService securitySer
                 // Occurs whenever client (GPDB) decides to end the connection
                 if (LOG.isDebugEnabled()) {
                     // Stacktrace in debug
-                    LOG.debug("Remote connection closed by GPDB", cae);
+                    LOG.debug(String.format("Remote connection closed by GPDB (segment %s)", context.getSegmentId()), cae);
```
denalex, line 163: same question about WARN level.

server/pxf-service/src/main/resources/application.properties
```diff
@@ -7,7 +7,10 @@ management.endpoint.shutdown.enabled=true
 management.health.probes.enabled=true
 
 server.port=${pxf.port:5888}
+server.shutdown=graceful
+spring.lifecycle.timeout-per-shutdown-phase=1s
```
denalex, line 11: do we only give the server 1s to fish up work and shutdown ? What about our thread pool, will threads have a chance to finish reading /writing ? that might take a while

server/pxf-service/src/scripts/pxf
```diff
@@ -214,6 +214,13 @@ function generateUserConfigs()
     setup_conf_directory "${PXF_CONF}/servers"
     setup_conf_directory "${PXF_CONF}/servers/default"
     setup_conf_directory "${PXF_CONF}/templates" "${PXF_HOME}/templates/user/templates" 'override'
+
+    if [[ ! -f "${PXF_CONF}/conf/pxf-log4j2.xml" ]]; then
+        # copy the new pxf-log4j2.xml if it doesn't exist. This allows us to
+        # initialize PXF 6.x on a PXF 5.x installation that had already been
```
denalex, line 220:
```suggestion
        # initialize PXF 6.x using a pre-existing PXF 5.x configuration
```

server/pxf-service/src/scripts/pxf-env-default.sh
```diff
@@ -47,6 +47,11 @@ export PXF_PORT=${PXF_PORT:=5888}
 # Memory
 export PXF_JVM_OPTS=${PXF_JVM_OPTS:='-Xmx2g -Xms1g'}
 
+# The number of milliseconds the server will wait, after accepting a
```
denalex, line 50:
```suggestion
# The time interval the server will wait, after accepting a
```
message was inconsistent with the value `5m`

denalex: Also, this is too verbose, compared to other properties, maybe just `Server connection timeout`

server/pxf-service/src/templates/user/conf/pxf-env.sh
```diff
@@ -19,6 +19,11 @@ PXF_CONF="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd )"
 # Memory
 # export PXF_JVM_OPTS="-Xmx2g -Xms1g"
 
+# The number of milliseconds the server will wait, after accepting a
+# connection, for the request URI line to be presented. Use a value of -1 to
+# indicate no (i.e. infinite) timeout.
+# export PXF_CONNECTION_TIMEOUT=5m
```
denalex, line 25: same comment

server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/BridgeResponse.java
frankgh: good point

server/pxf-service/src/main/resources/application.properties
frankgh: I wasn't sure about this one, I think what I wanted to achieve is to enable graceful shutdown, and then users can configure the timeout, but I didn't want to set it to a high number. Maybe we can add `# export SPRING_LIFECYCLE_TIMEOUT_PER_SHUTDOWN_PHASE=1s` in our `pxf-env.sh` and `pxf-env-default.sh` templates

server/pxf-service/src/scripts/pxf
```diff
@@ -214,6 +214,13 @@ function generateUserConfigs()
     setup_conf_directory "${PXF_CONF}/servers"
     setup_conf_directory "${PXF_CONF}/servers/default"
     setup_conf_directory "${PXF_CONF}/templates" "${PXF_HOME}/templates/user/templates" 'override'
+
+    if [[ ! -f "${PXF_CONF}/conf/pxf-log4j2.xml" ]]; then
+        # copy the new pxf-log4j2.xml if it doesn't exist. This allows us to
+        # initialize PXF 6.x using a pre-existing PXF 5.x configuration
+        # initialized
```
denalex, line 221: this line is no longer needed

server/pxf-service/src/main/resources/application.properties
denalex: seems like the default is 30 secs: https://docs.spring.io/spring/docs/3.2.4.RELEASE_to_4.0.0.M3/Spring%20Framework%203.2.4.RELEASE/org/springframework/context/support/DefaultLifecycleProcessor.html#setTimeoutPerShutdownPhase(long)

I think we should discuss whether we need a graceful shutdown at all, it is not like we have a user-facing app with sub-second response time and want to preserve their session. If for us graceful shutdown means not to accept any new work but let existing queries finish, this might take minutes if not hours.

PR #399, opened 2020-06-30
Prefix tag names with `release-`

oliverralbertini:
This allows being more selective on which tags to create releases on.
For example tags like docs-v1.2.3 will not trigger a release.

### DIFF
--- a/concourse/scripts/promote_pxf_artifacts.bash
+++ b/concourse/scripts/promote_pxf_artifacts.bash
@@ -77,9 +77,10 @@ git -C pxf_src remote set-url origin "${GIT_REMOTE_URL}"
 # avoid detached HEAD state from Concourse checking out a SHA
 git -C pxf_src checkout "${GIT_BRANCH}"
 
-echo "Creating new tag ${pxf_version}..."
+TAG=release-${pxf_version}
+echo "Creating new tag ${TAG}..."
 # create annotated tag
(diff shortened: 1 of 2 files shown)


oliverralbertini:
Wondering if the version file should contain `v` or not. Not sure what the convention is for that.

concourse/scripts/promote_pxf_artifacts.bash
```diff
@@ -77,9 +77,9 @@ git -C pxf_src remote set-url origin "${GIT_REMOTE_URL}"
 # avoid detached HEAD state from Concourse checking out a SHA
 git -C pxf_src checkout "${GIT_BRANCH}"
 
-echo "Creating new tag ${pxf_version}..."
+echo "Creating new tag v${pxf_version}..."
```
frankgh, line 80: maybe create a PREFIX variable here?

concourse/scripts/promote_pxf_artifacts.bash
```diff
@@ -77,9 +77,9 @@ git -C pxf_src remote set-url origin "${GIT_REMOTE_URL}"
...
 # create annotated tag
-git -C pxf_src tag -am "PXF Version ${pxf_version}" "${pxf_version}"
+git -C pxf_src tag -am "PXF Version v${pxf_version}" "v${pxf_version}"
```
frankgh, line 82: can we use `release-` instead of `v`?

PR #400, opened 2020-06-30
CVE-2020-8184: Bump rack version in book/Gemfile.lock to >= 2.1.4

frankgh:
Addresses CVE-2020-8184

> A reliance on cookies without validation/integrity check security vulnerability exists in rack < 2.2.3, rack < 2.1.4 that makes it is possible for an attacker to forge a secure or host-only cookie prefix.


frankgh:
thanks for the review @lisakowen

PR #401, opened 2020-07-01
Add `date` complex type to HdfsWritableAvroTest

oliverralbertini:
We convert it to a string in Avro, but it's good to verify that it
works.


oliverralbertini:
I will add more types in forthcoming commits. No need to review yet.

automation/src/test/java/org/greenplum/pxf/automation/features/avro/HdfsWritableAvroTest.java
```diff
@@ -53,15 +54,23 @@
             "type_enum_mood mood",
             "type_long_array BIGINT[]",
             "type_numeric_array NUMERIC(8,1)[]",
-            "type_string_array TEXT[]"
+            "type_string_array TEXT[]",
+            "type_date DATE",
+            "type_time TIME",
+            "type_timestamp TIMESTAMP",
+            "type_timestampz TIMESTAMP WITH TIME ZONE"
```
oliverralbertini, line 61: @lisakowen I've added these types: `date`, `time`, `timestamp`, and `timestamp with time zone`. Is there any other Greenplum type you think I should certify?

lisakowen: the only other one that i can think of right now is varchar.  but not sure how often one would use that type when writing to avro?

oliverralbertini: Added testing for char and varchar. (in the tests for primitive types).

We use a writable external table with char and varchar columns that write to Avro string types, and can read them back into a readable external table using the same types.

PR #405, opened 2020-07-02
docs - avro write smallint mapping

lisakowen:
docs for https://github.com/greenplum-db/pxf/pull/383

in this PR:
- edit out some spaces in the existing (now read) data type mapping table
- add write mapping table
- add not about smallint conversion to int when writing avro

### DIFF
--- a/docs/content/hdfs_avro.html.md.erb
+++ b/docs/content/hdfs_avro.html.md.erb
@@ -40,24 +40,49 @@ Avro supports both primitive and complex data types.
 
 To represent Avro primitive data types in Greenplum Database, map data values to Greenplum Database columns of the same type. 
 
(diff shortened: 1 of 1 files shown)


docs/content/hdfs_avro.html.md.erb
```diff
@@ -40,24 +40,49 @@ Avro supports both primitive and complex data types.
...
+PXF uses the following data type mapping when reading Avro data:
+
+| Avro Data Type     | PXF/Greenplum Data Type    |
+|--------------------|----------------------------|
 | boolean | boolean |
 | bytes | bytea |
 | double | double |
 | float | real |
-| int | int or smallint |
+| int | int |
```
dyozie, line 57: Should this be | int | integer | since the write table below uses Integer instead of the "int" alias for Greenplum?

PR #410, opened 2020-07-13
Allow skipping the header for *:text profiles

frankgh:
We introduce a new parameter "SKIP_HEADER_COUNT", which when provided
for *:text profiles, it will skip the N first lines of the first split
of each file. For example, SKIP_HEADER_COUNT=2 will skip the first 2
lines of each remote file and will only stream back starting from the
3rd line. If SKIP_HEADER_COUNT is greater than the number of lines of a
file, no data will be streamed back to the client.

### DIFF
--- a/server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/LineBreakAccessor.java
+++ b/server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/LineBreakAccessor.java
@@ -41,11 +41,12 @@ import java.net.URI;
  * A PXF Accessor for reading delimited plain text records.
  */
 public class LineBreakAccessor extends HdfsSplittableDataAccessor {
+    private int skipHeaderCount;
     private DataOutputStream dos;
     private FSDataOutputStream fsdos;
     private FileSystem fs;
     private Path file;
-    private CodecFactory codecFactory;
+    private final CodecFactory codecFactory;
 
     /**
      * Constructs a LineBreakAccessor.
@@ -56,17 +57,20 @@ public class LineBreakAccessor extends HdfsSplittableDataAccessor {
     }
 
     @Override
-    public void initialize(RequestContext requestContext) {
-        super.initialize(requestContext);
+    public void initialize(RequestContext context) {
+        super.initialize(context);
         ((TextInputFormat) inputFormat).configure(jobConf);
+        skipHeaderCount = context.getFragmentIndex() != 0 ? 0 :
+                context.getOption("SKIP_HEADER_COUNT", 0, true);
     }
 
     @Override
     protected Object getReader(JobConf jobConf, InputSplit split)
             throws IOException {
 
+        // when the file has header use LineReaderReader to read one line at at time
         // for HDFS, try to use ChunkRecordReader, if possible (not reading from encrypted zone)
-        if (hcfsType == HcfsType.HDFS) {
+        if (skipHeaderCount == 0 && hcfsType == HcfsType.HDFS) {
             try {
                 return new ChunkRecordReader(jobConf, (FileSplit) split);
             } catch (IncompatibleInputStreamException e) {
@@ -75,6 +79,16 @@ public class LineBreakAccessor extends HdfsSplittableDataAccessor {
             }
         }
         return new LineRecordReader(jobConf, (FileSplit) split);
+    }
+
+    @Override
+    public OneRow readNextObject() throws IOException {
+        while (skipHeaderCount > 0) {
+            if (super.readNextObject() == null)
+                return null;
+            skipHeaderCount--;
+        }
+        return super.readNextObject();
     }
 
     /**
@@ -95,21 +109,6 @@ public class LineBreakAccessor extends HdfsSplittableDataAccessor {
         // create output stream - do not allow overwriting existing file
         createOutputStream(file, codec);
         return true;
-    }
-
-    /*
-     * Creates output stream from given file. If compression codec is provided,
-     * wrap it around stream.
-     */
-    private void createOutputStream(Path file, CompressionCodec codec)
-            throws IOException {
-        fsdos = fs.create(file, false);
-        if (codec != null) {
-            dos = new DataOutputStream(codec.createOutputStream(fsdos));
-        } else {
-            dos = fsdos;
-        }
-
     }
 
     /**
@@ -139,5 +138,20 @@ public class LineBreakAccessor extends HdfsSplittableDataAccessor {
             fsdos.hsync();
             dos.close();
         }
+    }
+
+    /*
+     * Creates output stream from given file. If compression codec is provided,
+     * wrap it around stream.
+     */
+    private void createOutputStream(Path file, CompressionCodec codec)
+            throws IOException {
+        fsdos = fs.create(file, false);
+        if (codec != null) {
+            dos = new DataOutputStream(codec.createOutputStream(fsdos));
+        } else {
+            dos = fsdos;
+        }
+
     }
 }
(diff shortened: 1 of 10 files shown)


server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/LineBreakAccessor.java
```diff
@@ -56,17 +57,20 @@ public LineBreakAccessor() {
     }
 
     @Override
-    public void initialize(RequestContext requestContext) {
-        super.initialize(requestContext);
+    public void initialize(RequestContext context) {
+        super.initialize(context);
         ((TextInputFormat) inputFormat).configure(jobConf);
+        skipHeaderCount = context.getFragmentIndex() != 0 ? 0 :
```
denalex, line 63: I think we can not rely on fragment index because:
- not sure if fragment list coming in JSON guarantees order of splits
- isn't index maintained from 0 for each segment ?
- when reading multiple files, some beginning splits will not be the first fragments as other files would be read before them

I think instead we should be looking at the offset in the fragment metadata, if the offset is 0, then we're reading the beginning of the file and can apply the header logic.

frankgh: agreed, pushing that change now.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HdfsSplittableDataAccessor.java
```diff
@@ -80,7 +84,6 @@ public void initialize(RequestContext requestContext) {
     @Override
     public boolean openForRead() throws Exception {
         LinkedList<InputSplit> requestSplits = new LinkedList<>();
-        FileSplit fileSplit = HdfsUtilities.parseFileSplit(context);
```
denalex, line 83: this is confusing, if `openForRead` is used only for 1 Fragment and each Fragment is represented by a FileSplit, why are there iterators in this old logic ? Asking because if we can process multiple splits as the current logic implies, then we have to implement skipping lines inside `readNextObject` method.

frankgh: Yes, this is confusing. I have seen this code before when we did the refactor, and I wasn't sure why it was there. I took at stab at removing this code in a subsequent commit and it looks like it was only being used by Avro, but there wasn't anything in avro that would add splits to the iterator. So it looks safe to remove. You can take a look at this commit for the cleanup: https://github.com/greenplum-db/pxf/pull/410/commits/24003f9505aa679d861fc75277e26843f4705d17

frankgh: now waiting for pipelines to go green: https://ud.ci.gpdb.pivotal.io/teams/main/pipelines/dev:fguerrero-support-header-text

frankgh:
For this PR, I also looked at validation hooks in the external table framework, but unfortunately we only get access to URL validation. We could validate during query execution, but it might be too late, and might as well do validation at the server side if the `HEADER` parameter was provided during table creation.

```c
/*
 * ExtProtocolValidatorData is the node type that is passed as fmgr "context" info
 * when a function is called by the External Table protocol manager.
 */
typedef struct ExtProtocolValidatorData
{
	NodeTag				 type;			  /* see T_ExtProtocolValidatorData */
	List				*url_list;
	ValidatorDirection	 direction;  /* validating read or write? */
	char				*errmsg;		  /* the validation error upon return, if any */

} ExtProtocolValidatorData;
```

frankgh:
> Looks good overall. A bit concerned about pulling out the support for multiple splits and not understanding why it was there in the first place. Non-splittable files due to compression codecs ? Perhaps review the history of the file, maybe there will be some hints...

will do

> Also, would be nice to have an integration test with multiple files in a directory to make sure all of them got header lines stripped out 

yes, I'm adding it now

> (a test that would've probably failed with the original implementation in this PR).

I believe the fragmenter guarantees that the splits are in order. I looked at the code and it looks like they give you some guarantees, but I agree that this is not ideal. Here's the code: https://github.com/greenplum-db/pxf/blob/3f2e0dcb5867c9dabec1ea862f5e6929fbdcef44/server/pxf-api/src/main/java/org/greenplum/pxf/api/utilities/FragmentsResponseFormatter.java#L70

> Checked the history, there are no hints. But also noticed HiveAccessor uses fragmentIndex to determine when to skip the lines, wonder if that is a problem ?

we can update it to use the split info, which is available to us. Should we change it?

denalex:
yes, was digging through the same code now, made sure C side used `fragment->index` that came from the Fragmenter and found the same code in `FragmentsResponseFormatter.java` I guess if this is a requirement for the `Fragmenter` to properly mark the index, we can rely on it. The problem with unordered datasets (like JDBC), there are no indices for the fragments, so this can't be a universal requirement. The we are back to the scope of the same profile -- if "your" Fragmenter guarantees it, then you can use it. The problem is that knowledge is not auto-discoverable.

So I am a bit torn here, either way is actually fine. Please feel free to rollback to fragmentIndex here or update HiveAccessor to use offset. FragmentIndex is a bit clearer, actually, once we established the guarantee. Ideally these 2 accessors should have similar logic.

frankgh:
> yes, was digging through the same code now, made sure C side used `fragment->index` that came from the Fragmenter and found the same code in `FragmentsResponseFormatter.java` I guess if this is a requirement for the `Fragmenter` to properly mark the index, we can rely on it. The problem with unordered datasets (like JDBC), there are no indices for the fragments, so this can't be a universal requirement. The we are back to the scope of the same profile -- if "your" Fragmenter guarantees it, then you can use it. The problem is that knowledge is not auto-discoverable.

I agree with that the knowledge is not auto-discoverable and this might impact the way we start doing things in the (near) future with the refactor.

> So I am a bit torn here, either way is actually fine. Please feel free to rollback to fragmentIndex here or update HiveAccessor to use offset. FragmentIndex is a bit clearer, actually, once we established the guarantee. Ideally these 2 accessors should have similar logic.

I will go ahead and update `HiveAccessor` to rely on `FileSplit` information and remove the dependency to fragmentIndex

frankgh:
> > Looks good overall. A bit concerned about pulling out the support for multiple splits and not understanding why it was there in the first place. Non-splittable files due to compression codecs ? Perhaps review the history of the file, maybe there will be some hints...
> 
> will do

Looking at the history for this I can only see `SGA import` and `HAWQ-28. JavaDoc fixes for PXF`. Very old commits from 2015.  So no hints why this is the case. But I also encountered the same during the refactor work that we did and I couldn't figure out why it was done this way. The iterator is private, so no extending classes can modify it, so I'm not too concerned about this cleanup.

frankgh:
> > yes, was digging through the same code now, made sure C side used `fragment->index` that came from the Fragmenter and found the same code in `FragmentsResponseFormatter.java` I guess if this is a requirement for the `Fragmenter` to properly mark the index, we can rely on it. The problem with unordered datasets (like JDBC), there are no indices for the fragments, so this can't be a universal requirement. The we are back to the scope of the same profile -- if "your" Fragmenter guarantees it, then you can use it. The problem is that knowledge is not auto-discoverable.
> 
> I agree with that the knowledge is not auto-discoverable and this might impact the way we start doing things in the (near) future with the refactor.

It turns out that `fileSplit.getStart()` is not 0 for ORC files (it was 3 in one of the integration tests that we have). So I will go ahead and use the `fragmentIndex` field for both HiveAccessor + LineBreakAccessor

PR #415, opened 2020-07-30
Hive: Report the correct error message from HiveMetaStoreClientCompat…

frankgh:
…ibility1xx

PXF is not reporting the causing error message when accessing Hive
version 1.x. PXF is also not reporting the causing error message during
initialization of HiveClient. For example, when a RuntimeException is
thrown when the HiveMetaStoreClientCompatibility1xx failed to
instantiate with a MetaException cause. The reported error is
java.lang.RuntimeException: Unable to instantiate
org.apache.hadoop.hive.metastore.HiveMetaStoreClientCompatibility1xx,
however the error above does not report the original cause of the
exception, making debugging harder for users/administrators.

### DIFF
--- a/server/pxf-hive/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClientCompatibility1xx.java
+++ b/server/pxf-hive/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClientCompatibility1xx.java
@@ -58,7 +58,7 @@ public class HiveMetaStoreClientCompatibility1xx extends HiveMetaStoreClient imp
                     return new GetTableResult(table).getTable();
                 }
             } catch (MetaException | NoSuchObjectException ex) {
-                LOG.warn("Original exception not re-thrown", e);
+                LOG.debug("Original exception not re-thrown", e);
                 throw ex;
             } catch (Throwable t) {
                 LOG.warn("Unable to run compatibility for metastore client method get_table_req. Will rethrow original exception: ", t);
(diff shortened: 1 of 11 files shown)


server/pxf-hive/src/main/java/org/greenplum/pxf/plugins/hive/HiveClientWrapper.java
```diff
@@ -420,9 +420,28 @@ static HiveClientFactory getInstance() {
...
         IMetaStoreClient initHiveClient(HiveConf hiveConf) throws MetaException {
-            return RetryingMetaStoreClient.getProxy(hiveConf, new Class[]{HiveConf.class, HiveMetaHookLoader.class, Boolean.class},
-                    new Object[]{hiveConf, null, true}, null, HiveMetaStoreClientCompatibility1xx.class.getName()
-            );
+            try {
+                return RetryingMetaStoreClient.getProxy(hiveConf, new Class[]{HiveConf.class, HiveMetaHookLoader.class, Boolean.class},
+                        new Object[]{hiveConf, null, true}, null, HiveMetaStoreClientCompatibility1xx.class.getName()
+                );
+            } catch (RuntimeException ex) {
+                LOG.warn("RuntimeException occurred while instantiating the HiveMetaStoreClientCompatibility1xx class", ex);
```
denalex, line 428: we do not know for sure that RuntimeException here would be caused only by the failure to instantiate `HiveMetaStoreClientCompatibility1xx` , it can happen for other reasons as well, I think we should adjust the error message and log it only if we're changing the actual exception that gets propagated.

server/pxf-hive/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClientCompatibility1xx.java
```diff
@@ -56,6 +57,9 @@ public Table getTable(String dbname, String name) throws TException {
                     Table table = client.get_table(dbname, name);
                     return new GetTableResult(table).getTable();
                 }
+            } catch (MetaException | NoSuchObjectException ex) {
+                LOG.warn("Original exception not re-thrown", e);
```
denalex, line 61: this is probably debug-level log entry as it will happen on every call for incompatible Hive versions

frankgh:
thanks for the review, @denalex. I have addressed your comments

PR #416, opened 2020-08-05
Fix performance issues when writing wide CSV/TEXT rows

frankgh:
During writing of wide CSV/TEXT rows (1MB+/row) PXF performance would be
greatly degraded. The performance degradation occurs at
org.greenplum.pxf.api.io.Text#setCapacity. Once the original buffer size
was exceeded, the setCapacity call would increase the buffer size by 1
and perform a copy of bytes. This became extremely slow when writing
wide rows because org.greenplum.pxf.api.io.Text would consume from the
buffer until it encountered a line break delimiter (\n). In this commit,
we delay consuming from the InputStream until the writing to HCFS
occurs, we consume the InputStream during write. The downside of this
approach is that we don't report the number of tuples written to the
external system correctly anymore, as we will always report a single
tuple written.

### DIFF
--- a/server/pxf-service/src/main/java/org/greenplum/pxf/service/BridgeInputBuilder.java
+++ b/server/pxf-service/src/main/java/org/greenplum/pxf/service/BridgeInputBuilder.java
@@ -26,7 +26,6 @@ import org.greenplum.pxf.api.GPDBWritableMapper;
 import org.greenplum.pxf.api.OneField;
 import org.greenplum.pxf.api.io.DataType;
 import org.greenplum.pxf.api.io.GPDBWritable;
-import org.greenplum.pxf.api.io.Text;
 import org.greenplum.pxf.api.model.OutputFormat;
 import org.greenplum.pxf.api.model.RequestContext;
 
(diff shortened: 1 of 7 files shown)


server/pxf-service/src/main/java/org/greenplum/pxf/service/BridgeInputBuilder.java
```diff
@@ -45,9 +44,16 @@ public BridgeInputBuilder(RequestContext protocolData) {
 
     public List<OneField> makeInput(DataInput inputStream) throws Exception {
         if (protocolData.getOutputFormat() == OutputFormat.TEXT) {
-            Text txt = new Text();
-            txt.readFields(inputStream);
-            return Collections.singletonList(new OneField(DataType.BYTEA.getOID(), txt.getBytes()));
+            // Avoid copying the bytes from the inputStream directly. This
```
denalex, line 47: being inside this `if` means external table was defined with FORMAT "TEXT" or "CSV", not that the actual profile will be`:text`, right ? Maybe coincidentally this is only used for :text profiles, but considering we were thinking of getting rid of `gpdb_writable` I think in the longer term we need to be more selective here and look at the actual profile or resolver to be used.

frankgh: > being inside this `if` means external table was defined with FORMAT "TEXT" or "CSV", not that the actual profile will be`:text`, right ? 

That is correct

> Maybe coincidentally this is only used for :text profiles, but considering we were thinking of getting rid of `gpdb_writable` I think in the longer term we need to be more selective here and look at the actual profile or resolver to be used.

The last decision during the rewrite was to keep some binary format, because we did see a performance gain when doing binary vs text, even though we had in the past done some comparisons that indicated otherwise. And that was only for the read case, we still need to define what we'll do about the write case

PR #417, opened 2020-08-10
docs - add content for SKIP_HEADER_COUNT

lisakowen:
document the SKIP_HEADER_COUNT custom option that the user can specify when reading a CSV-format file from hadoop or an object store.  this option identifies the number of rows/lines that PXF skips when reading the first split of the file(s).

questions - want to verify the following:
1.  that this option is not applicable when reading via S3-select.
2.  we haven't exposed the *:csv profile.  so i assumed that this option was supported for *:text profiles when FORMAT 'CSV'.

### DIFF
--- a/docs/content/hdfs_text.html.md.erb
+++ b/docs/content/hdfs_text.html.md.erb
@@ -46,11 +46,11 @@ The specific keywords and values used in the Greenplum Database [CREATE EXTERNAL
 | PROFILE    | The `PROFILE` keyword must specify `hdfs:text`. |
 | SERVER=\<server_name\>    | The named server configuration that PXF uses to access the data. Optional; PXF uses the `default` server if not specified. |
 | IGNORE_MISSING_PATH=\<boolean\> | Specify the action to take when \<path-to-hdfs-file\> is missing or invalid. The default value is `false`, PXF returns an error in this situation. When the value is `true`, PXF ignores missing path errors and returns an empty fragment. |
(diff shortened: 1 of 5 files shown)


docs/content/hdfs_text.html.md.erb
```diff
@@ -46,10 +46,11 @@ The specific keywords and values used in the Greenplum Database [CREATE EXTERNAL
 | PROFILE    | The `PROFILE` keyword must specify `hdfs:text`. |
 | SERVER=\<server_name\>    | The named server configuration that PXF uses to access the data. Optional; PXF uses the `default` server if not specified. |
 | IGNORE_MISSING_PATH=\<boolean\> | Specify the action to take when \<path-to-hdfs-file\> is missing or invalid. The default value is `false`, PXF returns an error in this situation. When the value is `true`, PXF ignores missing path errors and returns an empty fragment. |
+| SKIP_HEADER_COUNT=\<numlines\> | Specify the number of header lines that PXF should skip in the first split of each \<hdfs-file\> before reading the data. The default value is 0, do not skip any lines. Supported for external tables that you create with `FORMAT` `'CSV'`. |
```
frankgh, line 49: I guess we also support it with TEXT format

lisakowen: ok, sounds like doesn't matter what the user identifies for  FORMAT (CSV or TEXT), PXF will skip the number of lines specified.  is this correct?  does this mean that the feature is not csv-format specific - i.e. the user can specify it on any format text file?

frankgh: yes, that's correct, but I 'm not sure if that would make sense for other text formats (i.e json)

PR #420, opened 2020-08-17
Build and support hadoop native libraries in PXF

frankgh:
Build Hadoop Native Libraries from source for all the supported
platforms for PXF and add these libraries to the build process. Add the
native libraries to the RPMs produced during the build process. Ensure
that the native libraries are added to the java.library.path during PXF
start up.

### DIFF
--- a/concourse/README.md
+++ b/concourse/README.md
@@ -9,7 +9,7 @@ engine for Python. This allows the generation of portions of the
 pipeline from common blocks of pipeline code. Logic (Python code) can
 be embedded to further manipulate the generated pipeline.
 
-# Deploy the `pxf-build` (release) pipeline
+## Deploy the `pxf-build` (release) pipeline
 
 To deploy the build pipeline for PXF, make sure PXF master branch is currently checked-out and run this command:
 
@@ -17,15 +17,24 @@ To deploy the build pipeline for PXF, make sure PXF master branch is currently c
 make -C "${HOME}/workspace/pxf/concourse" build
 ```
 
-# Deploy the `pxf-certification` (release) pipeline
+## Deploy the `pxf-certification` (release) pipeline
 
-To deploy the certifcation pipeline (forward compatibility) for PXF, make sure PXF master branch is currently checked-out and run this command:
+To deploy the certification pipeline (forward compatibility) for PXF, make sure PXF master branch is currently checked-out and run this command:
 
 ```shell script
 make -C "${HOME}/workspace/pxf/concourse" certification
 ```
 
-# Deploy the singlecluster pipeline
+## Deploy the `pxf-native` pipeline
+
+The `pxf-native` pipeline builds native components to be shipped with PXF, for
+example, it builds the Hadoop Native Libraries to be embedded in the RPM.
+
+```shell script
+make pxf-native
+```
+
+## Deploy the singlecluster pipeline
 
 The singlecluster pipeline generates the singlecluster tarball for CDH, HDP2,
 and HDP3. The generated tarballs are then published to an S3 and GCS bucket.
@@ -35,7 +44,7 @@ The produced tarballs can then be consumed in the pxf-build pipelines.
 make singlecluster
 ```
 
-# Deploy the cloudbuild pipeline
+## Deploy the cloudbuild pipeline
 
 ```shell
 fly -t ud set-pipeline \
@@ -46,13 +55,13 @@ fly -t ud set-pipeline \
     -v pxf-git-branch=master -p cloudbuild
 ```
 
-# Deploy the pull-request pipeline
+## Deploy the pull-request pipeline
 
 ```shell
 make -C "${HOME}/workspace/pxf/concourse" pr
 ```
 
-# Deploy the performance pipelines
+## Deploy the performance pipelines
 
 10G Performance pipeline:
 
@@ -76,7 +85,7 @@ make SCALE=50 perf
 make SCALE=500 perf
 ```
 
-# Deploy a PXF acceptance pipeline
+## Deploy a PXF acceptance pipeline
 Acceptance pipelines can be deployed for feature testing purposes.
 ```shell
 ./deploy dev master -a -n acceptance
@@ -90,7 +99,7 @@ After acceptance, the pipeline can be cleaned up as follows:
 fly -t ud dp -p acceptance
 ```
 
-# Deploy development PXF pipelines
+## Deploy development PXF pipelines
 
 The dev pipeline is an abbreviated version of the `pxf-build` pipeline.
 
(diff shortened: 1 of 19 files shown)


concourse/README.md
```diff
@@ -9,23 +9,32 @@ engine for Python. This allows the generation of portions of the
...
 ```
 
-# Deploy the singlecluster pipeline
+## Deploy the `pxf-native` pipeline
+
+The `pxf-native` pipeline builds native components to be shipped with PXF, for
+example, it builds the Hadoop Native Libraries to be embedded in the RPM.
+
+```shell script
+make pxf-native
```
denalex, line 34:
```suggestion
make -C "${HOME}/workspace/pxf/concourse" pxf-native
```

concourse/README.md
```diff
@@ -35,7 +44,7 @@ The produced tarballs can then be consumed in the pxf-build pipelines.
 make singlecluster
```
denalex, line 44:
```suggestion
make -C "${HOME}/workspace/pxf/concourse" singlecluster
```

concourse/docker/pxf-dev-base/cloudbuild.yaml
```diff
@@ -93,6 +117,18 @@ steps:
...
+  args:
+  - 'build'
+  - '--build-arg=BASE_IMAGE=gcr.io/$PROJECT_ID/gpdb-pxf-dev/gpdb6-centos7-test-pxf:$COMMIT_SHA'
+  - '--tag=gcr.io/$PROJECT_ID/gpdb-pxf-dev/gpdb6-centos7-build-native:$COMMIT_SHA'
+  - '-f'
+  - 'concourse/docker/pxf-native-base/gpdb6/centos7/Dockerfile'
+  - '/workspace/build/'
+  waitFor:
+    - gpdb6-centos7-test-pxf-image
+
```
denalex, line 131: PXF native build does not depend on gp version, so will we only need 1 `centos7` job/image ?

concourse/pipelines/perf_pipeline.yml
```diff
@@ -168,6 +168,14 @@ resources:
     json_key: ((pxf-storage-service-account-key))
     versioned_file: build-dependencies/pxf-build-dependencies.tar.gz
 
+- name: hadoop_gp6_tarball_rhel7
+  type: gcs
+  icon: google-drive
+  source:
+    bucket: data-gpdb-ud-pxf-build-resources
+    json_key: ((pxf-storage-service-account-key))
+    versioned_file: hadoop-native/hadoop-gp6.el7.tar.gz
```
denalex, line 177: how would Hadoop version play into naming / building of the artifacts ? When we bump Hadoop version, the old artifact should ideally still be there for going backwards with patches, etc.

concourse/pipelines/pxf-native_pipeline.yml
```diff
@@ -0,0 +1,141 @@
...
+  source:
+    branch: ((pxf-git-branch))
+    uri: ((pxf-git-remote))
+
+- name: hadoop_src
+  type: git
+  icon: git
+  source:
+    uri: https://github.com/apache/hadoop.git
+    tag_filter: rel/release-2.9.2
```
denalex, line 35: this should ideally be a variable and play into artifact names

concourse/pipelines/pxf-native_pipeline.yml
```diff
@@ -0,0 +1,141 @@
...
+
+- name: hadoop_gp6_tarball_rhel7
+  type: gcs
+  icon: google-drive
+  source:
+    bucket: data-gpdb-ud-pxf-build-resources
+    json_key: ((pxf-storage-service-account-key))
+    versioned_file: hadoop-native/hadoop-gp6.el7.tar.gz
+
+## ---------- Centos 6 Swimlane ----------
```
denalex, line 90: if this is here, please put similar comments on other swimlanes :-)

concourse/tasks/build.yml
```diff
@@ -9,6 +9,8 @@ inputs:
   optional: true
 - name: pxf_src
 - name: gpdb_package
+- name: hadoop-native-libraries
+  optional: true
```
denalex, line 13: why is this optional if we're building them ?

server/pxf-service/src/main/java/org/greenplum/pxf/service/servlet/ServletLifecycleListener.java
```diff
@@ -20,6 +20,14 @@
...
 
 
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.crypto.OpensslCipher;
+import org.apache.hadoop.io.compress.Lz4Codec;
+import org.apache.hadoop.io.compress.SnappyCodec;
+import org.apache.hadoop.io.compress.ZStandardCodec;
+import org.apache.hadoop.io.compress.bzip2.Bzip2Factory;
+import org.apache.hadoop.io.compress.zlib.ZlibFactory;
+import org.apache.hadoop.util.NativeCodeLoader;
```
denalex, line 30: I'd suggest to move all new logic to some kind of utility, too much specifics for a lifecycle listener

server/pxf-service/src/scripts/pxf-env-default.sh
```diff
@@ -50,6 +47,13 @@ export PXF_SHUTDOWN_PORT=${PXF_SHUTDOWN_PORT:=5889}
 # Memory
 export PXF_JVM_OPTS=${PXF_JVM_OPTS:='-Xmx2g -Xms1g'}
 
+# Path to native libraries
+if [[ -d ${PXF_HOME}/lib/shared/native ]]; then
+  export LD_LIBRARY_PATH=${PXF_HOME}/lib/shared/native:${LD_LIBRARY_PATH}
+
+  PXF_JVM_OPTS="-Djava.library.path=${PXF_HOME}/lib/shared/native ${PXF_JVM_OPTS}"
```
denalex, line 54: do we know if both are needed or only one way is sufficient ?

concourse/tasks/build.yml
frankgh: I think I was thinking about deb, when deb comes, we won't have native libraries for Ubuntu, so it might block them, but maybe it's something we can get done quickly before deb work

server/pxf-service/src/main/java/org/greenplum/pxf/service/servlet/ServletLifecycleListener.java
frankgh: good point

server/pxf-service/src/scripts/pxf-env-default.sh
frankgh: I've seen both being done in different places, I'm not sure if both are needed though.

concourse/docker/pxf-dev-base/cloudbuild.yaml
frankgh: The gpdb5 and gpdb6 image toolchain differs a lot, so I am not sure if when building native code, do we need to have the same toolchain for the library?

concourse/pipelines/perf_pipeline.yml
frankgh: that's a great point. I'll address it

concourse/pipelines/pxf-native_pipeline.yml
frankgh: agree, I didn't get to do it, but I will address it

frankgh:
closing in favor of documenting how to configure PXF with Hadoop Native Libraries cc. @lisakowen

PR #421, opened 2020-08-18
Remove threadSafe locks

frankgh:
Currently we allow the bzip2 codec in when writing text and Avro formats
to an external source.

However, we do not run concurrent threads for bzip2 since it's not
considered to be thread-safe. This has been fixed by now, so we
should create an automation test to read/write files using bzip2
compression in multi-threaded mode to have assurance that bzip2 is now
thread safe.

Earlier versions of CBZip2InputStream, the library used by hadoop to
decode bzip2 files, were not thread-safe. This shortcoming has been
addressed in version 2.5.0 of hadoop [1]. Because PXF bundles newer
hadoop libraries, it doesn't need to lock anymore, as these libraries
are thread-safe now.

[1] https://issues.apache.org/jira/browse/HADOOP-10614


server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/WritableResource.java
```diff
@@ -107,49 +107,26 @@ public WritableResource() {
...
      * @throws Exception in case of wrong request parameters, failure to
-     *             initialize bridge or to write data
+     *                   initialize bridge or to write data
      */
     @POST
     @Path("stream")
     @Consumes(MediaType.APPLICATION_OCTET_STREAM)
     public Response stream(@Context final ServletContext servletContext,
                            @Context HttpHeaders headers,
-                           @QueryParam("path") String path,
```
denalex, line 125: I recall there was a log somewhere trying to log this value, has it already been fixed elsewhere ?

frankgh: we fix it as part of this PR. The path no longer comes as a query string parameter, but it is part of the request headers as the DATA-DIR header

PR #422, opened 2020-08-18
docs - add deprecation notice to THREAD-SAFE custom option

lisakowen:
some docs for https://github.com/greenplum-db/pxf/pull/421.

the PR deprecates and ignores the THREAD-SAFE custom option.  the docs mentioned this option in the write sections of the text and sequencefile formats.

in this PR:
- note in descriptions of the THREAD-SAFE option that it is deprecated.

(the 5.15 release notes will mention the locking change, and the deprecation - this will be a separate PR.)

### DIFF
--- a/docs/content/hdfs_seqfile.html.md.erb
+++ b/docs/content/hdfs_seqfile.html.md.erb
@@ -72,7 +72,6 @@ The `hdfs:SequenceFile` profile supports the following custom options:
 | COMPRESSION_CODEC    | The compression codec Java class name. If this option is not provided, Greenplum Database performs no data compression. Supported compression codecs include:<br>`org.apache.hadoop.io.compress.DefaultCodec`<br>`org.apache.hadoop.io.compress.BZip2Codec`<br>`org.apache.hadoop.io.compress.GzipCodec` |
 | COMPRESSION_TYPE    | The compression type to employ; supported values are `RECORD` (the default) or `BLOCK`. |
 | DATA-SCHEMA    | The name of the writer serialization/deserialization class. The jar file in which this class resides must be in the PXF classpath. This option is required for the `hdfs:SequenceFile` profile and has no default value. |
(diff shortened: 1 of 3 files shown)


lisakowen:
interesting, we only mentioned it in the docs when discussing write compression options...

frankgh:
yeah, we had similar logic in the Bridge (read) and the WritableResource

docs/content/hdfs_seqfile.html.md.erb
```diff
@@ -72,7 +72,7 @@ The `hdfs:SequenceFile` profile supports the following custom options:
 | COMPRESSION_CODEC    | The compression codec Java class name. If this option is not provided, Greenplum Database performs no data compression. Supported compression codecs include:<br>`org.apache.hadoop.io.compress.DefaultCodec`<br>`org.apache.hadoop.io.compress.BZip2Codec`<br>`org.apache.hadoop.io.compress.GzipCodec` |
 | COMPRESSION_TYPE    | The compression type to employ; supported values are `RECORD` (the default) or `BLOCK`. |
 | DATA-SCHEMA    | The name of the writer serialization/deserialization class. The jar file in which this class resides must be in the PXF classpath. This option is required for the `hdfs:SequenceFile` profile and has no default value. |
-| THREAD-SAFE | Boolean value determining if a table query can run in multi-threaded mode. The default value is `TRUE`. Set this option to `FALSE` to handle all requests in a single thread for operations that are not thread-safe (for example, compression). |
+| THREAD-SAFE | Boolean value determining if a table query or write can run in multi-threaded mode. The default value is `TRUE`. *Deprecated*; PXF ignores this option, all operations are thread-safe. |
```
denalex, line 75: should we remove this altogether from this table as it serves no value going forward ? `Deprecated` means still working slated for removal in the future, it is no longer working here ? If we want to mention it, maybe have a special notice section ?

PR #426, opened 2020-08-20
Log information about loaded hadoop native libraries during startup

frankgh:
Log at INFO level information about hadoop native libraries and native
hadoop codecs available for PXF during startup. This allows displaying
important information for PXF that can help in debugging issues with
different compression codecs used for hadoop. Some compression codecs
require native libraries, for example snappy, while others have both
native and java-based implementations of the compression codecs.

### DIFF
--- a/server/pxf-service/src/main/java/org/greenplum/pxf/service/utilities/NativeLibraryChecker.java
+++ b/server/pxf-service/src/main/java/org/greenplum/pxf/service/utilities/NativeLibraryChecker.java
@@ -1,0 +1,84 @@
+package org.greenplum.pxf.service.utilities;
+
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.crypto.OpensslCipher;
+import org.apache.hadoop.io.compress.Lz4Codec;
+import org.apache.hadoop.io.compress.SnappyCodec;
+import org.apache.hadoop.io.compress.ZStandardCodec;
+import org.apache.hadoop.io.compress.bzip2.Bzip2Factory;
+import org.apache.hadoop.io.compress.zlib.ZlibFactory;
+import org.apache.hadoop.util.NativeCodeLoader;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Helper class that checks whether native libraries have been loaded and logs
+ * information about the loaded native libraries
+ */
+public class NativeLibraryChecker {
+
+    private static final Logger LOG = LoggerFactory.getLogger(NativeLibraryChecker.class);
+
+    /**
+     * Checks whether native libraries are loaded and logs details of the
+     * loaded libraries
+     */
+    public static void checkNativeLibraries() {
+        boolean nativeHadoopLoaded = NativeCodeLoader.isNativeCodeLoaded();
+
+        if (nativeHadoopLoaded) {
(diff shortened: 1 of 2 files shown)


server/pxf-service/src/main/java/org/greenplum/pxf/service/utilities/NativeLibraryChecker.java
```diff
@@ -0,0 +1,84 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(NativeLibraryChecker.class);
+
+    /**
+     * Checks whether native libraries are loaded and logs details of the
+     * loaded libraries
+     */
+    public static void checkNativeLibraries() {
+        boolean nativeHadoopLoaded = NativeCodeLoader.isNativeCodeLoaded();
+
+        if (nativeHadoopLoaded) {
```
denalex, line 29: I'd revert the condition, check if not loaded, log message and exit. The rest will go as main body instead of `else` branch. Pattern of early return.

server/pxf-service/src/main/java/org/greenplum/pxf/service/utilities/NativeLibraryChecker.java
```diff
@@ -0,0 +1,84 @@
...
+            Configuration conf = new Configuration();
```
denalex, line 30: any danger of establishing Hadoop configuration this early (recall we had an issue with Kerberos before, perhaps no longer a problem) ?

server/pxf-service/src/main/java/org/greenplum/pxf/service/utilities/NativeLibraryChecker.java
```diff
@@ -0,0 +1,84 @@
...
+                openSslLoaded = false;
+            } else {
+                openSslDetail = OpensslCipher.getLibraryName();
+                openSslLoaded = true;
+            }
+            lz4LibraryName = Lz4Codec.getLibraryName();
+            if (bzip2Loaded) {
+                bzip2LibraryName = Bzip2Factory.getLibraryName(conf);
+            }
+
```
denalex, line 71: wonder if this will actually load all `.so` in memory, even though we might not be using any of them ? Ideally we would be just checking whether we `could load them, if needed`. How large are they ? Which memory do they load into (non-heap I'd assume)

server/pxf-service/src/main/java/org/greenplum/pxf/service/utilities/NativeLibraryChecker.java
frankgh: I think the issue in Kerberos was with `UserGroupInformation` and not the `Configuration` itself. We have `Configuration` objects instantiated before any Kerberos logic inside the `SecurityServletFilter`

server/pxf-service/src/main/java/org/greenplum/pxf/service/utilities/NativeLibraryChecker.java
frankgh: Yes, this will force loading the classes into memory. It looks like they are loaded in a `static` block. I'm not sure what would be an alternate way to check, ideas?

frankgh:
closing in favor of providing a set of tools for diagnostics

PR #429, opened 2020-08-24
Add pipeline tasks to package OSL file inside RPMs

denalex:


concourse/pipelines/templates/build_pipeline-tpl.yml
```diff
@@ -1532,7 +1550,7 @@ jobs:
     <<: *slack_alert
 {% endif %}
 
-- name: Publish PXF-GP5 and PXF-GP6 Artifacts to GP-RelEng
+- name: Publish_PXF-GP5_and_PXF-GP6_Artifacts_to_GP-RelEng
```
bradfordb-vmware, line 1553: I am curious about the name change -- did you add the underscores to make it easier to interact with this job via the CLI? Or just preference/consistency?

concourse/pipelines/templates/build_pipeline-tpl.yml
```diff
@@ -272,7 +272,7 @@ resources:
     bucket: ((pxf-releng-gp5-drop-bucket))
     region_name: ((aws-region))
     secret_access_key: ((bucket-secret-access-key))
-    regexp: ((pxf-releng-gp5-drop-path))/pxf-gp5-(.*)-1.el6.x86_64.rpm
+    regexp: ((pxf-releng-gp5-drop-path))/pxf-gp5-(.*)-2.el6.x86_64.rpm
```
bradfordb-vmware, line 275: One suggesting for a future refactor: We could move this into a Concourse var (e.g., `pxf-rpm-release`) and then only have to work about changing it in one place. I'll add a chore to the icebox.

concourse/pipelines/templates/build_pipeline-tpl.yml
```diff
@@ -387,6 +387,24 @@ resources:
     password: ((pxf-cloudbuild-service-account-key))
 {% set gp_ver = None %}
 
+- name: rpmrebuild-centos6-image
```
bradfordb-vmware, line 390: Since this is a template and we have looping constructs, would it make sense to define this inside a template loop?

concourse/pipelines/templates/build_pipeline-tpl.yml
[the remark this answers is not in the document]
denalex: want to specifically ensure we are shipping and naming as `-2`

concourse/pipelines/templates/build_pipeline-tpl.yml
denalex: having spaces in the job name breaks an HTML link in the email, will see if this will be better

concourse/pipelines/templates/build_pipeline-tpl.yml
denalex: ok, makes sense, thanks @bradfordb-vmware

concourse/pipelines/templates/build_pipeline-tpl.yml
denalex: could be, I was on the fence there, was actually advising against excessive loop usage to improve readability. I think we mostly use loops on gp_ver, to a lesser extend on centos versions

PR #431, opened 2020-08-27
Add Make target for setting a dev version of the full build pipeline

bradfordb-vmware:


concourse/vars/build.dev.yml
```diff
@@ -0,0 +1,5 @@
+pxf-git-remote-ssh-url: git@github.com:pivotal/gp-pxf-dev.git
```
frankgh, line 1: I guess we have a `concourse/settings` that is kind of equivalent to `vars`. Should we consolidate?

concourse/pipelines/templates/build_pipeline-tpl.yml
```diff
@@ -1565,10 +1568,10 @@ jobs:
     image: google-cloud-sdk-slim-image
     file: pxf_src/concourse/tasks/get_pxf_release_artifacts.yml
     params:
-      GCS_ODP_PATH: prod/ODP
-      GCS_OSL_PATH: prod/OSL
+      GCS_ODP_PATH: [[pxf_releases_bucket_prefix]]/ODP
```
frankgh, line 1571: what's the difference between `[` and `(`?

denalex: [ are template generation variables, it will generate yams files with resolved variables. We usually use that for control (send slack or not) and gp_ver. Should these buckets be driven by values of the secrets rather than template substitutions ?

concourse/vars/build.dev.yml
denalex: probably. We have tings scattered around `pxf-secrets` and settings. I'd like to understand what belongs where -- anything non-credential goes to settings ? In that case we might need to do some migration.

bradfordb-vmware: I think consolidation is a great idea. I'll move this file to `concourse/settings/build.dev.yml`

concourse/pipelines/templates/build_pipeline-tpl.yml
bradfordb-vmware: We can change this to `((pxf-release-bucket-prefix))` and add them to `concourse/settings/build.{dev,prod}.yml`. Since these aren't credentials, I like to keep the settings close to the pipeline.

concourse/Makefile
```diff
@@ -46,28 +48,61 @@ set-build-pipeline:
...
 		--config "$${PIPELINE_FILE}" \
-		--load-vars-from=$(SECRETS_FILE) \
-		--load-vars-from=$(HOME)/workspace/gp-continuous-integration/secrets/gpdb_common-ci-secrets.yml \
+		--load-vars-from=$(SECRETS_HOME)/pxf-secrets.yml \
+		--load-vars-from=$(SECRETS_HOME)/gpdb_common-ci-secrets.yml \
 		--load-vars-from=$(HOME)/workspace/pxf/concourse/settings/pxf-multinode-params.yml \
-		--load-vars-from=$(HOME)/workspace/gp-continuous-integration/secrets/ccp_ci_secrets_ud.yml \
-		--load-vars-from=$(HOME)/workspace/gp-continuous-integration/secrets/ccp_ci_secrets_ud_kerberos.yml \
+		--load-vars-from=$(SECRETS_HOME)/ccp_ci_secrets_ud.yml \
+		--load-vars-from=$(SECRETS_HOME)/ccp_ci_secrets_ud_kerberos.yml \
```
denalex, line 67: I thought at some point we talked about copying all secrets, even common GP CI ones into PXF secrets file. Not sure if this is still desired

bradfordb-vmware: w/ @ashuka24: We've consolidated the separate CCP credential files into the single `pxf-secrets.yml` and have updated the make commands.

PR #433, opened 2020-08-28
Parquet Write: Fix physical and logical storage for DATE types

frankgh:
Hive stores DATE types as `int32` with logical type of `DATE` whereas
PXF is storing DATE types as `BINARY` with `STRING` logical type:

**Hive**
```
 $ java -jar parquet-tools-1.12.0-SNAPSHOT.jar schema /tmp/000000_0
message hive_schema {
  optional int32 dt (DATE);
}
```

**PXF**
```
 $ java -jar parquet-tools-1.12.0-SNAPSHOT.jar schema /tmp/1597879570-0000000014_2.0.snappy.parquet
message hive_schema {
  optional binary dt (STRING);
}
```

This commit fixes how PXF stores DATE types to make it consistent with
the way Hive stores DATE types.

### DIFF
--- a/server/pxf-api/src/main/java/org/greenplum/pxf/api/GreenplumDateTime.java
+++ b/server/pxf-api/src/main/java/org/greenplum/pxf/api/GreenplumDateTime.java
@@ -9,21 +9,33 @@ import java.time.temporal.ChronoField;
  */
 public class GreenplumDateTime {
 
+    public static final String DATE_FORMATTER_BASE_PATTERN = "yyyy-MM-dd";
+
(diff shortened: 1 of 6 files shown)


frankgh:
Thanks for the review @ashuka24 and @denalex 

> Looks good. So it already does the correct thing for reading `int32 (DATE)` fields ?

Yes! we had added support in the past to read date fields

> Did we obsolete reading `binary (utf8)` as DATE if Greenplum column name is DATE ?

If your type in Greenplum is date, and your values in parquet are BINARY UTF8, we still allow reading this value. This makes this version of PXF backwards compatible with Parquet files that were written with previous versions of PXF

PR #437, opened 2020-09-02
Build DEB package for pxf-gp6 on Ubuntu 18.04

bradfordb-vmware:


concourse/scripts/pxf_common.bash
```diff
@@ -155,15 +155,15 @@ function install_gpdb_package() {
 		fi
 		python_dir=python${python_version}/site-packages
 		export_pythonpath+=:/usr/lib/${python_dir}:/usr/lib64/${python_dir}
-	elif command -v apt; then
-		# install GPDB DEB, apt wants an absolute path
+	elif command -v apt-get; then
```
frankgh, line 158: I was reading that `apt` is a more user friendly command introduced to be a more user facing, whereas `apt-get` gives  your more low-level control during installation. Any reason we prefer one or the other here?

dev/install_greenplum.bash
```diff
@@ -29,8 +29,8 @@ else
     fi
 
     echo "Installing GPDB from ${LATEST_DEB} ..."
-    # apt wants a full path
-	sudo apt install -qq "${PWD}/${LATEST_DEB}"
+    # apt-get wants a full path
+	sudo apt-get install -qq "${PWD}/${LATEST_DEB}"
```
frankgh, line 33: minor, this line uses tab, vs the lines preceding it use spaces

package/control
```diff
@@ -0,0 +1,5 @@
+Package: pxf-gp6
+Version: %VERSION%
+Architecture: amd64
+Maintainer: Greenplum Unmanaged Data <pvtl-gp-ud@vmware.com>
```
frankgh, line 4: should we parameterize the `Maintainer` as well? I'm concerned about our team email address being in the wild

concourse/scripts/pxf_common.bash
bradfordb-vmware: From `man 8 apt`

> The `apt` command is meant to be pleasant for end users and does not need to be backward compatible like apt-get(8).

When you use `apt` in a script it prints the following

```sh
WARNING: apt does not have a stable CLI interface. Use with caution in scripts.
```

That's why I prefer using `apt-get` in scripts.

dev/install_greenplum.bash
bradfordb-vmware: Good catch.

concourse/scripts/generate_releng_email.bash
```diff
@@ -22,6 +22,7 @@ version=$(<pxf_shipit_file/version)
 pxf_gp5_el6_releng_url="${RELENG_GP5_DROP_URL}/pxf-gp5-${version}-2.el6.x86_64.rpm"
 pxf_gp5_el7_releng_url="${RELENG_GP5_DROP_URL}/pxf-gp5-${version}-2.el7.x86_64.rpm"
 pxf_gp6_el7_releng_url="${RELENG_GP6_DROP_URL}/pxf-gp6-${version}-2.el7.x86_64.rpm"
+pxf_gp6_ubuntu18_releng_url="${RELENG_GP6_DROP_URL}/pxf-gp6_${version}-2.ubuntu18.04_amd64.deb"
```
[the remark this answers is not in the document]
bradfordb-vmware: This was something that @ashuka24 and I discussed a bit, let me see if I can summarize:

1. The `rpmbuild` tool builds RPM files and names them following the RHEL convention.
2. Using `dpkg-deb --build`, it is up to us to name the DEB file and here we decided to follow the larger Debian convention of `<pkg-name>_<version>_<architecture>.deb`. This gives us roughly `pxf-gp6` as the package, `5.15.1-1.ubuntu18.04` as the _package_ version, and `amd64` as the architecture.

Note that neither of the PXF RPM or DEB naming convention matches with the Greenplum Server RPM/DEB naming pattern which is `greenplum-db-<version>-{rhel6|rhel7|ubuntut18.04}-{x86_64|amd64}.{rpm|deb}`, e.g.:

- greenplum-db-6.9.1-rhel6-x86_64.rpm
- greenplum-db-6.9.1-rhel7-x86_64.rpm
- greenplum-db-6.9.1-ubuntu18.04-amd64.deb

I have a _personal_ preference for matching the platform's convention but this is a weakly help opinion.

package/control
denalex: Does it need to have email there ? Can we just say `VMware, Inc.` ?

package/prerm
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+rm -f /usr/local/pxf-gp6/conf/pxf-private.classpath
+rm -rf /usr/local/pxf-gp6/pxf-service
+rm -rf /usr/local/pxf-gp6/run
```
denalex, line 5: how is this file integrated with install / upgrade from DEB ?

package/control
bradfordb-vmware: From the [Debian Polciy Manual](https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-maintainer) (which I am using as a stand-in for technical documentation because I don't think the format of the control file is specified elsewhere:

> The name must come first, then the email address inside angle brackets <> (in RFC822 format).

But it appears that `dpkg-deb --build` does not check or enforce this, though `lintian` does flag it as an error.

Makefile
```diff
@@ -97,6 +97,40 @@ rpm-tar: rpm
...
+	set -e ;\
+	GP_MAJOR_VERSION=$$(cat external-table/build/metadata/gp_major_version) ;\
+	PXF_MAIN_VERSION=$${PXF_VERSION//-SNAPSHOT/} ;\
+	if [[ $${PXF_VERSION} == *"-SNAPSHOT" ]]; then PXF_RELEASE=SNAPSHOT; else PXF_RELEASE=1; fi ;\
+	rm -rf build/debbuild ;\
+	mkdir -p build/debbuild/usr/local/pxf-gp$${GP_MAJOR_VERSION} ;\
+	mkdir -p build/debbuild/usr/local/pxf-gp$${GP_MAJOR_VERSION}/gpextable ;\
+	cp -a external-table/build/stage/* build/debbuild/usr/local/pxf-gp$${GP_MAJOR_VERSION}/gpextable ;\
+	cp -a cli/build/stage/pxf/* build/debbuild/usr/local/pxf-gp$${GP_MAJOR_VERSION} ;\
+	cp -a server/build/stage/pxf/* build/debbuild/usr/local/pxf-gp$${GP_MAJOR_VERSION} ;\
```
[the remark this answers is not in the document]
bradfordb-vmware: The `rpm` target does not depend on the top-level `stage` target, instead invoking the `stage` target in `external-table`, `server`, and `cli`. The `deb` target just copied this pattern.

If I am understanding you correctly, we could refactor to make `stage` a prerequisite target for both `rpm` and `deb?

package/prerm
bradfordb-vmware: This file is copied into the `DEBIAN/` sub-directory before building the package; `dpkg`/`apt` will call this before uninstalling this version of the package, both when the user is removing the package altogether or when updating to the next version.

This is mean to mirror the `%preun` section of the RPM spec.

Makefile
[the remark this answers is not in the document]
denalex: we can, I am not sure we should. I don't recall why it was not done to start with, one thing different is that `stage` doesn't seem to copy external table artifacts into `gpextable` for some reason

concourse/pipelines/templates/build_pipeline-tpl.yml
```diff
@@ -724,14 +706,10 @@ jobs:
       - Test PXF-GP6-HDP2 on OEL7
 {% set gp_ver = None %}
 {% set gp_ver = 6 %}
-{% if build_deb %}
       - Test PXF-GP[[gp_ver]]-HDP2 on Ubuntu18
-{% endif %}
 {% set gp_ver = 6 %}
-{% if build_deb %}
     - get: pxf_gp[[gp_ver]]_tarball_ubuntu18
       passed: [Test PXF-GP[[gp_ver]]-HDP2 on Ubuntu18]
```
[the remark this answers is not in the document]
bradfordb-vmware: After trying to make this change, I think this is a bit more entangled than I initially thought. If we hardcode the number 6 in the job name, we should probably also hardcode it in all `passed` conditions that refer it. But then there are resource names that we could hardcode the number in, and so on. Further complicating this refactor, there are several locations where we have `gp_ver` block around Ubuntu **and** RHEL7 related items, for example

```jinja
{% set gp_ver = 6 %}
- name: gpdb[[gp_ver]]-pxf-dev-ubuntu18-image
  type: registry-image
  icon: docker
  source:
    repository: gcr.io/data-gpdb-ud/gpdb-pxf-dev/gpdb[[gp_ver]]-ubuntu18.04-test-pxf
    tag: latest
    username: _json_key

- name: gpdb[[gp_ver]]-pxf-dev-centos7-mapr-server-image
  type: registry-image
  icon: docker
  source:
    repository: gcr.io/data-gpdb-ud/gpdb-pxf-dev/gpdb[[gp_ver]]-centos7-test-pxf-mapr
    tag: latest
    username: _json_key
{% set gp_ver = None %}
```

Do we hardcode the version in these RHEL7 locations?

Would it make sense to do this refactoring (i.e., removing unnecessary `[[gp_ver]]`) as a separate chore?

Makefile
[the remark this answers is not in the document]
bradfordb-vmware: Looking at the [`pxf` secript in the `server` directory][0], it looks like the `gpextable` directory needs to exist and contain the external table PXF extension.

If the [top-level `deb` target][1] depended on the [top-level `stage` target][2], then the DEB package would not contain `$PXF_HOME/gpextable` and the extension would not be installed. We would need to fix that in the top-level `stage` target before we could use it.

[0]: https://github.com/greenplum-db/pxf/blob/a0d6713cc4fe2847856e27d632da8fb389f2d76f/server/pxf-service/src/scripts/pxf#L369-L381
[1]: https://github.com/greenplum-db/pxf/blob/456bfe4181419aa30d3d1bdbfb2c52ce1a0b1665/Makefile#L100-L119
[2]: https://github.com/greenplum-db/pxf/blob/a0d6713cc4fe2847856e27d632da8fb389f2d76f/Makefile#L42-L56

concourse/scripts/generate_releng_email.bash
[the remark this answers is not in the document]
frankgh: I see, I would prefer to maintain consistency with the other artifacts we produce, it looks weird that we stop consistency. But I'll leave it up to you guys to decide what's best.

Makefile
```diff
@@ -142,5 +141,7 @@ help:
 	@echo	'  - test - runs tests for PXF Go CLI and server'
 	@echo	'  - install - install PXF external table extension, CLI and server'
 	@echo	'  - tar - bundle PXF external table extension, CLI, server and tomcat into a single tarball'
-	@echo	'  - rpm - create PXF RPM / DEB package'
-	@echo	'  - rpm-tar - bundle PXF RPM / DEB package along with helper scripts into a single tarball'
+	@echo	'  - rpm - create PXF RPM'
```
ashuka24, line 144:
```suggestion
	@echo	'  - rpm - create PXF RPM package'
```

concourse/pipelines/templates/build_pipeline-tpl.yml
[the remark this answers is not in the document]
frankgh: I think the idea is that we will eventually add 7 in the list of versions. Not sure if we are going to support ubuntu 18 with 7 though. That might be a data point to make a decision on hardcoding the gp_ver or not

package/control
bradfordb-vmware: w/ @ashuka24: We've moved the value of the maintainer field to a Concourse variables file.

concourse/scripts/generate_releng_email.bash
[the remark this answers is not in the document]
bradfordb-vmware: w/ @ashuka24: We've update the filename to be consistent with the greenplum server filename conventions: `pxf-gp6-<version>-ubunutu18.04-amd64.deb`.

PR #438, opened 2020-09-03
Update template_tool to use python3 instead of python2

bradfordb-vmware:
python2 has reached EOL and can longer be installed via homebrew.
Homebrew can be used to install python3 and then jinja2 can be installed
with

```
python3 -m pip install jinja2
```

### DIFF
--- a/concourse/template_tool
+++ b/concourse/template_tool
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python3
 """Simple tool for interpolating jinja vars in pipeline configs
 Example usage:
 template-tool --template external-table-tpl.yml --vars gpdb_type=5X_STABLE num_gpdb5_versions=4


denalex (commented):
this does not work on my laptop as python3 is not readily available:

```
/usr/bin/python --version
Python 2.7.16

→ which python2
/usr/local/bin/python2
```
Python3 is available here: /usr/local/Cellar/python/3.6.3/bin/python3 but not explicitly linked into PATH.

I can obviously link it, but this seems not a default setup.

bradfordb-vmware:
Since Homebrew removed the `python@2` formula, I think it does automatically link when you install `python@3` now.

bradfordb-vmware:
> Are there other locations where we use python2 that we could easily migrate to python3?

I did an `ag` search over the PXF repo and this was the only explicit python2 I found.

denalex (approved):
yes, upgrading python3 via Homebrew seems to have fixed the link, works now

ashuka24:
> > Are there other locations where we use python2 that we could easily migrate to python3?
> 
> I did an `ag` search over the PXF repo and this was the only explicit python2 I found.

Do we need to explicitly set it to python3? i see in a `concourse/deploy` we just have `#!/usr/bin/env python` without specifying the version

bradfordb-vmware:
> > > Are there other locations where we use python2 that we could easily migrate to python3?
> > 
> > 
> > I did an `ag` search over the PXF repo and this was the only explicit python2 I found.
> 
> Do we need to explicitly set it to python3? i see in a `concourse/deploy` we just have `#!/usr/bin/env python` without specifying the version

The default for Homebrew  is to not link the unversioned `python` when installing the `python@3` formula.

ashuka24:
> > > > Are there other locations where we use python2 that we could easily migrate to python3?
> > > 
> > > 
> > > I did an `ag` search over the PXF repo and this was the only explicit python2 I found.
> > 
> > 
> > Do we need to explicitly set it to python3? i see in a `concourse/deploy` we just have `#!/usr/bin/env python` without specifying the version
> 
> The default for Homebrew is to not link the unversioned `python` when installing the `python@3` formula.

in which case should we be updating the other locations to specify the exact python version?

frankgh (approved):

PR #439, opened 2020-09-05
Add lib/native directory in PXF_CONF

frankgh:
Add a new configuration directory, `$PXF_CONF/lib/native`, where users
can install native libraries for PXF (i.e. Hadoop native libraries).
Also, add an option to export the LD_LIBRARY_PATH to that location if
users configure their native dependencies in that location.

### DIFF
--- a/server/pxf-service/src/templates/user/conf/pxf-env.sh
+++ b/server/pxf-service/src/templates/user/conf/pxf-env.sh
@@ -59,5 +59,5 @@ PXF_CONF="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd )"
 # Dump heap on OutOfMemoryError, set to dump path to enable
 # export PXF_OOM_DUMP_PATH=/tmp/pxf_heap_dump
 
-# Uncomment below if you installed native libraries to be loaded by PXF
-# export LD_LIBRARY_PATH=${PXF_CONF}/lib/native${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}
+# Additional native libraries to be loaded by PXF
+# export LD_LIBRARY_PATH=
--- a/server/pxf-service/src/templates/user/conf/pxf-env.sh
+++ b/server/pxf-service/src/templates/user/conf/pxf-env.sh
@@ -60,4 +60,4 @@ PXF_CONF="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd )"
 # export PXF_OOM_DUMP_PATH=/tmp/pxf_heap_dump
 
 # Uncomment below if you installed native libraries to be loaded by PXF
-# export LD_LIBRARY_PATH=${PXF_CONF}/lib/native:${LD_LIBRARY_PATH}
+# export LD_LIBRARY_PATH=${PXF_CONF}/lib/native${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}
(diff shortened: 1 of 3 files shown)


lisakowen (approved):
looks good to me.

server/pxf-service/src/templates/user/conf/pxf-env.sh
```diff
@@ -58,3 +58,6 @@ PXF_CONF="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd )"
 
 # Dump heap on OutOfMemoryError, set to dump path to enable
 # export PXF_OOM_DUMP_PATH=/tmp/pxf_heap_dump
+
+# Uncomment below if you installed native libraries to be loaded by PXF
+# export LD_LIBRARY_PATH=${PXF_CONF}/lib/native:${LD_LIBRARY_PATH}
```
bradfordb-vmware, line 63: Are we sure that `$LD_LIBRARY_PATH` is non-empty at this point? If not, we need to use a different pattern: https://stackoverflow.com/questions/9631228/how-to-smart-append-ld-library-path-in-shell-when-nounset/9631350#9631350. Here's a anecdote of this bug in the wild.

frankgh: Thanks for the feedback, I updated the PR

lisakowen: do we want the examples in the docs to use the new pattern as well?

server/pxf-service/src/templates/user/conf/pxf-env.sh
```diff
@@ -58,3 +58,6 @@ PXF_CONF="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd )"
 
 # Dump heap on OutOfMemoryError, set to dump path to enable
 # export PXF_OOM_DUMP_PATH=/tmp/pxf_heap_dump
+
+# Uncomment below if you installed native libraries to be loaded by PXF
+# export LD_LIBRARY_PATH=${PXF_CONF}/lib/native${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}
```
denalex, line 63: if they uncomment it, it will cause `${PXF_CONF}/lib/native` be listed after `/usr/lib/hadoop/lib/native` since that's how it is defined in pxf-env-default.sh, should it be in a different order ?

frankgh: @denalex good point, `${PXF_CONF}/lib/native` should be the first in the search path.

We should document that by default we pick native libraries from this path : `/usr/lib/hadoop/lib/native`. But if uncommented, the priority is 
- `${PXF_CONF}/lib/native`
- `/usr/lib/hadoop/lib/native`
- `LD_LIBRARY_PATH`

@bradfordb-vmware  are there any side effects of having `LD_LIBRARY_PATH` point to a non-existent directory?

denalex: The way it is currently implemented (which is correct), if users uncomment, the order will be:

${PXF_CONF}/lib/native
LD_LIBRARY_PATH
/usr/lib/hadoop/lib/native

denalex (approved):

server/pxf-service/src/templates/user/conf/pxf-env.sh
denalex: Will tests on single cluster now use this automatically ? What about dataproc ? We should have at least some environments to use it so we know it is working fine. And some environments should not have native so we know it is working as well.

lisakowen: do we still want to allow users to specify their own path for native libraries?  and if so, how do we want them to modify the LD_LIBRARY_PATH setting in pxf-env.sh - add their library directory as the first element?

PR #448, opened 2020-09-11
Update install_pxf_package function to work with DEB packages

bradfordb-vmware:

### DIFF
--- a/concourse/scripts/pxf_common.bash
+++ b/concourse/scripts/pxf_common.bash
@@ -305,7 +305,7 @@ function install_pxf_tarball() {
 
 function install_pxf_package() {
 	if [[ ${TARGET_OS} == centos ]]; then
-		# install GPDB RPM
+		# install PXF RPM
 		pkg_file=$(find pxf_package -name 'pxf-gp*.x86_64.rpm')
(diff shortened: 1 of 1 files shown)


concourse/scripts/pxf_common.bash
```diff
@@ -305,15 +305,25 @@ function install_pxf_tarball() {
...
 		pkg_file=$(find pxf_package -name 'pxf-gp*.x86_64.rpm')
 		if [[ -z ${pkg_file} ]]; then
 			echo "Couldn't find PXF RPM file in pxf_package. Skipping install..."
 			return 1
 		fi
 		echo "Installing ${pkg_file}..."
 		rpm --quiet -ivh "${pkg_file}" >/dev/null
+	elif [[ ${TARGET_OS} == ubuntu ]]; then
+		# install PXF DEB
+		pkg_file=$(find pxf_package -name 'pxf-gp*.deb')
```
ashuka24, line 318: for the rpm, we seem to check that the architecture is `x86_64` should we do a similar check here and include `amd64` in the regex?

ashuka24 (approved):

denalex (approved):

PR #453, opened 2020-09-22
Release UGI if there is an error during the filter execution

frankgh:
When the filter execution fails, for example when accessing a
non-existing HCFS path, the fragmenter call will throw an
InvalidInputException. This causes all segments to leave the UGI entry
available, and the FileSystem associated to the UGI does not get closed
until the UGI entry is expired 15 minutes earlier. This commit eagerly
closes UGI resources when an exception occurs.

### DIFF
--- a/server/pxf-service/src/main/java/org/greenplum/pxf/service/servlet/SecurityServletFilter.java
+++ b/server/pxf-service/src/main/java/org/greenplum/pxf/service/servlet/SecurityServletFilter.java
@@ -147,6 +147,7 @@ public class SecurityServletFilter implements Filter {
             return true;
         };
 
+        boolean exceptionDetected = false;
         try {
             // Retrieve proxy user UGI from the UGI of the logged in user
             UserGroupInformation userGroupInformation = ugiCache
@@ -157,20 +158,25 @@ public class SecurityServletFilter implements Filter {
             // Execute the servlet chain as that user
             userGroupInformation.doAs(action);
         } catch (UndeclaredThrowableException ute) {
+            exceptionDetected = true;
             // unwrap the real exception thrown by the action
             throw new ServletException(ute.getCause());
         } catch (InterruptedException ie) {
+            exceptionDetected = true;
             throw new ServletException(ie);
         } finally {
             // Optimization to cleanup the cache if it is the last fragment
-            LOG.debug("Releasing proxy user for session: {}. {}",
-                    session, lastCallForSegment ? " Last fragment call" : "");
+            boolean releaseUgi = lastCallForSegment || exceptionDetected;
+            LOG.debug("Releasing UGI from cache for session: {}. {}",
+                    session, exceptionDetected
+                            ? " Exception while processing"
+                            : (lastCallForSegment ? " Processed last fragment for segment" : ""));
             try {
-                ugiCache.release(session, lastCallForSegment);
+                ugiCache.release(session, releaseUgi);
             } catch (Throwable t) {
-                LOG.error("Error releasing UGICache for session: {}", session, t);
+                LOG.error("Error releasing UGI from cache for session: {}", session, t);
             }
-            if (lastCallForSegment) {
+            if (releaseUgi) {
                 LOG.info("Finished processing {}", session);
             }
         }
--- a/server/pxf-service/src/main/java/org/greenplum/pxf/service/servlet/SecurityServletFilter.java
+++ b/server/pxf-service/src/main/java/org/greenplum/pxf/service/servlet/SecurityServletFilter.java
@@ -105,17 +105,16 @@ public class SecurityServletFilter implements Filter {
 
         final String serverName = StringUtils.defaultIfBlank(getHeaderValue(request, SERVER_HEADER, false), "default");
         final String configDirectory = StringUtils.defaultIfBlank(getHeaderValue(request, CONFIG_HEADER, false), serverName);
-
-        Configuration configuration = configurationFactory.initConfiguration(configDirectory, serverName, gpdbUser, null);
-
-        boolean isUserImpersonation = secureLogin.isUserImpersonationEnabled(configuration);
+        final Configuration configuration = configurationFactory.initConfiguration(configDirectory, serverName, gpdbUser, null);
+        final boolean isUserImpersonation = secureLogin.isUserImpersonationEnabled(configuration);
+        final boolean isSecurityEnabled = Utilities.isSecurityEnabled(configuration);
 
         // Establish the UGI for the login user or the Kerberos principal for the given server, if applicable
         UserGroupInformation loginUser = secureLogin.getLoginUser(serverName, configDirectory, configuration);
 
         String serviceUser = loginUser.getUserName();
 
-        if (!isUserImpersonation && Utilities.isSecurityEnabled(configuration)) {
+        if (!isUserImpersonation && isSecurityEnabled) {
             // When impersonation is disabled and security is enabled
             // we check whether the pxf.service.user.name property was provided
             // and if provided we use the value as the remote user instead of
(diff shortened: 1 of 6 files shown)


server/pxf-service/src/main/java/org/greenplum/pxf/service/servlet/SecurityServletFilter.java
```diff
@@ -101,7 +101,7 @@ public void doFilter(final ServletRequest request, final ServletResponse respons
         final String gpdbUser = getHeaderValue(request, USER_HEADER, true);
         final String transactionId = getHeaderValue(request, TRANSACTION_ID_HEADER, true);
         final Integer segmentId = getHeaderValueInt(request, SEGMENT_ID_HEADER, true);
-        final boolean lastCallForSegment = getHeaderValueBoolean(request, LAST_FRAGMENT_HEADER, false);
+        boolean releaseUgi = getHeaderValueBoolean(request, LAST_FRAGMENT_HEADER, false);
```
denalex, line 104: I'd prefer to leave this line as before to capture the semantic meaning of the incoming header. we can tie it to releaseUGI logic elsewhere

server/pxf-service/src/main/java/org/greenplum/pxf/service/servlet/SecurityServletFilter.java
```diff
@@ -157,20 +157,22 @@ public void doFilter(final ServletRequest request, final ServletResponse respons
...
             // unwrap the real exception thrown by the action
             throw new ServletException(ute.getCause());
         } catch (InterruptedException ie) {
+            releaseUgi = true;
             throw new ServletException(ie);
         } finally {
             // Optimization to cleanup the cache if it is the last fragment
             LOG.debug("Releasing proxy user for session: {}. {}",
-                    session, lastCallForSegment ? " Last fragment call" : "");
+                    session, releaseUgi ? " Last fragment call" : "");
```
denalex, line 169: this semantic is not clear now. It would log "Last fragment call" if that was indeed last request for fragment data. We don't want to have this printed on error, so maybe have 2 conditions for UGI release - last call and error ?

server/pxf-service/src/main/java/org/greenplum/pxf/service/servlet/SecurityServletFilter.java
frankgh: sure, makes sense

server/pxf-service/src/main/java/org/greenplum/pxf/service/servlet/SecurityServletFilter.java
frankgh: I've addressed this issue. I think this wasn't clear in the past either, so I have updated the log message

server/pxf-service/src/main/java/org/greenplum/pxf/service/servlet/SecurityServletFilter.java
```diff
@@ -157,20 +157,22 @@ public void doFilter(final ServletRequest request, final ServletResponse respons
             // Execute the servlet chain as that user
             userGroupInformation.doAs(action);
         } catch (UndeclaredThrowableException ute) {
+            releaseUgi = true;
```
denalex, line 160:
```suggestion
            exceptionDetected = true;
```

server/pxf-service/src/main/java/org/greenplum/pxf/service/servlet/SecurityServletFilter.java
```diff
@@ -157,20 +157,22 @@ public void doFilter(final ServletRequest request, final ServletResponse respons
...
             // unwrap the real exception thrown by the action
             throw new ServletException(ute.getCause());
         } catch (InterruptedException ie) {
+            releaseUgi = true;
```
denalex, line 164:
```suggestion
            exceptionDetected = true;
```

server/pxf-service/src/main/java/org/greenplum/pxf/service/servlet/SecurityServletFilter.java
```diff
@@ -147,6 +146,7 @@ public void doFilter(final ServletRequest request, final ServletResponse respons
             return true;
         };
 
+        boolean releaseUgi = lastCallForSegment;
```
denalex, line 149:
```suggestion
        boolean exceptionDetected = false;
```

server/pxf-service/src/main/java/org/greenplum/pxf/service/servlet/SecurityServletFilter.java
```diff
@@ -157,20 +157,22 @@ public void doFilter(final ServletRequest request, final ServletResponse respons
...
             userGroupInformation.doAs(action);
         } catch (UndeclaredThrowableException ute) {
+            releaseUgi = true;
             // unwrap the real exception thrown by the action
             throw new ServletException(ute.getCause());
         } catch (InterruptedException ie) {
+            releaseUgi = true;
             throw new ServletException(ie);
         } finally {
             // Optimization to cleanup the cache if it is the last fragment
```
denalex, line 167:
```suggestion
            boolean releaseUgi = lastCallForSegment || exceptionDetected;
            // Optimization to cleanup the cache if it is the last fragment
```

server/pxf-service/src/main/java/org/greenplum/pxf/service/servlet/SecurityServletFilter.java
```diff
@@ -157,20 +157,22 @@ public void doFilter(final ServletRequest request, final ServletResponse respons
...
             LOG.debug("Releasing proxy user for session: {}. {}",
```
denalex, line 168:
```suggestion
            LOG.debug("Releasing UGI from cache for session: {}. {}",
```

server/pxf-service/src/main/java/org/greenplum/pxf/service/servlet/SecurityServletFilter.java
```diff
@@ -157,20 +157,22 @@ public void doFilter(final ServletRequest request, final ServletResponse respons
...
-                    session, lastCallForSegment ? " Last fragment call" : "");
+                    session, releaseUgi ? " Cleaning UGI immediately from UGICache if there are no other references to this UGI entry." : "");
             try {
-                ugiCache.release(session, lastCallForSegment);
+                ugiCache.release(session, releaseUgi);
             } catch (Throwable t) {
                 LOG.error("Error releasing UGICache for session: {}", session, t);
```
denalex, line 173:
```suggestion
                LOG.error("Error releasing UGI from cache for session: {}", session, t);
```

server/pxf-service/src/main/java/org/greenplum/pxf/service/servlet/SecurityServletFilter.java
```diff
@@ -157,20 +157,22 @@ public void doFilter(final ServletRequest request, final ServletResponse respons
...
             // unwrap the real exception thrown by the action
             throw new ServletException(ute.getCause());
         } catch (InterruptedException ie) {
+            releaseUgi = true;
             throw new ServletException(ie);
         } finally {
             // Optimization to cleanup the cache if it is the last fragment
             LOG.debug("Releasing proxy user for session: {}. {}",
-                    session, lastCallForSegment ? " Last fragment call" : "");
+                    session, releaseUgi ? " Cleaning UGI immediately from UGICache if there are no other references to this UGI entry." : "");
```
denalex, line 169:
```suggestion
                    session, exceptionDetected ? " Exception while processing" : (lastCallForSegment ? " Processed last fragment for segment" : ""));
```

denalex: we don't need to explain in the DEBUG log statement, but we need to reflect which path our logic took for troubleshooting, so would like to know why we are calling this -- due to exception or receiving last fragment flag

denalex (approved):

PR #454, opened 2020-09-24
Publish PXF binary tarballs

frankgh:
Support an internal component tarball for new PXF releases that fully
support installation with no manual action. The
`Publish_PXF-GP5_and_PXF-GP6_Artifacts_to_GP-RelEng` job in the
`pxf-build` pipeline will now has two new tasks:
- Generate PXF binary tarball from PXF-GP5 on RHEL6 RPM
- Generate PXF binary tarball from PXF-GP6 on RHEL7 RPM

These tasks will produce a tarball with an installer script from a
PXF RPM. The binary tarballs need to be created for both GP5 (RHEL6) and
GP6 (RHEL7) only.

The script that is run by the task does the following:
- Takes as input a `licensed` PXF RPM
- Tars the content of the PXF RPM so it is laid out in $GPHOME upon
  extraction
- Tars the resulting tar from the previous step and an installer script
- Provides an `install_gpdb_component` executable bash script that
  installs PXF under GPHOME (GPHOME is mandatory)

The produced artifacts are then uploaded to Google Cloud Storage buckets
to be consumed by release engineering.

### DIFF
--- a/concourse/pipelines/templates/build_pipeline-tpl.yml
+++ b/concourse/pipelines/templates/build_pipeline-tpl.yml
@@ -260,21 +260,17 @@ resources:
 
 ## ---------- RelEng Drop Artifacts ----------
 - name: pxf_gp5_rhel6_release_to_releng
-  type: s3
+  type: gcs
   source:
-    access_key_id: ((bucket-access-key-id))
     bucket: ((pxf-releng-gp5-drop-bucket))
-    region_name: ((aws-region))
-    secret_access_key: ((bucket-secret-access-key))
+    json_key: ((concourse-gcs-resources-service-account-key))
     regexp: ((pxf-releng-gp5-drop-path))/pxf-gp5-(.*)-2.el6.x86_64.rpm
 
 - name: pxf_gp5_rhel7_release_to_releng
-  type: s3
+  type: gcs
   source:
-    access_key_id: ((bucket-access-key-id))
     bucket: ((pxf-releng-gp5-drop-bucket))
-    region_name: ((aws-region))
-    secret_access_key: ((bucket-secret-access-key))
+    json_key: ((concourse-gcs-resources-service-account-key))
     regexp: ((pxf-releng-gp5-drop-path))/pxf-gp5-(.*)-2.el7.x86_64.rpm
 
 - name: pxf_gp6_rhel7_release_to_releng
--- a/concourse/pipelines/templates/build_pipeline-tpl.yml
+++ b/concourse/pipelines/templates/build_pipeline-tpl.yml
@@ -305,6 +305,21 @@ resources:
     json_key: ((concourse-gcs-resources-service-account-key))
     regexp: ((pxf-releng-odp-drop-path))/((pxf-odp-file-prefix))-(.*)-ODP.tar.gz
 
+## ---------- RelEng Drop Artifacts for tarballs ----------
+- name: pxf_gp5_tarball_release_to_releng
+  type: gcs
+  source:
+    bucket: ((pxf-releng-gp6-drop-bucket))
+    json_key: ((concourse-gcs-resources-service-account-key))
+    regexp: ((pxf-releng-gp5-drop-path))/pxf-(.*).tar.gz
+
+- name: pxf_gp6_tarball_release_to_releng
+  type: gcs
+  source:
+    bucket: ((pxf-releng-gp6-drop-bucket))
+    json_key: ((concourse-gcs-resources-service-account-key))
+    regexp: ((pxf-releng-gp6-drop-path))/pxf-(.*).tar.gz
+
 ## ---------- PXF ShipIt resources ----------
 - name: pxf_shipit_file
   type: gcs
@@ -1563,6 +1578,18 @@ jobs:
     params:
       GP_VER: 6
       TARGET_OS_VERSION: ubuntu18.04
+  - task: Generate PXF binary tarball from PXF-GP5 on RHEL6 RPM
+    file: pxf_src/concourse/tasks/generate_tarball_from_rpm.yml
+    image: rpmrebuild-centos6-image
+    params:
+      GP_VER: 5
+      TARGET_OS_VERSION: 6
+  - task: Generate PXF binary tarball from PXF-GP6 on RHEL7 RPM
+    file: pxf_src/concourse/tasks/generate_tarball_from_rpm.yml
+    image: rpmrebuild-centos7-image
+    params:
+      GP_VER: 6
+      TARGET_OS_VERSION: 7
   - task: Generate Releng Email
     image: google-cloud-sdk-slim-image
     file: pxf_src/concourse/tasks/generate_releng_email.yml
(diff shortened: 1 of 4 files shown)


concourse/pipelines/templates/build_pipeline-tpl.yml
```diff
@@ -305,6 +301,21 @@ resources:
     json_key: ((concourse-gcs-resources-service-account-key))
     regexp: ((pxf-releng-odp-drop-path))/((pxf-odp-file-prefix))-(.*)-ODP.tar.gz
 
+## ---------- RelEng Drop Artifacts for MS ----------
```
denalex, line 304: no MS please on OSS code

concourse/scripts/generate_tarball_from_rpm.bash
```diff
@@ -0,0 +1,62 @@
...
+if [[ ${rpm_file_name##*/} =~ pxf-gp${GP_VER}-([0-9.]+)-2[\.-](.*\.(deb|rpm)) ]]; then
+  pxf_version=${BASH_REMATCH[1]}
+  suffix=${BASH_REMATCH[2]}
+  echo "Determined PXF version number to be '${pxf_version}' with suffix '${suffix}'..."
+else
+  echo "Couldn't determine version number from file named '${rpm_file_name}'..."
+  exit 1
+fi
+
+# install the new RPM, check that the OSL file is present
```
denalex, line 30: we're not checking OSL file here

concourse/scripts/generate_tarball_from_rpm.bash
```diff
@@ -0,0 +1,62 @@
...
+function fail() {
+  echo "Error: $1"
+  exit 1
+}
+
+# check if directory with artifacts is available
+[[ -d pxf_artifacts ]] || fail "pxf_artifacts directory not found"
+
+# check if the RPM for GP version and TARGET_OS_VERSION is available
+rpm_file_name=$(find pxf_artifacts -type f -name "pxf-gp${GP_VER}-*-2.el${TARGET_OS_VERSION}.x86_64.rpm")
```
denalex, line 17: if you expect files in `pxf_artifacts/licensed` below, why not search there ?

concourse/scripts/generate_tarball_from_rpm.bash
```diff
@@ -0,0 +1,62 @@
...
+[[ -f ${rpm_file_name} ]] || fail "pxf_artifacts/licensed/pxf-gp${GP_VER}-*-2.el${TARGET_OS_VERSION}.x86_64.rpm not found"
+
+# attempt to determine the PXF version
+if [[ ${rpm_file_name##*/} =~ pxf-gp${GP_VER}-([0-9.]+)-2[\.-](.*\.(deb|rpm)) ]]; then
```
denalex, line 21: we are not dealing with `deb` here

concourse/scripts/generate_tarball_from_rpm.bash
```diff
@@ -0,0 +1,62 @@
...
+  pxf_version=${BASH_REMATCH[1]}
+  suffix=${BASH_REMATCH[2]}
```
denalex, line 23: why even bother if you gonna install the RPM, the version is in the `version` file

concourse/scripts/generate_tarball_from_rpm.bash
```diff
@@ -0,0 +1,62 @@
...
+rm -rf /tmp/pxf_tarball_repackage/pxf/gpextable/
+
+# list staging directory
+echo "listing staging directory /tmp/pxf_tarball_repackage"
+ls -al /tmp/pxf_tarball_repackage
+
+# create the pxf.tar.gz that contains all files from the RPM installation
+echo "create the pxf tarball"
+mkdir -p /tmp/pxf_tarball
+tar -czf /tmp/pxf_tarball/pxf.tar.gz -C /tmp/pxf_tarball_repackage .
```
denalex, line 50: so we're packaging the OSL file as well ?

concourse/scripts/generate_tarball_from_rpm.bash
```diff
@@ -0,0 +1,62 @@
...
+
+# create the install_gpdb_component file
+cat > /tmp/pxf_tarball/install_gpdb_component <<EOF
+#!/bin/bash
+set -x
+: "\${GPHOME:?GPHOME must be set}"
+tar xvzf pxf.tar.gz -C \$GPHOME
```
denalex, line 57: wouldn't you want to determine base_dir and prefix `pxf.tar.gz` with it so that the script can be called from anywhere ?

concourse/scripts/generate_tarball_from_rpm.bash
frankgh: yes, it's required, checked with releng

concourse/scripts/generate_tarball_from_rpm.bash
frankgh: sure

concourse/scripts/generate_tarball_from_rpm.bash
bradfordb-vmware: @ashuka24 and I were working with bash regex earlier and the following pattern works:

```bash
if ! [[ ${rpm_file_name##*/} =~ pxf-gp${GP_VER}-([0-9.]+)-2[\.-](.*\.(deb|rpm)) ]]; then
  echo "Couldn't determine version number from file named '${rpm_file_name}'..."
  exit 1
fi
pxf_version=${BASH_REMATCH[1]}
suffix=${BASH_REMATCH[2]}
echo "Determined PXF version number to be '${pxf_version}' with suffix '${suffix}'..."
```

I feel like it is _marginally_ more readable by turning the `if`-condition into a guard clause.

frankgh: I went with Alex's suggestion of just getting the PXF version from the version file

frankgh: so no more if :(

concourse/scripts/generate_tarball_from_rpm.bash
```diff
@@ -0,0 +1,62 @@
...
+# create the pxf.tar.gz that contains all files from the RPM installation
+echo "create the pxf tarball"
+mkdir -p /tmp/pxf_tarball
+tar -czf /tmp/pxf_tarball/pxf.tar.gz -C /tmp/pxf_tarball_repackage .
+
+# create the install_gpdb_component file
+cat > /tmp/pxf_tarball/install_gpdb_component <<EOF
+#!/bin/bash
+set -x
+: "\${GPHOME:?GPHOME must be set}"
```
bradfordb-vmware, line 50: Have we confirmed w/ RelEng if `$GPHOME` will be set when this is called? I can't remember.

frankgh: it was on the story as a requirement:

> When install_gpdb_component is run using the standard process of installing internal components it shall:
> - Check is $GPHOME is set, fail otherwise
> - ...

frankgh: thanks for checking @bradfordb-vmware . GPHOME is set to run the script.

concourse/scripts/generate_tarball_from_rpm.bash
```diff
@@ -0,0 +1,62 @@
...
+  pxf_version=${BASH_REMATCH[1]}
+  suffix=${BASH_REMATCH[2]}
+  echo "Determined PXF version number to be '${pxf_version}' with suffix '${suffix}'..."
+else
+  echo "Couldn't determine version number from file named '${rpm_file_name}'..."
+  exit 1
+fi
+
+# install the new RPM, check that the OSL file is present
+rpm -ivh "$rpm_file_name"
```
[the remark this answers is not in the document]
frankgh: for now, I'd like to keep rpm install, but we can revisit if requirements change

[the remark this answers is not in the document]
frankgh: after some discussion, we think the `rpm2cpio` approach is more beneficial for the future

denalex (approved):

PR #456, opened 2020-09-30
Add support for Network File Storage (NFS) 

frankgh:
The NFS profile allows PXF reading data from Network File Systems. The
NFS has to be mounted on every segment host where PXF is running from.
PXF will then read/write data from the mounted NFS file.

For NFS, the `pxf.fs.basePath` property must be set to a valid path. The
`pxf.fs.basePath` property represents the base path to be used for
tables accessing that server.

Add automation test for the NFS profile using a multinode cluster


server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HcfsType.java
```diff
@@ -199,8 +213,23 @@ public String getDataUri(Configuration configuration, String path) {
...
+
+        String effectiveBasePath = StringUtils.removeStart(dataSource, "/");
+
+        if ("..".equals(effectiveBasePath) || StringUtils.contains(effectiveBasePath, "../")) {
+            // Disallow relative paths
+            throw new IllegalArgumentException(String
+                    .format("the provided path '%s' is invalid. Relative paths are not allowed by PXF", effectiveBasePath));
+        }
+
+        if (StringUtils.contains(effectiveBasePath, "$")) {
```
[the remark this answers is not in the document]
frankgh: yeah, but we are being conservative here. We make a big bet that nobody will actually have a path with `$` in the path name.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HcfsType.java
```diff
@@ -199,8 +215,23 @@ public String getDataUri(Configuration configuration, String path) {
      * @param dataSource The path to the data source
      * @return the normalized path to the data source
      */
-    public String normalizeDataSource(String dataSource) {
-        return StringUtils.removeStart(dataSource, "/");
+    public String validateAndNormalizeDataSource(String dataSource) {
+
+        String effectiveBasePath = StringUtils.removeStart(dataSource, "/");
```
frankgh, line 220: effectiveDataSource

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HcfsType.java
```diff
@@ -212,20 +243,37 @@ protected String getDataUriForPrefix(Configuration configuration, String dataSou
...
+     * Validates the basePath and normalizes it for the appropriate filesystem
+     *
+     * @param basePath the basePath as configured by the user
+     * @return the normalized basePath
+     */
+    protected String validateAndNormalizeBasePath(String basePath) {
+        if (StringUtils.isNotBlank(basePath)) {
+            if ("/".equals(basePath))
+                return "/";
+            return StringUtils.removeEnd(StringUtils.removeStart(basePath, "/"), "/") + "/";
```
denalex, line 271: this logic will still be correct for "/" input, right ? Then I think comparing is a premature optimization.

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/HcfsTypeTest.java
```diff
@@ -373,4 +374,144 @@ public void testHcfsGlobPattern() {
...
+        assertEquals("file:///my/base/path/foo/bar.txt", uri);
+
+        // no / preceding the basePath
+        configuration.set("pxf.fs.basePath", "my/base/path");
+        context.setProfileScheme("nfs");
+        uri = nfs.getDataUri(configuration, context);
+        assertEquals("file:///my/base/path/foo/bar.txt", uri);
+
+        // no / trailing the basePath
+        configuration.set("pxf.fs.basePath", "my/base/path");
```
denalex, line 459: this seems to be the same as the previous one

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/HcfsTypeTest.java
```diff
@@ -373,4 +374,144 @@ public void testHcfsGlobPattern() {
...
+        context.setDataSource("..");
+        HcfsType nfs = HcfsType.getHcfsType(configuration, context);
+        nfs.getDataUri(configuration, context);
+    }
+
+    @Test
+    public void testDataSourceWithTwoDotsInName() {
+        configuration.set("pxf.fs.basePath", "/some/base/path");
+        context.setProfileScheme("nfs");
+        context.setDataSource("a..txt");
```
denalex, line 499: let's also test `../` and `a/..`

server/pxf-service/src/main/resources/pxf-profiles-default.xml
```diff
@@ -749,4 +749,106 @@ under the License.
         </plugins>
         <protocol>wasbs</protocol>
     </profile>
+
+    <!-- NFS (Network File System profiles -->
```
denalex, line 753:
```suggestion
    <!-- NFS (Network File System) profiles -->
```

server/pxf-service/src/main/resources/pxf-profiles-default.xml
```diff
@@ -749,4 +749,106 @@ under the License.
...
+        <description>
+            Profile for accessing Sequence files serialized with a custom Writable class
+        </description>
+        <plugins>
+            <fragmenter>org.greenplum.pxf.plugins.hdfs.HdfsDataFragmenter</fragmenter>
+            <accessor>org.greenplum.pxf.plugins.hdfs.SequenceFileAccessor</accessor>
+            <resolver>org.greenplum.pxf.plugins.hdfs.WritableResolver</resolver>
+        </plugins>
+        <protocol>nfs</protocol>
+    </profile>
```
denalex, line 853: I wonder if anyone is using these SequenceFile profiles ? These are for MR jobs, I think. I don't think they will be relevant for NFS at all.

server/pxf-service/src/templates/user/templates/pxf-site.xml
```diff
@@ -35,4 +35,35 @@
...
+
+            Then PXF will use the :///mount/path/path/to/data URI to read
+            AVRO files. Users won't be able to access data at the root
+            directory of the filesystem.
+
+            This property can also be used for any Hadoop compatible file
+            system, for example S3.
+        ]]>
+        </description>
+    </property>
```
denalex, line 67: I think this is too much, especially the CDATA block. One paragraph should be enough, the rest can be found in documentation.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HcfsType.java
```diff
@@ -199,8 +213,23 @@ public String getDataUri(Configuration configuration, String path) {
      * @param dataSource The path to the data source
      * @return the normalized path to the data source
      */
-    public String normalizeDataSource(String dataSource) {
-        return StringUtils.removeStart(dataSource, "/");
+    public String validateAndNormalizeDataSource(String dataSource) {
+
+        String effectiveBasePath = StringUtils.removeStart(dataSource, "/");
+
+        if ("..".equals(effectiveBasePath) || StringUtils.contains(effectiveBasePath, "../")) {
```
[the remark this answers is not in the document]
denalex: it does, but we don't want to be super smart here as that location is reachable without using `..`

concourse/scripts/install_nfs.bash
```diff
@@ -0,0 +1,90 @@
...
+  chmod +x /tmp/install_and_configure_nfs_client.sh
+  scp /tmp/install_and_configure_nfs_client.sh "${MASTER_HOSTNAME}:~gpadmin"
+}
+
+# assumes only two segment hosts sdw1 and sdw2
+function run_nfs_installation() {
+
+  # install and configure the NFS server on master
+  ssh "centos@${MASTER_HOSTNAME}" "
+    echo 'install NFS server'
```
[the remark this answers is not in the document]
frankgh: I think this is a pattern we've been using for commands we run on master.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HcfsType.java
frankgh: yes, this is something I was doing somewhere else

server/pxf-service/src/main/resources/pxf-profiles-default.xml
frankgh: yeah, that makes sense

server/pxf-service/src/templates/user/templates/pxf-site.xml
frankgh: yeah, I figured this was going to be a lot, maybe @lisakowen can use it as a base for documentation

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HcfsType.java
```diff
@@ -212,20 +234,34 @@ protected String getDataUriForPrefix(Configuration configuration, String dataSou
...
+        String normalizedDataSource = validateAndNormalizeDataSource(dataSource);
 
         if (FILE_SCHEME.equals(defaultFS.getScheme())) {
             // if the defaultFS is file://, but enum is not FILE, use enum scheme only
-            uri = StringUtils.removeEnd(scheme, "://") + "://" + normalizedDataSource;
+            uri = StringUtils.removeEnd(scheme, "://") + "://" + basePath + normalizedDataSource;
         } else {
             // if the defaultFS is not file://, use it, instead of enum scheme and append user's path
-            uri = StringUtils.removeEnd(defaultFS.toString(), "/") + "/" + normalizedDataSource;
+            uri = StringUtils.removeEnd(defaultFS.toString(), "/") + basePath + "/" + normalizedDataSource;
```
denalex, line 255: normalized base path already has "/" at the end, unless it is empty, so will you end up with "//" here ?

frankgh: yes, this is incorrect

server/pxf-service/src/templates/user/templates/pxf-site.xml
```diff
@@ -35,4 +35,16 @@
         </description>
     </property>
     -->
+
+    <!--
+    <property>
+        <name>pxf.fs.basePath</name>
```
denalex, line 41: just a thought whether `pxf.fs.basePath` is the best user-visible name we can come up with. Not sure whether `pxf.file.base` is any better. In reality, it seems this property will impact other profiles like hdfs, s3, etc (although it is not mandatory there)

denalex (approved):

server/pxf-service/src/main/resources/pxf-profiles-default.xml
```diff
@@ -749,4 +749,75 @@ under the License.
         </plugins>
         <protocol>wasbs</protocol>
     </profile>
+
+    <!-- File profiles for file systems mounted on the Greenplum cluster -->
+    <profile>
+        <name>file:text</name>
+        <description>This profile is suitable for using when reading delimited single line records
+            from plain text, tab-delimited, files on Azure Data Lake
```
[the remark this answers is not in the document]
frankgh: good catch! thanks

PR #457, opened 2020-10-03
Convert pxf-cli to use go modules instead of go dep

lavjain:


frankgh (approved):
It looks good to me, maybe let's get another +1 to merge this PR

bradfordb-vmware (approved):
Ship It!

PR #460, opened 2020-10-13
Avro: fixing NullPointerException for writing NULL values for SMALLINT and BYTEA columns

apiminov:


server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/AvroResolver.java
```diff
@@ -141,10 +141,10 @@ public OneRow setFields(List<OneField> record) {
         for (OneField field : record) {
             if (field.type == DataType.BYTEA.getOID()) {
                 // Avro does not seem to understand regular byte arrays
-                field.val = ByteBuffer.wrap((byte[]) field.val);
+                field.val = field.val != null ? ByteBuffer.wrap((byte[]) field.val) : null;
```
frankgh, line 144: maybe we can have a single check for non-null once (between lines 141 and 142) instead

frankgh (approved):
Thanks for the PR
1

frankgh:
@lisakowen can we track this fix for documentation? Thanks

PR #461, opened 2020-10-14
Increase default maximumPoolSize property in jdbc-site.xml

frankgh:
The default value for jdbc.pool.property.maximumPoolSize is causing
problems in multiple users. The out-of-the-box experience for PXF with
JDBC is not great because an error is encountered during writes for
example when we have a large amount of data. The same is true for reads
that transfer large amounts of data.

This error is encountered on segment hosts that have more than 5
segments per host. A typical use case for PXF is residing on a host with
8 segments per hosts. This issue will surface right away on those
segment hosts. In some cases users have 12 segments per host. Increasing
the maximumPoolSize to 15 seems like a new sensible default that will
allow the out-of-the-box experience be smoother for PXF.

### DIFF
--- a/server/pxf-service/src/templates/user/templates/jdbc-site.xml
+++ b/server/pxf-service/src/templates/user/templates/jdbc-site.xml
@@ -34,7 +34,7 @@
 
     <property>
         <name>jdbc.pool.property.maximumPoolSize</name>
-        <value>5</value>
+        <value>15</value>
         <description>The maximum number of actual connections to the database backend</description>
     </property>
     <property>
--- a/server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
+++ b/server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
@@ -334,7 +334,7 @@ public class JdbcBasePlugin extends BasePlugin {
         if (isConnectionPoolUsed) {
             poolConfiguration = new Properties();
             // for PXF upgrades where jdbc-site template has not been updated, make sure there're sensible defaults
-            poolConfiguration.setProperty("maximumPoolSize", "5");
+            poolConfiguration.setProperty("maximumPoolSize", "15");
             poolConfiguration.setProperty("connectionTimeout", "30000");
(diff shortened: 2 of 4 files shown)


server/pxf-service/src/templates/user/templates/jdbc-site.xml
```diff
@@ -34,7 +34,7 @@
 
     <property>
         <name>jdbc.pool.property.maximumPoolSize</name>
-        <value>5</value>
+        <value>15</value>
```
denalex, line 37: this will not take effect on pre-existing RPM installations after upgrade. we should release doc this, @lisakowen fyi

denalex (approved):

server/pxf-service/src/templates/user/templates/jdbc-site.xml
frankgh: yes, anyone that has already configured the JDBC connector using the `jdbc-site.xml` will either manually have to modify their `jdbc.pool.property.maximumPoolSize` value to 15 in case they want to have the defaults that PXF provides

lisakowen: thanks for calling this out.  i was going to ask about upgrade implication for all of the *-xite.xml changes in 5.16:  the commenting out of pxf.service.user.name in pxf-site.xml, the new pxf.fs.basepath property in pxf-site.xml, and this.

yes, we will release note each of these changes.  and upgrade should probably identify each and provide info to the user as to if/when they might want to change their current server definitions.

denalex: this will NOT impact not only `anyone that has already configured the JDBC connector`, but everyone who already had a previous version installed (already ran `pxf init`, as we copy the templates to PXF_CONF/templates).

frankgh: Actually, we [overwrite templates](https://github.com/greenplum-db/pxf/blob/master/server/pxf-service/src/scripts/pxf#L261) during `init` inside `$PXF_CONF/templates`, and you are required to re-init anyway. So this will not affect people that already had their templates inside PXF_CONF/templates because we have to overwrite them anyway.

PR #464, opened 2020-10-15
Remove old CI script and pipelines

frankgh:
The deploy script was replaced vi the Makefile. Remove the deploy script
and pipelines that are no longer used


ashuka24 (approved):
LGTM. Can we also archive/clean out the corresponding pipelines in our CI?

frankgh:
> LGTM. Can we also archive/clean out the corresponding pipelines in our CI?

I think we can start cleaning out old pipelines

PR #465, opened 2020-10-15
Gpdb7 docker images

frankgh:
Prior to the RPM work, we had the `pg_regress` pipeline to build and test FDW. Since that time, the pipeline has been broken. This is the first step to restore the functionality we had for FDW.

### DIFF
--- a/concourse/docker/pxf-dev-base/README.md
+++ b/concourse/docker/pxf-dev-base/README.md
@@ -4,6 +4,8 @@ The generated images are the base images for building and testing PXF.
 This guide assumes the PXF repository lives under the `~/workspace/pxf`
 directory. The `cloudbuild.yaml` file produces the following docker images:
 
+## Greenplum 5 Images
+
 ### Docker gpdb5-centos6-test-pxf-image image
 
 Build this image for Greenplum 5 running on CentOS 6. Run the following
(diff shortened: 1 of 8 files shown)


concourse/docker/pxf-dev-base/README.md
```diff
@@ -68,3 +72,31 @@ following command to build the image:
...
+Build this image for Greenplum 7 running on Ubuntu 18.04. Run the following
+command to build the image:
+
+    pushd ~/workspace/pxf/concourse/docker/pxf-dev-base/
+    docker build \
+      --build-arg=BASE_IMAGE=pivotaldata/gpdb7-ubuntu18.04-test:latest \
+      --tag=gpdb7-ubuntu18.04-test-pxf \
+      -f ~/workspace/pxf/concourse/docker/pxf-dev-base/gpdb7/ubuntu18.04/Dockerfile \
+      .
+    popd
```
denalex, line 102: why do we have these commands here if we have cloudbuild to build these images ?

denalex (approved):

concourse/docker/pxf-dev-base/README.md
frankgh: well in case anyone wants to build them locally. I actually build them locally using this commands occasionally

denalex: then maybe first give a `docker pull gcr.io/...` command to pull from GCR and mention that if you want to build locally instead, use the following. The danger of local build is that you might sendup developing / testing on smith that is not used in CI

frankgh: sure, let me do that

PR #469, opened 2020-10-22
Specify Hive schema column names and types in HiveAccessor when creating RecordReaders

denalex:
When PXF accesses data via Hive profile, for example Hive tables stored as Parquet files, we need to explicitly specify the Hive schema column names and types in HiveAccessor when requesting a RecordReader so that ParquetRecordReader presents data to `HiveParquetSerDe` in the order of the Hive schema definition that is default order for Hive SerDes.


frankgh (approved):
Verified on my local, it also works correctly with the test case we had originally.

PR #470, opened 2020-10-22
Add Support for reading ORC without Hive

frankgh:
Support reading ORC files directly from blob storage without accessing
the Hive metastore.


server/pxf-api/src/main/java/org/greenplum/pxf/api/filter/InOperatorTransformer.java
```diff
@@ -0,0 +1,86 @@
...
+        return node;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public Node visit(Node node, int level) {
+
+        if (node instanceof OperatorNode) {
```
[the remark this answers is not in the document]
frankgh: I originally had the code like that, but then other implementations of `TreeVisitor` follow this pattern. So, I preferred to keep it like that for consistency across the code. I also prefer early returns.

server/pxf-api/src/main/java/org/greenplum/pxf/api/filter/InOperatorTransformer.java
```diff
@@ -0,0 +1,86 @@
...
+                    && operatorNode.getRight() instanceof CollectionOperandNode) {
+
+                ColumnIndexOperandNode columnNode = (ColumnIndexOperandNode) operatorNode.getLeft();
+                CollectionOperandNode collectionOperandNode = (CollectionOperandNode) operatorNode.getRight();
+                List<String> data = collectionOperandNode.getData();
+                DataType type = collectionOperandNode.getDataType().getTypeElem() != null
+                        ? collectionOperandNode.getDataType().getTypeElem()
+                        : collectionOperandNode.getDataType();
+
+                // Transform the IN operator into a chain of ORs
```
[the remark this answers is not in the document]
frankgh: data size can be at least 1 up to n. When the size is 1 then we just transform in to eq, otherwise we return a chain of ORs

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/filter/BPCharOperatorTransformer.java
```diff
@@ -0,0 +1,86 @@
...
+            if (!operator.isLogical()
+                    && operatorNode.getLeft() instanceof ColumnIndexOperandNode
+                    && operatorNode.getRight() instanceof ScalarOperandNode) {
+                ScalarOperandNode scalarOperandNode = (ScalarOperandNode) operatorNode.getRight();
+
+                if (scalarOperandNode.getDataType() == DataType.BPCHAR) {
+                    String value = scalarOperandNode.getValue();
+
+                    /* Determine whether the string has whitespace at the end */
+                    if (value.length() > 0 && value.charAt(value.length() - 1) == ' ') {
```
denalex, line 40: don't we need to do this multiple times if there are multiple blank space paddings ?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedAccessor.java
```diff
@@ -0,0 +1,250 @@
...
+        // Get the record filter in case of predicate push-down
+        SearchArgument searchArgument = getSearchArgument(context.getFilterString(), schema);
+
+        // Build the reader options
+        Reader.Options options = fileReader
+                .options()
+                .schema(readSchema)
+                .positionalEvolutionLevel(0)
+                .range(fileSplit.getStart(), fileSplit.getLength())
+                .searchArgument(searchArgument, new String[]{});
```
denalex, line 105: so I wonder why HiveVectorizedORC is not using filters, just omission ?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedAccessor.java
```diff
@@ -0,0 +1,250 @@
...
+     * Reads the next batch for the current fragment
+     *
+     * @return the next batch in OneRow format, the key is the batch number, and data is the batch
+     * @throws IOException when reading of the next batch occurs
+     */
+    @Override
+    public OneRow readNextObject() throws IOException {
+        final Instant start = Instant.now();
+        final boolean hasNextBatch = recordReader.nextBatch(batch);
+        totalReadTimeInNanos += Duration.between(start, Instant.now()).toNanos();
```
denalex, line 124: not sure how valuable timekeeping is here, as the reader might not actually read the whole batch in memory in `nextBatch()` or will it ?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedAccessor.java
```diff
@@ -0,0 +1,250 @@
...
+    }
+
+    /**
+     * Given the column descriptors that we receive from Greenplum, builds
+     * the read schema that will perform column projection
+     *
+     * @param originalSchema the original schema for the ORC file
+     * @return the read schema
+     */
+    private TypeDescription buildReadSchema(TypeDescription originalSchema) {
```
denalex, line 213: so, will we support the cases where GP has more columns than ORC and will fill with NULL and the other one where ORC has more columns and we will ignore them ? What about when diff ORC files have a bit different schemas ?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedMappingFunctions.java
```diff
@@ -0,0 +1,286 @@
...
+        return result;
+    }
+
+    public static OneField[] floatMapper(VectorizedRowBatch batch, ColumnVector columnVector, int oid) {
+        DoubleColumnVector dcv = (DoubleColumnVector) columnVector;
+        if (dcv == null)
+            return getNullResultSet(oid, batch.size);
+
+        OneField[] result = new OneField[batch.size];
+        int m = dcv.isRepeating ? 0 : 1;
```
denalex, line 131: repeating means all the values of this column in a batch are the same ?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedMappingFunctions.java
```diff
@@ -0,0 +1,286 @@
...
+        int rowId;
+        Double value;
+        for (int rowIndex = 0; rowIndex < batch.size; rowIndex++) {
+            rowId = m * rowIndex;
+            value = (dcv.noNulls || !dcv.isNull[rowId])
+                    ? dcv.vector[rowId]
+                    : null;
+            result[rowIndex] = new OneField(oid, value);
+        }
+        return result;
```
denalex, line 160: the body of these methods is all the same, other than types, would generics work here: `Mapper<T>`, `DoubleMapper extends Mapper<Double>`, etc ?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedMappingFunctions.java
```diff
@@ -0,0 +1,286 @@
...
+    }
+
+    public static OneField[] getNullResultSet(int oid, int size) {
+        OneField[] result = new OneField[size];
+        for (int i = 0; i < result.length; i++)
+            result[i] = new OneField(oid, null);
```
denalex, line 267: we can pre-define OneField(oid, null) as singletons for a given oid and then use `Arrays.fill` to fill the array. That will save a lot of object creation on null values.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedResolver.java
```diff
@@ -0,0 +1,272 @@
...
+        int batchSize = vectorizedBatch.size;
+
+        // The resolved batch returns a list of the list of OneField that
+        // matches the size of the batch. Every internal list, has a list of
+        // OneFields with size the number of columns
+        List<List<OneField>> resolvedBatch = new ArrayList<>(batchSize);
+
+        // Initialize the internal lists
+        for (int i = 0; i < batchSize; i++) {
+            resolvedBatch.add(new ArrayList<>(columnDescriptors.size()));
```
denalex, line 126: wonder if we can reuse these between batches

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedResolver.java
```diff
@@ -0,0 +1,272 @@
...
+                } else {
+                    throw new UnsupportedTypeException(
+                            String.format("Unable to resolve column '%s' with category '%s'. Only primitive types are supported.",
+                                    readSchema.getFieldNames().get(columnIndex), orcColumn.getCategory()));
+                }
+            }
+
+            // oneFields is the array of fields for the current column we are
+            // processing. We need to add it to the corresponding list
+            for (int i = 0; i < batchSize; i++) {
```
denalex, line 163:
```suggestion
            for (int row = 0; row < batchSize; row++) {
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedResolver.java
```diff
@@ -0,0 +1,272 @@
...
+                resolvedBatch.get(i).add(oneFields[i]);
```
denalex, line 164:
```suggestion
                resolvedBatch.get(row).add(oneFieldsForRow[row]);
```

server/pxf-hive/src/main/java/org/apache/hadoop/hive/ql/io/orc/PxfRecordReaderImpl.java
```diff
@@ -0,0 +1,50 @@
...
+
+import org.apache.hadoop.hive.ql.exec.vector.ColumnVector;
+import org.apache.hadoop.hive.ql.exec.vector.DateColumnVector;
+import org.apache.orc.Reader;
+import org.apache.orc.TypeDescription;
+
+import java.io.IOException;
+
+/**
+ * This class fixes an issue introduced in ORC core library version 1.5.9
```
denalex, line 11: will this be fixed in ORC, is there a JIRA to refer to ?

server/pxf-service/src/templates/conf/pxf-log4j2.xml
```diff
@@ -34,6 +34,7 @@
     <Logger name="com.google.cloud.hadoop.fs.gcs.GoogleHadoopFileSystemBase" level="warn"/>
     <Logger name="org.apache.hadoop" level="warn"/>
     <Logger name="org.apache.parquet" level="warn"/>
+    <Logger name="org.apache.orc" level="error"/>
```
denalex, line 37: some excessive WARN messages during normal use ?

denalex (approved):

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedAccessor.java
frankgh: most likely, good question. I guess we can support it.

server/pxf-service/src/templates/conf/pxf-log4j2.xml
frankgh: yes, I had to quiet down the log messages, otherwise my log was getting filled up with orc warn messages

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedAccessor.java
frankgh: we have tests where ORC have different schemas, so we support both GP has more columns than ORC and we fill with null: https://github.com/greenplum-db/pxf/pull/470/files#diff-f4945312d8b51ad817a8d7882795a2e32575ac346d2d8567a6dda9f4a4334a1dR74 
We also support reading a subset of columns from an ORC file (column projection).

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedMappingFunctions.java
frankgh: correct, this is from java docs
> True if same value repeats for whole column vector.
> If so, vector[0] holds the repeating value.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedMappingFunctions.java
frankgh: I didn't want to address this here, I wanted to solve this at the refactor level, where we determine what type we need for the wire format.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedMappingFunctions.java
frankgh: good idea!

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedResolver.java
frankgh: good point

server/pxf-hive/src/main/java/org/apache/hadoop/hive/ql/io/orc/PxfRecordReaderImpl.java
frankgh: as the time of this fix, there's no jira. But I assume that this will be addressed at some point since newer ORC libraries break Hive functionality

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetFileAccessor.java
```diff
@@ -322,16 +315,17 @@ public void closeForWrite() throws IOException, InterruptedException {
         List<ColumnDescriptor> tupleDescription = context.getTupleDescription();
         ParquetRecordFilterBuilder filterBuilder = new ParquetRecordFilterBuilder(
                 tupleDescription, originalFieldsMap);
-        TreeVisitor pruner = new ParquetOperatorPrunerAndTransformer(
+        TreeVisitor pruner = new ParquetOperatorPruner(
```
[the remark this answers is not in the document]
frankgh: I will open a separate PR for the changes here

server/pxf-hdfs/build.gradle
```diff
@@ -62,6 +62,13 @@ dependencies {
     implementation("com.fasterxml.woodstox:woodstox-core:5.0.3") { transitive = false }
     implementation("org.codehaus.woodstox:stax2-api:3.1.4")      { transitive = false }
 
+    // ORC dependencies
+    implementation("org.apache.orc:orc-core:${orcVersion}")      { transitive = false }
+    implementation("org.apache.orc:orc-shims:${orcVersion}")     { transitive = false }
+    implementation("org.apache.hive:hive-common:${hiveVersion}") { transitive = false }
+    implementation("com.google.protobuf:protobuf-java:2.5.0")    { transitive = false }
```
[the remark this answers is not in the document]
frankgh: where do you suggest we pin it?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedAccessor.java
frankgh: there are two cases, when the stripe needs to be read, and when the stripe is in memory. Depending on the case, timing will vary, but I think it's still worth keeping track. What would an alternative be for this?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetFileAccessor.java
```diff
@@ -322,16 +315,17 @@ public void closeForWrite() throws IOException, InterruptedException {
...
         try {
             // Parse the filter string into a expression tree Node
             Node root = new FilterParser().parse(filterString);
-            // Prune the parsed tree with valid supported operators and then
+            // Transform IN operators into a chain of ORs, then
+            // prune the parsed tree with valid supported operators and then
             // traverse the pruned tree with the ParquetRecordFilterBuilder to
             // produce a record filter for parquet
-            TRAVERSER.traverse(root, pruner, filterBuilder);
+            TRAVERSER.traverse(root, IN_OPERATOR_TRANSFORMER, pruner, BPCHAR_TRANSFORMER, filterBuilder);
```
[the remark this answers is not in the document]
frankgh: The logic in `ParquetOperatorPrunerAndTransformer` was split into `ParquetOperatorPruner` and `BPCharOperatorTransformer`. This effectively preserves the same logic as before. We need the BPCHAR transformer here because, depending on the implementation, the char fields in parquet files will either be padded to the width of the field or the second case is when the char field is trimmed. So for that reason we need `BPCHAR_TRANSFORMER`

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/filter/BPCharOperatorTransformer.java
frankgh: I have addressed this in a separate PR.

PR #476, opened 2020-11-02
RPM: Install the $PXF_HOME/run directory during %posttran

frankgh:
PXF v5 RPM installation removes the run directory during the %preun
step. The lack of run directory prevents PXF v6+ from starting up. We
fix the issue by installing the `$PXF_HOME/run` directory during the
%posttran step. %posttrans of the new package is the only step that runs
after the %preun of the old package.

### DIFF
--- a/package/pxf-gp5.spec
+++ b/package/pxf-gp5.spec
@@ -61,7 +61,7 @@ sed -i "s|module_pathname =.*|module_pathname = '%{prefix}/gpextable/pxf'|g" %{p
 %config(noreplace) %{prefix}/conf/pxf-log4j2.xml
 %config(noreplace) %{prefix}/conf/pxf-profiles.xml
 
-%preun
+%pre
 # cleanup files and directories created by 'pxf init' command
 # only applies for old installations (pre 6.0.0)
 %__rm -f %{prefix}/conf/pxf-private.classpath
--- a/package/pxf-gp5.spec
+++ b/package/pxf-gp5.spec
@@ -66,3 +66,10 @@ sed -i "s|module_pathname =.*|module_pathname = '%{prefix}/gpextable/pxf'|g" %{p
 # only applies for old installations (pre 6.0.0)
 %__rm -f %{prefix}/conf/pxf-private.classpath
 %__rm -rf %{prefix}/pxf-service
+
+%posttrans
+# PXF v5 RPM installation removes the run directory during the %preun step.
+# The lack of run directory prevents PXF v6+ from starting up.
+# %posttrans of the new package is the only step that runs after the %preun
+# of the old package
+%{__install} -d -m 700 %{prefix}/run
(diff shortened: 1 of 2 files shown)


package/pxf-gp5.spec
```diff
@@ -66,3 +66,10 @@ sed -i "s|module_pathname =.*|module_pathname = '%{prefix}/gpextable/pxf'|g" %{p
 # only applies for old installations (pre 6.0.0)
 %__rm -f %{prefix}/conf/pxf-private.classpath
 %__rm -rf %{prefix}/pxf-service
+
+%posttrans
+# PXF v5 RPM installation removes the run directory during the %preun step.
+# The lack of run directory prevents PXF v6+ from starting up.
+# %posttrans of the new package is the only step that runs after the %preun
+# of the old package
+%{__install} -d -m 700 %{prefix}/run
```
denalex, line 75: is this directory now also provisioned as part of installation of this RPM ? What happens on a clean install ?

denalex, line 75: also, lines 64-68 -- `%preun` seems to be only for old stuff, should this be now in `%prein` or `%postin` to clean up before installing new RPM ?

denalex (commented):
Is there anything similar to do for Debian ?

package/pxf-gp5.spec
bradfordb-vmware: Is there a reason we're creating this directory in a scriptlet? If this directory is needed by pxf, should we have it be be created in the build step above? This will ensure that the package manager sees this directory as being owned by the package.

I don't know if PXF supports SLES11 (or plans to), but I don't think `%posttrans` works with the version of RPM there.

frankgh: yes, makes sense

frankgh: @denalex yes, all the directories we need for PXF to start are provided as part of the RPM installation. That includes the `run` directory. On a clean install, the `run` directory is created.

frankgh:
@denalex there's no issue in Debian. This seems to be an RPM specific issue.

package/pxf-gp5.spec
denalex: even if we create it as part of the spec, chichis what @bradfordb-vmware surest and what we should do, the current PX5 `%postun` will remove it, so we need to re-create it in the scriptlet again, unless there's a better way.

frankgh: spoke to @bradfordb-vmware and we agreed that the `%posttrans` approach should be fine, so long SLES11 is not something we need to worry about. To answer Bradford's other question, the RPM includes the `run` directory, so this should be seen by the package manager as well. We verified that with this fix, running an `rpm -V pxf-gp6` did not show the `run` directory as missing. The `run` directory was shown as missing in the RPM created in the spring-boot branch

denalex (approved):

PR #478, opened 2020-11-03
docs - add info about new init --skip-register option

lisakowen:
docs changes for https://github.com/greenplum-db/pxf/pull/477

### DIFF
--- a/docs/content/ref/pxf.html.md.erb
+++ b/docs/content/ref/pxf.html.md.erb
@@ -47,7 +47,7 @@ You can copy the extension files from a PXF installation on the host to the Gree
 <dd>Display the `pxf` management utility help message and then exit.</dd>
 
 <dt>init</dt>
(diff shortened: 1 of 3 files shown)


frankgh (approved):
Looks good to me! Thanks Lisa

dyozie (approved):

docs/content/ref/pxf.html.md.erb
```diff
@@ -75,11 +75,14 @@ You can copy the extension files from a PXF installation on the host to the Gree
 
 ## <a id="options"></a>Options
 
-The `pxf init` command takes the following option:
+The `pxf init` command takes the following options:
 
 <dt>&#8211;y </dt>
 <dd>Do not prompt, use the default `$PXF_CONF` directory location if the environment variable is not set.</dd>
 
+<dt>&#8211;&#8211;skip-register </dt>
```
denalex, line 83: this is a bit inconsistent `-y` and `--skip-register`. I believe `-s` will also work, should we just document the short one ?

frankgh: well, we have `-d` and `--delete` for sync. I think it makes sense in that context

frankgh: I prefer `--skip-register` since it's self explanatory

denalex (approved):

PR #479, opened 2020-11-03
Add PXF CLI test to multinode test

frankgh:
After the RPM work was merged, the CLI test pipeline stopped working.
This PR restores the CLI tests, but it adds the test as a task in the
multinode impersonation job. This way, it reuses the setup, which is
expensive, and then the automation tests run after the CLI tests are
successful.


concourse/scripts/cli/common.sh
```diff
@@ -1,8 +1,13 @@
 #!/usr/bin/env bash
 
 export PXF_CONF=~gpadmin/pxf
-GPHOME=/usr/local/greenplum-db-devel
-export PATH=$PATH:${GPHOME}/pxf/bin
+export GPHOME=/usr/local/greenplum-db
+export PXF_HOME=$(find /usr/local/ -name "pxf-gp*" -type d)
```
bradfordb-vmware, line 5: I don't think we need to use `find` here, wouldn't `export PXF_HOME="/usr/local/pxf-gp*" work? If you wanted to be really defensive, you could use glob like `/usr/local/pxf-gp?` or `/usr/local/pxf-gp[56]`.

bradfordb-vmware (approved):

concourse/pipelines/templates/build_pipeline-tpl.yml
```diff
@@ -1189,6 +1189,7 @@ jobs:
             segments_per_host: 4
             instance_type: n1-standard-4
             ccp_reap_minutes: 120
+            standby_master: true
```
ashuka24, line 1192: why do we need the standby master?

ashuka24 (approved):

concourse/pipelines/templates/build_pipeline-tpl.yml
frankgh: The PXF CLI tests make sure that standby master is also being tested

concourse/scripts/cli/common.sh
```diff
@@ -1,8 +1,13 @@
 #!/usr/bin/env bash
 
 export PXF_CONF=~gpadmin/pxf
-GPHOME=/usr/local/greenplum-db-devel
-export PATH=$PATH:${GPHOME}/pxf/bin
+export GPHOME=/usr/local/greenplum-db
+export PXF_HOME=/usr/local/pxf-gp?
```
denalex, line 5: is `?` supposed to be there ?

concourse/pipelines/templates/dev_build_pipeline-tpl.yml
```diff
@@ -1074,6 +1075,12 @@ jobs:
       GP_VER: [[gp_ver]]
       KERBEROS: ((kerberos-enabled))
       PXF_JVM_OPTS: ((pxf-jvm-opts))
+  - task: Test PXF CLI
+    image: gpdb[[gp_ver]]-pxf-dev-centos7-image
+    file: pxf_src/concourse/tasks/test_pxf_cli.yml
+    params:
+      TARGET_OS: centos
+      TARGET_OS_VERSION: 7
```
denalex, line 1083: does this assume CLI tests leave PXF installation / configuration in a correct state for the downstream feature suite to work ? If yes, that would be a burden for future CLI test developers not aware of this coupling. If not, should the feature suite cleanup everything before starting ?

frankgh: that's a concern I have in the back of my mind. I was considering moving this task to run after the automation test task. That way we don't need to worry about the state of PXF before running the automation tests

concourse/scripts/cli/common.sh
frankgh: it turns out glob expansion doesn't work when defining env variables

concourse/pipelines/templates/dev_build_pipeline-tpl.yml
frankgh: ok, moved the test to run as the last task.

concourse/scripts/cli/common.sh
frankgh: I couldn't get globs working here. Reverted back to find

denalex (approved):

PR #480, opened 2020-11-03
Enable predicate pushdown for Hive profile when accessing Parquet backed tables

denalex:
When PXF accessed data managed via Hive and accessed via Hive profile, predicate pushdown only worked for RC/ORC backed tables. This PR enables predicate pushdown to be used for Parquet-backed tables. Not all operators and datatypes are supported with Hive-Parquet, but those that are supported can bring significant performance improvements.


server/pxf-api/src/main/java/org/greenplum/pxf/api/filter/BaseTreePruner.java
```diff
@@ -0,0 +1,55 @@
...
+    @Override
+    public Node after(Node node, int level) {
+        if (node instanceof OperatorNode) {
+            OperatorNode operatorNode = (OperatorNode) node;
+            Operator operator = operatorNode.getOperator();
+            int childCount = operatorNode.childCount();
+            if (AND == operator && childCount == 1) {
+                Node promoted = Optional.ofNullable(operatorNode.getLeft()).orElse(operatorNode.getRight());
+                LOG.debug("Child {} was promoted higher in the tree", promoted);
+                // AND need at least two children. If the operator has a
```
frankgh, line 37: I realize I wrote this:
```suggestion
                // AND needs at least two children. If the operator has a
```

server/pxf-api/src/main/java/org/greenplum/pxf/api/filter/BaseTreePruner.java
```diff
@@ -0,0 +1,55 @@
...
+                // single child node left, we promote the child one level up
+                // the tree
+                return promoted;
+            } else if (OR == operator && childCount <= 1) {
+                LOG.debug("Child with operator {} will be pruned because it has {} children", operator, childCount);
+                operatorNode.setLeft(null);
+                operatorNode.setRight(null);
+                // OR need two or more children
```
frankgh, line 45:
```suggestion
                // OR needs two or more children
```

server/pxf-api/src/main/java/org/greenplum/pxf/api/filter/SupportedDataTypePruner.java
```diff
@@ -0,0 +1,66 @@
+package org.greenplum.pxf.api.filter;
+
+import org.greenplum.pxf.api.io.DataType;
+import org.greenplum.pxf.api.utilities.ColumnDescriptor;
+
+import java.util.EnumSet;
+import java.util.List;
+
+/**
+ * A tree pruner that removes operator nodes for non-supported column data types.
```
frankgh, line 10: these are Greenplum data types
```suggestion
 * A tree pruner that removes operator nodes for non-supported Greenplum column data types.
```

server/pxf-api/src/main/java/org/greenplum/pxf/api/filter/SupportedDataTypePruner.java
```diff
@@ -0,0 +1,66 @@
...
+
+    private final List<ColumnDescriptor> columnDescriptors;
+    private final EnumSet<DataType> supportedDataTypes;
+
+    /**
+     * Constructor, assumes all possible predicate pushdown datatypes will be supported
+     *
+     * @param columnDescriptors  the list of column descriptors for the table
+     */
+    public SupportedDataTypePruner(List<ColumnDescriptor> columnDescriptors) {
```
frankgh, line 22: I wonder if we need this constructor. Why have a pruner that won't prune any types. It seems like if we have such pruner, it will do unnecessary work

server/pxf-api/src/main/java/org/greenplum/pxf/api/filter/SupportedDataTypePruner.java
```diff
@@ -0,0 +1,66 @@
...
+     */
+    @Override
+    public Node visit(Node node, int level) {
+        if (node instanceof OperatorNode) {
+            OperatorNode operatorNode = (OperatorNode) node;
+            if (!operatorNode.getOperator().isLogical()) {
+                DataType datatype = getColumnType(operatorNode);
+                if (!supportedDataTypes.contains(datatype)) {
+                    // prune the operator node if its operand is a column of unsupported type
+                    LOG.debug("DataType oid={} is not supported", datatype.getOID());
```
frankgh, line 49: can we also log the name here? mostly for convenience when debugging

server/pxf-hive/src/main/java/org/greenplum/pxf/plugins/hive/HiveAccessor.java
```diff
@@ -84,13 +92,91 @@
...
+            EnumSet.of(
+                    Operator.NOOP,
+                    Operator.LESS_THAN,
+                    Operator.GREATER_THAN,
+                    Operator.LESS_THAN_OR_EQUAL,
+                    Operator.GREATER_THAN_OR_EQUAL,
+                    Operator.EQUALS,
+                    Operator.NOT_EQUALS,
+                    //Operator.IS_NULL,
+                    //Operator.IS_NOT_NULL,
```
frankgh, line 119: Should we remove these? I personally like keeping them here commented out because we can quickly see what's supported and what's not.

server/pxf-hive/src/main/java/org/greenplum/pxf/plugins/hive/HiveAccessor.java
```diff
@@ -123,8 +209,29 @@ public HiveAccessor() {
      */
     @Override
     public void initialize(RequestContext context) {
+
         super.initialize(context);
-        HiveMetadata metadata;
+
+        // determine if predicate pushdown is allowed by configuration
+        if (configuration.get(PXF_PPD_HIVE, "true").equalsIgnoreCase("true")) {
+            isPredicatePushdownAllowed = true;
```
frankgh, line 217: Should we explicitly set the flag to prevent any second guessing here?
```suggestion
        isPredicatePushdownAllowed = configuration.get(PXF_PPD_HIVE, "true").equalsIgnoreCase("true");
        if (isPredicatePushdownAllowed) {
```

server/pxf-hive/src/main/java/org/greenplum/pxf/plugins/hive/HiveAccessor.java
```diff
@@ -498,6 +609,61 @@ protected void addColumns() {
...
+
+        // Parse the filter string into a expression tree Node
+        String filterStr = context.getFilterString();
+        Node root = new FilterParser().parse(filterStr);
+
+        // Prune the parsed tree with valid supported datatypes and operators and then
+        // traverse the pruned tree with the searchArgumentBuilder to produce a SearchArgument
+        TRAVERSER.traverse(
+                root,
+                new SupportedDataTypePruner(context.getTupleDescription(), getSupportedDatatypesForPushdown()),
```
frankgh, line 634: it looks like parquet is our lower bound in terms of supported datatype and operators. I wonder if we can do something better in the future to determine the underlying file types (or maybe we can choose based on the input format). We don't need to address it here.

server/pxf-hive/src/main/java/org/greenplum/pxf/plugins/hive/HiveAccessor.java
```diff
@@ -498,6 +609,61 @@ protected void addColumns() {
...
+                new SupportedOperatorPruner(getSupportedOperatorsForPushdown()),
+                searchArgumentBuilder);
+
+        String kryoString = toKryo(searchArgumentBuilder.getFilterBuilder().build());
```
frankgh, line 638: I wonder if we should generalize `HiveUtilities#serializeProperties` and then use that method to get the `byte[]` resulting from serialization. We would need a better name for `serializeProperties`
```suggestion
        String kryoString = Base64.encodeBase64String(HiveUtilities.serializeProperties(searchArgumentBuilder.getFilterBuilder().build()));
```

server/pxf-hive/src/main/java/org/greenplum/pxf/plugins/hive/HiveSearchArgumentBuilder.java
```diff
@@ -46,14 +46,14 @@
  * ..endNot
  * endAnd
  */
-public class HiveORCSearchArgumentBuilder implements TreeVisitor {
```
frankgh, line 49: this will change when ORC lands

server/pxf-api/src/test/java/org/greenplum/pxf/api/filter/SupportedDataTypePrunerTest.java
```diff
@@ -0,0 +1,520 @@
...
+import org.greenplum.pxf.api.utilities.ColumnDescriptor;
+import org.junit.Test;
+
+import java.util.Arrays;
+import java.util.EnumSet;
+import java.util.List;
+
+import static org.junit.Assert.assertEquals;
+
+public class SupportedDataTypePrunerTest {
```
frankgh, line 13: good set of test cases!

server/pxf-hive/src/test/java/org/greenplum/pxf/plugins/hive/HiveParquetFilterPushDownTest.java
```diff
@@ -0,0 +1,919 @@
...
+
+    @Test
+    public void testInOperationFilter() throws Exception {
+        // a16 in (11, 12)
+        int[] expectedRows = {11, 12, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25};
+        context.setFilterString("a16m1007s2d11s2d12o10");
+        assertRowsReturned(expectedRows);
+    }
+
+    private Map<String, Type> getOriginalFieldsMap(MessageType originalSchema) {
```
frankgh, line 706: this seems to be unused

server/pxf-hive/src/test/java/org/greenplum/pxf/plugins/hive/HiveParquetFilterPushDownTest.java
```diff
@@ -0,0 +1,919 @@
...
+    private static final String[] COL13 = {"USD", "USD", "USD", "USD", "USD", "USD", "USD", "EUR", "UAH", "USD", "UAH", "EUR", "USD", "UAH", "USD", "USD", "EUR", "USD", "USD", "UAH", "USD", "EUR", "EUR", null, "USD"};
+    private static final String[] COL14 = {"1.23456", "1.23456", "-1.23456", "123456789.1", "1E-12", "1234.889", "0.0001", "45678.00002", "23457.1", "45678.00002", "0.123456789", "0.123456789", "0.123456789", "0.123456789", null, "0.123456789", "0.123456789", "0.123456789", "0.123456789", "0.123456789", "0.123456789", "0.123456789", "0.123456789", "0.123456789", "0.123456789"};
+    private static final String[] COL15 = {"0", "123.45", "-1.45", "0.25", "-0.25", "999.99", "-999.99", "1", "-1", "789", "-789", "0.99", "-0.99", "1.99", null, "-1.99", "15.99", "-15.99", "-299.99", "299.99", "555.55", "0.15", "3.89", "3.14", "8"};
+    private static final String[] COL16 = {"0.12345", "-0.12345", "12345678.90123", "-12345678.90123", "99999999", "-99999999", "-99999999.99999", "99999999.99999", "0", "1", "-1", "0.9", "-0.9", "45", null, "-45", "3.14159", "-3.14159", "2.71828", "-2.71828", "45.99999", "-45.99999", "450.45001", "0.00001", "-0.00001"};
```
frankgh, line 55: In the native Parquet reader, we expect Double, it looks like hive resolves to string for these columns?

frankgh (approved):
It looks good. I added some minor comments.

server/pxf-hive/src/main/java/org/greenplum/pxf/plugins/hive/HiveAccessor.java
denalex: I left them on purpose to see which ones are not supported from the overall set, will leave as is.

server/pxf-hive/src/main/java/org/greenplum/pxf/plugins/hive/HiveAccessor.java
denalex: yes, makes sense

server/pxf-hive/src/main/java/org/greenplum/pxf/plugins/hive/HiveAccessor.java
```diff
@@ -657,13 +659,6 @@ JobConf getJobConf() {
         return jobConf;
     }
 
-    private String toKryo(SearchArgument sarg) {
-        Output out = new Output(KRYO_BUFFER_SIZE, KRYO_MAX_BUFFER_SIZE);
```
frankgh: we no longer need lines 106-107

frankgh (approved):

server/pxf-hive/src/test/java/org/greenplum/pxf/plugins/hive/HiveParquetFilterPushDownTest.java
denalex: yes, those are numeric / decimals so are represented by a String. It is not correct to treat them as Double as that would cause precision loss, maybe we should review native Parquet for this ?

PR #482, opened 2020-11-05
docs - add upgrade steps for pxf v5.16

lisakowen:
some upgrade steps for the user.  see the 5th step in Step 2 section on doc review site.

doc review site link:  http://docs-lisa-pxfv516-relupg.cfapps.io/pxf/5-16/using/upgrade_pxf_rpm.html

question:  wasn't sure about the upgrade steps.  for Hive* profile changes...  not clear to me if this should be specific to Hive accessing Parquet


frankgh (approved):

docs/content/upgrade_pxf_rpm.html.md.erb
```diff
@@ -88,6 +88,13 @@ After you install the new version of PXF, perform the following procedure:
 
         3. Review any SQL scripts that you may have created that reference the PXF external table, and update column names if required.
 
+4. **If you are upgrading from PXF version 5.15.x or earlier**:
+
+    1. The `pxf.service.user.name` property in the `pxf-site.xml` template file is now commented out by default. Keep this in mind when you configure new PXF servers.
```
denalex, line 93: not sure what a user should do with `Keep this in mind` instruction. what would change in their workflow when they create new servers ?

docs/content/upgrade_pxf_rpm.html.md.erb
```diff
@@ -88,6 +88,13 @@ After you install the new version of PXF, perform the following procedure:
...
+    2. The default value for the `jdbc.pool.property.maximumPoolSize` property is now `15`. If you have previously configured a JDBC server(s) and want that server to use the new default value, you must manually change the property value in the server's `jdbc-site.xml` file.
+    3. PXF 5.16 disallows specifying relative paths and environment variables in the `CREATE EXTERNAL TABLE` `LOCATION` clause file path. If you previously created any external tables that specified a relative path or environment variable, you must drop each external table, and then re-create it without these constructs.
+    4. Filter pushdown is enabled by default for queries on external tables that specify the `Hive`, `HiveRC`, or `HiveORC` profiles. If you have previously created an external table that specifies one these profiles and queries are failing with PXF v5.16+, try disabling filter pushdown: drop the external table and re-create it, specifying the `&PPD=false` option in the `LOCATION` clause.
```
denalex, line 96: we should also advise them to change the property value in `pxf-site.xml` if they do not want to modify the tables.

docs/content/upgrade_pxf_rpm.html.md.erb
lisakowen: i was trying to call out that there is a difference in behaviour, but i guess that i don't know what that really is.   previously, the property value was `${user.name}`.  could the admin have been relying on that?  and now that the property is unset, they might need to change it?

denalex (approved):

docs/content/upgrade_pxf_rpm.html.md.erb
```diff
@@ -88,6 +88,25 @@ After you install the new version of PXF, perform the following procedure:
...
 
+4. **If you are upgrading from PXF version 5.15.x or earlier**:
+
+    1. The `pxf.service.user.name` property in the `pxf-site.xml` template file is now commented out by default. Keep this in mind when you configure new PXF servers.
+    2. The default value for the `jdbc.pool.property.maximumPoolSize` property is now `15`. If you have previously configured a JDBC server(s) and want that server to use the new default value, you must manually change the property value in the server's `jdbc-site.xml` file.
+    3. PXF 5.16 disallows specifying relative paths and environment variables in the `CREATE EXTERNAL TABLE` `LOCATION` clause file path. If you previously created any external tables that specified a relative path or environment variable, you must drop each external table, and then re-create it without these constructs.
+    4. Filter pushdown is enabled by default for queries on external tables that specify the `Hive`, `HiveRC`, or `HiveORC` profiles. If you have previously created an external table that specifies one these profiles and queries are failing with PXF v5.16+, you can disable filter pushdown at the external table-level or at the server level:
```
dyozie, line 96: one these -> one of these

dyozie (approved):

PR #486, opened 2020-11-11
Serialize Fragment metadata using Kryo

frankgh:
In PXF 5.16.0, we introduced Kryo to serialize Fragment's userData for
Hive profiles. The original version of boot, combines Fragment metadata
and userData into metadata. It also serializes metadata into a escaped
JSON, that is then deserialized by the HttpRequestParser during the
bridge call. This introduces a regression in the Hive metadata
optimization. To reduce the payload size, we now serialize metadata
using kryo instead of JSON. This commit fixes the regression introduced
by the original version of boot.


server/pxf-api/src/main/java/org/greenplum/pxf/api/utilities/FragmentMetadata.java
```diff
@@ -19,16 +19,9 @@
...
 package org.greenplum.pxf.api.utilities;
 
-import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
-
 /**
  * Interface that represents Fragment metadata. Each profile can implement it's
  * own metadata object
  */
-@JsonIgnoreProperties(value={ "className" }, allowGetters=true)
 public interface FragmentMetadata {
```
denalex, line 26: if this becomes a marker interface only, do we still need it ?

server/pxf-api/src/main/java/org/greenplum/pxf/api/utilities/SerializationService.java
```diff
@@ -0,0 +1,40 @@
...
+    public SerializationService() {
+        // A simple factory that creates kryo objects
+        KryoFactory factory = Kryo::new;
+        kryoPool = new KryoPool.Builder(factory).softReferences().build();
+    }
+
+    /**
+     * By default, kryo pool uses ConcurrentLinkedQueue which is unbounded. To facilitate reuse of
+     * kryo object call releaseKryo() after done using the kryo instance. The class loader for the
+     * kryo instance will be set to current thread's context class loader.
```
denalex, line 22: let's mention the soft references ensure the instances in the queue should get deleted when there's a GC memory pressure

server/pxf-api/src/main/java/org/greenplum/pxf/api/utilities/SerializationService.java
```diff
@@ -0,0 +1,40 @@
...
+        return kryo;
+    }
+
+    /**
+     * Release kryo instance back to the pool.
+     *
+     * @param kryo - kryo instance to be released
+     */
+    public void releaseKryo(Kryo kryo) {
+        kryoPool.release(kryo);
```
denalex, line 38: are we sure we don't need to reset the classloader back to what it was before ?

server/pxf-api/src/test/java/org/greenplum/pxf/api/utilities/FragmentMetadataSerDeTest.java
```diff
@@ -36,60 +35,48 @@ public void testSerialize() throws JsonProcessingException {
...
+        assertEquals("\"AQDPAW9yZy5ncmVlbnBsdW0ucHhmLmFwaS51dGlsaXRpZXMuRnJhZ21lbnRNZXRhZGF0YVNlckRlVGVzdCRUZXN0RnJhZ21lbnRNZXRhZGF0YQEBdGVz9AoUAQFqYXZhLnNxbC5EYXTlAYC70tClLgEEZm9v\"",
                 mapper.writeValueAsString(testMetadata));
     }
 
     @Test
-    public void testDeserialize() throws JsonProcessingException {
+    public void testDeserialize() {
 
-        String metadataJson = "{\"path\": \"deserialize me\", \"className\": \"org.greenplum.pxf.api.examples.DemoFragmentMetadata\" }";
+        String metadataJson = "\"AQBvcmcuZ3JlZW5wbHVtLnB4Zi5hcGkuZXhhbXBsZXMuRGVtb0ZyYWdtZW50TWV0YWRhdOEBAWFi4w==\"";
```
denalex, line 49: this is no longer a JSON (variable name) :)

server/pxf-hive/src/main/java/org/greenplum/pxf/plugins/hive/HiveAccessor.java
```diff
@@ -657,9 +656,17 @@ JobConf getJobConf() {
...
+    /**
+     * Serializes an object into a Base64 encoded String using Kryo serialization
+     *
+     * @param object the object to serialize
+     * @return the serialized object as a String
+     */
+    private String toKryoString(Object object) {
+        Output out = new Output(4 * 1024, 10 * 1024 * 1024);
+        hiveUtilities.getKryo().writeObject(out, object);
+        out.close();
```
denalex, line 669: do we need to release kryo here ?

server/pxf-hive/src/main/java/org/greenplum/pxf/plugins/hive/utilities/HiveUtilities.java
```diff
@@ -57,7 +56,6 @@
     // serializer is thread safe.
     private final ThreadLocal<Kryo> kryo = ThreadLocal.withInitial(() -> {
```
denalex, line 57: are we still using thread locals ? I thought we now use KryoPool ?

server/pxf-api/src/main/java/org/greenplum/pxf/api/utilities/FragmentMetadata.java
frankgh: yeah, this is the type we use in  `RequestContext` for fragment metadata

server/pxf-api/src/main/java/org/greenplum/pxf/api/utilities/SerializationService.java
frankgh: It's not necessary, the classloader will always be set when we borrow the kryo object. No need to have the kryo classloader set when they are in the pool waiting to be consumed

server/pxf-hive/src/main/java/org/greenplum/pxf/plugins/hive/utilities/HiveUtilities.java
frankgh: we should use Kryo from the pool here, I'll fix it

server/pxf-api/src/main/java/org/greenplum/pxf/api/utilities/FragmentMetadataSerDe.java
```diff
@@ -1,55 +1,56 @@
...
-    @SuppressWarnings({"unchecked", "rawtypes"})
-    public FragmentMetadata deserialize(String json) throws JsonProcessingException {
-        JsonNode node = mapper.readTree(json);
-        String className = node.get(CLASSNAME).textValue();
-
-        Class klass = getObjectClass(className);
-        return (FragmentMetadata) mapper.readValue(json, klass);
+    public void serialize(FragmentMetadata value, JsonGenerator gen, SerializerProvider provider)
+            throws IOException {
+        Output out = new Output(4 * 1024, 10 * 1024 * 1024);
```
bradfordb-vmware, line 34: Should we use named constants here? I am assuming this is 4kb and 10MB but I could be wrong.

server/pxf-api/src/main/java/org/greenplum/pxf/api/utilities/SerializationService.java
denalex: yea, I guess it would be needed if we were to unload the web app so that they can be GC-ed, but probably not that important for Boot, and maybe even setting the class loader in the first place is not necessary as all the class loaders will be the same, no ?

frankgh: maybe, but I think it's okay to keep it as is

server/pxf-hive/src/main/java/org/greenplum/pxf/plugins/hive/HiveAccessor.java
frankgh: done

denalex (approved):

server/pxf-api/src/main/java/org/greenplum/pxf/api/utilities/FragmentMetadataSerDe.java
```diff
@@ -1,55 +1,56 @@
...
+        Kryo kryo = serializationService.borrowKryo();
+        try {
+            kryo.writeClassAndObject(out, value);
+            out.close();
+            // Serialized fragment metadata is base64 encoded
```
bradfordb-vmware, line 39: Is the encoding done by the `writeClassAndObject()` call or the `writeBinary()` call? It seems odd that serialize doesn't have an explicit call to encode to Base64 but that deserialize does have a call to decode.

server/pxf-hbase/src/main/java/org/greenplum/pxf/plugins/hbase/HBaseFragmentMetadata.java
```diff
@@ -12,23 +11,20 @@
  * Fragment metadata for HBase profiles
  */
 @Getter
+@NoArgsConstructor
 public class HBaseFragmentMetadata implements FragmentMetadata {
 
-    private final byte[] startKey;
+    private byte[] startKey;
```
bradfordb-vmware, line 17: It doesn't look like the class as setters generated. Did we need to remove the `final` modifiers because of how kyro does serialization and deserialization?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HcfsFragmentMetadata.java
```diff
@@ -1,27 +1,24 @@
...
+import lombok.NoArgsConstructor;
 import org.apache.hadoop.mapred.FileSplit;
 import org.greenplum.pxf.api.utilities.FragmentMetadata;
 
+@NoArgsConstructor
 public class HcfsFragmentMetadata implements FragmentMetadata {
 
     @Getter
-    private final long start;
+    protected long start;
```
bradfordb-vmware, line 12: Echoing my comment from HBaseFragmentMetadata above.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/HdfsUtilities.java
```diff
@@ -53,18 +44,16 @@
...
      *
-     * @param context request input data
+     * @param file     the file name for the split
+     * @param metadata the fragment metadataa
      * @return FileSplit with fragment metadata
      */
-    public static FileSplit parseFileSplit(RequestContext context) {
-        HcfsFragmentMetadata metadata = context.getFragmentMetadata();
-
+    public static FileSplit parseFileSplit(String file, HcfsFragmentMetadata metadata) {
```
bradfordb-vmware, line 51: Context question: did some other code change require a change in this methods signature?

server/pxf-api/src/main/java/org/greenplum/pxf/api/utilities/FragmentMetadataSerDe.java
frankgh: I think we were using named constants, not sure at what point that changed

server/pxf-hbase/src/main/java/org/greenplum/pxf/plugins/hbase/HBaseFragmentMetadata.java
frankgh: we need to remove final because of the `@NoArgsConstructor` which is required by kryo

server/pxf-api/src/main/java/org/greenplum/pxf/api/utilities/FragmentMetadataSerDe.java
frankgh: encoding is done by jackson, so that means it's done in `writeBinary`

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/HdfsUtilities.java
frankgh: we only need two pieces of information from the RequestContext to build the FileSplit, so we just pass what we need.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HcfsFragmentMetadata.java
frankgh: again, we need it for the constructor with no arguments

PR #490, opened 2020-11-13
Update pxf-build-base dependency building after Spring Boot merge

frankgh:


concourse/docker/pxf-build-base/cloudbuild.yaml
```diff
@@ -5,40 +5,108 @@
...
-# Push the pxf-build-dependencies.tar.gz tarball to Google Cloud Storage
-- name: 'gcr.io/cloud-builders/gsutil'
-  id: pxf-build-dependencies-tarball
-  args: ['cp', '/workspace/build/pxf-build-dependencies.tar.gz', 'gs://${_PXF_BUILD_BUCKET}/build-dependencies/pxf-build-dependencies.tar.gz']
-
-# Push the pxf-automation-dependencies.tar.gz tarball to Google Cloud Storage
-- name: 'gcr.io/cloud-builders/gsutil'
-  id: pxf-automation-dependencies-tarball
-  args: ['cp', '/workspace/build/pxf-automation-dependencies.tar.gz', 'gs://${_PXF_BUILD_BUCKET}/automation-dependencies/pxf-automation-dependencies.tar.gz']
+  # download and extract the existing dependencies
```
denalex, line 8: how would we solve the problem of getting rid of obsolete versions / dependencies, if we always download existing ones to start with ? Not essential, just wonder if we can think of smth.

concourse/docker/pxf-build-base/cloudbuild.yaml
```diff
@@ -5,40 +5,108 @@
...
+  - name: 'gcr.io/$PROJECT_ID/tar'
+    id: untar-pxf-automation-dependencies
+    args: [ '-xzf', 'pxf-automation-dependencies.tar.gz' ]
+    waitFor: [ 'retrieve-pxf-automation-dependencies' ]
+
+  # use gradle image with JDK8 to build the project
+  - name: gradle:6.6.1-jdk8
+    id: gradle-build
+    entrypoint: gradle
+    args: [ '--gradle-user-home=/workspace/.gradle', '-Dorg.gradle.daemon=false', '-b', './server/build.gradle', 'test', 'stage' ]
```
denalex, line 31: I think this is ok for capturing dependencies, otherwise we probably should be using the Makefile to build the "official" product ?

denalex (approved):

concourse/docker/pxf-build-base/cloudbuild.yaml
```diff
@@ -5,40 +5,108 @@
...
+
+  - name: golang:1.15.3
+    id: golang-build
+    entrypoint: 'bash'
+    args:
+      - '-c'
+      - |
+        mkdir -p ./.go-dep-cached-sources
+        cp ./.go-dep-cached-sources ./cli/go/pkg/dep/sources
+        go get github.com/golang/dep/cmd/dep
```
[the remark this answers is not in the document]
frankgh: correct. we don't need `dep` after that PR comes in

concourse/docker/pxf-build-base/cloudbuild.yaml
frankgh: good point, something to think about

concourse/docker/pxf-build-base/cloudbuild.yaml
frankgh: correct, this is only to capture dependencies.

concourse/docker/pxf-build-base/cloudbuild.yaml
```diff
@@ -5,40 +5,108 @@
...
+  - name: 'gcr.io/$PROJECT_ID/tar'
+    id: untar-pxf-build-dependencies
+    args: [ '-xzf', 'pxf-build-dependencies.tar.gz' ]
+    waitFor: [ 'retrieve-pxf-build-dependencies' ]
+
+  - name: 'gcr.io/cloud-builders/gsutil'
+    id: retrieve-pxf-automation-dependencies
+    args: [ 'cp', 'gs://${_PXF_BUILD_BUCKET}/automation-dependencies/pxf-automation-dependencies.tar.gz', 'pxf-automation-dependencies.tar.gz' ]
+    waitFor: [ '-' ]
+  - name: 'gcr.io/$PROJECT_ID/tar'
```
bradfordb-vmware, line 22: Is this a one of image that we've built? Would it be better for us to use an upstream base image like alpine or bash? One less thing that we need to own.

concourse/docker/pxf-build-base/cloudbuild.yaml
```diff
@@ -5,40 +5,108 @@
...
+    waitFor: [ 'retrieve-pxf-automation-dependencies' ]
+
+  # use gradle image with JDK8 to build the project
+  - name: gradle:6.6.1-jdk8
+    id: gradle-build
+    entrypoint: gradle
+    args: [ '--gradle-user-home=/workspace/.gradle', '-Dorg.gradle.daemon=false', '-b', './server/build.gradle', 'test', 'stage' ]
+    waitFor: [ 'untar-pxf-build-dependencies' ]
+
+  - name: golang:1.15.3
```
bradfordb-vmware, line 34: It looks like we are also updating the version of golang used to build PXF cli. Is that intentional?

concourse/docker/pxf-build-base/cloudbuild.yaml
bradfordb-vmware: This looks like a bit of a bootstrap paradox -- we're using a tarball that is created by this cloudbuild. I know that this already existed from the previous workflow where it was created by scratch from the Dockerfile, but it took me a minute and looking at the diff to understand where it was coming from.

Maybe it's worth keeping the Dockerfile that creates the tarball from scratch to avoid this confusion and address Alex's point about old deps.

bradfordb-vmware (approved):

concourse/docker/pxf-build-base/cloudbuild.yaml
frankgh: this is an image that we built. I think Alex has context on it.

concourse/docker/pxf-build-base/cloudbuild.yaml
frankgh: we are actually matching the golang version we use to build: https://github.com/greenplum-db/pxf/blob/master/concourse/docker/pxf-dev-base/gpdb7/ubuntu18.04/Dockerfile#L7

concourse/docker/pxf-build-base/cloudbuild.yaml
frankgh: this is the pattern we use across different projects, I kept it consistent for now, but we should address the issue mentioned above at some point

ashuka24 (approved):

PR #491, opened 2020-11-13
Remove invalid GemFireXD profile

frankgh:


ashuka24 (approved):
Why is the profile now invalid? Or was it always invalid prior to the Springboot merge?

frankgh:
@ashuka24 we used to have a gemfire connector for PXF, but it was removed. Apparently, there are some remnants of that removal that were not taken care of.

denalex (approved):
wow !

PR #495, opened 2020-11-17
Hive profile names now split "protocol" and "format"

frankgh:
The hive profile names do not follow the "protocol":"format" naming
convention introduced for HCFS profiles. This commit adds Hive profile
names with the new format. For example, `HiveORC` becomes `hive:orc`.

### DIFF
--- a/automation/src/main/java/org/greenplum/pxf/automation/structures/tables/utils/TableFactory.java
+++ b/automation/src/main/java/org/greenplum/pxf/automation/structures/tables/utils/TableFactory.java
@@ -159,7 +159,7 @@ public abstract class TableFactory {
                                                                  HBaseTable hbaseTable) {
         ReadableExternalTable exTable = new ReadableExternalTable(tableName,
                 fields, hbaseTable.getName(), "CUSTOM");
-        exTable.setProfile("hbaase");
+        exTable.setProfile("hbase");
         exTable.setFormatter("pxfwritable_import");
         return exTable;
     }
@@ -398,7 +398,7 @@ public abstract class TableFactory {
             userParameters.add(customParameters);
         }
         exTable.setUserParameters(userParameters.toArray(new String[userParameters.size()]));
-        exTable.setProfile("Jdbc");
+        exTable.setProfile("jdbc");
         exTable.setFormatter("pxfwritable_import");
 
         return exTable;
@@ -434,7 +434,7 @@ public abstract class TableFactory {
             userParameters.add(customParameters);
         }
         exTable.setUserParameters(userParameters.toArray(new String[userParameters.size()]));
-        exTable.setProfile("Jdbc");
+        exTable.setProfile("jdbc");
         exTable.setFormatter("pxfwritable_export");
 
         return exTable;
--- a/automation/src/main/java/org/greenplum/pxf/automation/structures/tables/utils/TableFactory.java
+++ b/automation/src/main/java/org/greenplum/pxf/automation/structures/tables/utils/TableFactory.java
@@ -38,7 +38,7 @@ public abstract class TableFactory {
                 fields, hiveTable.getName(), "CUSTOM");
 
         if (useProfile) {
-            exTable.setProfile("Hive");
+            exTable.setProfile("hive");
         } else {
             exTable.setFragmenter("org.greenplum.pxf.plugins.hive.HiveDataFragmenter");
             exTable.setAccessor("org.greenplum.pxf.plugins.hive.HiveAccessor");
@@ -159,7 +159,7 @@ public abstract class TableFactory {
                                                                  HBaseTable hbaseTable) {
         ReadableExternalTable exTable = new ReadableExternalTable(tableName,
                 fields, hbaseTable.getName(), "CUSTOM");
-        exTable.setProfile("HBase");
+        exTable.setProfile("hbaase");
         exTable.setFormatter("pxfwritable_import");
         return exTable;
     }
--- a/server/pxf-hive/src/main/java/org/greenplum/pxf/plugins/hive/HiveProtocolHandler.java
+++ b/server/pxf-hive/src/main/java/org/greenplum/pxf/plugins/hive/HiveProtocolHandler.java
@@ -1,0 +1,48 @@
+package org.greenplum.pxf.plugins.hive;
+
+import org.greenplum.pxf.api.model.ProtocolHandler;
+import org.greenplum.pxf.api.model.RequestContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Implementation of ProtocolHandler for "hive" protocol.
+ */
+public class HiveProtocolHandler implements ProtocolHandler {
+
+    private static final Logger LOG = LoggerFactory.getLogger(HiveProtocolHandler.class);
+    private static final String HIVE_VECTORIZED_ORC_ACCESSOR = HiveORCVectorizedAccessor.class.getName();
+    private static final String HIVE_VECTORIZED_ORC_RESOLVER = HiveORCVectorizedResolver.class.getName();
+    private static final String OPTION_VECTORIZE = "VECTORIZE";
+
+    @Override
+    public String getFragmenterClassName(RequestContext context) {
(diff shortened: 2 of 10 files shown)


server/pxf-service/src/main/resources/pxf-profiles-default.xml
```diff
@@ -37,7 +37,7 @@ under the License.
         </plugins>
     </profile>
     <profile>
-        <name>Hive</name>
+        <name>hive</name>
```
lisakowen, line 40: looks like this replaces the `Hive` profile name with `hive`, and adds new `hive:xxx` profiles for the others?  will the user still be able to use `Hive`?

are we going to want to deprecate the old `Hive*` profile names in 6.0.0?

frankgh: This PR preserves the old names. The old names are cheap in terms of memory footprint, but we should at some point remove them.

denalex: Let's make them case-insensitive in the code and lower-cases in documentation / this file. This will apply to Hive, Jdbc, HBase, etc. The old one-word capitalized names will still work, but will be effectively deprecated. Composite ones like `HiveORC` will be `deprecated`, meaning they will still work for now, but will not be documented.

server/pxf-service/src/main/resources/pxf-profiles-default.xml
```diff
@@ -119,6 +171,19 @@ under the License.
             <outputFormat>org.greenplum.pxf.api.io.GPDBWritable</outputFormat>
         </plugins>
     </profile>
+    <profile>
+        <name>hive:vectorizedorc</name>
```
denalex, line 175: should we do `hive:orc:vectorized` ? Or better yet give it as a user option `VECTORIZE=true` since other profiles might also want to support vectorized execution or we can even auto-apply it if necessary ?

frankgh: the latter would require a code change, let's discuss further

server/pxf-service/src/main/resources/pxf-profiles-default.xml
```diff
@@ -73,6 +73,25 @@ under the License.
             <mapping option="ppd" property="pxf.ppd.hive"/>
         </optionMappings>
     </profile>
+    <profile>
+        <name>hive:rc</name>
+        <description>This profile is suitable only for Hive tables stored in RC files and serialized
+            with either the ColumnarSerDe or the LazyBinaryColumnarSerDe. It is much faster than the
+            general purpose Hive profile. DELIMITER parameter is mandatory. Supports both
+            GPDBWritable and TEXT output formats, as specified in FORMAT header parameter. Primary
```
ashuka24, line 81:
```suggestion
            GPDBWritable and TEXT output formats, as specified in FORMAT header parameter. Primarily
```

server/pxf-service/src/main/resources/pxf-profiles-default.xml
```diff
@@ -88,6 +107,21 @@ under the License.
             <outputFormat>org.greenplum.pxf.api.io.Text</outputFormat>
         </plugins>
     </profile>
+    <profile>
+        <name>hive:text</name>
+        <description>This profile is suitable only for Hive tables stored as Text files. It is much
+            faster than the general purpose Hive profile. DELIMITER parameter is mandatory. Supports
+            both GPDBWritable and TEXT output formats, as specified in FORMAT header parameter.
+            Primary optimized for TEXT output format.
```
ashuka24, line 115:
```suggestion
            Primarily optimized for TEXT output format.

File: server/pxf-service/src/main/resources/pxf-profiles-default.xml [hunk 5]
```diff
@@ -119,6 +171,19 @@ under the License.
             <outputFormat>org.greenplum.pxf.api.io.GPDBWritable</outputFormat>
         </plugins>
     </profile>
+    <profile>
+        <name>hive:vectorizedorc</name>
+        <description>This profile is same as HiveORC profile, but operates on batches of rows
```
ashuka24, line 176: shouldn't this be `hive:orc` now instead of using the `HiveORC` convention?

server/pxf-service/src/main/resources/pxf-profiles-default.xml
bradfordb-vmware: @denalex's comment make sense and I agree. A follow-up question -- should we log some of kind of warning/notice in addition to marking it deprecated in docs?

automation/src/main/java/org/greenplum/pxf/automation/structures/tables/utils/TableFactory.java
```diff
@@ -159,7 +159,7 @@ public static ReadableExternalTable getPxfHBaseReadableTable(String tableName,
                                                                  HBaseTable hbaseTable) {
         ReadableExternalTable exTable = new ReadableExternalTable(tableName,
                 fields, hbaseTable.getName(), "CUSTOM");
-        exTable.setProfile("HBase");
+        exTable.setProfile("hbaase");
```
frankgh, line 162:
```suggestion
        exTable.setProfile("hbase");
```

server/pxf-service/src/test/java/org/greenplum/pxf/service/HttpRequestParserTest.java
```diff
@@ -432,7 +431,25 @@ public void protocolIsSetWhenProfileIsSpecified() {
...
+        assertEquals("test-protocol", context.getProfileScheme());
+    }
+
+    @Test
+    public void profileIsSetInLowerCase() {
+        parameters.add("X-GP-OPTIONS-PROFILE", "teST-pROFile");
+
+        RequestContext context = parser.parseRequest(parameters, RequestType.FRAGMENTER);
 
+        assertEquals("test-profile", context.getProfile());
```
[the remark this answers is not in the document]
denalex: yes, we do now

server/pxf-hive/src/main/java/org/greenplum/pxf/plugins/hive/HiveProtocolHandler.java
```diff
@@ -0,0 +1,48 @@
...
+        return resolver;
+    }
+
+    /**
+     * Determines whether the user has requested to use vectorized ORC accessor / resolver
+     * @param context request context
+     * @return true if vectorized ORC accessor and resolver will need to be used
+     */
+    private boolean useHiveVectorizedORC(RequestContext context) {
+        return ("hive:orc".equals(context.getProfile()) && context.getOption(OPTION_VECTORIZE, false));
```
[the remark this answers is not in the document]
denalex: we always lowercase the profile now and options are stored in a case-insensitive map

server/pxf-hive/src/test/java/org/greenplum/pxf/plugins/hive/HiveProtocolHandlerTest.java
```diff
@@ -0,0 +1,64 @@
...
+import org.mockito.junit.jupiter.MockitoExtension;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mockito.Mockito.when;
+
+@ExtendWith(MockitoExtension.class)
+public class HiveProtocolHandlerTest {
+
+    @Mock
+    RequestContext context;
```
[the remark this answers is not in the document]
denalex: we can (and do in many places), but using mocks for RequestContext is the proper unit testing technique that does not rely on internal implementation of the RequestContext.

frankgh (commented):
+1 for some reason i'm not able to "Approve" this PR

server/pxf-hive/src/test/java/org/greenplum/pxf/plugins/hive/HiveProtocolHandlerTest.java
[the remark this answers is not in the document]
frankgh: either way it's fine, but since RequestContext is a pojo, we should be able to use it a real object.

PR #496, opened 2020-11-17
Add PXF upgrade test

frankgh:
Test upgrading from the Latest released version of PXF 5 to the PXF 6
version in the multinode environment.


concourse/scripts/pxf_common.bash
```diff
@@ -442,6 +451,15 @@ function start_hadoop_services() {
 	fi
 }
 
+function init_pxf() {
```
ashuka24, line 454: Why was this not required before?

frankgh: we used to init for PXF 5.x, but we no longer do it. We are bringing the init step back for PXF 5.x installations

concourse/scripts/install_pxf.bash
```diff
@@ -186,35 +227,55 @@ function run_pxf_installer_scripts() {
...
-	local SCP_FILES=(pxf_tarball cluster_env_files/*)
+	mkdir -p /tmp/pxf_installer/
+	if [[ -d pxf_tarball ]]; then
+		if [[ ${PXF_COMPONENT} == true ]]; then
+			mkdir -p /tmp/pxf_inflate
+			tar -xzf pxf_tarball/pxf-*.tar.gz -C /tmp/pxf_inflate
+			cp /tmp/pxf_inflate/pxf*/* /tmp/pxf_installer/
+		else
+			cp pxf_tarball/pxf.tar.gz /tmp/pxf_installer
+			cat << EOF > /tmp/pxf_installer/install_component
```
bradfordb-vmware, line 263: If we change this `cat <<-EOF > /tmp/pxf_installer/install_component`, then we can indent the here-document with tabs and they will be stripped. This would allow indenting in a natural fashion, although we would have a mix of tabs and spaces in the whole file.

bradfordb-vmware (approved):

PR #498, opened 2020-11-18
Support pushing predicates of type varchar

frankgh:
This commit adds support for pushing predicates for varchar types. For
predicates on varchar types the node is of type `RelabelType` which
wraps the Var that we need for the predicate.


denalex (commented):
any ways to test this in C or Java ? We have some fragmenter that echoes filters back, can we make sure it does it for a varchar column ?

frankgh:
@denalex good point, let me find those tests.

frankgh:
@denalex done!

denalex (approved):

ashuka24 (approved):

PR #499, opened 2020-11-18
Docker: Remove go dep from Dockerfiles

frankgh:


README.md
```diff
@@ -91,7 +91,6 @@ To build PXF, you must have:
     Once you have installed Go, you will need the `dep` and `ginkgo` tools, which install Go dependencies and run Go tests,
```
bradfordb-vmware, line 91: We can get rid of this reference to `dep`.

cli/go/src/pxf-cli/Makefile
```diff
@@ -25,6 +26,10 @@ help:
 all: test stage install
 
 build:
+	@if [ -d $(HOME)/.go-dep-cached-sources ] && [ ! -d ${GO_MOD_CACHE}/mod ]; then \
```
bradfordb-vmware, line 29: Small suggestion: In make, either `$(foo)` or `${foo}` is a valid reference. One convention that I have seen in practice is exclusively using `$(foo)` when foo is a variable defined by make (e.g., `$(GO_MOD_CACHE)`) and using `${foo}` when foo is an environment variable (e.g., `${HOME}`). It can be small visual clue where the variable is coming from.

bradfordb-vmware (approved):

README.md
frankgh: good catch

cli/go/src/pxf-cli/Makefile
```diff
@@ -25,12 +26,16 @@ help:
 all: test stage install
 
 build:
-	go build -v ${LDFLAGS}
-	go install ${LDFLAGS}
+	@if [ -d ${HOME}/.go-dep-cached-sources ] && [ ! -d $(GO_MOD_CACHE)/mod ]; then \
+		mkdir -p $(GO_MOD_CACHE); \
+		ln -s ${HOME}/.go-dep-cached-sources $(GO_MOD_CACHE)/mod; \
```
denalex, line 31: Is this for Ci to pick dependencies from `pxf-build-dependencies` extracted tarball ? Will it work for both dep and mod ? I guess I don't like `.go-dep-cached-sources` name if those are supposed to be modules, but if we have to support both PXF 5 and 6 with the same dependency cache artifacts, then it's ok.

PR #501, opened 2020-11-19
Refactored CLI multi-node tests

denalex:


concourse/pipelines/templates/build_pipeline-tpl.yml
```diff
@@ -1404,6 +1402,14 @@ jobs:
       GROUP: security,multiClusterSecurity
       PXF_JVM_OPTS: ((pxf-jvm-opts))
       TARGET_OS: centos
+  - task: Test PXF CLI
```
frankgh, line 1405: do we need to test it here too? If we do, we will need to provision standby master

concourse/pipelines/templates/dev_build_pipeline-tpl.yml
```diff
@@ -1227,6 +1225,15 @@ jobs:
       GROUP: security,multiClusterSecurity
       PXF_JVM_OPTS: ((pxf-jvm-opts))
       TARGET_OS: centos
+  - task: Test PXF CLI
```
frankgh, line 1228: I would not add it here. Any reason to test it here?

frankgh (approved):

concourse/pipelines/templates/build_pipeline-tpl.yml
denalex: we need it here since this setup uses default PXF_BASE and the other one uses external in ~gphome/pxf. I added standby in the latest commit.

concourse/pipelines/templates/dev_build_pipeline-tpl.yml
denalex: we need it here since this setup uses default PXF_BASE and the other one uses external in ~gphome/pxf. I added standby in the latest commit.

PR #510, opened 2020-12-03
Build the pxf-tools tarball

frankgh:
PXF regression tests require some tools, for example the avro tools jar,
to perform data preparation. We build the pxf-tools tarball and add all
the jars needed for data preparation, we then push it to the
dependencies bucket. This tarball can then be later consumed by the
regression test suite.


concourse/docker/pxf-build-base/cloudbuild.yaml
```diff
@@ -98,8 +98,21 @@ steps:
         mkdir -p /root/pxf/servers/db-session-params/ && cp ./server/pxf-service/src/templates/templates/jdbc-site.xml /root/pxf/servers/db-session-params/
         mkdir -p /root/pxf/servers/db-hive/ && cp ./server/pxf-service/src/templates/templates/jdbc-site.xml /root/pxf/servers/db-hive/
         GPHD_ROOT=/tmp PXF_HOME=/tmp/pxf mvn -B -e -Dpxf.lib=/workspace/automation_tmp_lib -Dtest=HdfsSmokeTest -Djava.awt.headless=true -Dmaven.repo.local=/workspace/.m2/repository -f ./automation/pom.xml -U test || true 
+        # create pxf-tools directory and copy jars needed by pxf
```
denalex, line 101: are they needed by PXF or by regression suite ? If the latter, maybe call it `regression-tools.jar` ?

denalex (approved):

concourse/docker/pxf-build-base/cloudbuild.yaml
frankgh: makes sense

bradfordb-vmware (approved):

PR #512, opened 2020-12-07
Port some external table commits to FDW

frankgh:
- PXF: Encode header values for custom headers
- FDW: Add disable_ppd option ( similar to `gp_external_enable_filter_pushdown`)
- Support pushing predicates of type varchar


frankgh:
not porting some commits, because there are more changes that need to come in to FDW before we port those commits.

PR #513, opened 2020-12-08
Separate out the parquet prune and transform logic

frankgh:
The `ParquetOperatorPrunerAndTransformer` functionality is overloaded. We
would like to extract out the Prune and Transform functionalities into
their own classes to separate out functionality and for a cleaner
implementation.


frankgh:
@denalex This PR takes care of BPChar. Let me know if this addresses your feedback on the ORC PR.

server/pxf-api/src/main/java/org/greenplum/pxf/api/model/BasePlugin.java
```diff
@@ -31,4 +33,27 @@ public void setRequestContext(RequestContext context) {
...
+     * in nanoseconds
+     *
+     * @param totalRowsRead        the total number of rows read
+     * @param totalReadTimeInNanos the total nanoseconds it took to read the file
+     */
+    protected void logReadStats(long totalRowsRead, long totalReadTimeInNanos) {
+        if (LOG.isDebugEnabled()) {
+            final long millis = TimeUnit.NANOSECONDS.toMillis(totalReadTimeInNanos);
+            long tuplesPerNanos = totalReadTimeInNanos == 0 ? 0 : totalRowsRead / totalReadTimeInNanos;
+            LOG.debug("{}-{}: Read TOTAL of {} rows from file {} on server {} in {} ms. Average speed: {} tuples/nanoseconds",
```
denalex, line 49: will this almost always be 0 as we unlikely will reach 1tuple/ns speed and the variable is not float. Why not measure in `tuples / ms` units ?

frankgh: sure, let me change that

PR #515, opened 2020-12-09
Use Lombok for RequestContext

frankgh:
- Simplify RequestContext by using Lombok.
- Remove deprecated X-GP-DATA-FRAGMENT, use X-GP-FRAGMENT-INDEX instead.


server/pxf-api/src/main/java/org/greenplum/pxf/api/model/RequestContext.java
```diff
@@ -63,25 +68,101 @@ public void setRequestType(RequestType requestType) {
     }
 
     // ----- NAMED PROPERTIES -----
+
+    /**
+     * The fully-qualified class name for the java class that was defined as
+     * Accessor.
+     */
     private String accessor;
```
denalex, line 76:
```suggestion
    private String accessorClassName;
```
should we do this and for other components as well (fragmented / resolver) ?

server/pxf-api/src/main/java/org/greenplum/pxf/api/model/RequestContext.java
```diff
@@ -36,6 +40,7 @@
  * Common configuration available to all PXF plugins. Represents input data
  * coming from client applications, such as GPDB.
  */
+@Data
```
denalex, line 43: nice, even though we technically do not need ` @RequiredArgsConstructor` that comes with it. Any way to suppress it ?

server/pxf-api/src/main/java/org/greenplum/pxf/api/model/RequestContext.java
```diff
@@ -63,25 +68,101 @@ public void setRequestType(RequestType requestType) {
...
+     */
+    @Setter(AccessLevel.NONE)
     private String config;
+
+    /**
+     * The server configuration associated to the server that this request is
+     * accessing
+     */
     private Configuration configuration;
-    private int dataFragment = -1; /* should be deprecated */
```
denalex, line 70: so this was not used at all ? the write case would not have any fragments, default will change from -1 to 0, hopefully no existing logic impact ?

server/pxf-api/src/main/java/org/greenplum/pxf/api/model/RequestContext.java
```diff
@@ -94,15 +175,46 @@ public void setRequestType(RequestType requestType) {
      */
     private int numAttrsProjected;
 
+    /**
+     * The plugin configuration
+     */
+    private PluginConf pluginConf;
```
denalex, line 181: should this be transient to exclude from toString / hashcode methods ?

server/pxf-api/src/main/java/org/greenplum/pxf/api/model/RequestContext.java
```diff
@@ -94,15 +175,46 @@ public void setRequestType(RequestType requestType) {
...
+    /**
+     * The scheme defined at the profile level
+     */
     private String profileScheme;
 
-    // The protocol defined at the foreign data wrapper (FDW) level
+    /**
+     * The protocol defined at the foreign data wrapper (FDW) level
+     */
+    @Getter(AccessLevel.NONE)
```
denalex, line 201: why restrict getters ?

server/pxf-api/src/main/java/org/greenplum/pxf/api/model/RequestContext.java
```diff
@@ -118,26 +230,62 @@ public void setRequestType(RequestType requestType) {
...
+    /**
+     * The contents of pxf_remote_service_login set in Greenplum. Should the
+     * user set it to an empty string this function will return null.
+     */
     private String remoteLogin;
+
+    /**
+     * The contents of pxf_remote_service_secret set in Greenplum. Should the
+     * user set it to an empty string this function will return null.
+     */
```
denalex, line 242: wow, what are these for ?

server/pxf-api/src/main/java/org/greenplum/pxf/api/model/RequestContext.java
```diff
@@ -118,26 +230,62 @@ public void setRequestType(RequestType requestType) {
...
+
+    /**
+     * The transaction ID for the current Greenplum query.
+     */
     private String transactionId;
+
     /**
      * The name of the server to access. The name will be used to build
-     * a path for the config files (i.e. $PXF_BASE/servers/$serverName/*.xml)
+     * a path for the config files (i.e. $PXF_CONF/servers/$serverName/*.xml)
```
denalex, line 257: still PXF_BASE, no ?

denalex (approved):
Looks good overall. Not so sure about using @Data as we need to mark a few props as `transient` to be excluded from class-level methods and generated toString and dashcode/equals will have too many members.

Also, I believe we wanted to dismantle this with the Session concept shared across requests, perhaps for refactor only.

server/pxf-api/src/main/java/org/greenplum/pxf/api/model/RequestContext.java
frankgh: good catch, I'll replace it with `@Getter` `@Setter`.

server/pxf-api/src/main/java/org/greenplum/pxf/api/model/RequestContext.java
frankgh: I do not want to make any changes that will impact other classes. This will make us change getters for accessor, fragmenter, resolver.

server/pxf-api/src/main/java/org/greenplum/pxf/api/model/RequestContext.java
frankgh: this was not used at all, but we had the same information here: 
https://github.com/greenplum-db/pxf/pull/515/files#diff-f4bcdfde3d8615e2a52da19be244339ca0e9b457df836d33d469f9a01a0d5148L228
https://github.com/greenplum-db/pxf/pull/515/files#diff-f4bcdfde3d8615e2a52da19be244339ca0e9b457df836d33d469f9a01a0d5148L226

server/pxf-api/src/main/java/org/greenplum/pxf/api/model/RequestContext.java
frankgh: these should not be included anymore after I remove `@Data` and use `@Getter`, `@Setter` instead.

server/pxf-api/src/main/java/org/greenplum/pxf/api/model/RequestContext.java
frankgh: we have a `getProtocol` implementation down below that has some logic:
```java
    public String getProtocol() {
        return StringUtils.isNotBlank(protocol) ? protocol : inferProtocolName();
    }
```

server/pxf-api/src/main/java/org/greenplum/pxf/api/model/RequestContext.java
frankgh: no idea, but I did not remove them, maybe we should clean up some of this code at some point

server/pxf-api/src/main/java/org/greenplum/pxf/api/model/RequestContext.java
frankgh: good catch, thanks!

PR #518, opened 2020-12-15
Upgrade to Spring Boot version 2.4.1

frankgh:


frankgh:
Closing because spring boot 2.4.1 has a bug that prevents initialization of the application when specifying the application.properties file: https://github.com/spring-projects/spring-boot/issues/24515

PR #529, opened 2021-01-28
Upgrade to Spring Boot version 2.4.2

frankgh:


concourse/scripts/initialize_gpdb.bash
```diff
@@ -10,7 +10,9 @@ data_dirs=(~gpadmin/data{1..3}/primary)
 dirs=(~gpadmin/{gpconfigs,data/master} "${data_dirs[@]}")
 mkdir -p "${dirs[@]}"
 
+# cdw is the new name for the coordinator host (previously master)
 sed -e "s/MASTER_HOSTNAME=mdw/MASTER_HOSTNAME=\$(hostname -f)/g" \
```
ashuka24, line 14: if cdw is the new hostname, do we still need the command in line 14 that changes the value for mdw? Or did we leave it in to keep it backwards compatible?

frankgh: yes! we still need it for gp6 and gp5, but gp7 needs cdw

denalex (approved):

PR #530, opened 2021-02-01
Remove fragmenter call

frankgh:
Move the fragmentation logic to the Bridge. We no longer expect a
fragmentation call from the c side, and move the fragmentation
allocation for each segment to the server.

- Remove UGICache
- Remove the `pxf.metadata.cache-enabled` property
- Remove metadata serialization.
- Remove last fragment call flag


server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/BridgeResource.java
```diff
@@ -65,16 +68,16 @@ public BridgeResource(BridgeFactory bridgeFactory, SecurityService securityServi
...
     @GetMapping(value = "/Bridge", produces = MediaType.APPLICATION_OCTET_STREAM_VALUE)
     public ResponseEntity<StreamingResponseBody> read(
-            @RequestHeader MultiValueMap<String, String> headers) throws IOException, InterruptedException {
+            @RequestHeader MultiValueMap<String, String> headers) throws Throwable {
 
         RequestContext context = parseRequest(headers);
-        Bridge bridge = securityService.doAs(context, false,
-                () -> bridgeFactory.getBridge(context));
+
+        List<Fragment> fragments = fragmenterService.getFragmentsForSegment(context);
```
denalex, line 75: don't we need UGI set for this call ?

server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/BridgeResource.java
```diff
@@ -65,16 +68,16 @@ public BridgeResource(BridgeFactory bridgeFactory, SecurityService securityServi
...
 
         // Create a streaming class which will iterate over the records and put
         // them on the output stream
         StreamingResponseBody response =
-                new BridgeResponse(securityService, bridge, context);
+                new BridgeResponse(bridgeFactory, securityService, parser, context, fragments, headers);
```
denalex, line 80: I don't think we should be passing parser and headers out of this function. Ideally parser has to parse the request into Context that has all the required info for further processing.

server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/BridgeResponse.java
```diff
@@ -2,59 +2,117 @@
...
+                        !StringUtils.equalsIgnoreCase(profile, context.getProfile())) {
+
+                    LOG.debug("Fragment {} of resource {} using profile: {}", fragment.getIndex(), dataDir, profile);
+
+                    // Remember the value of the original profile defined in
+                    // the table LOCATION, so we can restore it later to its
+                    // original value
+                    List<String> profileHeaderValue = headers.get(PROFILE_KEY);
+                    headers.set(PROFILE_KEY, profile);
+                    context = parser.parseRequest(headers, context.getRequestType());
```
denalex, line 77: I'd rather not re-parse the request, as mentioned above

server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/BridgeResponse.java
```diff
@@ -2,59 +2,117 @@
...
-                ++recordCount;
+                    // Restore the original value of the profile if it was
+                    // present
+                    headers.remove(PROFILE_KEY);
+                    if (profileHeaderValue != null) {
+                        headers.addAll(PROFILE_KEY, profileHeaderValue);
+                    }
+                }
+                context.setDataSource(fragment.getSourceName());
+                context.setFragmentIndex(fragment.getIndex());
```
denalex, line 88: what do we need index for (other than logging / tracking) ?

server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/BridgeResponse.java
```diff
@@ -2,59 +2,117 @@
...
+                context.setFragmentMetadata(fragment.getMetadata());
+
+                Bridge bridge = bridgeFactory.getBridge(context);
```
denalex, line 91: let's extract his and try block into a separate method

server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/BridgeResponse.java
```diff
@@ -2,59 +2,117 @@
...
+                    LOG.debug("Finished streaming fragment {} of resource {}, {} records.", fragment.getIndex(), dataDir, recordCount);
+                } finally {
+                    LOG.debug("Stopped streaming fragment {} of resource {}, {} records.", fragment.getIndex(), dataDir, recordCount);
+                    try {
+                        bridge.endIteration();
+                    } catch (Exception e) {
+                        LOG.warn("Ignoring error encountered during bridge.endIteration()", e);
+                    }
+                }
+                fragments.set(i, null);
```
denalex, line 114: please comment why we're doing this

server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/BridgeResponse.java
```diff
@@ -2,59 +2,117 @@
...
+                    }
+                    Writable record;
+                    DataOutputStream dos = new DataOutputStream(out);
+
+                    LOG.debug("Starting streaming fragment {} of resource {}", fragment.getIndex(), dataDir);
+                    while ((record = bridge.getNext()) != null) {
+                        record.write(dos);
+                        ++recordCount;
+                    }
+                    LOG.debug("Finished streaming fragment {} of resource {}, {} records.", fragment.getIndex(), dataDir, recordCount);
```
denalex, line 105: should we `dos.flush()` since we are not closing it and it will go out of scope ? Or will that unnecessarily flush the underlying stream where we could've written more data from the next fragment ?

server/pxf-service/src/main/java/org/greenplum/pxf/service/security/BaseSecurityService.java
```diff
@@ -76,20 +74,19 @@ public BaseSecurityService(SecureLogin secureLogin, UGICache ugiCache) {
...
         String remoteUser = (isUserImpersonation ? gpdbUser : serviceUser);
-
-        SessionId session = new SessionId(
-                segmentId,
-                transactionId,
-                remoteUser,
-                serverName,
-                isSecurityEnabled,
-                loginUser);
+        String session = remoteUser + ":" + transactionId + ":" + segmentId + ":" + serverName;
```
denalex, line 77: let's use common tracking, this is no longer a cache key

server/pxf-service/src/main/java/org/greenplum/pxf/service/security/BaseSecurityService.java
```diff
@@ -105,13 +102,15 @@ public BaseSecurityService(SecureLogin secureLogin, UGICache ugiCache) {
             throw new IOException(ie);
         } finally {
             // Optimization to cleanup the cache if it is the last fragment
-            boolean releaseUgi = lastCallForSegment || exceptionDetected;
+            boolean releaseUgi = shouldReleaseUgi || exceptionDetected;
             LOG.debug("Releasing UGI from cache for session: {}. {}",
```
denalex, line 106: please update the comment, there is no more UGI cache

server/pxf-service/src/main/java/org/greenplum/pxf/service/security/BaseSecurityService.java
```diff
@@ -47,7 +45,7 @@ public BaseSecurityService(SecureLogin secureLogin, UGICache ugiCache) {
      * @param action  the action to be executed
      * @throws IOException when an IO error occurs
      */
-    public <T> T doAs(RequestContext context, final boolean lastCallForSegment, PrivilegedExceptionAction<T> action) throws IOException {
+    public <T> T doAs(RequestContext context, final boolean shouldReleaseUgi, PrivilegedExceptionAction<T> action) throws IOException {
```
denalex, line 48: would we ever send `shouldReleaseUgi=false` now ?

server/pxf-service/src/main/java/org/greenplum/pxf/service/security/SecurityService.java
```diff
@@ -29,14 +29,14 @@
     /**
      * Do as the user provide in the context
      *
-     * @param context            the request context
-     * @param lastCallForSegment true if it is the last call for a segment
-     * @param action             the action to perform
-     * @param <T>                the type
+     * @param context          the request context
+     * @param shouldReleaseUgi true if it is the last call for a segment
```
denalex, line 33: is last call description still valid ?

server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/BridgeResponse.java
frankgh: I would prefer not to flush anything. Let's let the underlying `OutputStream` do any flushing if needed. Also, we do not want to eagerly flush, when there's an exception, tomcat will replace the response buffer with the error buffer, as long as the response buffer has not been flushed.

server/pxf-service/src/main/java/org/greenplum/pxf/service/security/BaseSecurityService.java
frankgh: yes, as the code stands right now, but I am going to refactor it so we no longer need it.

server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/BridgeResponse.java
frankgh: We use it to determine whether we need to skip the header in a few profiles: `*:text` some hive profiles.
```
        skipHeaderCount = context.getFragmentIndex() == 0
                ? context.getOption("SKIP_HEADER_COUNT", 0, true)
                : 0;
```

server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/BridgeResource.java
frankgh: `FragmenterService` uses the `SecurityService`, but I am doing a refactor of this code so we don't need it anymore

server/pxf-service/src/main/java/org/greenplum/pxf/service/FragmenterService.java
```diff
@@ -0,0 +1,226 @@
...
+        }
+    }
+
+    /**
+     * Updates the fragments' indexes so that it is incremented by sourceName.
+     * (E.g.: {"a", 0}, {"a", 1}, {"b", 0} ... )
+     *
+     * @param fragments fragments to be updated
+     */
+    private void updateFragmentIndex(List<Fragment> fragments) {
```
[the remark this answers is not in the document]
frankgh: yes, we need to update it to the correct fragment index, because some profiles rely on the correct index

server/pxf-service/src/main/java/org/greenplum/pxf/service/FragmenterService.java
```diff
@@ -0,0 +1,226 @@
...
+                    fragmenterCacheFactory.getCache().size(),
+                    fragmenterCacheFactory.getCache().stats().toString());
+        }
+
+        LOG.debug("FRAGMENTER started for path \"{}\"", path);
+
+        List<Fragment> fragments;
+
+        try {
+            // We can't support lambdas here because asm version doesn't support it
```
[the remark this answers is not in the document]
frankgh: the comment is no longer relevant, removing

external-table/src/pxfbridge.c
```diff
@@ -100,28 +95,11 @@ gpbridge_export_start(gphadoop_context *context)
 int
 gpbridge_read(gphadoop_context *context, char *databuf, int datalen)
 {
-	size_t		n = 0;
+	size_t		n = fill_buffer(context, databuf, datalen);
```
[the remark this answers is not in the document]
frankgh: what do you mean by just at the end?

external-table/src/pxfbridge.c
```diff
@@ -19,6 +19,8 @@
  */
 
 #include "pxfbridge.h"
+#include "pxfheaders.h"
```
ashuka24, line 22: why is this header suddenly needed?

external-table/src/pxfbridge.h
```diff
@@ -22,11 +22,12 @@
 #define _PXFBRIDGE_H
 
 #include "libchurl.h"
-#include "pxffragment.h"
 #include "pxfuriparser.h"
 
 #include "cdb/cdbvars.h"
+#include "nodes/execnodes.h"
```
ashuka24, line 28: why are these header files needed now?

server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/BridgeResponse.java
```diff
@@ -14,47 +21,97 @@
...
-            Writable record;
             DataOutputStream dos = new DataOutputStream(out);
+            RequestContext context = this.context;
+            for (int i = 0; i < fragments.size(); i++) {
+                Fragment fragment = fragments.get(i);
+                String profile = fragment.getProfile();
+                restoreOriginalValues = false;
+                if (StringUtils.isNotBlank(profile) &&
+                        !StringUtils.equalsIgnoreCase(profile, context.getProfile())) {
+                    restoreOriginalValues = true;
```
ashuka24, line 80: what would happen if we did not restore the original values? can you add the reasoning as a comment?

external-table/src/pxfbridge.c
frankgh: We need it for `PxfInputData` in line 175 of `pxfbridge.c`. Before this import was transitively coming from the `pxffragment.h` import.

external-table/src/pxfbridge.h
frankgh: `nodes/execnodes.h` is needed for the `ProjectionInfo` in the `gphadoop_context` struct. This was also being transitively coming from `pxffragment`

server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/BridgeResponse.java
```diff
@@ -130,4 +129,44 @@ private Void writeToInternal(OutputStream out) throws IOException {
...
     }
+
+    private int processFragment(DataOutputStream dos, RequestContext context, Fragment fragment) throws Exception {
+        Writable record;
+        int recordCount = 0;
+        Bridge bridge = bridgeFactory.getBridge(context);
+        try {
+            if (!bridge.beginIteration()) return 0;
+
+            LOG.debug("Starting streaming fragment {} of resource {}", fragment.getIndex(), dataDir);
```
denalex, line 142: I'd add context.getId() to all these log statements here

server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/WritableResource.java
```diff
@@ -115,11 +115,11 @@ public WritableResource(BridgeFactory bridgeFactory, SecurityService securitySer
                                          HttpServletRequest request) throws Exception {
 
         RequestContext context = parseRequest(headers);
-        Bridge bridge = securityService.doAs(context, false,
-                () -> bridgeFactory.getBridge(context));
         InputStream inputStream = request.getInputStream();
 
         PrivilegedExceptionAction<Long> action = () -> {
+            Bridge bridge = bridgeFactory.getBridge(context);
```
denalex, line 121: can we follow the same pattern and have an internal method so only method invocation will go into lambda for better readability ?

server/pxf-service/src/main/java/org/greenplum/pxf/service/security/BaseSecurityService.java
```diff
@@ -102,20 +100,14 @@ public BaseSecurityService(SecureLogin secureLogin, UGIProvider ugiProvider) {
             throw new IOException(ie);
         } finally {
             // Optimization to cleanup the cache if it is the last fragment
-            boolean releaseUgi = shouldReleaseUgi || exceptionDetected;
-            LOG.debug("Releasing UGI from cache for session: {}. {}",
-                    session, exceptionDetected
-                            ? " Exception while processing"
-                            : (shouldReleaseUgi ? " Processed last fragment for segment" : ""));
+            LOG.debug("{} Releasing UGI resources. {}",
```
denalex, line 103: comment above is no longer relevant

server/pxf-service/src/main/java/org/greenplum/pxf/service/security/UGIProvider.java
```diff
@@ -39,9 +39,8 @@
      * @param effectiveUser the name of the user that we want to impersonate
      * @param loginUser     the UGI of the login user (or Kerberos principal)
      * @return a {@link UserGroupInformation} for impersonation.
-     * @throws IOException
      */
-    UserGroupInformation createProxyUGI(String effectiveUser, UserGroupInformation loginUser) throws IOException {
+    UserGroupInformation createProxyUGI(String effectiveUser, UserGroupInformation loginUser) {
```
denalex, line 43: should this be called `createProxyUser` for consistency with the method below ?

server/pxf-service/src/test/java/org/greenplum/pxf/service/security/BaseSecurityServiceTest.java
```diff
@@ -92,60 +92,51 @@ public void determineRemoteUser_IsServiceUser_NoKerberos_NoImpersonation_Service
...
-    public void doesNotCleanTheUGICacheOnNonLastCalls() throws Exception {
-        expectScenario("login-user", false, false, false);
-        service.doAs(context, false, EMPTY_ACTION);
-        verifyScenario("login-user", false, false);
-        verify(mockUGIProvider, never()).destroy(any(UserGroupInformation.class));
-
-    }
-
     @Test
     public void tellsTheUGICacheToCleanItselfOnTheLastCallForASegment() throws Exception {
```
denalex, line 137: update method name and comment on line 134

server/pxf-service/src/main/java/org/greenplum/pxf/service/FragmenterService.java
```diff
@@ -84,38 +71,54 @@ public FragmenterService(FragmenterCacheFactory fragmenterCacheFactory,
...
                         return fragmentList;
                     });
         } catch (UncheckedExecutionException | ExecutionException e) {
+            IOException exception;
             // Unwrap the error
-            if (e.getCause() != null)
-                throw e.getCause();
-            throw e;
+            if (e.getCause() != null) {
+                exception = e.getCause() instanceof IOException ? (IOException) e.getCause() : new IOException(e.getCause());
```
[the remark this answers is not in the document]
frankgh: I refactored this code, we no longer need to do this

denalex (approved):

PR #533, opened 2021-02-03
Support for Apache portable runtime (APR)

frankgh:
- Build the Apache Tomcat Native libraries
- Add APR listener, allow disabling APR through PXF configuration
- Drop PXF support for GP5 on RHEL6


Makefile
```diff
@@ -74,7 +74,11 @@ stage:
 	cp -a cli/build/stage/* build/stage/$${PXF_PACKAGE_NAME} ;\
 	cp -a server/build/stage/* build/stage/$${PXF_PACKAGE_NAME} ;\
 	echo $$(git rev-parse --verify HEAD) > build/stage/$${PXF_PACKAGE_NAME}/pxf/commit.sha ;\
-	cp package/install_binary build/stage/$${PXF_PACKAGE_NAME}/install_component
+	cp package/install_binary build/stage/$${PXF_PACKAGE_NAME}/install_component ;\
+	if [ -n "$(NATIVE_LIBRARY_PATH)" ] && [ -f "$(NATIVE_LIBRARY_PATH)/libtcnative-1.so" ]; then \
+		mkdir -p build/stage/$${PXF_PACKAGE_NAME}/pxf/application/native && \
+		cp $(NATIVE_LIBRARY_PATH)/libtcnative-1.so build/stage/$${PXF_PACKAGE_NAME}/pxf/application/native/; \
```
bradfordb-vmware, line 80: I am not overly familiar with the Tomcat Native library, but usually shared-objects have multiple paths:

```console
$ ls -l /lib/x86_64-linux-gnu/libz.so*
lrwxrwxrwx 1 root root     36 Feb 24  2020 /lib/x86_64-linux-gnu/libz.so -> /lib/x86_64-linux-gnu/libz.so.1.2.11
lrwxrwxrwx 1 root root     14 Feb 24  2020 /lib/x86_64-linux-gnu/libz.so.1 -> libz.so.1.2.11
-rw-r--r-- 1 root root 113088 Feb 24  2020 /lib/x86_64-linux-gnu/libz.so.1.2.11
```
We don't need to copy all of them, but we do need to make sure we copy the one that is needed by whatever is linked against it:

```console
$ ldd $(command -v vim)
...	
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007f72448cc000)
...
```

Have we verified that this is the only shared object we need?

concourse/pipelines/perf_pipeline.yml
```diff
@@ -143,6 +143,14 @@ resources:
     json_key: ((pxf-storage-service-account-key))
     versioned_file: singlecluster/HDP2/singlecluster-HDP2.tar.gz
 
+- name: tomcat-native
+  type: gcs
+  icon: google-drive
+  source:
+    bucket: data-gpdb-ud-pxf-build-resources
+    json_key: ((pxf-storage-service-account-key))
+    regexp: tomcat-native/tomcat-native-(.*)-src.tar.gz
```
bradfordb-vmware, line 152: Have we documented where we get this source artifact from? Do we have a process for updating the version?

concourse/pipelines/templates/dev_build_pipeline-tpl.yml
```diff
@@ -398,6 +326,10 @@ jobs:
       resource: gpdb[[gp_ver]]_rhel7_rpm_latest-0
     - get: gpdb[[gp_ver]]-pxf-dev-centos7-image
     - get: pxf-build-dependencies
+    - get: tomcat-native
+  - task: Build Tomcat Native on RHEL7
+    image: gpdb[[gp_ver]]-pxf-dev-centos7-image
+    file: pxf_src/concourse/tasks/build_tomcat_native.yml
```
bradfordb-vmware, line 332: I am assuming that `tomcat-native` doesn't depend on PXF or GPDB. Could we build it once and save the built artifact in blob store?

concourse/pipelines/templates/dev_build_pipeline-tpl.yml
```diff
@@ -490,6 +422,10 @@ jobs:
       resource: gpdb[[gp_ver]]_ubuntu18_deb_latest-0
     - get: gpdb[[gp_ver]]-pxf-dev-ubuntu18-image
     - get: pxf-build-dependencies
+    - get: tomcat-native
+  - task: Build Tomcat Native on Ubuntu18
```
bradfordb-vmware, line 426: It looks like libtcnative-1 1.2.16 is available in Ubuntu's package archive. Could we use that one?

concourse/scripts/build_tomcat_native.bash
```diff
@@ -0,0 +1,45 @@
...
+elif command -v dpkg; then
+  dpkg -l "*apr*"
+else
+  echo "Unsupported operating system ${TARGET_OS}. Exiting..."
+  exit 1
+fi
+
+# Sources JAVA_HOME
+source ~/.pxfrc
+echo "JAVA_HOME=$JAVA_HOME"
```
bradfordb-vmware, line 29: Does `JAVA_HOME` need to be set when building libtcnative? If it does, does that mean that the shared-object depends on the version of the JVM being used to run PXF?

server/pxf-service/src/main/java/org/greenplum/pxf/service/PxfConfiguration.java
```diff
@@ -100,4 +104,15 @@ public ThreadPoolTaskExecutor pxfApplicationTaskExecutor(PxfServerProperties pxf
 
         return builder.build(PxfThreadPoolTaskExecutor.class);
     }
+
+    @Bean
+    public ServletWebServerFactory servletContainer(PxfServerProperties properties) {
+        TomcatServletWebServerFactory container = new TomcatServletWebServerFactory();
+        if (!properties.isAprDisabled()) {
```
bradfordb-vmware, line 111: Does this property need to be named `isAprDisabled`? This reads like a double negative.

server/pxf-service/src/scripts/pxf
```diff
@@ -45,7 +45,7 @@ export PXF_BASE=${PXF_BASE:=$PXF_HOME}
 [[ -f ${PXF_BASE}/conf/pxf-env.sh ]] && source "${PXF_BASE}/conf/pxf-env.sh"
 
 # Path to HDFS native libraries
-export LD_LIBRARY_PATH=${LD_LIBRARY_PATH:+$LD_LIBRARY_PATH:}${PXF_BASE}/lib/native:/usr/lib/hadoop/lib/native
+export LD_LIBRARY_PATH=${LD_LIBRARY_PATH:+$LD_LIBRARY_PATH:}${PXF_HOME}/application/native:${PXF_BASE}/lib/native:/usr/lib/hadoop/lib/native
```
bradfordb-vmware, line 48: Is there a reason that we put the libtcnative shared-object in a different location that other native libraries?

concourse/pipelines/perf_pipeline.yml
frankgh: it's in tracker, maybe we should expose the steps in some README in source?

concourse/pipelines/templates/dev_build_pipeline-tpl.yml
frankgh: It's cheap to build, and I didn't want to have a dedicated pipeline for that, ~17 secs in the pipeline. I think it's fine to build it every tiime.

server/pxf-service/src/scripts/pxf
frankgh: `${PXF_BASE}/lib/native` is intended for user provided libraries, but we can circle back on this

server/pxf-service/src/main/java/org/greenplum/pxf/service/PxfConfiguration.java
frankgh: The property can be renamed to `pxf.apr-enabled` and default to true.

Makefile
frankgh: what `command` can I use?

concourse/docker/pxf-dev-base/cloudbuild.yaml
```diff
@@ -199,7 +174,6 @@ steps:
 
 # Push images to Cloud Build to Container Registry
 images:
-- 'gcr.io/$PROJECT_ID/gpdb-pxf-dev/gpdb5-centos6-test-pxf:$COMMIT_SHA'
```
denalex, line 202: would dropping this make it hard for us to ship patches for PXF 5.x ?

concourse/pipelines/templates/dev_build_pipeline-tpl.yml
denalex: I kinda second @bradfordb-vmware comment, can we drive that build via cloudbuild as we do for other dependencies?

concourse/scripts/build_tomcat_native.bash
```diff
@@ -0,0 +1,45 @@
...
+openssl version
+
+# Check APR version
+echo "Checking APR version"
+if command -v rpm; then
+  rpm -qa | grep -i apr
+elif command -v dpkg; then
+  dpkg -l "*apr*"
+else
+  echo "Unsupported operating system ${TARGET_OS}. Exiting..."
```
denalex, line 23: is this still gonna work on Mac ?

concourse/scripts/build_tomcat_native.bash
```diff
@@ -0,0 +1,45 @@
...
+BUILD=${PWD}
+
+# Apache Tomcat Native Library requires:
+# 1. OpenSSL version 1.0.2 or higher
+# 2. APR version 1.4.0 or higher
+# 3. JAVA_HOME to be set
+
+# Check OpenSSL version
+echo "Checking OpenSSL version"
+openssl version
```
denalex, line 14: we are not checking the return value to make sure it's a proper version ?

concourse/pipelines/perf_pipeline.yml
denalex: they release patches regularly, we need to know which version of Native we are building / bundling for every PXF release and we need to integrate with license finder. As I mentioned elsewhere, maybe have a cloudbuild trigger that gets source from the web, builds and pushes the artifact.

denalex (commented):
I think it's be much better to break this PR into 2:
* removal of Centos6 support for GP5
* actual Tomcat native stuff

concourse/scripts/build_tomcat_native.bash
frankgh: the build makes sure the minimum required version, that's why CentOS6  was failing

concourse/scripts/build_tomcat_native.bash
frankgh: not this script specifically, but libapr is supported for macos

concourse/docker/pxf-dev-base/cloudbuild.yaml
frankgh: we are not removing the images, but we can keep building them on the 5.16.0 branch

frankgh:
Closing in favor of https://github.com/greenplum-db/pxf/pull/540 and https://github.com/greenplum-db/pxf/pull/536

PR #535, opened 2021-02-04
Allow short names for text compression codecs

ashuka24:
In Parquet and Avro we allow the user to specify COMPRESSION_CODEC by
short names like snappy, gzip, bzip2, deflate, etc. With this change, we
can allow the user to specify supported codecs for text data using short
names.

This commit also does some refactoring and removes the CodeFactory and 
CodeFactoryTest classes in favor of the HdfsSplittableDataAccessorTest class.


server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HdfsSplittableDataAccessor.java
```diff
@@ -126,4 +129,21 @@ public void closeForRead() throws Exception {
...
     }
+
+    /**
+     * Helper routine to get compression codec by class name or alias.
+     *
+     * @param name codec name
+     * @param conf configuration used
+     * @return generated CompressionCodec
+     */
+    protected CompressionCodec getCodec(String name, Configuration conf) {
```
denalex, line 140: actually, this class already has `protected JobConf jobConf` so we don't need to take a part here and can just reuse the class field we already have access to.

frankgh (commented):
Should we support the `uncompressed` codec short name (meaning no codec), similar to the  `COMPRESSION_CODEC` options we support for parquet and avro?

https://greenplum.docs.pivotal.io/pxf/5-16/using/hdfs_avro.html#profile_cet
https://greenplum.docs.pivotal.io/pxf/5-16/using/hdfs_parquet.html

denalex:
For Parquet we would write with `snappy` by default if codec is not specified. For text we write uncompressed by default, so I don't think we really need that option, but we can add it just for consistency.

frankgh (approved):

lisakowen:
a few questions about this PR:
1.  we only document the compression codec alias names for parquet and avro.  should we deprecate the use of the complete codec java class name for text and have 6.0.0 document only the short/alias names?
2.  does this PR also change the sequencefile profile to support compression codec short names?
3.  the user needs to provide the codec when they create a writable external table.  when using an external table to read compressed data:  a.  the user does not need to provide the codec, correct?  b.  for a given profile, can we read only the codecs that we support for write?

PR #536, opened 2021-02-04
Drop PXF support for GP5 on RHEL6

frankgh:


denalex (approved):
looks good, but please check PR pipeline for this PR.

frankgh:
This PR https://github.com/greenplum-db/pxf/pull/537 fixes the gpdb7 pipeline. GP5 on RHEL6 fails because we are removing rhel6 support

PR #538, opened 2021-02-05
Add custom rule for HADOOP_SECURITY_AUTH_TO_LOCAL

frankgh:
Starting with Hadoop 2.10.0, the "DEFAULT" rule will throw an exception
when no rules are applied while getting the principal name translation
into operating system user name. See
org.apache.hadoop.security.authentication.util.KerberosName#getShortName
We add a default rule that will return the service name as the short
name, i.e. gpadmin/_HOST@REALM will map to gpadmin.

The second commit fixes an issue with minio during write, when the parent
directory has not been created. We add the
`Options.CreateOpts.createParent()` option to make sure the parent directory
is created during write.

### DIFF
--- a/server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/SequenceFileAccessor.java
+++ b/server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/SequenceFileAccessor.java
@@ -23,6 +23,7 @@ package org.greenplum.pxf.plugins.hdfs;
 import org.apache.hadoop.fs.CreateFlag;
 import org.apache.hadoop.fs.FileContext;
 import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Options;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.io.LongWritable;
 import org.apache.hadoop.io.SequenceFile;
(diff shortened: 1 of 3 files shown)


server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/SequenceFileAccessor.java
```diff
@@ -177,7 +178,8 @@ public boolean writeNextObject(OneRow onerow) throws IOException {
             // create writer - do not allow overwriting existing file
             writer = SequenceFile.createWriter(fc, configuration, file, keyClass,
                     valueClass, compressionType, codec,
-                    new SequenceFile.Metadata(), EnumSet.of(CreateFlag.CREATE));
+                    new SequenceFile.Metadata(), EnumSet.of(CreateFlag.CREATE),
+                    Options.CreateOpts.createParent());
```
ashuka24, line 182: Will it always be the case that the parent directory does not exist? Or does `createParent()` do a create if not exists?

denalex (approved):

ashuka24 (approved):

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/SequenceFileAccessor.java
frankgh: it will create the parent directory if it doesn't exist. This is only an issue for minio starting with hadoop 2.10.1 libraries

PR #540, opened 2021-02-06
Support for Apache portable runtime (APR)

frankgh:
- Java: Add Apache Portable Runtime (APR) as default listener
- Cloud Build: Use cloudbuild to produce `tomcat-native` libraries for CentOS7 GP5, GP6, GP7 and Ubuntu 18 GP6, GP7
- Bundle tomcat native library in RPM/DEB files
- Add `application/native` to the LD_LIBRARY_PATH


server/pxf-service/src/main/resources/application.properties
```diff
@@ -28,11 +28,14 @@ pxf.tomcat.max-header-count=30000
 # TODO: spring_boot_todo what value should we set here
 spring.mvc.async.request-timeout=86400000
 
+# Enable/disable the Apache Portable Runtime (APR) based Native library for Tomcat
+pxf.apr-enabled=true
```
denalex, line 32: should we assume that for developers this will be false as they will not have access to the native libs (unless they explicitly build it or download it from the bucket ? We might have diff grade profiles and use `native` profile in CI build jobs to enable this.

frankgh: yes, I'll add a `dev` profile and disable APR by default for dev profiles

concourse/docker/tomcat-native/cloudbuild.yaml
```diff
@@ -0,0 +1,169 @@
...
+    args:
+      - '-c'
+      - |
+        ls
+        cd apr
+        git checkout 1.4.0
+    waitFor: [ 'git-tomcat-native' ]
+
+  # build tomcat-native for Greenplum 5 on CentOS 7
+  - name: 'gcr.io/$PROJECT_ID/gpdb-pxf-dev/gpdb5-centos7-test-pxf:latest'
```
bradfordb-vmware, line 43: We might want to provide some more info here about why we need multiple tomcat-artifacts for RHEL7. I don't think tomcat-native depends on GPDB, so my initial thought was to suggest building once for RHEL7 and re-using for the different GPDBs. However, in GP5 we vendor SSL which tomcat-native does link against. It might be worth pointing out that we (indirectly) do depend on the GPDB version.

bradfordb-vmware (approved):

concourse/docker/tomcat-native/cloudbuild.yaml
frankgh: done, added a note at the beginning of this file

server/pxf-service/src/main/resources/application.properties
frankgh: by default, APR is enabled, but we fallback to NIO if the native library is not present

concourse/docker/tomcat-native/README.md
```diff
@@ -0,0 +1,98 @@
...
+  /usr/local/etc/openssl@1.1/certs
+
+and run
+  /usr/local/opt/openssl@1.1/bin/c_rehash
+
+openssl@1.1 is keg-only, which means it was not symlinked into /usr/local,
+because macOS provides LibreSSL.
+
+If you need to have openssl@1.1 first in your PATH, run:
+  echo 'export PATH="/usr/local/opt/openssl@1.1/bin:$PATH"' >> /Users/fguerrero/.bash_profile
```
ashuka24, line 67:
```suggestion
  echo 'export PATH="/usr/local/opt/openssl@1.1/bin:$PATH"' >> /Users/<username>/.bash_profile

**Review by ashuka24 — APPROVED** [review]

File: concourse/docker/tomcat-native/README.md [hunk 4]
```diff
@@ -0,0 +1,98 @@
...
+```shell
+cd ~/workspace/tomcat-native/native
+sh buildconf --with-apr=${HOME}/workspace/apr
+./configure --with-ssl=/usr/local/Cellar/openssl@1.1/1.1.1i
+make
+```
+
+This should produce the native libraries inside `.libs/`.
+
+Finally, copy all the files from `.libs` to your `$PXF_BASE/lib/native`.
```
denalex, line 94: should we advise them to copy those files somewhere else and set NATIVE_LIBRARY_PATH to that location so it can be used when running `make stage` etc ?

server/pxf-service/src/main/java/org/greenplum/pxf/service/PxfConfiguration.java
```diff
@@ -100,4 +106,29 @@ public ThreadPoolTaskExecutor pxfApplicationTaskExecutor(PxfServerProperties pxf
...
+        try {
+            Library.initialize(null);
+            nativeLibraryPresent = true;
+        } catch (LibraryNotFoundError e) {
+            LOG.warn("Tomcat Native Library is not present", e);
+        } catch (Exception e) {
+            LOG.error("Error while loading Tomcat Native Library", e);
+        }
+        if (nativeLibraryPresent) {
+            LOG.debug("Using tomcat native library");
```
denalex, line 127:
```suggestion
            LOG.info("Using Apache Tomcat Native library and APR protocol");
```

server/pxf-service/src/main/java/org/greenplum/pxf/service/PxfConfiguration.java
```diff
@@ -100,4 +106,29 @@ public ThreadPoolTaskExecutor pxfApplicationTaskExecutor(PxfServerProperties pxf
...
+            return container;
+
+        boolean nativeLibraryPresent = false;
+        try {
+            Library.initialize(null);
+            nativeLibraryPresent = true;
+        } catch (LibraryNotFoundError e) {
+            LOG.warn("Tomcat Native Library is not present", e);
+        } catch (Exception e) {
+            LOG.error("Error while loading Tomcat Native Library", e);
```
denalex, line 124:
```suggestion
            LOG.warn("Error while loading Apache Tomcat Native library, APR protocol will not be used", e);
```

server/pxf-service/src/main/java/org/greenplum/pxf/service/PxfConfiguration.java
```diff
@@ -100,4 +106,29 @@ public ThreadPoolTaskExecutor pxfApplicationTaskExecutor(PxfServerProperties pxf
...
+
+        if (!properties.isAprEnabled())
+            return container;
+
+        boolean nativeLibraryPresent = false;
+        try {
+            Library.initialize(null);
+            nativeLibraryPresent = true;
+        } catch (LibraryNotFoundError e) {
+            LOG.warn("Tomcat Native Library is not present", e);
```
denalex, line 122:
```suggestion
            LOG.warn("Apache Tomcat Native library is not present, APR protocol will not be used", e);
```

server/pxf-service/src/main/java/org/greenplum/pxf/service/PxfConfiguration.java
```diff
@@ -100,4 +106,29 @@ public ThreadPoolTaskExecutor pxfApplicationTaskExecutor(PxfServerProperties pxf
...
+        } catch (Exception e) {
+            LOG.error("Error while loading Tomcat Native Library", e);
+        }
+        if (nativeLibraryPresent) {
+            LOG.debug("Using tomcat native library");
+            LifecycleListener aprLifecycle = new AprLifecycleListener();
+            container.setProtocol("org.apache.coyote.http11.Http11AprProtocol");
```
denalex, line 129: I wonder if instead we should do `setUseAprConnector(true)` which should internally check for library presence and fallback, if not found: https://tomcat.apache.org/tomcat-8.5-doc/config/http.html

server/pxf-service/src/main/resources/application.properties
denalex: should we name it `useAprConnector` to align with tomcat property ?

server/pxf-service/src/main/java/org/greenplum/pxf/service/PxfConfiguration.java
denalex: Also, maybe it is now auto-hooked up ? See https://github.com/spring-projects/spring-boot/commit/22c9364237c4b576ae6792fac84766683ed2818e and https://github.com/spring-projects/spring-boot/issues/16040 , meaning if the library is detected, it will be used automatically ?

Also, it seems APR is getting deprecated in Tomcat 10.1.x - https://github.com/apache/tomcat/commit/72323cb76307698b66b0253a34bfbb4a9ea034fe
I have asked Tomcat experts for advice.

concourse/docker/tomcat-native/README.md
frankgh: I would prefer not to, it doesn't seem like a good workflow for developers. I will add both options, in my case, I have them in `$PXF_BASE/lib/native`

server/pxf-service/src/main/java/org/greenplum/pxf/service/PxfConfiguration.java
frankgh: I am unable to have the APR loaded automatically, let's take a look together today

frankgh: Regarding deprecation, it might be a while before we get to Tomcat 10. By that time we might decide to switch to reactor-netty or some other transport protocol. In the meantime, it is my opinion that we should take advantage of APR, which was especially noticeable in the refactor branch.

PR #542, opened 2021-02-08
Support different database encodings

frankgh:
- Pass the data encoding and database enconding from the client to server.
- Add tests


automation/src/main/java/org/greenplum/pxf/automation/components/gpdb/Gpdb.java
```diff
@@ -52,7 +52,17 @@ public void init() throws Exception {
 		connect();
 
 		if (!checkDataBaseExists(getDb())) {
-			createDataBase(getDb(), true);
+			String encoding = getEncoding();
+			String lcCollate = getLcCollate();
+			String lcCType = getLcCType();
+
+			if (StringUtils.isNotBlank(encoding)
```
ashuka24, line 59: is it possible to create with only a subset of the three? or are all three always going to be needed?

automation/src/test/java/org/greenplum/pxf/automation/features/multiserver/MultiServerTest.java
```diff
@@ -70,8 +70,6 @@ public void beforeClass() throws Exception {
      * Before every method determine default hdfs data Path, default data, and
      * default external table structure. Each case change it according to it
      * needs.
-     *
```
ashuka24, line 73: why did we get rid of this comment, it looks like we do throw an exception here?

server/pxf-api/src/main/java/org/greenplum/pxf/api/model/RequestContext.java
```diff
@@ -282,6 +284,16 @@ public void setRequestType(RequestType requestType) {
      */
     private String user;
 
+    /**
+     * The encoding of the source data
+     */
+    private Charset dataEncoding = StandardCharsets.UTF_8;
```
ashuka24, line 290: why do we hard code this to `UTF_8` here?

automation/src/main/java/org/greenplum/pxf/automation/components/gpdb/Gpdb.java
frankgh: actually yes, I need to change this for Greenplum 5

automation/src/test/java/org/greenplum/pxf/automation/features/multiserver/MultiServerTest.java
frankgh: it's a javadoc issue, you are supposed to enter a description when using `@throws` in javadocs. But it's not necessary to have these for testing code

server/pxf-api/src/main/java/org/greenplum/pxf/api/model/RequestContext.java
frankgh: We are defaulting to `UTF8` rather. We parse the headers from the request, and set the appropriate encoding values.

server/pxf-api/src/main/java/org/greenplum/pxf/api/io/GPDBWritable.java
```diff
@@ -376,68 +379,72 @@ public void write(DataOutput out) throws IOException {
...
+                 * For text format, we must convert encoding first.
+                 */
                 if (!coldbtype.isVarLength()) {
                     colLength[i] = coldbtype.getTypeLength();
                 } else if (!isTextForm(colType[i])) {
                     colLength[i] = ((byte[]) colValue[i]).length;
+                } else if (colType[i] == DataType.TEXT.getOID()
+                        || colType[i] == DataType.VARCHAR.getOID()
+                        || colType[i] == DataType.BPCHAR.getOID()) {
+                    colLength[i] = ((String) colValue[i]).getBytes(databaseEncoding).length;
```
denalex, line 401: what about date / timestamps -- they can also be localized, does not matter here ?

server/pxf-api/src/main/java/org/greenplum/pxf/api/io/GPDBWritable.java
```diff
@@ -376,68 +379,72 @@ public void write(DataOutput out) throws IOException {
...
                 } else {
-                    colLength[i] = ((String) colValue[i]).getBytes(CHARSET).length;
+                    colLength[i] = ((String) colValue[i]).getBytes(StandardCharsets.UTF_8).length;
```
denalex, line 403: why not always use database encoding ?

server/pxf-api/src/main/java/org/greenplum/pxf/api/io/GPDBWritable.java
```diff
@@ -458,24 +465,33 @@ public void write(DataOutput out) throws IOException {
...
+                        out.writeInt(colLength[i]);
+                        out.write(data);
+                        break;
+                    }
+
+                    /* For text format, add 4byte length header. string is already '\0' terminated */
                     default: {
                         out.writeInt(colLength[i]);
-                        byte[] data = ((String) colValue[i]).getBytes(CHARSET);
+                        byte[] data = ((String) colValue[i]).getBytes(StandardCharsets.UTF_8);
```
denalex, line 486: same questions as above

server/pxf-api/src/main/java/org/greenplum/pxf/api/utilities/CharsetUtils.java
```diff
@@ -0,0 +1,214 @@
...
+import java.util.HashMap;
+import java.util.Locale;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Source Apache Tika:
+ * https://github.com/apache/tika/blob/master/tika-core/src/main/java/org/apache/tika/utils/CharsetUtils.java
+ */
```
denalex, line 36: can we declare Tika as dependency and use the existing class rather than copy & pasting the code ?

server/pxf-api/src/main/java/org/greenplum/pxf/api/utilities/CharsetUtils.java
```diff
@@ -0,0 +1,214 @@
...
+@Component
+public class CharsetUtils {
```
denalex, line 38: why do we need all this logic in addition to what Java already provides ?

server/pxf-service/src/main/java/org/greenplum/pxf/service/HttpRequestParser.java
```diff
@@ -94,6 +98,16 @@ public RequestContext parseRequest(MultiValueMap<String, String> requestHeaders,
             LOG.info("Original query has filter, but it was not propagated to PXF");
         }
 
+        String sourceEncoding = params.removeOptionalProperty("DATA-ENCODING");
+        if (StringUtils.isNotBlank(sourceEncoding)) {
+            context.setDataEncoding(charsetUtils.forName(sourceEncoding));
+        }
+
+        String targetEncoding = params.removeOptionalProperty("DATABASE-ENCODING");
```
denalex, line 106: why is this optional, wouldn't we always now send database encoding ?

server/pxf-service/src/main/java/org/greenplum/pxf/service/HttpRequestParser.java
```diff
@@ -94,6 +98,16 @@ public RequestContext parseRequest(MultiValueMap<String, String> requestHeaders,
             LOG.info("Original query has filter, but it was not propagated to PXF");
         }
 
+        String sourceEncoding = params.removeOptionalProperty("DATA-ENCODING");
+        if (StringUtils.isNotBlank(sourceEncoding)) {
+            context.setDataEncoding(charsetUtils.forName(sourceEncoding));
```
denalex, line 103: if not specified, seems like we will assume UTF-8. So even if the DB, system and data are all in the same encoding (like WIN1251), we will still require users to specify it in the external table definition ? Alternatively we can default data encoding to be the same as database encoding, unless specified explicitly. Not sure what Greenplum conventions are for this.

automation/src/test/java/org/greenplum/pxf/automation/features/general/DatabaseEncodingTest.java
```diff
@@ -0,0 +1,130 @@
...
+import java.util.List;
+
+import static java.lang.Thread.sleep;
+
+/**
+ * Tests non-UTF8 database encodings
+ **/
+public class DatabaseEncodingTest extends BaseFeature {
+
+    private static final String[] PARQUET_TABLE_COLUMNS = new String[]{"a    TEXT"};
```
denalex, line 20: wonder if we need all other datatypes (chars, varchars) and date/timestamps/arrays to test for multi-locale support.

server/pxf-api/src/main/java/org/greenplum/pxf/api/utilities/CharsetUtils.java
frankgh: java will not resolve the values we get from Greenplum (i.e "WIN1251") to a Charset

server/pxf-api/src/main/java/org/greenplum/pxf/api/io/GPDBWritable.java
frankgh: yeah, that's  valid point, I will change it

server/pxf-service/src/main/java/org/greenplum/pxf/service/HttpRequestParser.java
frankgh: yeah, it is required.

server/pxf-service/src/main/java/org/greenplum/pxf/service/HttpRequestParser.java
frankgh: actually, this information is always being passed from the client to the server. So I will make it required.

frankgh: it looks like the convention for Greenplum is that the table encoding will use the client encoding (which might be different from the database encoding).

server/pxf-api/src/main/java/org/greenplum/pxf/api/utilities/CharsetUtils.java
frankgh: yeah, we can. But we only need a single piece of functionality. let's discuss today

server/pxf-api/src/main/java/org/greenplum/pxf/api/utilities/CharsetUtils.java
```diff
@@ -1,214 +1,92 @@
...
+                // We prefer KOI8-U, since it is a superset of KOI8-R.
+                put("KOI8", new String[]{"KOI8_U", "KOI8_R"});
+                // If the database isn't encoding-aware then we can't have
+                // any preferred encodings.
+                put("UNKNOWN", new String[0]);
+                // The following encodings do not have a java equivalent
+                put("MULE_INTERNAL", new String[0]);
+                put("LATIN6", new String[0]);
+                put("LATIN8", new String[0]);
+                put("LATIN10", new String[0]);
```
denalex, line 61: what's the point of adding unsupported ones to the map when we gonna throw an exception anyways, we can have it just commented out for documentation completeness.

frankgh: I've addressed that.

server/pxf-api/src/main/java/org/greenplum/pxf/api/utilities/CharsetUtils.java
```diff
@@ -0,0 +1,103 @@
...
+
+@Component
+public class CharsetUtils {
+
+    private static final Logger LOG = LoggerFactory.getLogger(CharsetUtils.class);
+
+    private static final Map<String, String[]> KNOWN_ENCODINGS = Collections.unmodifiableMap(
+            new HashMap<String, String[]>() {{
+                //Note: this list should match the set of supported server
+                // encodings found in backend/util/mb/encnames.c
```
ashuka24, line 22: is this list a fairly stable list? how likely are we to update this list again in the future?

ashuka24 (approved):

server/pxf-api/src/main/java/org/greenplum/pxf/api/utilities/CharsetUtils.java
frankgh: This list corresponds to PG 12 (GP 7) and PG 9.4 (GP 6). In PG 8.3 (GP 5), we had 1 less encoding in the list. So yes, the list is pretty stable, and when the GP 8 time comes, we need to make sure new encodings are covered

bradfordb-vmware: I am not familiar with either of Postgres' or Java's encodings but looking at the mentioned file (`src/backend/util/mb/encnames.c`) it is not obvious how I would map from the Postgres source to this file. Is this documented somewhere?

server/pxf-api/src/test/java/org/greenplum/pxf/api/utilities/CharsetUtilsTest.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    CharsetUtils utils;
+
+    @BeforeEach
+    void setup() {
+        utils = new CharsetUtils();
+    }
+
+    @MethodSource("unsupportedEncodings")
```
bradfordb-vmware, line 83: :+1:

bradfordb-vmware (approved):

server/pxf-api/src/main/java/org/greenplum/pxf/api/utilities/CharsetUtils.java
frankgh: Postgres encodings are documented here: https://www.postgresql.org/docs/12/multibyte.html. Java encodings are documented in the postgres jdbc driver, but there are a few missing there.

PR #544, opened 2021-02-08
Fix different encoding when using LineRecordReader 

frankgh:
LineRecordReader produces a org.apache.hadoop.io.Text value, that is
being converted to String by PXF. However, during the String conversion
the encoding information is not taken into account. In this commit we
skip conversion to String and just pass org.apache.hadoop.io.Text and
then finally just byte[]. This allows us to more efficiently transport
the data from external source to Greenplum without worrying about
converting to the appropriate encoding. The encoding conversion is
handled by Greenplum later.

Restore automation test testing different encoding

### DIFF
--- a/server/pxf-api/src/main/java/org/greenplum/pxf/api/io/BufferWritable.java
+++ b/server/pxf-api/src/main/java/org/greenplum/pxf/api/io/BufferWritable.java
@@ -8,9 +8,9 @@ package org.greenplum.pxf.api.io;
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
- * 
+ *
  *   http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing,
  * software distributed under the License is distributed on an
  * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
@@ -31,7 +31,8 @@ import java.lang.UnsupportedOperationException;
  */
 public class BufferWritable implements Writable {
 
-    byte[] buf = null;
+    byte[] buf;
+    int length;
 
     /**
      * Constructs a BufferWritable. Copies the buffer reference and not the
@@ -39,10 +40,25 @@ public class BufferWritable implements Writable {
      * through the Bridge framework without copying the data each time the
      * buffer is passed between the Bridge objects.
      *
-     * @param inBuf buffer
+     * @param inBuf buffer reference
      */
     public BufferWritable(byte[] inBuf) {
-        buf = inBuf;
+        this(inBuf, inBuf.length);
+    }
+
+    /**
+     * Constructs a BufferWritable. Copies the buffer reference (not the
+     * actual bytes) and the length of bytes.
+     * This class is used when we intend to transport a buffer through the
+     * Bridge framework without copying the data each time the buffer is passed
+     * between the Bridge objects.
+     *
+     * @param inBuf  buffer reference
+     * @param length the length of data within the buffer
+     */
+    public BufferWritable(byte[] inBuf, int length) {
+        this.buf = inBuf;
+        this.length = length;
     }
 
     /**
@@ -55,7 +71,7 @@ public class BufferWritable implements Writable {
     public void write(DataOutput out) throws IOException {
         if (buf == null)
             throw new IOException("BufferWritable was not set");
-        out.write(buf);
+        out.write(buf, 0, length);
     }
 
     /**
@@ -84,6 +100,7 @@ public class BufferWritable implements Writable {
     public void append(byte[] app) {
         if (buf == null) {
             buf = app;
+            length = buf.length;
             return;
         }
         if (app == null) {
@@ -94,5 +111,6 @@ public class BufferWritable implements Writable {
         System.arraycopy(buf, 0, newbuf, 0, buf.length);
         System.arraycopy(app, 0, newbuf, buf.length, app.length);
         buf = newbuf;
+        length = newbuf.length;
     }
 }
--- a/server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/StringPassResolver.java
+++ b/server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/StringPassResolver.java
@@ -62,7 +62,7 @@ public class StringPassResolver extends BasePlugin implements Resolver {
         if (data instanceof ChunkWritable) {
             record.add(new OneField(BYTEA.getOID(), ((ChunkWritable) data).box));
         } else {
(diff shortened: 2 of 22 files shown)


server/pxf-service/src/main/java/org/greenplum/pxf/service/BridgeOutputBuilder.java
```diff
@@ -305,10 +304,16 @@ void fillText(List<OneField> recFields) throws BadRecordException {
                 output = new BufferWritable((byte[]) val);
             }
         } else {
-            String textRec = (recFields.size() == 1 && val instanceof String) ?
-                    val + greenplumCSV.getNewline() :
-                    fieldListToCSVString(recFields);
-            output = new Text(textRec);
+
+            if (recFields.size() == 1 && val instanceof org.apache.hadoop.io.Text) {
+                org.apache.hadoop.io.Text text = (org.apache.hadoop.io.Text) val;
```
ashuka24, line 309: why must we cast it here?

server/pxf-api/src/main/java/org/greenplum/pxf/api/io/BufferWritable.java
```diff
@@ -42,7 +43,22 @@
...
+     * Constructs a BufferWritable. Copies the buffer reference and not the
+     * actual bytes and the length of bytes.
+     * This class is used when we intend to transport a buffer through the
+     * Bridge framework without copying the data each time the buffer is passed
+     * between the Bridge objects.
+     *
+     * @param inBuf  buffer
+     * @param length the length of data within the buffer
+     */
+    public BufferWritable(byte[] inBuf, int length) {
```
denalex, line 59: not sure where this new constructor is really used, since it seems we always want to write the whole buffer, no ?

server/pxf-api/src/main/java/org/greenplum/pxf/api/io/BufferWritable.java
```diff
@@ -94,5 +111,6 @@ public void append(byte[] app) {
         System.arraycopy(buf, 0, newbuf, 0, buf.length);
         System.arraycopy(app, 0, newbuf, buf.length, app.length);
         buf = newbuf;
+        length = newbuf.length;
```
denalex, line 114: what's the point to keep track of length, when `byte[].length` already has it ?

server/pxf-service/src/main/java/org/greenplum/pxf/service/BridgeOutputBuilder.java
```diff
@@ -81,13 +80,13 @@
      *                information
      */
     public BridgeOutputBuilder(RequestContext context) {
-        this.context = context;
         columnDescriptors = context.getTupleDescription();
         greenplumCSV = context.getGreenplumCSV();
+        newline = context.getGreenplumCSV().getNewline().getBytes();
```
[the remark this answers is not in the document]
frankgh: Greenplum (and CSV in general) only supports 3 types of linebreaks NL, CRNL, CR.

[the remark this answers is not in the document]
frankgh: we don't need encoding

server/pxf-service/src/main/java/org/greenplum/pxf/service/BridgeOutputBuilder.java
```diff
@@ -305,10 +304,16 @@ void fillText(List<OneField> recFields) throws BadRecordException {
...
         } else {
-            String textRec = (recFields.size() == 1 && val instanceof String) ?
-                    val + greenplumCSV.getNewline() :
-                    fieldListToCSVString(recFields);
-            output = new Text(textRec);
+
+            if (recFields.size() == 1 && val instanceof org.apache.hadoop.io.Text) {
+                org.apache.hadoop.io.Text text = (org.apache.hadoop.io.Text) val;
+                text.append(newline, 0, newline.length);
+                output = new BufferWritable(text.getBytes(), text.getLength());
```
frankgh, line 311: here we use the constructor with the length.

server/pxf-api/src/main/java/org/greenplum/pxf/api/io/BufferWritable.java
frankgh: I added this constructor because we only need to write a subset of the byte[]

server/pxf-api/src/main/java/org/greenplum/pxf/api/io/BufferWritable.java
frankgh: for the case where we don't write the whole length of the byte[]

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/StringPassResolver.java
```diff
@@ -62,7 +62,7 @@
         if (data instanceof ChunkWritable) {
             record.add(new OneField(BYTEA.getOID(), ((ChunkWritable) data).box));
         } else {
-            record.add(new OneField(VARCHAR.getOID(), data.toString()));
+            record.add(new OneField(VARCHAR.getOID(), data));
```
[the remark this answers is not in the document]
frankgh: because we'd need to create a copy of the byte[] here, I prefer passing the reference  of `Text` which internally has a byte[].

server/pxf-service/src/main/java/org/greenplum/pxf/service/BridgeOutputBuilder.java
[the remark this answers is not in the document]
frankgh: By looking at the implementation of `LineRecordReader` we only read bytes from the external source until we find a line break. That byte[] is wrapped in a `Text` object, which is what we ship around.

server/pxf-service/src/main/java/org/greenplum/pxf/service/BridgeOutputBuilder.java
frankgh: to be able to append the newline to the internal byte[]

server/pxf-service/src/main/java/org/greenplum/pxf/service/BridgeOutputBuilder.java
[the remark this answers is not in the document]
frankgh: From Greenplum documentation 
```
              [NEWLINE [ AS ] 'LF' | 'CR' | 'CRLF']
```

server/pxf-service/src/main/java/org/greenplum/pxf/service/BridgeOutputBuilder.java
frankgh: @denalex we use the constructor with the length here note that `text.getLength()` is different from the size of the byte[] used internally. `text.getBytes().length` might be different from `text.getLength()`

denalex (approved):

PR #546, opened 2021-02-08
added Prometheus metrics endpoint

denalex:
This adds the endpoint of `http://localhost:5888/actuator/prometheus` to expose the PXF Micrometer metrics in Prometheus format.

### DIFF
--- a/server/pxf-service/src/main/java/org/greenplum/pxf/service/PxfConfiguration.java
+++ b/server/pxf-service/src/main/java/org/greenplum/pxf/service/PxfConfiguration.java
@@ -1,12 +1,10 @@
 package org.greenplum.pxf.service;
 
-import io.micrometer.core.instrument.MeterRegistry;
 import org.greenplum.pxf.api.configuration.PxfServerProperties;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.ListableBeanFactory;
 import org.springframework.beans.factory.ObjectProvider;
-import org.springframework.boot.actuate.autoconfigure.metrics.MeterRegistryCustomizer;
 import org.springframework.boot.autoconfigure.task.TaskExecutionProperties;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.boot.task.TaskExecutorBuilder;
@@ -101,10 +99,5 @@ public class PxfConfiguration implements WebMvcConfigurer {
                 shutdown.getAwaitTerminationPeriod());
 
         return builder.build(PxfThreadPoolTaskExecutor.class);
-    }
-
-    @Bean
-    MeterRegistryCustomizer<MeterRegistry> metricsCommonTags() {
-        return registry -> registry.config().commonTags("application", "pxf-service");
     }
 }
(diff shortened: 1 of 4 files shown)


ashuka24 (approved):

server/pxf-service/src/main/java/org/greenplum/pxf/service/PxfConfiguration.java
```diff
@@ -100,4 +102,9 @@ public ThreadPoolTaskExecutor pxfApplicationTaskExecutor(PxfServerProperties pxf
 
         return builder.build(PxfThreadPoolTaskExecutor.class);
     }
+
+    @Bean
+    MeterRegistryCustomizer<MeterRegistry> metricsCommonTags() {
+        return registry -> registry.config().commonTags("application", "pxf-service");
```
frankgh, line 108: should we add the `hostname` to the common tags?

frankgh: also, I would prefer to move this to `server/pxf-service/src/main/resources/application.properties` 

```
management.metrics.tags.application=pxf-service
```

denalex: hostname is already available (when viewing via Grafana dashboard, at least)

where did you see you can do `management.metrics.tags.application=pxf-service` ? We can autowire the property value and set the tag based on the value, which is possible, but we still have to do MeterRegistryCustomizer, no ?

frankgh: somewhere in the documentation, that's how I was able to get it working on my local. We don't need to do `MeterRegistryCustomizer` if it's only for the application name

frankgh: https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#production-ready-metrics-common-tags

denalex: thanks, updated

denalex: The following metrics will be available from the `actuator/metrics` endpoint:
```
http.server.requests                
jvm.buffer.count                    
jvm.buffer.memory.used              
jvm.buffer.total.capacity           
jvm.classes.loaded                  
jvm.classes.unloaded                
jvm.gc.live.data.size               
jvm.gc.max.data.size                
jvm.gc.memory.allocated             
jvm.gc.memory.promoted              
jvm.gc.pause                        
jvm.memory.committed                
jvm.memory.max                      
jvm.memory.used                     
jvm.threads.daemon                  
jvm.threads.live                    
jvm.threads.peak                    
jvm.threads.states                  
log4j2.events                       
process.cpu.usage                   
process.files.max                   
process.files.open                  
process.start.time                  
process.uptime                      
system.cpu.count                    
system.cpu.usage                    
system.load.average.1m              
tomcat.cache.access                 
tomcat.cache.hit                    
tomcat.connections.config.max       
tomcat.connections.current          
tomcat.connections.keepalive.current
tomcat.global.error                 
tomcat.global.received              
tomcat.global.request               
tomcat.global.request.max           
tomcat.global.sent                  
tomcat.servlet.error                
tomcat.servlet.request              
tomcat.servlet.request.max          
tomcat.sessions.active.current      
tomcat.sessions.active.max          
tomcat.sessions.alive.max           
tomcat.sessions.created             
tomcat.sessions.expired             
tomcat.sessions.rejected            
tomcat.threads.busy                 
tomcat.threads.config.max           
tomcat.threads.current              
```

frankgh (approved):

PR #547, opened 2021-02-09
Automation: add error that reproduces the segmentation fault

frankgh:
When an unterminated quoted field at the end of the file is being read
from a PXF external table with SEGMENT REJECT LIMIT for the table
definition, a segmentation fault was being thrown by PXF. This test
makes sure the segmentation fault error no longer occurs.

### DIFF
--- a/automation/src/test/java/org/greenplum/pxf/automation/features/hdfs/HdfsReadableTextTest.java
+++ b/automation/src/test/java/org/greenplum/pxf/automation/features/hdfs/HdfsReadableTextTest.java
@@ -22,6 +22,7 @@ import org.testng.annotations.Test;
 
 import java.io.File;
 import java.nio.charset.StandardCharsets;
+import java.util.List;
 
 import static org.greenplum.pxf.automation.features.tpch.LineItem.LINEITEM_SCHEMA;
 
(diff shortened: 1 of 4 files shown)


automation/src/test/java/org/greenplum/pxf/automation/features/hdfs/HdfsReadableTextTest.java
```diff
@@ -595,6 +600,32 @@ public void negativeBadTextData() throws Exception {
...
+     * When an unterminated quoted field at the end of the file is being read
+     * from a PXF external table with SEGMENT REJECT LIMIT for the table
+     * definition, a segmentation fault was being thrown by PXF. This test
+     * makes sure that the segfault does not occur.
+     */
+    @Test(groups = {"features", "gpdb", "hcfs", "security"})
+    public void unterminatedQuotedFieldAtEndOfFile() throws Exception {
+
+        Table smallDataTable = getSmallData("foo", 3);
+        List<String> firstRow = smallDataTable.getData().get(1);
```
denalex, line 613:
```suggestion
        List<String> secondRow = smallDataTable.getData().get(1);
```

denalex (approved):
thanks for adding the test. with the enhancement on C side, there will no longer be seg fault, but do we need to take a look at formatter code in C to maybe fix the issue that unterminated row causes there ?

frankgh:
We can look at the issue on the c-side

PR #551, opened 2021-02-16
Stop using jcenter and replace with maven central

frankgh:
Bintray is being shutdown, which includes jcenter hosting for java
artifacts. Using maven central

### DIFF
--- a/automation/pom.xml
+++ b/automation/pom.xml
@@ -26,8 +26,7 @@
 
     <repositories>
         <repository>
-            <id>pivotal-releases</id>
-            <name>pivotal-releases</name>
+            <id>test-dependencies</id>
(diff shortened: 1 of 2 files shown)


automation/pom.xml
```diff
@@ -26,8 +26,9 @@
 
     <repositories>
         <repository>
-            <id>jcenter</id>
-            <url>https://jcenter.bintray.com/</url>
+            <id>pivotal-releases</id>
+            <name>pivotal-releases</name>
```
denalex, line 30: I don't think we need `<name>` attribute and I would not call it `pivotal-releases`, maybe `test-dependencies` not to imply that it contains releases from Pivotal.

denalex (approved):

automation/pom.xml
frankgh: done

PR #552, opened 2021-02-16
Add PXF version headers to requests

bradfordb-vmware:
The server now checks for PXF version headers in RequestParser

- if headers are missing, we return an error message which provides a
  hint to upgrade the PXF exentsion
- if extension version does not match server version, we log a warning
- the values from the version headers are stored as fields on the
  RequestContext

Rename existing PXF_VERSION macro to PXF_PROTOCOL_VERSION to better
distinguish from the extension version. The PXF external table extension
now includes two additional HTTP headers:

- X-GP-PXF-VERSION
- X-GP-MIN-PXF-VERSION

[#176749393]


external-table/Makefile
```diff
@@ -7,6 +7,10 @@ REGRESS    = setup pxf pxfinvalid
 PG_CONFIG = pg_config
 PGXS := $(shell $(PG_CONFIG) --pgxs)
 
+PXF_EXTENSION_VERSION := $(shell cat ../version)
+MIN_PXF_EXTENSION_VERSION := 6.0.0-SNAPSHOT
+PG_CPPFLAGS := -DPXF_EXTENSION_VERSION=\"$(PXF_EXTENSION_VERSION)\" -DMIN_PXF_EXTENSION_VERSION=\"$(MIN_PXF_EXTENSION_VERSION)\"
+
```
denalex, line 13: another standing question is whether / when / how we should be versioning PXF C extension in `pxf.control` file. Changes there are very rare, but if we start having a catalog table with PXF, then it will become important. Right now that version is set to `1.0`

external-table/Makefile
```diff
@@ -7,6 +7,10 @@ REGRESS    = setup pxf pxfinvalid
 PG_CONFIG = pg_config
 PGXS := $(shell $(PG_CONFIG) --pgxs)
 
+PXF_EXTENSION_VERSION := $(shell cat ../version)
```
frankgh, line 10: this is confusing, this is actually not the `PXF_EXTENSION_VERSION` but rather the PXF server version that we used to compile the extension. I would rename this to `PXF-VERSION` --> version of PXF the .so is built with

external-table/Makefile
```diff
@@ -7,6 +7,10 @@ REGRESS    = setup pxf pxfinvalid
...
+MIN_PXF_EXTENSION_VERSION := 6.0.0-SNAPSHOT
```
frankgh, line 11: same here: it should probably be `MIN-PXF-VERSION` --> min version of PXF (server specifically) that is required by the .so file

external-table/src/pxfuriparser.h
```diff
@@ -26,10 +26,10 @@
 
 /*
  * Path constants for accessing PXF.
- * All PXF's resources are under /PXF_SERVICE_PREFIX/PXF_VERSION/...
+ * All PXF's resources are under /PXF_SERVICE_PREFIX/PXF_PROTOCOL_VERSION/...
  */
 #define PXF_SERVICE_PREFIX "pxf"
-#define PXF_VERSION "v15"		/* PXF version */
+#define PXF_PROTOCOL_VERSION "v15"		/* PXF version */
```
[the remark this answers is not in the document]
frankgh: I think a lot of the issues that @bradfordb-vmware and @ashuka24 are seeing because of breaking changes in the API, should go away after we bump this version

server/pxf-api/src/main/java/org/greenplum/pxf/api/model/RequestContext.java
```diff
@@ -282,6 +282,16 @@ public void setRequestType(RequestType requestType) {
      */
     private String user;
 
+    /**
+     * The version of the postgres extension client
+     */
+    private String extensionVersion;
```
frankgh, line 288: this is not the version of the extension, but rather the PXF Server version used to build the extension. the extension version has been 1.0 for a long time now

server/pxf-service/src/main/java/org/greenplum/pxf/service/HttpRequestParser.java
```diff
@@ -39,6 +40,17 @@
...
     private final PluginConf pluginConf;
+    private BuildProperties buildProperties;
+
+    /**
+     * Sets the {@link BuildProperties} object
+     *
+     * @param buildProperties the build properties object
+     */
+    @Autowired
+    public void setBuildProperties(BuildProperties buildProperties) {
```
frankgh, line 51: I think the preferred way to autowire these is through constructor parameters.

external-table/src/pxfbridge.c
```diff
@@ -136,7 +136,7 @@ build_uri_for_read(gphadoop_context *context)
 {
 	resetStringInfo(&context->uri);
 	appendStringInfo(&context->uri, "http://%s/%s/%s/Bridge/",
-					 get_authority(), PXF_SERVICE_PREFIX, PXF_VERSION);
+					 get_authority(), PXF_SERVICE_PREFIX, PXF_PROTOCOL_VERSION);
```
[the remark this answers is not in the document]
bradfordb-vmware: Should we make this `X-GP-PXF-API-VERSION` instead of `X-GP-API-VERSION`? I think the `X-GP-` prefix is common to PXF and gpfdist.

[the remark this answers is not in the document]
denalex: I'd say use `X-GP-API-VERSION` as this is the PXF extension, so it is clear this is PXF API we mean. Also, no other option names have PXF in it.

bradfordb-vmware:
This PR is still very much a work in progress, closing for now.

PR #553, opened 2021-02-16
Hardcode replicas in fragment

frankgh:


server/pxf-api/src/main/java/org/greenplum/pxf/api/utilities/FragmentsResponseFormatter.java
```diff
@@ -90,36 +85,10 @@ private static void updateFragmentIndex(List<Fragment> fragments) {
...
-                    convertedIp = addr.getHostAddress();
-                    hostToIpMap.put(host, convertedIp);
-                }
-
-                /* update IPs array */
-                ips[index] = convertedIp;
-                ++index;
-            }
-            fragment.setReplicas(ips);
+            fragment.setReplicas(new String[]{"127.0.0.1"});
```
ashuka24, line 91: why do we want to hardcode this here?

frankgh: this information is not used anywhere, and it's expensive to perform reverse dns lookups, so we save a little bit of time by skip doing unnecessary work

denalex (approved):

server/pxf-api/src/main/java/org/greenplum/pxf/api/utilities/FragmentsResponseFormatter.java
frankgh: I added a comment explaining why we are doing this.

PR #554, opened 2021-02-17
Fix different encoding when using LineRecordReader (PXF 5.16 backport)

frankgh:
LineRecordReader produces a org.apache.hadoop.io.Text value, that is
being converted to String by PXF. However, during the String conversion
the encoding information is not taken into account. In this commit we
skip conversion to String and just pass org.apache.hadoop.io.Text and
then finally just byte[]. This allows us to more efficiently transport
the data from external source to Greenplum without worrying about
converting to the appropriate encoding. The encoding conversion is
handled by Greenplum later. Backport of 67e008b89f6719e196f2c0893e2c986ddec6d98d

Restore automation test testing different encoding. Backport of 9eea79637d8d741bfb60b4cc5211cd50f65c1f8d


denalex (approved):
are we not changing existing behavior by now respecting newline definitions on CSV formats, can this potentially break existing workloads ?

frankgh:
@denalex  we are not changing existing behavior in 5.16. We only change behavior starting in 6.0

PR #561, opened 2021-02-19
Add base_branch for the pxf PR pipeline

frankgh:
PR's against the 5.16 branch should not be run through this PR pipeline.

### DIFF
--- a/concourse/pipelines/templates/pr_pipeline-tpl.yml
+++ b/concourse/pipelines/templates/pr_pipeline-tpl.yml
@@ -40,7 +40,7 @@ resources:
   type: pull-request
   source:
     repository: greenplum-db/pxf
-    base_branch: "master"
+    base_branch: {{pxf-git-branch}}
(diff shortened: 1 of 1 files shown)


frankgh:
I will open a follow up PR to add a new PR pipeline for `branch-5.16`.

concourse/pipelines/templates/pr_pipeline-tpl.yml
```diff
@@ -40,6 +40,7 @@ resources:
   type: pull-request
   source:
     repository: greenplum-db/pxf
+    base_branch: "master"
```
bradfordb-vmware, line 43: Minor nit: do we want to parameterize this value? The branch is already available in the Makefile.

bradfordb-vmware (approved):

concourse/pipelines/templates/pr_pipeline-tpl.yml
frankgh: good idea

PR #562, opened 2021-02-19
Add a PXF 5 PR pipeline

frankgh:


concourse/Makefile
```diff
@@ -7,7 +7,7 @@ BRANCH ?= $(shell git rev-parse --abbrev-ref HEAD)
 SHELL                        = /bin/bash
 BUILD_PIPELINE_NAME         ?= pxf-5-build
 DEV_BUILD_PIPELINE_NAME     ?= dev:$(USER)-$(BRANCH)
-PR_BUILD_PIPELINE_NAME      ?= pxf_pr_pipeline
+PR_BUILD_PIPELINE_NAME      ?= pxf_5_pr_pipeline
```
bradfordb-vmware, line 10: Minor nit: the base branch is `branch-5.16` and I think we plan to only support bug fixes/patch releases. Do we want to name the pipeline `pxf_5.16_pr_pipeline`?

concourse/Makefile
```diff
@@ -7,7 +7,7 @@ BRANCH ?= $(shell git rev-parse --abbrev-ref HEAD)
...
 CERTIFICATION_PIPELINE_NAME ?= pxf-certification
```
bradfordb-vmware, line 11: I don't know if we plan to have set pipelines for certification/perf/etc off this branch. Do we want to rename these now on the off chance that we do end up needing them? Or de want to "poison" those targets to prevent accidentally setting them over the main pipelines?

bradfordb-vmware (approved):

concourse/Makefile
frankgh: well, this assumes that we will not release a 5.17, which is desired, but not necessarily decided yet

concourse/Makefile
frankgh: we have a 5 build pipeline, which has already been renamed. We don't have a certification pipeline yet, but I think we probably should have it for other pipelines

frankgh: Adding the `pxf-5-certification` pipeline in a different PR

PR #563, opened 2021-02-19
Stop using jcenter and replace with maven central (Backport PXF 5)

frankgh:
Bintray is being shutdown, which includes jcenter hosting for java
artifacts. Using maven central


bradfordb-vmware (approved):
Verified that repo.pivotal.io can be reached from the public internet.

PR #548, opened 2021-02-09
Update pxf cli cluster tool to support PXF on master

ashuka24:
This PR modifies pxf cluster start, stop, restart and status to support the master as well as the standby master host.

PR #566, opened 2021-02-19
docs - pxf service now runs on master/standby

lisakowen:
misc doc updates for https://github.com/greenplum-db/pxf/pull/548.  in this PR:
- update pxf [cluster] ref pages to note that master/standby also affected by start/stop/restart/status
- remove "segment" from certain references to "segment host", replace with "Greenplum" in some cases

questions:
1.  is the architecture still the same?  i.e. even though we are running a pxf service on master/standby, the segment hosts handle the data movement?  what does the PXF service running on the master do?
2.  i assume that the following are required only on/for the segment hosts, and didn't update "segment host" references:
    - avro schema file 
    - generate kerberos principals/keytabs
    - nfs mount points
    - jdbc partitioning specifics
3.   the "memory and threading" page has many references to segment host.  will the oom/thread config still apply only to segment hosts
4.   the troubleshooting page identifies some errors that reference segment host.  we need to verify that the error message currently listed still exist in pxf v6.0.0;  i plan to deal with this page in a separate PR.


bradfordb-vmware (approved):
Looks good to me.

frankgh (approved):

ashuka24 (approved):

lisakowen:
i updated the (relevant) remaining segment host occurrences.  i am not totally certain about the changes to the kerberos section.  the user is supposed to generate principals and keytabs now for all greenplum systems, right?

denalex (approved):

denalex:
> i updated the (relevant) remaining segment host occurrences. i am not totally certain about the changes to the kerberos section. the user is supposed to generate principals and keytabs now for all greenplum systems, right?

right, they will have to generate principals / keytabs for master and standby now as well -- maybe also important to mention in upgrade notes

docs/content/pxf_kerbhdfs.html.md.erb
```diff
@@ -292,6 +292,6 @@ When you configure PXF for secure HDFS using an MIT Kerberos KDC server, you wil
 9. Synchronize the PXF configuration to your Greenplum Database cluster:
 
     ``` shell
-    gpadmin@seghost$ pxf cluster sync
+    gpadmin@gpmaster$ pxf cluster sync
```
ashuka24, line 295: in some of the other pr's it looks like we are using `gphost` instead of `gpmaster`. is that change also needed here?

ashuka24 (approved):

docs/content/pxf_kerbhdfs.html.md.erb
lisakowen: i tried to use gphost when the host it could be run on was generic (i.e. any of the greenplum hosts).  and gpmaster when it should explicitly be run on the master, like the cluster sync call.  let me know if you notice anything different.

lisakowen:
merging, will address any further comments in a new PR.

docs/content/pxf_kerbhdfs.html.md.erb
```diff
@@ -184,13 +184,13 @@ When you configure PXF for secure HDFS using an MIT Kerberos KDC server, you wil
     root@kdc-server$ 
     ```
 
-2. Distribute the `/etc/krb5.conf` Kerberos configuration file on the KDC server host to **each** segment host in your Greenplum Database cluster if not already present. For example:
+2. Distribute the `/etc/krb5.conf` Kerberos configuration file on the KDC server host to **each** host in your Greenplum Database cluster if not already present. For example:
 
     ``` shell
-    root@kdc-server$ scp /etc/krb5.conf seghost:/etc/krb5.conf
+    root@kdc-server$ scp /etc/krb5.conf gphost:/etc/krb5.conf
```
dyozie, line 190: Should "gphost" be variablized as \<gphost> here?

PR #567, opened 2021-02-20
refactored REST / service layer

denalex:
Refactored REST resource logic into a resource and Spring services that can be mocked.


server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/PxfLegacyResource.java
```diff
@@ -0,0 +1,91 @@
...
+        this.writeService = writeService;
+    }
+
+    /**
+     * REST endpoint for read data requests.
+     *
+     * @param headers http headers from request that carry all parameters
+     * @return response object containing stream that will output records
+     */
+    @GetMapping(value = "/Bridge", produces = MediaType.APPLICATION_OCTET_STREAM_VALUE)
```
frankgh, line 56: maybe add the fragmenter call too?

server/pxf-service/src/test/java/org/greenplum/pxf/service/rest/PxfResourceIT.java
```diff
@@ -0,0 +1,96 @@
+package org.greenplum.pxf.service.rest;
```
frankgh, line 1: i wonder if we should create an it project `pxf-service/src/it/java/org...`

frankgh (commented):
I started reviewing but noticed this is a Draft, let me know when it's ready for review

server/pxf-service/src/main/java/org/greenplum/pxf/service/controller/ReadResponseImpl.java
```diff
@@ -25,7 +23,10 @@
 import java.util.List;
 import java.util.Map;
 
-public class BridgeResponse implements StreamingResponseBody {
+/**
+ * Implementation of the ReadResponse.
+ */
+public class ReadResponseImpl implements ReadResponse {
```
frankgh, line 29: I wonder if by implementing a different interface we won't be using the `async` thread pool? With `StreamingResponseBody`, we were using the `async` pool

denalex: we still use `StreamingResponseBody` in the Resource by wrapping `ReadResponse` into a lambda. This way `StreamingResponseBody` which is a very specific Spring MVC interface does not leak into the services layer.

server/pxf-service/src/test/java/org/greenplum/pxf/service/rest/PxfResourceIT.java
denalex: I think we can still keep some IT tests like these alongside the regular tests in the respective packages if they test REST endpoints / metrics. Once we have e2e scenarios to test, we can have an `integration` directory under `pxf-service/src/test/o/g/p/s/integration`. Finally, we can have a separate gradle profile to run them selectively based on the pattern `*IT`. But all of this is not needed in this PR, IMO.

bradfordb-vmware (approved):

server/pxf-service/src/main/java/org/greenplum/pxf/service/controller/BaseServiceImpl.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    /**
+     * Returns a new Bridge instance based on the current context.
+     *
+     * @param context request context
+     * @return an instance of the bridge to use
+     */
+    protected Bridge getBridge(RequestContext context) {
+        return bridgeFactory.getBridge(context);
```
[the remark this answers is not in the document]
denalex: not sure what you mean by collapse, but I extracted this call from subclasses and pushed here to make `bridgeFactory` a private final member of the base class, subclasses just pass it to `super(..)` in their constructors and forget about it.

server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/PxfLegacyResource.java
```diff
@@ -0,0 +1,102 @@
...
+
+    /**
+     * REST endpoint for getting a list of fragments.
+     * @param headers http headers from request that carry all parameters
+     * @return response
+     */
+    @GetMapping(value = "/Fragmenter/getFragments", produces = MediaType.APPLICATION_JSON_VALUE)
+    public ResponseEntity<String> getFragments(@RequestHeader MultiValueMap<String, String> headers) {
+        throw new PxfRuntimeException("getFragments API (v15) is no longer supported by the server",
+                "Upgrade PXF client library, did you run 'pxf register' ?");
```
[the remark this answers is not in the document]
denalex: So, should we embed the hint into the message itself ? `getFragments API (v15) is no longer supported by the server, upgrade PXF extension (run 'pxf [cluster] register' and then 'ALTER EXTENSION pxf UPDATE')"

frankgh (approved):

PR #569, opened 2021-02-24
Bump PXF external-table extension to 2.0 and test upgrade

frankgh:
Add a test that creates a database `testupgrade` before upgrading to PXF 6. This will surface an issue where the current extension will point to the old `.so` library location that was originally registered in the `pg_proc` catalog table. When installing the new extension, we need to run a `ALTER EXTENSION pxf UPDATE` to register the new extension (`pxf.so`) location. To be able to run an `ALTER EXTENSION` we need to bump up the version of the PXF external-table extension. The upgrade script will recreate the pxf functions with the updated `MODULE_PATHNAME`.

### DIFF
--- a/concourse/scripts/pxf_common.bash
+++ b/concourse/scripts/pxf_common.bash
@@ -8,6 +8,7 @@ PXF_VERSION=${PXF_VERSION:=6}
 PROXY_USER=${PROXY_USER:-pxfuser}
 PROTOCOL=${PROTOCOL:-}
 GOOGLE_PROJECT_ID=${GOOGLE_PROJECT_ID:-data-gpdb-ud}
+PXF_SRC=$(find /tmp/build -name pxf_src -type d)
 
 # on purpose do not call this PXF_CONF|PXF_BASE so that it is not set during pxf operations
 if [[ ${PXF_VERSION} == 5 ]]; then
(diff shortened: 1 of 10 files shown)


frankgh:
After discussing offline with @denalex we decided not to run the smoke test for 5, but rather create the pxf extension for the `pxfautomation` database. This would give us equivalent testing

concourse/scripts/pxf_common.bash
```diff
@@ -8,6 +8,7 @@ PXF_VERSION=${PXF_VERSION:=6}
 PROXY_USER=${PROXY_USER:-pxfuser}
 PROTOCOL=${PROTOCOL:-}
 GOOGLE_PROJECT_ID=${GOOGLE_PROJECT_ID:-data-gpdb-ud}
+PXF_SRC=$(find /tmp/build -name pxf_src -type d)
```
denalex, line 11: was this really necessary ? `/tmp/build` is very specific to concourse

denalex (approved):

concourse/scripts/pxf_common.bash
frankgh: not really, but we had some relative paths which were less intuitive in my opinion, for example

```
	echo "$export_pythonpath" >> "${PXF_COMMON_SRC_DIR}/../../automation/tinc/main/tinc_env.sh"
```

and `find /tmp/build` is all over our scripts anyway, so we are very concourse-specific. I can revert this since it was for a different purpose that we are not using anymore

frankgh: actually, we were doing this before in this script anyway, so I will just keep it like this:

```
			PXF_SRC_DIR=$(find /tmp/build/ -name pxf_src)
```

PR #570, opened 2021-02-25
Add tracing to PXF with Spring Sleuth

frankgh:
Add X-B3-TraceId and X-B3-SpanId to client and log the traces on the server logs

### DIFF
--- a/server/gradle.properties
+++ b/server/gradle.properties
@@ -27,6 +27,7 @@ pxfProtocolVersion=v15
 parquetVersion=1.11.1
 awsJavaSdk=1.11.801
 spring_boot_version=2.4.3
+springCloudVersion=2020.0.1
 org.gradle.daemon=true
 org.gradle.parallel=false
 # ORC versions 1.6.x have breaking changes with
--- a/server/pxf-service/src/main/resources/application.properties
+++ b/server/pxf-service/src/main/resources/application.properties
@@ -8,12 +8,17 @@ management.endpoints.web.exposure.include=health,info,shutdown,metrics,prometheu
 management.endpoint.shutdown.enabled=true
 management.endpoint.health.probes.enabled=true
 
+spring.application.name=pxf-service
+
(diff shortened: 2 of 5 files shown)


server/pxf-service/src/main/resources/application.properties
```diff
@@ -8,13 +8,18 @@ management.endpoints.web.exposure.include=health,info,shutdown,metrics,prometheu
...
+
 # common tags applied to all metrics
 management.metrics.tags.application=pxf-service
 
 spring.profiles.active=default
 
 server.port=${pxf.port:5888}
 
+# sleuth
+spring.sleuth.trace-id128=true
```
bradfordb-vmware, line 21: Do we want to leave a comment here explaining why we decided to use 128-bit trade IDs?

bradfordb-vmware (approved):

server/gradle.properties
```diff
@@ -27,6 +27,7 @@ pxfProtocolVersion=v15
 parquetVersion=1.11.1
 awsJavaSdk=1.11.801
 spring_boot_version=2.4.3
+springCloudVersion=2020.0.1
```
ashuka24, line 30: we seem to be inconsistent with using underscores, camelCase vs . between words. Is there a method to the madness that I'm missing or would it be possible to make these all consistent?

server/pxf-service/src/main/resources/application.properties
```diff
@@ -8,13 +8,18 @@ management.endpoints.web.exposure.include=health,info,shutdown,metrics,prometheu
 management.endpoint.shutdown.enabled=true
 management.endpoint.health.probes.enabled=true
 
+spring.application.name=pxf-service
```
ashuka24, line 11: why is this needed now and not needed before?

frankgh:
Closing for the moment until we concretely define use cases, and ways for users to consume tracing information.

PR #571, opened 2021-02-25
Add PXF version header to requests

bradfordb-vmware:
The server now checks for a PXF API version header in RequestParser

- if the header is missing, an error message is returned which provides
  a hint to upgrade the PXF exentsion
- if the API version reported by the extension does not match with the
  API version on the server, an error message with the hint to upgrade
  the extension is returned
  - in future revisions of the API, the server may support clients
    (e.g., extension, Spark connector, etc.) using an older API version
- the value of the API version being used by the extension is stored as
  a field on the RequestContext

[#176749393]


server/pxf-service/src/main/resources/application.properties
```diff
@@ -40,6 +40,9 @@ pxf.task.pool.core-size=8
 pxf.task.pool.max-size=${pxf.max.threads:200}
 pxf.task.pool.queue-capacity=0
 
+# this must match with `PXF_API_VERSION` in external-table/Makefile
+pxf.api.version=16
```
frankgh, line 44: I'm not sure if this is the right approach, anyone is able to change this value by setting `pxf.api.version` in the `pxf-application.properties` file. And PXF would stop working. I think this value should be built into the application (build time value), and it should not be allowed to change. Maybe Spring Boot approach is not the right approach for this, but rather some approach with gradle maybe?

server/pxf-service/src/main/java/org/greenplum/pxf/service/HttpRequestParser.java
```diff
@@ -457,6 +472,22 @@ private String removeProperty(String property) {
...
+         *
+         * @param property the lookup property key
+         * @param errMsgHint hint for user to include in error message
+         * @throws IllegalArgumentException if property key is missing
+         */
+        private String removeProperty(String property, String errMsgHint) {
+            try {
+                return removeProperty(property);
+            } catch (IllegalArgumentException e) {
+                throw new IllegalArgumentException(String.format("%s. %s", e.getMessage(), errMsgHint));
```
frankgh, line 487: Maybe use `PxfRuntimeException`, which supports hints, and the hints are displayed by the pxf extension in the psql session (or any client session).

external-table/Makefile
```diff
@@ -7,6 +7,10 @@ REGRESS    = setup pxf pxfinvalid
 PG_CONFIG = pg_config
 PGXS := $(shell $(PG_CONFIG) --pgxs)
 
+# the value of PXF_API_VERSION should be kept in sync with `pxf.api.version` in application.properties
+PXF_API_VERSION := 16
```
frankgh, line 11: I wonder if we should read this value from a file (assuming it's easy to do it from the Makefile)

external-table/src/pxfbridge.c
```diff
@@ -135,8 +135,8 @@ static void
 build_uri_for_read(gphadoop_context *context)
 {
 	resetStringInfo(&context->uri);
-	appendStringInfo(&context->uri, "http://%s/%s/%s/Bridge/",
-					 get_authority(), PXF_SERVICE_PREFIX, PXF_VERSION);
+	appendStringInfo(&context->uri, "http://%s/%s/v15/Bridge/",
```
frankgh, line 138: why v15? shouldn't we use the new endpoints? The legacy endpoints should return an error (like the legacy `getFragments` endpoint).

external-table/src/pxfbridge.c
```diff
@@ -152,8 +152,8 @@ build_uri_for_read(gphadoop_context *context)
 static void
 build_uri_for_write(gphadoop_context *context)
 {
-	appendStringInfo(&context->uri, "http://%s/%s/%s/Writable/stream",
-					 get_authority(), PXF_SERVICE_PREFIX, PXF_VERSION);
+	appendStringInfo(&context->uri, "http://%s/%s/v15/Writable/stream",
```
frankgh, line 155: same here, we should use the new endpoints

external-table/test/pxfbridge_test.c
```diff
@@ -119,8 +119,8 @@ test_gpbridge_import_start(void **state)
 
 	initStringInfo(&expected_uri);
 	appendStringInfo(&expected_uri,
-					 "http://%s/%s/%s/Bridge/",
-					 AUTHORITY, PXF_SERVICE_PREFIX, PXF_VERSION);
+					 "http://%s/%s/v15/Bridge/",
```
frankgh, line 122: are these tests working?

server/pxf-service/src/main/java/org/greenplum/pxf/service/HttpRequestParser.java
```diff
@@ -65,6 +68,14 @@ public RequestContext parseRequest(MultiValueMap<String, String> requestHeaders,
 
         // fill the Request-scoped RequestContext with parsed values
 
+        context.setExtensionApiVersion(
+                params.removeProperty("PXF-API-VERSION", "Ensure PXF extension has been updated to the latest version."));
```
frankgh, line 72: I think there is a similar hint in getFragments from the legacy endpoints, maybe make this message the same for consistentcy?

server/pxf-service/src/main/java/org/greenplum/pxf/service/HttpRequestParser.java
```diff
@@ -65,6 +68,14 @@ public RequestContext parseRequest(MultiValueMap<String, String> requestHeaders,
...
+
+        if (!PxfApiVersionChecker.isCompatible(context.getExtensionApiVersion(), getServerApiVersion())) {
+            LOG.warn("server API version: {}. extension API version: {}", getServerApiVersion(), context.getExtensionApiVersion());
```
frankgh, line 75: no need to log if we are throwing an exception, maybe add the version numbers to the exception itself?

server/pxf-service/src/main/java/org/greenplum/pxf/service/PxfApiVersionChecker.java
```diff
@@ -0,0 +1,10 @@
+package org.greenplum.pxf.service;
+
+import org.apache.commons.lang.StringUtils;
+
+public class PxfApiVersionChecker {
+
+    public static boolean isCompatible(String version1, String version2) {
```
frankgh, line 7: should we move this to `Utilities` with a descriptive name? If this class evolves into more complexity maybe it's better to keep it here.

external-table/Makefile
denalex: yes, should use similar mechanism of how build version is propagated

external-table/src/pxfbridge.c
denalex: yes, please rebase off master, use new endpoints and make old ones obsolete as part of this PR.

server/pxf-api/src/main/java/org/greenplum/pxf/api/model/RequestContext.java
```diff
@@ -294,6 +294,11 @@ public void setRequestType(RequestType requestType) {
      */
     private Charset databaseEncoding;
 
+    /**
+     * The version of the PXF API the postgres extension implements
+     */
+    private String extensionApiVersion;
```
denalex, line 300:
```suggestion
    private String clientApiVersion;
```

server/pxf-service/src/main/java/org/greenplum/pxf/service/HttpRequestParser.java
```diff
@@ -65,6 +68,14 @@ public RequestContext parseRequest(MultiValueMap<String, String> requestHeaders,
 
         // fill the Request-scoped RequestContext with parsed values
 
+        context.setExtensionApiVersion(
```
denalex, line 71:
```suggestion
        context.setClientApiVersion(
```

server/pxf-service/src/main/java/org/greenplum/pxf/service/HttpRequestParser.java
denalex: true, but if the message will be presented to a client that does not support hints, we still need the hint inside the main message.

server/pxf-service/src/main/java/org/greenplum/pxf/service/HttpRequestParser.java
```diff
@@ -65,6 +68,14 @@ public RequestContext parseRequest(MultiValueMap<String, String> requestHeaders,
...
+                params.removeProperty("PXF-API-VERSION", "Ensure PXF extension has been updated to the latest version."));
+
+        if (!PxfApiVersionChecker.isCompatible(context.getExtensionApiVersion(), getServerApiVersion())) {
```
denalex, line 74: using static methods is an anti-pattern in Spring, and it is very hard to unit test, define a `@Component` service and inject it here.

server/pxf-service/src/main/java/org/greenplum/pxf/service/PxfApiVersionChecker.java
```diff
@@ -0,0 +1,10 @@
+package org.greenplum.pxf.service;
+
+import org.apache.commons.lang.StringUtils;
+
```
denalex, line 4:
```suggestion
@Component
```

server/pxf-service/src/main/java/org/greenplum/pxf/service/PxfApiVersionChecker.java
```diff
@@ -0,0 +1,10 @@
...
+public class PxfApiVersionChecker {
+
+    public static boolean isCompatible(String version1, String version2) {
+        return StringUtils.equals(version1, version2);
```
denalex, line 8:
```suggestion
        return StringUtils.equalsIgnoreCase(version1, version2);
```
just in case

server/pxf-service/src/test/java/org/greenplum/pxf/service/PxfApiVersionCheckerTest.java
```diff
@@ -0,0 +1,30 @@
...
+                Arguments.of(null, "1.0.0", false),
+                Arguments.of("1.0.0", null, false),
+                Arguments.of(null, null, true)
+        );
+    }
+
+    @ParameterizedTest
+    @MethodSource("provideVersions")
+    public void testApiVersionsMatch(String extensionApiVersion, String serverApiVersion, boolean expected) {
+        assertEquals(expected, PxfApiVersionChecker.isCompatible(extensionApiVersion, serverApiVersion));
```
denalex, line 28: while this is cool, having 7 assert statements would result in less lines of code and complexity

server/pxf-service/src/main/java/org/greenplum/pxf/service/HttpRequestParser.java
frankgh: 5 clients will hit the legacy endpoints (where the hints should be inlined), but this error will be presented to 6+ clients that support hints

denalex: yes, good point

server/pxf-service/src/main/resources/application.properties
ashuka24: My understanding is that they would not be able to override it with a value from pxf-application.properties unless we had something similar to what we do in line 40 for `pxf.max.threads`? Is this an incorrect assumption?

bradfordb-vmware: > My understanding is that they would not be able to override it with a value from pxf-application.properties unless we had something similar to what we do in line 40 for `pxf.max.threads`? Is this an incorrect assumption?

I had the same assumption too but it is incorrect. The property name "mapping" is more for presenting a unified interface to the user.

external-table/test/pxfbridge_test.c
bradfordb-vmware: No, I don't think these tests have worked since PXF was moved from the GPDB repo. We can revert the changes in this file for now. We can open a follow-up PR to remove these tests altogether.

server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/PxfLegacyResource.java
```diff
@@ -93,14 +84,6 @@ public PxfLegacyResource(RequestParser<MultiValueMap<String, String>> parser,
...
-
-        // parse incoming HTTP request
-        RequestContext context = parser.parseRequest(headers, RequestType.WRITE_BRIDGE);
-
-        // write data and get a response message
-        String returnMsg = writeService.writeData(context, request.getInputStream());
-
-        // send the response to the client
-        return new ResponseEntity<>(returnMsg, HttpStatus.OK);
+        throw new PxfRuntimeException(String.format(ERROR_MESSAGE_TEMPLATE, "stream"), ERROR_MESSAGE_HINT);
```
frankgh, line 87: maybe use `Writable` or `Writable/stream` here instead, otherwise it might be difficult for a user reading the message what endpoint we are talking about

server/gradle.properties
```diff
@@ -23,7 +23,7 @@ hiveVersion=2.3.7
 hiveStorageApiVersion=2.7.2
 hbaseVersion=1.3.2
 junitVersion=4.11
-pxfProtocolVersion=v15
+api_version=0
```
denalex, line 26: maybe moving this line to the top right after version one will be better

server/pxf-service/src/main/java/org/greenplum/pxf/service/HttpRequestParser.java
```diff
@@ -426,6 +452,22 @@ private String removeProperty(String property) {
             return result;
         }
 
+        /**
+         * Returns the value to which the specified property is mapped and
+         * removes it from the map
+         *
+         * @param property the lookup property key
+         * @param errMsgHint hint for user to include in error message
+         * @throws IllegalArgumentException if property key is missing
```
denalex, line 461: that doc statement is not accurate, as this method will  wrap IllegalArgumentException

server/pxf-service/src/test/java/org/greenplum/pxf/service/HttpRequestParserTest.java
```diff
@@ -58,10 +60,16 @@
...
+    private BuildProperties mockBuildProperties;
+    private PxfApiVersionChecker apiVersionChecker;
 
     @BeforeEach
     public void setUp() {
         mockPluginConf = mock(PluginConf.class);
+        mockBuildProperties = mock(BuildProperties.class);
+        when(mockBuildProperties.get("pxfApiVersion")).thenReturn("16");
+
+        apiVersionChecker = new PxfApiVersionChecker();
```
denalex, line 72: strictly speaking, this needs to be mocked as well

server/pxf-service/src/test/java/org/greenplum/pxf/service/HttpRequestParserTest.java
```diff
@@ -625,6 +635,23 @@ public void testMissingResolver() {
...
+    }
+
+    @Test
+    public void testDifferentPxfApiVersions() {
+        parameters.set("X-GP-PXF-API-VERSION", "15");
+        Exception e = assertThrows(PxfRuntimeException.class,
+                () -> parser.parseRequest(parameters, RequestType.READ_BRIDGE));
+        assertEquals("API version mismatch; server implements v16 and client implements v15", e.getMessage());
+    }
+
```
denalex, line 654: I think we should also add a positive test case that asserts the getApiVersion() returns 16.

server/pxf-service/src/test/java/org/greenplum/pxf/service/PxfApiVersionCheckerTest.java
```diff
@@ -0,0 +1,29 @@
...
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+public class PxfApiVersionCheckerTest {
+
+    private PxfApiVersionChecker apiVersionChecker;
+
+    @BeforeEach
+    public void setUp() {
```
denalex, line 13: with only one test method you do not need setUp()

server/pxf-service/src/test/java/org/greenplum/pxf/service/rest/PxfResourceIT.java
```diff
@@ -76,30 +76,28 @@ public void testLegacyFragmenterEndpoint() throws Exception {
...
     @Test
     public void testLegacyBridgeEndpoint() throws Exception {
-        //TODO: legacy endpoint should throw 500 exception with a hint, validate error message
         when(mockParser.parseRequest(any(), eq(RequestContext.RequestType.READ_BRIDGE))).thenReturn(mockContext);
 
-        ResultActions result = mvc.perform(get("/pxf/v15/Bridge")).andExpect(status().isOk());
+        ResultActions result = mvc.perform(get("/pxf/v15/Bridge")).andExpect(status().is5xxServerError());
         Thread.sleep(200);
-        result.andExpect(content().string("Hello from read!"));
+        result.andExpect(content().string("upgrade PXF extension (run 'pxf [cluster] register' and then 'ALTER EXTENSION pxf UPDATE')"));
```
denalex, line 90: no reason to sleep now for this case, there's no async processing

server/pxf-service/src/test/java/org/greenplum/pxf/service/rest/PxfResourceIT.java
```diff
@@ -76,30 +76,28 @@ public void testLegacyFragmenterEndpoint() throws Exception {
...
         result.andExpect(content().string("upgrade PXF extension (run 'pxf [cluster] register' and then 'ALTER EXTENSION pxf UPDATE')"));
     }
 
     @Test
     public void testLegacyBridgeEndpoint() throws Exception {
-        //TODO: legacy endpoint should throw 500 exception with a hint, validate error message
         when(mockParser.parseRequest(any(), eq(RequestContext.RequestType.READ_BRIDGE))).thenReturn(mockContext);
 
-        ResultActions result = mvc.perform(get("/pxf/v15/Bridge")).andExpect(status().isOk());
+        ResultActions result = mvc.perform(get("/pxf/v15/Bridge")).andExpect(status().is5xxServerError());
```
denalex, line 88: can we get specific 500 (server error) instead of 5xx series ?

server/gradle.properties
frankgh: should we use camel case for consistency?

server/pxf-service/src/main/java/org/greenplum/pxf/service/HttpRequestParser.java
```diff
@@ -77,6 +91,14 @@ public RequestContext parseRequest(MultiValueMap<String, String> requestHeaders,
 
         // fill the Request-scoped RequestContext with parsed values
 
+        context.setClientApiVersion(
+                params.removeProperty("PXF-API-VERSION", ERROR_MESSAGE_HINT));
+
+        if (!apiVersionChecker.isCompatible(context.getClientApiVersion(), getServerApiVersion())) {
+            throw new PxfRuntimeException(
+                    String.format(ERROR_MESSAGE_TEMPLATE, getServerApiVersion(), context.getClientApiVersion()));
```
frankgh, line 99: let's add the hint to the exception here:
```suggestion
            throw new PxfRuntimeException(
                    String.format(ERROR_MESSAGE_TEMPLATE, getServerApiVersion(), context.getClientApiVersion()), ERROR_MESSAGE_HINT);
```

server/pxf-service/src/main/java/org/greenplum/pxf/service/PxfApiVersionChecker.java
```diff
@@ -0,0 +1,12 @@
+package org.greenplum.pxf.service;
+
+import org.apache.commons.lang.StringUtils;
+import org.springframework.stereotype.Component;
+
+@Component
```
frankgh, line 6: maybe a javadoc here and the method?

server/pxf-service/src/main/java/org/greenplum/pxf/service/HttpRequestParser.java
```diff
@@ -249,7 +249,7 @@ public RequestContext parseRequest(MultiValueMap<String, String> requestHeaders,
         return context;
     }
 
-    private String getServerApiVersion() {
+    public String getServerApiVersion() {
```
frankgh, line 252: why public?

frankgh: oh, is this for testing? maybe just make it package private?

frankgh (approved):

server/pxf-service/src/test/java/org/greenplum/pxf/service/HttpRequestParserTest.java
```diff
@@ -625,6 +637,31 @@ public void testMissingResolver() {
...
+    @Test
+    public void testGetApiVersion() {
+        when(mockBuildProperties.get("pxfApiVersion")).thenReturn("42");
+
+        assertEquals("42", parser.getServerApiVersion());
+    }
+
```
denalex, line 664: I guess what I meant is to `assertEquals("16", context.getClientApiVersion)` after the request is parsed.

denalex (approved):

PR #572, opened 2021-02-26
Added custom tags for MVC metrics

denalex:
PXF is now capable of applying custom tags to metrics reported by Spring MVC. These tags are: `user`, `segment`, `profile`, `server` and will be applied to metrics such as `http.server.requests`.

However, due to a problem with Prometheus integration (https://github.com/micrometer-metrics/micrometer/issues/2399), these tags have to be applied to all endpoints, so the actuator endpoints that receive calls not from Greenplum, will have these tags applied with the values `unknown`.


server/pxf-service/src/main/java/org/greenplum/pxf/service/PxfConfiguration.java
```diff
@@ -100,4 +112,42 @@ public ThreadPoolTaskExecutor pxfApplicationTaskExecutor(PxfServerProperties pxf
...
+     * @return the {@link WebMvcTagsContributor} instance
+     */
+    @Bean
+    public WebMvcTagsContributor webMvcTagsContributor() {
+        return new WebMvcTagsContributor() {
+            @Override
+            public Iterable<Tag> getTags(HttpServletRequest request, HttpServletResponse response, Object handler, Throwable exception) {
+                // add tags only if enabled and the request is for PXF service endpoints (not actuator ones)
+                if (!customTagsEnabled || StringUtils.isBlank(request.getHeader("X-GP-USER"))) {
+                    return EMPTY_TAGS;
```
frankgh, line 128: why not just `Tags.empty()` which is probably already an object that was built once, no?

server/pxf-service/src/main/java/org/greenplum/pxf/service/PxfConfiguration.java
```diff
@@ -100,4 +112,42 @@ public ThreadPoolTaskExecutor pxfApplicationTaskExecutor(PxfServerProperties pxf
...
+                tags = tags.and("user", request.getHeader("X-GP-USER"));
+                tags = addTag("segment", request.getHeader("X-GP-SEGMENT-ID"), tags, null);
+                tags = addTag("profile", request.getHeader("X-GP-OPTIONS-PROFILE"), tags, null);
+                tags = addTag("server", request.getHeader("X-GP-OPTIONS-SERVER"), tags, "default");
+                return tags;
+            }
+
+            @Override
+            public Iterable<Tag> getLongRequestTags(HttpServletRequest request, Object handler) {
+                return new ArrayList<>();
```
frankgh, line 140: did you try with long requests tags ? maybe there is no issue if we mark our controller as long request?

server/pxf-service/src/test/java/org/greenplum/pxf/service/rest/PxfResourceIT.java
```diff
@@ -51,11 +53,6 @@
     @Mock
     private RequestContext mockContext;
 
-    @BeforeAll
-    public static void init() {
-        System.setProperty("pxf.logdir", "/tmp");
```
frankgh, line 56: yay! did you want to default pxf.logdir to /tmp as part of this PR?

server/pxf-service/src/main/java/org/greenplum/pxf/service/PxfConfiguration.java
frankgh: Actually I think our requests are long requests, and we should probably mark them as such, and we should probably implement this method instead. Both read and write are considered long requests according to the documentation. All other requests (actuator, health) won't have the headers anyway, so nothing to add there.

server/pxf-service/src/main/java/org/greenplum/pxf/service/PxfConfiguration.java
```diff
@@ -27,23 +36,26 @@
...
 public class PxfConfiguration implements WebMvcConfigurer {
 
-    private static final Logger LOG = LoggerFactory.getLogger(PxfConfiguration.class);
-
     /**
      * Bean name of PXF's {@link TaskExecutor}.
      */
     public static final String PXF_RESPONSE_STREAM_TASK_EXECUTOR = "pxfResponseStreamTaskExecutor";
+    private static final Logger LOG = LoggerFactory.getLogger(PxfConfiguration.class);
+    private static final Tags EMPTY_TAGS = Tags.empty();
```
frankgh, line 44: We don't need to do this: https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/main/java/io/micrometer/core/instrument/Tags.java#L275

server/pxf-service/src/main/java/org/greenplum/pxf/service/PxfConfiguration.java
denalex: I don't yet understand the difference, I can't believe the only difference would be some arbitrary duration boundary, need more research there.

server/pxf-service/src/test/java/org/greenplum/pxf/service/rest/PxfResourceIT.java
denalex: only defaulted for my tests, I can rebase and remove this altogether and default to /tmp in `application.properties`

server/pxf-service/src/main/java/org/greenplum/pxf/service/PxfConfiguration.java
```diff
@@ -100,4 +107,45 @@ public ThreadPoolTaskExecutor pxfApplicationTaskExecutor(PxfServerProperties pxf
...
+            }
+
+            @Override
+            public Iterable<Tag> getLongRequestTags(HttpServletRequest request, Object handler) {
+                return new ArrayList<>();
+            }
+
+            private Tags addTag(String tag, String value, Tags tags, String defaultValue) {
+                value = StringUtils.defaultIfBlank(value, defaultValue);
+                if (StringUtils.isNotBlank(value)) {
```
frankgh, line 143: this check is no longer necessary

server/pxf-service/src/main/java/org/greenplum/pxf/service/PxfConfiguration.java
```diff
@@ -100,4 +107,45 @@ public ThreadPoolTaskExecutor pxfApplicationTaskExecutor(PxfServerProperties pxf
...
+                tags = addTag("user", request.getHeader("X-GP-USER"), tags, UNKNOWN_VALUE);
+                tags = addTag("segment", request.getHeader("X-GP-SEGMENT-ID"), tags, UNKNOWN_VALUE);
+                tags = addTag("profile", request.getHeader("X-GP-OPTIONS-PROFILE"), tags, UNKNOWN_VALUE);
+                tags = addTag("server", request.getHeader("X-GP-OPTIONS-SERVER"), tags, defaultServer);
+                return tags;
+            }
+
+            @Override
+            public Iterable<Tag> getLongRequestTags(HttpServletRequest request, Object handler) {
+                return new ArrayList<>();
```
frankgh, line 138: can we return `Tags.empty()` here, if we are not going to use it?

frankgh (approved):

server/pxf-service/src/main/java/org/greenplum/pxf/service/PxfConfiguration.java
denalex: ah, yes, thank you for the catch, now I can remove the private method !

bradfordb-vmware (approved):

frankgh (approved):
it looks good, but there seems there's a unit test failing in CI

PR #573, opened 2021-02-26
Add trace and table headers to the request

frankgh:
Add X-B3-TraceId, X-B3-SpanId and X-GP-TABLE-NAME headers

### DIFF
--- a/external-table/src/pxfheaders.c
+++ b/external-table/src/pxfheaders.c
@@ -173,7 +173,7 @@ build_http_headers(PxfInputData *input)
 	else
 		churl_headers_append(headers, "X-GP-HAS-FILTER", "0");
 
-	// Add trace id = xid : filterstr : namespace_oid : tablename : user
+	// Add trace id = xid : filterstr : schema : tablename : user
 	traceId = GetTraceId(ev.GP_XID, filterstr, relnamespace, relname, ev.GP_USER);
 	churl_headers_append(headers, "X-B3-TraceId", traceId);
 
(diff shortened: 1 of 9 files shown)


external-table/src/pxfheaders.c
```diff
@@ -164,6 +173,15 @@ build_http_headers(PxfInputData *input)
 	else
 		churl_headers_append(headers, "X-GP-HAS-FILTER", "0");
 
+	// Add trace id = xid : filterstr : namespace_oid : tablename : user
```
bradfordb-vmware, line 176: Should `namespace_oid` by `namespace`?

external-table/src/pxfutils.c
```diff
@@ -116,4 +119,80 @@ get_pxf_port(void)
...
+{
+	char	   *traceId,
+			   *md5Hash;
+
+	traceId = psprintf("%s:%s:%s:%s:%s", xid, filter, relnamespace, relname, user);
+	elog(DEBUG3, "GetTraceId: generated traceId %s", traceId);
+
+	md5Hash = palloc0(33);
+
+	if (!pg_md5_hash(traceId, strlen(traceId), md5Hash))
```
bradfordb-vmware, line 138: `pg_md5_hash` returns false on failure; is it safe for us to just return NULL and continue?

bradfordb-vmware (approved):

external-table/src/pxfutils.c
frankgh: I think so, if we fail it should be fine. Sleuth already does sampling of traces, so it's okay if some trace headers are missing because on the server side some traces will be dropped anyway

PR #574, opened 2021-02-27
add application property for configuring logging level

bradfordb-vmware:
The application property `pxf.log.level` is added for setting the value
of Spring Boot's `logging.level.root` property. This allows a user to
change the logging level of the PXF service by editing the property in
pxf-application.properties or setting an environment variable
(`PXF_LOG_LEVEL`) when (re-)starting PXF.

### DIFF
--- a/server/pxf-service/src/templates/conf/pxf-application.properties
+++ b/server/pxf-service/src/templates/conf/pxf-application.properties
@@ -16,3 +16,7 @@
 # pxf.task.pool.core-size=8
 # pxf.task.pool.queue-capacity=0
 # pxf.task.pool.max-size=200
+
+# Logging
+# To enable debug logging, uncomment and change `info` to `debug` here
+# pxf.log.level=info
--- a/server/pxf-service/src/templates/conf/pxf-log4j2.xml
+++ b/server/pxf-service/src/templates/conf/pxf-log4j2.xml
@@ -6,6 +6,7 @@
     <Property name="LOG_DATEFORMAT_PATTERN">yyyy-MM-dd HH:mm:ss.SSS</Property>
     <Property name="CONSOLE_LOG_PATTERN">%clr{%d{${LOG_DATEFORMAT_PATTERN}}}{faint} %clr{${LOG_LEVEL_PATTERN}} %clr{%pid}{magenta} %clr{---}{faint} %clr{[%15.15t]}{faint} %clr{%-40.40c{1.}}{cyan} %clr{:}{faint} %m%n${sys:LOG_EXCEPTION_CONVERSION_WORD}</Property>
     <Property name="FILE_LOG_PATTERN">%d{${LOG_DATEFORMAT_PATTERN}} ${LOG_LEVEL_PATTERN} %pid --- [%t] %-40.40c{1.} : %m%n${sys:LOG_EXCEPTION_CONVERSION_WORD}</Property>
+    <Property name="PXF_LOG_LEVEL">${bundle:pxf-application:pxf.log.level}</Property>
 </Properties>
 <Appenders>
     <Console name="Console" target="SYSTEM_OUT" follow="true">
@@ -24,8 +25,7 @@
         <AppenderRef ref="RollingFile"/>
     </Root>
 
-    <!-- Uncomment to see DEBUG messages from PXF service -->
-    <!-- <Logger name="org.greenplum.pxf" level="debug"/> -->
+    <Logger name="org.greenplum.pxf" level="${env:PXF_LOG_LEVEL:-${sys:PXF_LOG_LEVEL:-info}}"/>
 
     <!-- The levels below are tuned to provide minimal output, change to INFO or DEBUG to troubleshoot -->
     <Logger name="com.google.cloud.hadoop.fs.gcs.GoogleHadoopFileSystemBase" level="warn"/>


server/pxf-service/src/templates/conf/pxf-application.properties
```diff
@@ -15,4 +15,5 @@
 # pxf.task.pool.allow-core-thread-timeout=false
 # pxf.task.pool.core-size=8
 # pxf.task.pool.queue-capacity=0
-# pxf.task.pool.max-size=200
\ No newline at end of file
+# pxf.task.pool.max-size=200
+# pxf.log.level=info
```
frankgh, line 19: maybe let's add a comment and default to debug

server/pxf-service/src/main/resources/application.properties
```diff
@@ -42,4 +42,5 @@ pxf.task.pool.queue-capacity=0
 
 # logging
 logging.file.name=${pxf.logdir:/var/log}/pxf-service.log
-logging.file.path=${pxf.logdir:/var/log}
\ No newline at end of file
+logging.file.path=${pxf.logdir:/var/log}
+logging.level.root=${pxf.log.level:info}
```
frankgh, line 46: I just tried this and this produces way too much output. Ideally, we would like this setting to only apply for the `org.greenplum.pxf` package

server/pxf-service/src/templates/conf/pxf-application.properties
frankgh: a user-facing comment

bradfordb-vmware: Are you saying that we should make the default log level `DEBUG`? Or are you asking that we make just this commented out property be `DEBUG`?

frankgh: commented out property to be DEBUG

server/pxf-service/src/templates/conf/pxf-application.properties
```diff
@@ -15,4 +15,8 @@
 # pxf.task.pool.allow-core-thread-timeout=false
 # pxf.task.pool.core-size=8
 # pxf.task.pool.queue-capacity=0
-# pxf.task.pool.max-size=200
\ No newline at end of file
+# pxf.task.pool.max-size=200
+
+# Logging
+# To enable debug logging, uncomment and change `info` to `debug` here
+#pxf.log.level=info
```
frankgh, line 22: since this is user facing let's add the space.
```suggestion
# pxf.log.level=info
```

frankgh: I'm torn between info and debug here as commented out. For the other properties we have the default value, not the desired value. @denalex any input here?

ashuka24: i would stay consistent and leave the default value commented out, there is already a comment saying to change it to debug for further logging.

ashuka24 (approved):

frankgh (approved):

server/pxf-service/src/main/resources/application.properties
```diff
@@ -42,4 +42,5 @@ pxf.task.pool.queue-capacity=0
 
 # logging
 logging.file.name=${pxf.logdir:/tmp}/pxf-service.log
-logging.file.path=${pxf.logdir:/tmp}
\ No newline at end of file
+logging.file.path=${pxf.logdir:/tmp}
+logging.level.org.greenplum.pxf=${pxf.log.level:info}
```
denalex, line 46: All logging levels are configured in `pxf-logj2.xml` file, so why not consume that property there, where instead of current:
```
<!-- <Logger name="org.greenplum.pxf" level="debug"/> -->
```
it will become
```
<Logger name="org.greenplum.pxf" level="${pxf.log.level:info}"/>
```

frankgh: yeah, this is what I was thinking about too

bradfordb-vmware: With my testing, this approach isn't working:

```console
$ cat $PXF_BASE/conf/pxf-application.properties
...
pxf.log.level=debug

$ grep -wc DEBUG $PXF_BASE/logs/pxf-service.log
0

$ psql -c 'SELECT * FROM ext_hdfs_readable;'
...

$ grep -wc DEBUG $PXF_BASE/logs/pxf-service.log
0

$ cat $PXF_BASE/conf/pxf-log4j2.xml
...
<Logger name="org.greenplum.pxf" level="${pxf.log.level:info}"/>
```

frankgh: try adding a property

```
    <Property name="PXF_LOG_LEVEL">${bundle:pxf-application:pxf.log.level}</Property>
````

and then 

```
    <Logger name="org.greenplum.pxf" level="${sys:PXF_LOG_LEVEL}"/>
````

Comment (bradfordb-vmware) [reply server/pxf-service/src/main/resources/application.properties:46]: That works but only if you change/set the log level via the application.properties; setting the environment variable `PXF_LOG_LEVEL` doesn't work.

Comment (frankgh) [reply server/pxf-service/src/main/resources/application.properties:46]: yeah, good point. Maybe this other approach is becoming too complex

**Review by frankgh — APPROVED** [review]

File: server/pxf-service/src/templates/conf/pxf-log4j2.xml [hunk 5]
```diff
@@ -25,7 +26,7 @@
     </Root>
 
     <!-- Uncomment to see DEBUG messages from PXF service -->
-    <!-- <Logger name="org.greenplum.pxf" level="debug"/> -->
+    <Logger name="org.greenplum.pxf" level="${env:PXF_LOG_LEVEL:-${sys:PXF_LOG_LEVEL:-info}}"/>
```
denalex, line 29: the comment above should probably be removed now or its text needs to be updated

denalex (approved):

PR #577, opened 2021-03-02
Enhance MDC with PXF context

frankgh:
Log the PXF sessionId (xid : server) and the segmentId in LOG entries
where the information is available for logging. Add the
`PxfContextMdcLogEnhancerDecorator` and `PxfContextMdcLogEnhancerFilter`
classes to support logging the PXF context when available.

### DIFF
--- a/server/pxf-service/src/main/java/org/greenplum/pxf/service/spring/PxfContextMdcLogEnhancerFilter.java
+++ b/server/pxf-service/src/main/java/org/greenplum/pxf/service/spring/PxfContextMdcLogEnhancerFilter.java
@@ -1,5 +1,6 @@
 package org.greenplum.pxf.service.spring;
 
+import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang.StringUtils;
 import org.greenplum.pxf.service.HttpHeaderDecoder;
 import org.slf4j.MDC;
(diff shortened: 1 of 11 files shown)


server/pxf-service/src/main/java/org/greenplum/pxf/service/spring/PxfContextMdcLogEnhancerFilter.java
```diff
@@ -0,0 +1,72 @@
...
+     */
+    private void insertIntoMDC(HttpServletRequest request) {
+        boolean encoded = decoder.areHeadersEncoded(request);
+        String xid = decoder.getHeaderValue("X-GP-XID", request, encoded);
+        if (StringUtils.isBlank(xid)) {
+            return; // Not a PXF extension request
+        }
+
+        // xid : server
+        String serverName = StringUtils.defaultString(decoder.getHeaderValue("X-GP-OPTIONS-SERVER", request, encoded), "default");
```
denalex, line 58: I think I was using `defaultIfBlank` to take care or empty header values

server/pxf-service/src/main/java/org/greenplum/pxf/service/spring/PxfContextMdcLogEnhancerFilter.java
```diff
@@ -0,0 +1,72 @@
...
+        String sessionId = String.format("%s:%s", xid, serverName);
+        MDC.put(MDC_SESSION_ID, sessionId);
+        MDC.put(MDC_SEGMENT_ID, decoder.getHeaderValue("X-GP-SEGMENT-ID", request, encoded));
```
denalex, line 61: can we do `log.debug` with what we set here ?

denalex (approved):

PR #579, opened 2021-03-04
report "fragments.sent" PXF metric

denalex:
PXF will report `fragment.sent` Timer metric to Spring Boot Metric Registry. It will measure the time it takes to send data for a single fragment. The metric will also calculate the total number of invocations, which means the system will know a total number of fragments served. The metric will have a standard set of tags: `user`, `segment`, `profile`, `server` and the additional tag `outcome` with values either `success` or `error`, depending on whether all data for the fragment has been sent without errors.

The metric will be available from the `/actuator/metrics/fragments.sent` endpoint, e.g.:
```
{"name":"fragments.sent","description":null,"baseUnit":"seconds","measurements":[{"statistic":"COUNT","value":6.0},{"statistic":"TOTAL_TIME","value":0.039},{"statistic":"MAX","value":0.012}],"availableTags":[{"tag":"server","values":["default"]},{"tag":"application","values":["pxf-service"]},{"tag":"segment","values":["0","1","2"]},{"tag":"profile","values":["unknown"]},{"tag":"user","values":["adenissov"]},{"tag":"outcome","values":["success"]}]}
```

The metric will be also available from the `/actuator/prometheus` endpoint, e.g.:
```
# HELP fragments_sent_seconds  
# TYPE fragments_sent_seconds summary
fragments_sent_seconds_count{application="pxf-service",outcome="success",profile="unknown",segment="1",server="default",user="adenissov",} 2.0
fragments_sent_seconds_sum{application="pxf-service",outcome="success",profile="unknown",segment="1",server="default",user="adenissov",} 0.013
fragments_sent_seconds_count{application="pxf-service",outcome="success",profile="unknown",segment="2",server="default",user="adenissov",} 2.0
fragments_sent_seconds_sum{application="pxf-service",outcome="success",profile="unknown",segment="2",server="default",user="adenissov",} 0.013
fragments_sent_seconds_count{application="pxf-service",outcome="success",profile="unknown",segment="0",server="default",user="adenissov",} 2.0
fragments_sent_seconds_sum{application="pxf-service",outcome="success",profile="unknown",segment="0",server="default",user="adenissov",} 0.013
# HELP fragments_sent_seconds_max  
# TYPE fragments_sent_seconds_max gauge
fragments_sent_seconds_max{application="pxf-service",outcome="success",profile="unknown",segment="1",server="default",user="adenissov",} 0.012
fragments_sent_seconds_max{application="pxf-service",outcome="success",profile="unknown",segment="2",server="default",user="adenissov",} 0.012
fragments_sent_seconds_max{application="pxf-service",outcome="success",profile="unknown",segment="0",server="default",user="adenissov",} 0.012
```

### DIFF
--- a/server/pxf-service/src/main/java/org/greenplum/pxf/service/MetricsReporter.java
+++ b/server/pxf-service/src/main/java/org/greenplum/pxf/service/MetricsReporter.java
@@ -1,0 +1,129 @@
+package org.greenplum.pxf.service;
+
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tags;
+import io.micrometer.core.instrument.Timer;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang.StringUtils;
+import org.greenplum.pxf.api.model.RequestContext;
+import org.springframework.core.env.Environment;
+import org.springframework.stereotype.Component;
+
+import java.time.Duration;
+
+/**
+ * Service responsible for submitting metrics to MeterRegistry.
+ */
+@Component
+@Slf4j
+public class MetricsReporter {
+
+    private static final String UNKNOWN_VALUE = "unknown";
+    private static final Tags SUCCESS_TAG = Tags.of("outcome", "success");
+    private static final Tags ERROR_TAG = Tags.of("outcome", "error");
+
+    private final MeterRegistry registry;
+    private final Environment env;
+
+    /**
+     * Creates a new instance.
+     *
+     * @param registry meter registry to submit metrics to
+     * @param env      environment which serves as the source of property values
+     */
+    public MetricsReporter(MeterRegistry registry, Environment env) {
+        this.registry = registry;
+        this.env = env;
+    }
+
+    /**
+     * Reports timer metric with a given name and duration to the registry. Applies custom tags before reporting.
+     *
+     * @param metric   metric to apply
+     * @param duration duration measured by the metric
+     * @param context  request context
+     */
+    public void reportTimer(PxfMetric metric, Duration duration, RequestContext context) {
+        reportTimer(metric, duration, context, null);
+    }
+
+    /**
+     * Reports timer metric, duration and outcome of the timed operation to the registry.
+     * Applies custom tags before reporting and adds an outcome tag.
+     *
+     * @param metric   metric to apply
+     * @param duration duration measured by the metric
+     * @param context  request context
+     * @param success  true if timed operation was successful, false otherwise
+     */
+    public void reportTimer(PxfMetric metric, Duration duration, RequestContext context, boolean success) {
+        reportTimer(metric, duration, context, success ? SUCCESS_TAG : ERROR_TAG);
+    }
+
+    /**
+     * Reports timer metric with a given name, duration and additional tags to the registry.
+     * Applies custom tags before reporting and adds a success outcome tag.
+     *
+     * @param metric
+     * @param duration
+     * @param context
+     * @param extraTags
+     */
+    private void reportTimer(PxfMetric metric, Duration duration, RequestContext context, Tags extraTags) {
+        String metricName = metric.getMetricName();
+        long durationMs = duration.toMillis();
+        if (env.getProperty(metric.getEnabledPropertyName(), Boolean.class, Boolean.FALSE)) {
+            Tags tags = (extraTags == null) ? getTags(context) : getTags(context).and(extraTags);
+            Timer timer = Timer.builder(metric.getMetricName()).tags(tags).register(registry);
+            timer.record(duration);
+            if (log.isTraceEnabled()) {
+                log.trace("Reported timer {}{} with duration={}ms", metricName, tags, durationMs);
+            }
+        } else {
+            log.trace("Skipping reporting metric {} with duration {}ms", metricName, durationMs);
+        }
+    }
+
+    /**
+     * Produces a set of custom tags with values from the provided request context.
+     *
+     * @param context request context
+     * @return collection of custom tags
+     */
+    private Tags getTags(RequestContext context) {
+        return Tags.empty()
+                .and("user", StringUtils.defaultIfBlank(context.getUser(), UNKNOWN_VALUE))
+                .and("segment", String.valueOf(context.getSegmentId()))
+                .and("profile", StringUtils.defaultIfBlank(context.getProfile(), UNKNOWN_VALUE))
+                .and("server", StringUtils.defaultIfBlank(context.getServerName(), "default"));
+    }
+
+    /**
+     * Enum that has information about all custom metrics for PXF.
+     */
+    public enum PxfMetric {
+        FRAGMENTS_SENT("fragments.sent", "pxf.metrics.fragments.enabled");
+
+        private final String metricName;
+        private final String enabledPropertyName;
+
+        /**
+         * Creates a new instance.
+         *
+         * @param metricName          name of the metric as will be seen by the registry
+         * @param enabledPropertyName property name that manages whether reporting of this metric is turned on / off
+         */
+        PxfMetric(String metricName, String enabledPropertyName) {
+            this.metricName = metricName;
+            this.enabledPropertyName = enabledPropertyName;
+        }
+
+        public String getMetricName() {
+            return metricName;
+        }
+
+        public String getEnabledPropertyName() {
+            return enabledPropertyName;
+        }
+    }
+}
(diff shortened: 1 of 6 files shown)


server/pxf-service/src/main/java/org/greenplum/pxf/service/controller/ReadServiceImpl.java
```diff
@@ -127,26 +132,33 @@ private OperationStats writeStream(RequestContext context, OutputStream outputSt
...
         } finally {
-            log.debug("{} Stopped streaming fragment {} of resource {}, {} records.",
-                    context.getId(), fragment.getIndex(), context.getDataSource(), recordCount);
             try {
                 bridge.endIteration();
             } catch (Exception e) {
                 log.warn("{} Ignoring error encountered during bridge.endIteration()", context.getId(), e);
             }
+            Duration duration = Duration.between(startTime, Instant.now());
+            log.debug("{} Finished streaming fragment {} of resource {}, {} records in {} ms.",
```
bradfordb-vmware, line 159: Minor nit: Sometimes we will omit the log message `Starting streaming...` but we always log the message `Finished streaming...`. Should we move the starting log message?

server/pxf-service/src/test/java/org/greenplum/pxf/service/MetricsReporterTest.java
```diff
@@ -0,0 +1,112 @@
...
+    public void setup() {
+        // it is not possible to mock the registry as Timer.Builder calls a package-level function that cannot be mocked.
+        // so using an actual registry for verification
+        registry = new SimpleMeterRegistry();
+        reporter = new MetricsReporter(registry, mockEnvironment);
+    }
+
+    @Test
+    public void testFragmentsSentMetricDisabled() {
+        when(mockEnvironment.getProperty("pxf.metrics.fragments.enabled", Boolean.class, Boolean.FALSE)).thenReturn(false);
```
bradfordb-vmware, line 44: Would it make sense to have two helper methods:

```java
private void enableFragmentMetrics() {
    when(mockEnvironment.getProperty("pxf.metrics.fragments.enabled", Boolean.class, Boolean.FALSE)).thenReturn(true);

}

private void disableFragmentMetrics() {
    when(mockEnvironment.getProperty("pxf.metrics.fragments.enabled", Boolean.class, Boolean.FALSE)).thenReturn(false);
}
```

Might make the test a little easier to read plus we change the property name, its a little easier to change?

bradfordb-vmware (approved):
Two small comments for your consideration but otherwise looks good.

server/pxf-service/src/main/java/org/greenplum/pxf/service/MetricsReporter.java
```diff
@@ -0,0 +1,129 @@
...
+     * @param extraTags
+     */
+    private void reportTimer(PxfMetric metric, Duration duration, RequestContext context, Tags extraTags) {
+        String metricName = metric.getMetricName();
+        long durationMs = duration.toMillis();
+        if (env.getProperty(metric.getEnabledPropertyName(), Boolean.class, Boolean.FALSE)) {
+            Tags tags = (extraTags == null) ? getTags(context) : getTags(context).and(extraTags);
+            Timer timer = Timer.builder(metric.getMetricName()).tags(tags).register(registry);
+            timer.record(duration);
+            if (log.isTraceEnabled()) {
```
frankgh, line 79: is this if necessary?

server/pxf-service/src/test/java/org/greenplum/pxf/service/MetricsReporterTest.java
```diff
@@ -0,0 +1,112 @@
...
+import static org.mockito.Mockito.when;
+
+@ExtendWith(MockitoExtension.class)
+public class MetricsReporterTest {
+
+    private MeterRegistry registry;
+    private MetricsReporter reporter;
+
+    @Mock
+    private RequestContext mockContext;
```
frankgh, line 30: Don't we prefer to use concrete objects for `RequestContext`?

frankgh (approved):

server/pxf-service/src/main/java/org/greenplum/pxf/service/MetricsReporter.java
denalex: I prefer to have this in case we need to troubleshoot an actual deployment that doesn't report or reports inconsistent metrics. Since this orthogonal to the main business flow, I set it at `trace` level, so it doesn't show up in DEBUG log when troubleshooting other issues.

server/pxf-service/src/main/java/org/greenplum/pxf/service/controller/ReadServiceImpl.java
denalex: yes, I'm a bit bothered by the asymmetry as well, but what is a better suggestion ? Currently it can be:
<nothing> -- Finished (if exception during getBridge)
Skipping  -- Finished (if bridge did not open)
Starting -- Finished in all other cases

server/pxf-service/src/test/java/org/greenplum/pxf/service/MetricsReporterTest.java
denalex: I now prefer to use mocks for simplicity and consistency and using actual objects only if they cannot be mocked.

server/pxf-service/src/test/java/org/greenplum/pxf/service/MetricsReporterTest.java
denalex: I usually don't do this, especially for 1-liners -- prefer the test cases "raw" and "dumb" to avoid chances of having a mistake in test logic abstraction, unless there's a lot of common code shared by a lot of test scenarios.

But I will implement your suggestion, if you believe it aids readability, which is equally important.

server/pxf-service/src/main/java/org/greenplum/pxf/service/controller/ReadServiceImpl.java
bradfordb-vmware: Maybe something like

```java
log.debug("{} Processing fragment {} of resource {}", ...);
if (!bridge.beginIteration()) {
   log.debug("{} Skipping streaming fragment {} of resource {}", ...);
} else {
    log.debug("{} Streaming fragment {} of resource {}", ...);
}
...
log.debug("{} Finished processing fragment {} of resource {}, {} records in {} ms.",...);
```

server/pxf-service/src/main/java/org/greenplum/pxf/service/MetricsReporter.java
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    /**
+     * Produces a set of custom tags with values from the provided request context.
+     *
+     * @param context request context
+     * @return collection of custom tags
+     */
+    private Tags getTags(RequestContext context) {
+        return Tags.empty()
```
ashuka24, line 94: why do we get `Tags.empty()` before adding all the custom tags?

server/pxf-service/src/main/java/org/greenplum/pxf/service/MetricsReporter.java
```diff
@@ -0,0 +1,129 @@
...
+                .and("user", StringUtils.defaultIfBlank(context.getUser(), UNKNOWN_VALUE))
+                .and("segment", String.valueOf(context.getSegmentId()))
```
ashuka24, line 96: is there ever a case where `context.getSegmentId()` would return null/empty?

server/pxf-service/src/main/java/org/greenplum/pxf/service/MetricsReporter.java
denalex: that's how their "fluent" API works, to work with `.and` you need `Tags` instance first

server/pxf-service/src/main/java/org/greenplum/pxf/service/MetricsReporter.java
denalex: we define it as `int` in the `RequestContext`, so it cannot be null / empty, the request will fail to parse in that case.

server/pxf-service/src/main/java/org/greenplum/pxf/service/controller/ReadServiceImpl.java
denalex: hmm, in most cases this will print Processing and Starting right after each other with the same context, polluting the log file with no additional value, I prefer to have only one "starting" message. What if I only change `Finished streaming` wording to `Finish processing` ?

`<nothing>` -- Finished processing            : (if exception during getBridge) <-- never happens
Skipping streaming -- Finished processing : (if bridge did not open)
Starting streaming -- Finished processing  : in all other cases

bradfordb-vmware: That sounds good.

frankgh (approved):

PR #581, opened 2021-03-05
Fix PXF startup when PID file is exists and another process acquire PID

frankgh:
An existing PID file can prevent a PXF startup, if the PID has been
recycled and has been acquired by a new process. This commit validates
that the process name contains `[p]xf-app`, adding more validation for
the check.

### DIFF
--- a/server/pxf-service/src/scripts/pxf
+++ b/server/pxf-service/src/scripts/pxf
@@ -158,7 +158,7 @@ function checkWebapp()
 
 # determines whether the application is running
 function isRunning() {
-    ps -p "$1" &> /dev/null
+    ps -o pid,command -p "$1" | grep '[p]xf-app' &> /dev/null
 }
 
 # creates the runtime directory structure inside $PXF_BASE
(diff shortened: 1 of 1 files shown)


server/pxf-service/src/scripts/pxf
```diff
@@ -158,7 +158,7 @@ function checkWebapp()
 
 # determines whether the application is running
 function isRunning() {
-    ps -p "$1" &> /dev/null
+    ps -o pid,command -p "$1" | grep '[p]xf-app' &> /dev/null
```
denalex, line 161: what's the importance of [p], why not just `grep 'pxf-app'` ?

denalex (approved):

server/pxf-service/src/scripts/pxf
frankgh: from a comment from Oliver: # grep with regex to avoid catching grep process itself 
I'll add the comment here as well.

frankgh: even though it shouldn't be caught because we specify the pid here

denalex: but here you're narrowing down with pid, so grep itself will never be caught anyways

PR #583, opened 2021-03-05
Add metrics for reporting records sent from PXF to gpdb

ashuka24:
This commit adds a counter for tracking the number of records PXF sends
out. We report on frequency determined by
pxf.metrics.records.report-frequency.

Co-authored-by: Alex Denissov <adenissov@vmware.com>

### DIFF
--- a/server/pxf-service/src/main/java/org/greenplum/pxf/service/MetricsReporter.java
+++ b/server/pxf-service/src/main/java/org/greenplum/pxf/service/MetricsReporter.java
@@ -1,5 +1,6 @@
 package org.greenplum.pxf.service;
 
+import io.micrometer.core.instrument.Counter;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.Tags;
 import io.micrometer.core.instrument.Timer;
@@ -72,16 +73,52 @@ public class MetricsReporter {
     private void reportTimer(PxfMetric metric, Duration duration, RequestContext context, Tags extraTags) {
         String metricName = metric.getMetricName();
         long durationMs = duration.toMillis();
-        if (env.getProperty(metric.getEnabledPropertyName(), Boolean.class, Boolean.FALSE)) {
-            Tags tags = (extraTags == null) ? getTags(context) : getTags(context).and(extraTags);
-            Timer timer = Timer.builder(metric.getMetricName()).tags(tags).register(registry);
-            timer.record(duration);
-            if (log.isTraceEnabled()) {
-                log.trace("Reported timer {}{} with duration={}ms", metricName, tags, durationMs);
-            }
-        } else {
-            log.trace("Skipping reporting metric {} with duration {}ms", metricName, durationMs);
+        if (!env.getProperty(metric.getEnabledPropertyName(), Boolean.class, Boolean.FALSE)) {
+            log.trace("Skipping reporting metric {} with duration={}ms", metricName, durationMs);
+            return;
         }
+
+        Tags tags = (extraTags == null) ? getTags(context) : getTags(context).and(extraTags);
+        Timer timer = Timer.builder(metricName).tags(tags).register(registry);
+        timer.record(duration);
+        if (log.isTraceEnabled()) {
+            log.trace("Reported timer {}{} with duration={}ms", metricName, tags, durationMs);
+        }
+    }
+
+    /**
+     * Reports counter metric with a given name and the increment to the registry.
+     * Reports with any tags given by the context and adds a success outcome tag.
+     *
+     * @param metric
+     * @param increment
+     * @param context
+     */
+    public void reportCounter(PxfMetric metric, long increment, RequestContext context) {
+        String metricName = metric.getMetricName();
+        if (!env.getProperty(metric.getEnabledPropertyName(), Boolean.class, Boolean.FALSE)) {
+            log.trace("Skipping reporting metric {} with increment={}", metricName, increment);
+            return;
+        }
+
+        Double incrementCount = Long.valueOf(increment).doubleValue();
+        Tags tags = getTags(context);
+        Counter counter = Counter.builder(metricName).tags(tags).register(registry);
+        counter.increment(incrementCount);
+        if (log.isTraceEnabled()) {
+            log.trace("Reported counter {}{} with increment={}", metricName, tags, increment);
+        }
+    }
+
+    /**
+     * Pulls the value for reporting frequency for the given metric from the environment.
+     * If no value found, the default reporting frequency is 1000.
+     *
+     * @param metric
+     */
+    public int getReportFrequency(PxfMetric metric) {
+        String metricFrequencyName = metric.getFrequencyName();
+        return env.getProperty(metricFrequencyName, Integer.class, 1000);
     }
 
     /**
@@ -102,20 +139,24 @@ public class MetricsReporter {
      * Enum that has information about all custom metrics for PXF.
      */
     public enum PxfMetric {
-        FRAGMENTS_SENT("fragments.sent", "pxf.metrics.fragments.enabled");
+        FRAGMENTS_SENT("fragments.sent", "pxf.metrics.fragments.enabled", null),
+        RECORDS_SENT("records.sent", "pxf.metrics.records.enabled", "pxf.metrics.records.report-frequency");
 
         private final String metricName;
         private final String enabledPropertyName;
+        private final String frequencyName;
 
         /**
          * Creates a new instance.
          *
          * @param metricName          name of the metric as will be seen by the registry
          * @param enabledPropertyName property name that manages whether reporting of this metric is turned on / off
+         * @param frequencyName           property name that manages how often we report the metric
          */
-        PxfMetric(String metricName, String enabledPropertyName) {
+        PxfMetric(String metricName, String enabledPropertyName, String frequencyName) {
             this.metricName = metricName;
             this.enabledPropertyName = enabledPropertyName;
+            this.frequencyName = frequencyName;
         }
 
         public String getMetricName() {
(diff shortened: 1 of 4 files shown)


server/pxf-service/src/main/java/org/greenplum/pxf/service/MetricsReporter.java
```diff
@@ -74,16 +75,36 @@ private void reportTimer(PxfMetric metric, Duration duration, RequestContext con
...
             if (log.isTraceEnabled()) {
                 log.trace("Reported timer {}{} with duration={}ms", metricName, tags, durationMs);
             }
         } else {
-            log.trace("Skipping reporting metric {} with duration {}ms", metricName, durationMs);
+            log.trace("Skipping reporting metric {} with duration={}ms", metricName, durationMs);
         }
     }
 
+    public void reportCounter(PxfMetric metric, long increment, RequestContext context) {
```
frankgh, line 88: let's add javadocs here

server/pxf-service/src/main/java/org/greenplum/pxf/service/MetricsReporter.java
```diff
@@ -74,16 +75,36 @@ private void reportTimer(PxfMetric metric, Duration duration, RequestContext con
...
+        String metricName = metric.getMetricName();
+        Double incrementCount = Long.valueOf(increment).doubleValue();
```
frankgh, line 90: should this be inside the if statement?

server/pxf-service/src/main/java/org/greenplum/pxf/service/controller/ReadServiceImpl.java
```diff
@@ -74,6 +74,9 @@ private OperationStats writeStream(RequestContext context, OutputStream outputSt
         String originalResolver = context.getResolver();
         String originalProfileScheme = context.getProfileScheme();
 
+        int recordReportFrequency = metricsReporter.getReportFrequency(MetricsReporter.PxfMetric.RECORDS_SENT);
+        log.info("We report every {} records written.", recordReportFrequency);
```
frankgh, line 78:
```suggestion
        log.info("Report every {} records written.", recordReportFrequency);
```

frankgh: do we really want this at `info` level?

server/pxf-service/src/main/java/org/greenplum/pxf/service/controller/ReadServiceImpl.java
```diff
@@ -146,16 +149,28 @@ private int processFragment(DataOutputStream dos, RequestContext context, Fragme
                 while ((record = bridge.getNext()) != null) {
                     record.write(dos);
                     ++recordCount;
+                    // report records based off the recordReportFrequency
+                    if (recordCount % recordReportFrequency == 0)
+                    {
```
frankgh, line 154: this bracket should be in the previous line

server/pxf-service/src/main/java/org/greenplum/pxf/service/controller/ReadServiceImpl.java
```diff
@@ -146,16 +149,28 @@ private int processFragment(DataOutputStream dos, RequestContext context, Fragme
...
+                        metricsReporter.reportCounter(MetricsReporter.PxfMetric.RECORDS_SENT, recordReportFrequency, context);
+                    }
+                }
+                // report the remaining records that have yet to be reported
+                int remainder = recordCount % recordReportFrequency;
+                if (remainder != 0)
+                {
```
frankgh, line 161: bracket in the prvious line

server/pxf-service/src/main/resources/application.properties
```diff
@@ -13,6 +13,8 @@ management.metrics.tags.application=pxf-service
 
 # PXF-specific metrics
 pxf.metrics.fragments.enabled=true
+pxf.metrics.records.enabled=true
+pxf.metrics.records.report-frequency=1000
```
frankgh, line 17: will this make sense in the refactor? In the refactor we can report per batch, our default batch size (as of now) is 1024, so we can always report after the batch has been consumed

server/pxf-service/src/main/java/org/greenplum/pxf/service/MetricsReporter.java
```diff
@@ -74,16 +75,36 @@ private void reportTimer(PxfMetric metric, Duration duration, RequestContext con
...
+        if (env.getProperty(metric.getEnabledPropertyName(), Boolean.class, Boolean.FALSE)) {
+            Tags tags = getTags(context);
+            Counter counter = Counter.builder(metricName).tags(tags).register(registry);
```
frankgh, line 93: I'm not sure how the micrometer API works, but shouldn't this be built once for the JVM?

server/pxf-service/src/main/java/org/greenplum/pxf/service/MetricsReporter.java
```diff
@@ -74,16 +75,36 @@ private void reportTimer(PxfMetric metric, Duration duration, RequestContext con
...
+            counter.increment(incrementCount);
+            if (log.isTraceEnabled()) {
+                log.trace("Reported counter {}{} with increment={}", metricName, tags, increment);
+            }
+        } else {
+            log.trace("Skipping reporting metric {} with increment={}", metricName, increment);
+        }
+    }
+
+    public int getReportFrequency(PxfMetric metric) {
```
frankgh, line 103: let's add the javadoc here also

server/pxf-service/src/main/java/org/greenplum/pxf/service/controller/ReadServiceImpl.java
ashuka24: Oops. I meant to take this out before creating the PR. If we think it may be helpful, we can keep it in and change it to debug level?

server/pxf-service/src/main/java/org/greenplum/pxf/service/MetricsReporter.java
ashuka24: The values are used in the else case as well so we cannot move it inside the if statement.

server/pxf-service/src/main/java/org/greenplum/pxf/service/MetricsReporter.java
ashuka24: We cannot register it once per JVM because the tag values will be different for every request. However, when we register the counter, the registry will either return an existing counter with the matching tags, or we will create a new one with the appropriate tags.

server/pxf-service/src/main/resources/application.properties
ashuka24: we can revisit it when we merge refactor. maybe we will not need this at that time.

server/pxf-service/src/main/java/org/greenplum/pxf/service/MetricsReporter.java
```diff
@@ -102,20 +137,24 @@ private Tags getTags(RequestContext context) {
      * Enum that has information about all custom metrics for PXF.
      */
     public enum PxfMetric {
-        FRAGMENTS_SENT("fragments.sent", "pxf.metrics.fragments.enabled");
+        FRAGMENTS_SENT("fragments.sent", "pxf.metrics.fragments.enabled", null),
```
bradfordb-vmware, line 140: It seems odd having all metrics with a frequency field but only used for some of them.

server/pxf-service/src/main/java/org/greenplum/pxf/service/controller/ReadServiceImpl.java
```diff
@@ -74,6 +74,9 @@ private OperationStats writeStream(RequestContext context, OutputStream outputSt
         String originalResolver = context.getResolver();
         String originalProfileScheme = context.getProfileScheme();
 
+        int recordReportFrequency = metricsReporter.getReportFrequency(MetricsReporter.PxfMetric.RECORDS_SENT);
```
bradfordb-vmware, line 77: The report frequency is a property of the metric and not the metrics reporter, but we're querying this property through the reporter? At first flance, I would think that

```java
int recordReportFrequency = MetricsReporter.PxfMetric.RECORDS_SENT.getReportFrequency();
```

would seem more intuitive.

server/pxf-service/src/main/java/org/greenplum/pxf/service/MetricsReporter.java
```diff
@@ -74,16 +75,50 @@ private void reportTimer(PxfMetric metric, Duration duration, RequestContext con
...
 
+    /**
+     * Reports counter metric with a given name and the increment to the registry.
+     * Reports with any tags given by the context and adds a success outcome tag.
+     *
+     * @param metric
+     * @param increment
+     * @param context
+     */
+    public void reportCounter(PxfMetric metric, long increment, RequestContext context) {
```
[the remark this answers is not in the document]
ashuka24: For timer, we keep track of the outcome of the operation it is attached to and we store the outcome in the extra tags.

For the counter we do not. We are using the counter for tracking tuples sent/received where the outcome does not matter.

server/pxf-service/src/main/java/org/greenplum/pxf/service/controller/ReadServiceImpl.java
ashuka24: Doing `MetricsReporter.PxfMetric.RECORDS_SENT.getReportFrequency();` would mean that the ENUM is resolving the value instead of just being a holder of metadata. It seems weird for an enum to resolve a value and a bit counter intuitive to me.

server/pxf-service/src/main/java/org/greenplum/pxf/service/MetricsReporter.java
denalex: some metrics that report on every occurrence do not need frequency setting, so they do not have the property for that and hence the name of the property is `null`.

bradfordb-vmware (approved):

PR #586, opened 2021-03-08
Log the number of records received by PXF

ashuka24:
In conjunction with reporting the number of records sent by PXF, we now
report the number of records PXF receives.

Co-authored-by: Alex Denissov <adenissov@vmware.com>


server/pxf-service/src/main/java/org/greenplum/pxf/service/controller/WriteServiceImpl.java
```diff
@@ -61,12 +61,24 @@ private OperationStats readStream(RequestContext context, InputStream inputStrea
         long recordCount = 0;
         IOException ex = null;
 
+        long recordReportFrequency = metricsReporter.getReportFrequency(MetricsReporter.PxfMetric.RECORDS_RECEIVED);
+        log.info("We report every {} records received.", recordReportFrequency);
```
bradfordb-vmware, line 65: Let's remove this.

PR #600, opened 2021-03-18
Log error messages with context

denalex:


server/pxf-service/src/main/java/org/greenplum/pxf/service/controller/BaseServiceImpl.java
```diff
@@ -54,42 +56,53 @@ protected BaseServiceImpl(String serviceName,
...
-        // Initialize the configuration for this request
-        Configuration configuration = configurationFactory.
-                initConfiguration(
-                        context.getConfig(),
-                        context.getServerName(),
-                        context.getUser(),
-                        context.getAdditionalConfigProps());
-        context.setConfiguration(configuration);
-
         Instant startTime = Instant.now();
```
ashuka24, line 65: why is the startTime calculated outside the catch if we only use it inside the try?

server/pxf-service/src/main/java/org/greenplum/pxf/service/controller/BaseServiceImpl.java
```diff
@@ -54,42 +56,53 @@ protected BaseServiceImpl(String serviceName,
...
+                        byteCount,
+                        String.format("%.2f", byteRate));
+            } else {
+                log.info("{} completed", context.getId());
+            }
+            return stats;
+        } catch (PxfIOException | PxfRuntimeException | Error e) {
+            log.error(e.getMessage(), e);
+            throw e;
+        } catch (Exception e) {
```
ashuka24, line 102: i noticed that we do this same catch logic in various places. is there a way that we can remove the duplication?

bradfordb-vmware: We looked at this as a team and have pushed some changes to address this.

denalex:
closing in favor of PR#604 https://github.com/greenplum-db/pxf/pull/604

PR #603, opened 2021-03-23
Docs: misc updates

mperezfuster:
Corrected a couple of typos and added pxf.log.level parameter to the pxf-application.properties table
I don't think it needs tech review but feel free to add any other reviewers if needed.


lisakowen (approved):
thanks!

ashuka24 (approved):

PR #604, opened 2021-03-23
Log error messages with context

denalex:


server/pxf-service/src/main/java/org/greenplum/pxf/service/security/BaseSecurityService.java
```diff
@@ -55,58 +54,52 @@ public BaseSecurityService(SecureLogin secureLogin, UGIProvider ugiProvider) {
...
-            LOG.debug("{} Releasing UGI resources. {}",
-                    contextId, exceptionDetected ? " Exception while processing" : "");
+            LOG.debug("Releasing UGI resources. {}", exceptionDetected ? " Exception while processing" : "");
             try {
                 if (userGroupInformation != null) {
                     ugiProvider.destroy(userGroupInformation);
                 }
             } catch (Throwable t) {
-                LOG.error("{} Error releasing UGI resources", contextId, t);
+                LOG.warn("Error releasing UGI resources, ignored.", t);
```
ashuka24, line 102: why do we go from logged error to ignored error listed as warn?

server/pxf-service/src/main/java/org/greenplum/pxf/service/security/BaseSecurityService.java
```diff
@@ -55,58 +54,52 @@ public BaseSecurityService(SecureLogin secureLogin, UGIProvider ugiProvider) {
...
-            throw new IOException(ute.getCause());
-        } catch (InterruptedException ie) {
+        } catch (Exception e) {
             exceptionDetected = true;
-            throw new IOException(ie);
+            throw e;
         } finally {
-            LOG.debug("{} Releasing UGI resources. {}",
-                    contextId, exceptionDetected ? " Exception while processing" : "");
+            LOG.debug("Releasing UGI resources. {}", exceptionDetected ? " Exception while processing" : "");
```
ashuka24, line 96: adding a period for consistency
```suggestion
            LOG.debug("Releasing UGI resources. {}", exceptionDetected ? " Exception while processing." : "");
```

server/pxf-service/src/main/java/org/greenplum/pxf/service/security/BaseSecurityService.java
denalex: failure during destroying UGI when the request is complete does not impact the outcome seen by user, so this is not an error (smith that prevented successful outcome), but rather a warning about meeting an unexpected condition. We did not throw it before either, just changing the log level here to `warn` and stating that it is ignored by the logic flow.

bradfordb-vmware (approved):

PR #605, opened 2021-03-24
PXF async thread names start with 'pxf-response-' instead of 'pxf-task-'

denalex:


bradfordb-vmware (approved):

lisakowen (approved):
we do not discuss the pxf.task.thread-name-prefix in the docs, as it is not included in the pxf-application.properties file.  so no docs impact that i am aware of.

ashuka24 (approved):

mperezfuster (approved):

PR #617, opened 2021-03-31
 Update PXF build pipeline for PXF 6+ releases

ashuka24:


khuddlefish (approved):

version
```diff
@@ -1 +1 @@
-6.0.1-SNAPSHOT
+6.0.1
```
bradfordb-vmware, line 1: This was from our dev run and needs to be reverted.

denalex (approved):
looks good, but needs a better commit message

PR #616, opened 2021-03-30
Pin the shipit file for the pxf-5-build pipeline

ashuka24:

PR #620, opened 2021-04-15
Upload second copy of ODP and OSL for inclusion GP5 releases

bradfordb-vmware:
The PXF release process currently uploads ODP and OSL artifacts to a
GP6-specific location in blob store and directs the RelEng team to use
the same artifacts for PXF-GP5 releases. The RelEng team would prefer
that the ODP and OSL artifacts get uploaded to GP5 and GP6 locations.

Instead of adding additional template variables for GP5 and GP6 paths
for ODP and OSL, the existing template variables are updated to remove
the GP major version from the path. This is appended in the CI pipeline
instead. Since all release artifacts for RelEng are stored in the same
bucket, the template variables for the bucket name are simplified as
well.

PXF ODP and OSL artifacts for both GP5 and GP6 are now stored in Google
Cloud Storage so there isn't much benefit to parameterizing the protocol
(i.e., `s3` vs `gs`).

This commit allows for using a common set of template variables files for the
build pipelines on master and branch-5.16.


bradfordb-vmware:
This is split from #616 to separate the pipeline changes for pinning the regex pattern for shipit files from the changes needed for including a separate ODP & OSL for GP5 & GP6.

denalex (approved):

ashuka24 (approved):

PR #625, opened 2021-05-04
Update terraform concourse resource to version 0.11.15

bradfordb-vmware:
HashiCorp rotated their GPG key on or about 2021-04-22 and this means
that older version of terraform are no longer able to download and
verify provider plugins. The upstream maintainer of the
terraform-resource for Concourse has not published a updated container
image for the latest 0.11.x release of terraform (0.11.15).

This PR updates the CI pipelines to use an internally published version
of the Concourse resource that includes terraform 0.11.15.

[0]: https://discuss.hashicorp.com/t/hcsec-2021-12-codecov-security-event-and-hashicorp-gpg-key-exposure/23512

### DIFF
--- a/concourse/pipelines/longevity_pipeline.yml
+++ b/concourse/pipelines/longevity_pipeline.yml
@@ -17,10 +17,12 @@ ccp_destroy_anchor: &ccp_destroy
 ## ======================================================================
 resource_types:
 - name: terraform
-  type: docker-image
+  type: registry-image
   source:
-    repository: ljfranklin/terraform-resource
-    tag: 0.11.14
+    repository: gcr.io/data-gpdb-ud/terraform-resource
+    tag: 0.11.15
+    username: _json_key
+    password: ((pxf-cloudbuild-service-account-key))
 
 - name: gcs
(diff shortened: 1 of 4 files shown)


concourse/pipelines/longevity_pipeline.yml
```diff
@@ -16,11 +16,13 @@ ccp_destroy_anchor: &ccp_destroy
 ## RESOURCE TYPES
 ## ======================================================================
 resource_types:
-- name: terraform
-  type: docker-image
+- name: data-gpdb-ud-terraform
```
denalex, line 19: I wouldn't even change the name here, less changes downstream

denalex (approved):
Thank you for addressing this.

concourse/pipelines/longevity_pipeline.yml
bradfordb-vmware: That's what we tried initially; unfortunately this caused Concourse to try and fetch the wrong/non-existent container image. The simplest way to "bust the cache" was to rename the resource.

PR #630, opened 2021-05-08
Fix the inconsistency between rows count in external table and ORC file

lchx1010:
As we know, use the PXF HDFS connector `hdfs:orc` profile to read ORC-format data and read 1024 rows of data at a time.

However, when the rows of data in ORC file is greater than 1024, it will result in the inconsistency between the number of rows in the external table of GPDB and the ORC file. In this case, the number of rows in the external table is always a multiple of 1024.

For example:

| ORC Data (row) | external table(row) | multiple(externalTable/1024) |
| -------------- | ------------------- | ---------------------------- |
| 1025           | 2048                | 2                            |
| 2050           | 3072                | 3                            |

Reason: When the last read data is less than 1024 rows, such as 30 rows, only the first 30 rows of data in cachedBatch are cleared, and the subsequent data are not cleared, but are still written to the external table.


ashuka24 (approved):
Thank you so much for identifying and resolving the issue! Unit tests have been added in a follow up PR

PR #635, opened 2021-05-17
Update README and setup for automation tests

ashuka24:
This commit does the following:
- Update README
- Add --no-psqlrc flag to for easier testing
- Add function that creates the pxf extension when running automation
  testing

Co-authored-by: Bradford Boyle <bradfordb@vmware.com>

### DIFF
--- a/automation/src/main/java/org/greenplum/pxf/automation/components/gpdb/Gpdb.java
+++ b/automation/src/main/java/org/greenplum/pxf/automation/components/gpdb/Gpdb.java
@@ -73,6 +73,9 @@ public class Gpdb extends DbSystemObject {
 
 		connect();
 
+		// Create the extensions if they don't exist
+		createExtension("pxf", true);
+
 		ReportUtils.stopLevel(report);
 	}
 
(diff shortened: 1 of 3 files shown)


automation/tincrepo/main/mpp/lib/PSQL.py
```diff
@@ -54,7 +54,7 @@ def __init__(self, sql_file = None, sql_cmd = None, out_file = None, output_to_f
             if not os.path.exists(sql_file):
                 raise PSQLException('SQL file %s does not exist. ' %sql_file)
 
-            cmd_str = '%s psql %s %s %s %s %s -f %s' \
+            cmd_str = '%s psql %s %s %s %s %s --no-psqlrc -f %s' \
```
denalex, line 57: why was this needed ?

automation/src/main/java/org/greenplum/pxf/automation/components/gpdb/Gpdb.java
```diff
@@ -71,6 +71,9 @@ public void init() throws Exception {
 
 		connect();
 
+		// Create the extensions if they don't exist
+		createExtension("pxf", true);
```
denalex, line 75: Usually we do it once per local setup, but guess this does not hurt. Did we run this via dev pipeline, especially on multinode clusters with upgrade tasks to make sure it doesn't impact the tests ?

denalex (approved):

automation/tincrepo/main/mpp/lib/PSQL.py
bradfordb-vmware: If you have any customization in your local psqlrc file (for example always printing timing information) that can cause automation to fail when running locally. The printed timing info shows up in the diff resulting in a tinc failure.

automation/src/main/java/org/greenplum/pxf/automation/components/gpdb/Gpdb.java
bradfordb-vmware: I don't know if a dev pipeline has been set but I can make sure one is run before merging.

PR #636, opened 2021-05-18
Add support for reading and writing arrays in AVRO

ashuka24:
This PR adds support for reading and writing arrays to and from avro files through PXF.


server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/avro/AvroUtilities.java
```diff
@@ -161,6 +285,135 @@ private Schema getFieldSchema(DataType type, String colName) {
...
+    }
+
+    private Schema firstNotNullSchema(List<Schema> types) {
+        for (Schema schema : types) {
+            if (schema.getType() != Schema.Type.NULL) {
+                return schema;
+            }
+        }
+
+        throw new PxfRuntimeException("Avro union only contains null types");
```
denalex, line 414:
```suggestion
        throw new PxfRuntimeException("Avro union schema only contains null types");
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/avro/AvroUtilities.java
```diff
@@ -161,6 +285,135 @@ private Schema getFieldSchema(DataType type, String colName) {
...
+            } else {
+                break;
+            }
+        }
+
+        return num % 2 == 0;
+    }
+
+    private ByteBuffer parseEscapeFormat(String value) {
+        List<Byte> byteList = new ArrayList<>();
```
denalex, line 364: This can be very inefficient if we have to create a Byte object for every byte, Can we use byte[] or ByteBuffer ?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/avro/AvroUtilities.java
```diff
@@ -161,6 +285,135 @@ private Schema getFieldSchema(DataType type, String colName) {
...
+        int index = 0, length = value.length();
+        if (value.charAt(0) == '"') {
+            index = 1;
+            length -= 1;
```
denalex, line 368: don't we want to double-check that the last char is also a `"` before decrementing length by 1 ?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/avro/AvroUtilities.java
```diff
@@ -161,6 +285,135 @@ private Schema getFieldSchema(DataType type, String colName) {
...
+                    index += 3;
+                }
+            } else if (value.charAt(index) == '\\') {
+                byteList.add((byte) value.charAt(index+1));
+                index += 2;
+            } else {
+                byteList.add((byte) value.charAt(index));
+                index++;
+            }
+        }
```
denalex, line 388: we need some explanation of what the above is doing

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/avro/AvroUtilities.java
```diff
@@ -161,6 +285,135 @@ private Schema getFieldSchema(DataType type, String colName) {
...
+
+        String[] subStrings = new String[posList.size() - 1];
+        for (int i = 0; i < posList.size() - 1; i++) {
+            subStrings[i] = new String(value, posList.get(i) + 1, posList.get(i+1) - posList.get(i) - 1);
+        }
+        return subStrings;
+    }
+
+    private boolean isQuote(char[] value, int index) {
```
denalex, line 350: probably not named appropriately, this seems to be checking whether the character is NOT escaped ?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/avro/AvroUtilities.java
```diff
@@ -161,6 +285,135 @@ private Schema getFieldSchema(DataType type, String colName) {
...
+
+    /**
+     * Split the outer-most Postgres array into its constituent elements
+     *
+     * Re-used from GPHDFS, commit 3b0bfdc169
+     * @param value string with a Postgres array
+     * @param delimiter character separating array elements in value
+     * @return array of strings for each element
+     */
+    private String[] getArraySplits(char[] value, char delimiter) {
```
denalex, line 301: how does it work with different database encodings ? byte[] to String requires proper Charset, not sure about char[]

denalex, line 301: all these methods can be public (or package) so they can be unit tested

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/avro/AvroUtilitiesTest.java
```diff
@@ -404,6 +409,152 @@ public void testObtainSchema_OnWrite_WithUserProvidedSchema_Json_NotFound_Spaces
...
+        Schema arraySchema = Schema.createArray(Schema.create(Schema.Type.INT));
+        Object result = avroUtilities.decodeString(arraySchema, "{1,2,3}", true, ',');
+        assertEquals(Arrays.asList(1, 2, 3), result);
+    }
+
+    @Test
+    public void testDecodeIntegerArrayWithNulls() {
+        Schema arraySchema = Schema.createArray(Schema.create(Schema.Type.INT));
+        Object result = avroUtilities.decodeString(arraySchema, "{1,NULL,3}", true, ',');
+        assertEquals(Arrays.asList(1, null, 3), result);
```
denalex, line 423: wouldn't avro schema have to be with union of null and int to support this ?

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/avro/AvroUtilitiesTest.java
```diff
@@ -404,6 +409,152 @@ public void testObtainSchema_OnWrite_WithUserProvidedSchema_Json_NotFound_Spaces
...
+    }
+
+    @Test
+    public void testDecodeStringMultiDimensionalArrayWithPXFGeneratedSchema() {
+        // pxf generated schemas only produce one-dimensional arrays
+        Schema schema = Schema.createArray(Schema.create(Schema.Type.INT));
+        String value = "{{1,2},{3,4}}";
+
+        Exception e = assertThrows(NumberFormatException.class, () -> avroUtilities.decodeString(schema, value, true,','));
+        assertEquals("For input string: \"{1,2}\"", e.getMessage());
```
denalex, line 546: it would be better if we can detect multidimensional array (depth >0 ?) and provide an error message that informs users of the problem and gives them the hint to provide the schema and not rely on PXF in this case

server/pxf-service/src/main/java/org/greenplum/pxf/service/BridgeOutputBuilder.java
```diff
@@ -430,10 +432,21 @@ void fillOneGPDBWritableField(OneField oneField, int colIdx)
                 case NUMERIC:
                 case TIMESTAMP:
                 case DATE:
```
denalex, line 434: can we put a comment here smth like "Resolvers are expected to serialize arrays into Strings with GPDB array encoding"

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/avro/AvroUtilities.java
```diff
@@ -161,6 +285,135 @@ private Schema getFieldSchema(DataType type, String colName) {
...
+                    }
+                    break;
+                default:
+                    break;
+            }
+        }
+
+        String[] subStrings = new String[posList.size() - 1];
+        for (int i = 0; i < posList.size() - 1; i++) {
+            subStrings[i] = new String(value, posList.get(i) + 1, posList.get(i+1) - posList.get(i) - 1);
```
bradfordb-vmware, line 345: w/ @denalex: In reviewing this code, it looks like this might not remove the Postgres array quotes and it should. Need to add a test case for this.

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/avro/AvroUtilitiesTest.java
bradfordb-vmware: I think we need to add a similar test in AvroResolver.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/AvroResolver.java
```diff
@@ -111,6 +113,12 @@ public void afterPropertiesSet() {
         avroRecord = makeAvroRecord(row.getData(), avroRecord);
         List<OneField> record = new LinkedList<>();
 
+        int avroFieldsSize = fields.size();
+        int numGpdbCols = context.getColumns();
+        if (avroFieldsSize != numGpdbCols) {
```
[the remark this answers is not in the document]
bradfordb-vmware: Yes. When reading data into Greenplum, [`BridgeOutputBuilder`](https://github.com/greenplum-db/pxf/blob/master/server/pxf-service/src/main/java/org/greenplum/pxf/service/BridgeOutputBuilder.java#L233-L236) will throw an exception if number of `OneFields` doesn't equal the number of GPDB columns.

When writing data from Greenplum to Avro, if the number of fields in the Avro schema is less than the number of GPDB columns, then an `ArrayIndexOutOfBoundsException` would thrown inside of [`AvroResolver`](https://github.com/greenplum-db/pxf/blob/master/server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/AvroResolver.java#L151). If the number of fields in the Avro schema is more than the number of GPDB columns, then an NPE _can_ be thrown in [`AvroFileAccessor`](https://github.com/greenplum-db/pxf/blob/master/server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/AvroFileAccessor.java#L196-L200) but that depends on if the Avro schema allows nulls for those extra fields.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/AvroResolver.java
```diff
@@ -191,19 +228,30 @@ GenericRecord makeAvroRecord(Object obj, GenericRecord reuseRecord)
...
         switch (fieldType) {
             case ARRAY:
                 if (fieldValue == null) {
-                    return addOneFieldToRecord(record, DataType.TEXT, null);
+                    return addOneFieldToRecord(record, gpdbColType, null);
                 }
                 List<OneField> listRecord = new LinkedList<>();
-                ret = setArrayField(listRecord, fieldValue, fieldSchema);
-                addOneFieldToRecord(record, DataType.TEXT, String.format("[%s]",
+                ret = setArrayField(listRecord, fieldValue, fieldSchema, gpdbColType);
```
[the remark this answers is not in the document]
bradfordb-vmware: w/ @ashuka24: We agree that the existing implementation has some overhead with wrapping array element values in `OneField` stored on a `List` versus using an array if element types (e.g. `int[]` or `string[]`).

We've spent some time looking at ways to refactor this and we'd like to propose splitting this off into a separate chore for the following reasons:

1. This cleanup is somewhat orthogonal to the array feature and since this PR is already rather large, it would easier to review the refactor separately.
2. The possible approached we sketched out would touch a lot of the helper methods and adds risk for potential regressions in functionality and warrants a careful review of the code changes.

PR #638, opened 2021-05-19
report exception class if there's no exception message

denalex:
When an exception that does not have an error message (such as NPE) happens in PXF, Greenplum displays:

```
ERROR:  PXF server error : No message available  (seg0 slice1 127.0.0.1:6003 pid=93107)
HINT:  Check the PXF logs located in the '/opt/greenplum-pxf/logs' directory on host 'abcd' or 'set client_min_messages=LOG' for additional details.
```

This commit adds the exception class (such as `java.lang.NullPointerException`) in place of a missing message.
Then this is what Greenplum user will see:
```
ERROR:  PXF server error : java.lang.NullPointerException  (seg0 slice1 127.0.0.1:6003 pid=93107)
HINT:  Check the PXF logs located in the '/opt/greenplum-pxf/logs' directory on host 'abcd' or 'set client_min_messages=LOG' for additional details.
```

### DIFF
--- a/server/pxf-api/src/main/java/org/greenplum/pxf/api/error/PxfRuntimeException.java
+++ b/server/pxf-api/src/main/java/org/greenplum/pxf/api/error/PxfRuntimeException.java
@@ -1,6 +1,7 @@
 package org.greenplum.pxf.api.error;
 
 import lombok.Getter;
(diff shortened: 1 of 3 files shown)


bradfordb-vmware (approved):
Nice usability improvement.

PR #642, opened 2021-05-28
Run JSON Functions integration test on Greenplum 6+ only

denalex:


automation/src/main/java/org/greenplum/pxf/automation/components/gpdb/Gpdb.java
```diff
@@ -51,6 +50,14 @@ public void init() throws Exception {
 
 		connect();
 
+		String pgVersion = dbConnection.getMetaData().getDatabaseProductVersion();
+		if (pgVersion.startsWith("8.3.")) {
+			version = 5; // Greenplum 5
+		} else if (pgVersion.startsWith("9.4.")) {
+			version = 6; // Greenplum 6
```
ashuka24, line 57: Is this going to be problematic code moving forward with GPDB7? Is there a way we can future proof this a little since we have code below that says `if (version >= 6)`

denalex: Yes, it will fail with GP7 once we start running tests against it since we don't yet know which PG version GP7 will be based off. We can default `version` to 7, but then it will be the same problem with 8. I suggest we solve that problem once we start running GP7-based pipelines, unless you have a better idea.

ashuka24: Could we get the version based off the Greenplum version instead of the Postgres version? 

If you were to call `select version()` in psql it gives both the Postgres version as well as the Greenplum version is there a way that we could use those values instead?

denalex: good idea, update the code

automation/src/main/java/org/greenplum/pxf/automation/components/gpdb/Gpdb.java
```diff
@@ -49,14 +51,7 @@ public void init() throws Exception {
...
 		connect();
-
-		String pgVersion = dbConnection.getMetaData().getDatabaseProductVersion();
-		if (pgVersion.startsWith("8.3.")) {
-			version = 5; // Greenplum 5
-		} else if (pgVersion.startsWith("9.4.")) {
-			version = 6; // Greenplum 6
-		}
-		ReportUtils.report(report, getClass(), "Determined Greenplum version: " + version + " from Postgres version: " + pgVersion);
+		version = determineVersion();
```
ashuka24, line 54: since it is possible for `determineVersion` to throw an exception, is it something we should handle?

ashuka24 (approved):

automation/src/main/java/org/greenplum/pxf/automation/components/gpdb/Gpdb.java
denalex: we should not handle it, if exception happens it will fail the test case that is using this which is what we want

PR #643, opened 2021-05-28
Update automation to run on either JDK8 or JDK11

bradfordb-vmware:


ashuka24 (commented):
I believe there was a set up step involving the pxf-extra.jar file that needs to be added.

automation/README.Linux.md
```diff
@@ -0,0 +1,80 @@
...
+Since automation reads standard out to parse the return codes of commands it executes, these extra escape codes cause it to incorrectly interpret the output.
+To resolve this issue, will explicitly set the value of `TERM` when running commands.
+You can then control which terminal capabilities are used for the SSH connection.
+I've found that explicitly setting `TERM=dumb` to work best.
+
+```bash
+alias make='TERM=dumb make'
+```
+
+It is not recommended to export this environment variable with this value since your terminal will have virtually no capabilities (no color prompt, ctrl+L won't clear the screen, etc.).
```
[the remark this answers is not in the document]
bradfordb-vmware: I've updated the PR so that the automation framework automatically sets `TERM` to `dumb` when running so no need for aliasing or mucking with environment variables.

automation/README.Linux.md
```diff
@@ -0,0 +1,65 @@
+# Running Automation on Linux
+
+**Note:** This information was documented based on the steps taken to get automation running on a Debian Bullseye system.
+
+## SSH Setup
+
+The PXF automation project uses an old SSH2 Java library that does not support newer key exchange algorithms (`KexAlgorithms`).
```
denalex, line 7: should we update SSH Java library if it will solve the problem ?

automation/Makefile
```diff
@@ -100,7 +100,9 @@ symlink_pxf_jars: check-env
 				cp "$${jar}" "$${link}"; \
 			fi; \
 		done; \
-		touch $(PXF_TMP_LIB)/pxf-extras.jar && \
+		empty_dir=$$(mktemp -d) && \
+		jar cf $(PXF_TMP_LIB)/pxf-extras.jar -C $${empty_dir} . && \
+		rmdir $${empty_dir} && \
```
denalex, line 105: not sure why this is needed

automation/src/test/resources/sut/default.xml
```diff
@@ -97,5 +97,15 @@
...
         <PGDATABASE></PGDATABASE>
+        <!--
+        The way automation runs test is that it connects via SSH to the localhost, spawning an instance of your login
+        shell (e.g., `bash` or `zsh`). Somehow, the `TERM` environment variable gets set to or defaults to
+        `xterm-256color`, which results in extraneous ANSI escape codes printing to standard out. Since automation reads
+        standard out to parse the return codes of commands it executes, these extra escape codes cause it to incorrectly
+        interpret the output. To resolve this issue, automation explicitly set the value of `TERM` when running
+        commands. The `dumb` terminal capabilities assumes that simplest set of terminal capabilities and appears to be
+        a safe default.
+        -->
```
denalex, line 108: while interesting, this comment pollutes other information in the file. I suggest making this a commit comment and whoever is interested as to why this parameter is included can use `git blame` to find the commit and read the description.

automation/Makefile
bradfordb-vmware: When using JDK11, creating a zero-length file for pxf-extras.jar causes the following error:

``console
[ERROR] error reading /home/bboyle/automation_tmp_lib/pxf-extras.jar; zip file is empty
[ERROR] /home/bboyle/workspace/pxf/automation/src/main/java/org/greenplum/pxf/automation/testplugin/FilterPrinterAccessor.java:[1,1] cannot access org.greenplum.pxf.automation.testplugin
  zip END header not found
```

It looks like the pxf-extras.jar needs to be a valid jar/zipfile so this creates one.

automation/README.Linux.md
bradfordb-vmware: There really isn't a newer version of the library that we use: [ch.ethz.ganymed:ganymed-ssh2](https://search.maven.org/artifact/ch.ethz.ganymed/ganymed-ssh2). We are currently using version `261` which was published in 2013. That last version published was `262` which was published in 2014. I did try upgrading to that version but it did not contain support for new key exchange algorithms,

ashuka24 (approved):

denalex (approved):

PR #644, opened 2021-06-01
Allow configuring connection timeout for data uploads to PXF

denalex:
Data can be read from PXF (via readable external tables) or written to PXF (via writable external tables). PXF uses Spring Boot with embedded Tomcat server to process these requests. The Tomcat server will set underlying socket timeout based on the `connectionTimeout` property which we currently expose in `pxf-application.properties` as `pxf.connection.timeout`. The same timeout is set for both `read` and `write` operations. However, when there is a lot of data to write to PXF, it might be necessary to set a longer timeout for `write` operations only.

This PR does a few things:
1. Sets `disableUploadTimeout` Tomcat property to `false`, thus making possible setting a separate timeout for `write` operations.
2. This effectively makes `pxf.connection.timeout` property manage the timeout for `read` (download) operations only. The default value continues to be set as `5 minutes`.
3. Introduces a new `pxf.connection.upload-timeout` property that manages the timeout for `write` (upload) operations only.
4. Logs into `pxf-service.log` file the actual parameters the Tomcat server has been initialized with when PXF starts.

Users that use PXF 6.0.0 or PXF 6.0.1 and set the custom value for `pxf.connection.timeout` to change the timeout for upload operations will now have to set this value for the new `pxf.connection.upload-timeout` property instead.

### DIFF
--- a/server/pxf-api/src/main/java/org/greenplum/pxf/api/configuration/PxfServerProperties.java
+++ b/server/pxf-api/src/main/java/org/greenplum/pxf/api/configuration/PxfServerProperties.java
@@ -5,6 +5,8 @@ import lombok.Setter;
 import org.springframework.boot.autoconfigure.task.TaskExecutionProperties;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 
+import java.time.Duration;
+
 /**
  * Configuration properties for PXF.
  */
@@ -47,6 +49,17 @@ public class PxfServerProperties {
          * Maximum number of headers allowed in the request
          */
         private int maxHeaderCount = 30000;
+
+        /**
+         * Whether upload requests will use the same read timeout as connectionTimeout
+         */
+        private boolean disableUploadTimeout = true; // default Tomcat setting
+
+        /**
+         * Timeout for reading data from upload requests, if disableUploadTimeout is set to false.
+         */
+        private Duration connectionUploadTimeout = Duration.ofMinutes(5); // 5 min is default Tomcat setting
+
     }
 
     public void setBase(String base) {
--- a/server/pxf-service/src/main/resources/application.properties
+++ b/server/pxf-service/src/main/resources/application.properties
@@ -36,7 +36,7 @@ server.tomcat.connection-timeout=${pxf.connection.timeout:5m}
 server.tomcat.mbeanregistry.enabled=true
 pxf.tomcat.max-header-count=30000
 pxf.tomcat.disable-upload-timeout=false
-pxf.tomcat.connection-upload-timeout=${pxf.connection.upload-timeout:${pxf.connection.timeout:5m}}
+pxf.tomcat.connection-upload-timeout=${pxf.connection.upload-timeout:5m}
 
 # timeout (ms) for the request - 1 day
 # TODO: spring_boot_todo what value should we set here
--- a/server/pxf-service/src/main/resources/application.properties
+++ b/server/pxf-service/src/main/resources/application.properties
@@ -35,6 +35,8 @@ server.tomcat.accept-count=100
 server.tomcat.connection-timeout=${pxf.connection.timeout:5m}
 server.tomcat.mbeanregistry.enabled=true
 pxf.tomcat.max-header-count=30000
+pxf.tomcat.disable-upload-timeout=false
+pxf.tomcat.connection-upload-timeout=${pxf.connection.upload-timeout:${pxf.connection.timeout:5m}}
 
 # timeout (ms) for the request - 1 day
 # TODO: spring_boot_todo what value should we set here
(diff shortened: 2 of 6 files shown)


server/pxf-api/src/main/java/org/greenplum/pxf/api/configuration/PxfServerProperties.java
```diff
@@ -47,6 +49,17 @@
          * Maximum number of headers allowed in the request
          */
         private int maxHeaderCount = 30000;
+
+        /**
+         * Whether upload requests will use the same read timeout as connectionTimeout
+         */
+        private boolean disableUploadTimeout = true; // default Tomcat setting
```
ashuka24, line 56: I'm a little confused. The PR comment says that the default setting for `disableUploadTimeout` should be false, yet here we set it to be true and down in `application.properties` we set it to be false. Can you explain why it is true here?

denalex: I set them here in Java code to the same values that Tomcat would be configured with if we didn't have the property listed in `application.properties` file. Then we put the "custom" value in the properties file, to override the default value. The value of "custom" might change (well, not that much with booleans), the default Tomcat will not change.

server/pxf-service/src/main/resources/application.properties
```diff
@@ -35,6 +35,8 @@ server.tomcat.accept-count=100
 server.tomcat.connection-timeout=${pxf.connection.timeout:5m}
 server.tomcat.mbeanregistry.enabled=true
 pxf.tomcat.max-header-count=30000
+pxf.tomcat.disable-upload-timeout=false
+pxf.tomcat.connection-upload-timeout=${pxf.connection.upload-timeout:${pxf.connection.timeout:5m}}
```
lisakowen, line 39: does this mean that if the property is not set, upload timeout is the value set for pxf.connection.timeout, and if that is not set it is 5 minutes?  doesn't pxf.connection.timeout have its own default?

bradfordb-vmware: It looks like we use `pxf.conection.timeout` in line 35 to set the value of `server.tomcat.connection-timeout`; if no value has been given for `pxf.conection.timeout` then we use a default of 5 minutes. I think that means that `pxf.connection.timeout` effectively has a default of 5 minutes.

server/pxf-service/src/main/resources/application.properties
```diff
@@ -35,6 +35,8 @@ server.tomcat.accept-count=100
 server.tomcat.connection-timeout=${pxf.connection.timeout:5m}
```
bradfordb-vmware, line 35: For the connection timeout, it looks like we can customize it via config files only, while for customizing the upload timeout we have to do that via `PxfTomcatCustomizer`. Is this a limitation from Boot?

server/pxf-service/src/main/resources/application.properties
denalex: 1. correct - the precedence as coded right now is the following:
- if the synthetic property `pxf.connection.upload-timeout` is explicitly set in `pxf-application.properties` then that value is used
- if that property is not set, but the synthetic property `pxf.connection.timeout` is explicitly set in `pxf-application.properties` then that value is used 
- otherwise a default of 5 minutes is used
By `synthetic` I mean a property that is invented by us (starts with `pxf.`) and does not have a direct impact on the system other than to hold a value to be applied to some other property.
2. `pxf.connection.timeout` has its own default only when it is uncommented, otherwise the property does not even exist. Its value is used to set `server.tomcat.connection-timeout` and now this property as well.

server/pxf-service/src/main/resources/application.properties
denalex: yes, in a way. They exposed a subset of Tomcat properties via `server.tomcat.` namespace, but not all of them. See here for details: https://docs.spring.io/spring-boot/docs/2.4.6/reference/html/appendix-application-properties.html#common-application-properties-server

server/pxf-service/src/main/resources/application.properties
```diff
@@ -35,6 +35,8 @@ server.tomcat.accept-count=100
...
 server.tomcat.mbeanregistry.enabled=true
 pxf.tomcat.max-header-count=30000
+pxf.tomcat.disable-upload-timeout=false
```
bradfordb-vmware, line 38: It looks like we always set `disableUploadTimeout` to false and this is not something that is readily changed by an end user. In that case, I think I would be in favor of fully decoupling the parameters `pxf.connection.timeout` and `pxf.connection.upload-timeout`. I think two independent settings is easier to understand than two that are dependent some of the time. I also think this maps very cleanly to read heavy workflows vs write heavy workflows.

denalex: yes, this is the dilemma. Do we expose independent timeouts with PXF prop names (which we kinda do, except the existing property does not have `read` or `download` qualifiers) and make it work with Tomcat by setting appropriate knobs or do we expose Tomcat configs and dependencies that users might be familiar with, in which case they expect 1 prop to cover both unless more tweaking is done. I have hidden 1 tweak and made it work as Tomcat user would kinda expect, but the resulting logic is now probably more confusing.

bradfordb-vmware (approved):

PR #646, opened 2021-06-03
pxf-json: add support for reading JSON arrays as text arrays

bradfordb-vmware:


server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/PgUtilities.java
```diff
@@ -120,6 +125,17 @@
 
     }
 
+    private static final CharSequenceTranslator UNESCAPE_ARRAY_ELEMENT;
+    static {
+        final Map<CharSequence, CharSequence> unescapeArrayElementMap = new HashMap<>();
+        unescapeArrayElementMap.put("\\\"", "\"");
+        unescapeArrayElementMap.put("\\\\", "\\");
+
+        UNESCAPE_ARRAY_ELEMENT = new LookupTranslator(
```
denalex, line 134: I wonder about the performance of map lookup based translator. Can we run a one-off bulk test of old vs new code ?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/PgUtilities.java
```diff
@@ -144,7 +160,17 @@ public String unescapeArrayElement(String str) {
...
-        return StringEscapeUtils.unescapeJava(str.substring(beginIndx, endIndex));
+        return UNESCAPE_ARRAY_ELEMENT.translate(str.substring(beginIndx, endIndex));
+    }
+
+    private static final CharSequenceTranslator ESCAPE_ARRAY_ELEMENT;
+    static {
+        final Map<CharSequence, CharSequence> escapeArrayElementMap = new HashMap<>();
+        escapeArrayElementMap.put("\"", "\\\"");
+        escapeArrayElementMap.put("\\", "\\\\");
+        ESCAPE_ARRAY_ELEMENT = new LookupTranslator(
```
denalex, line 171: maybe a better name for this is `ESCAPE_TRANSLATOR` and the other one `UNESCAPE_TRANSLATOR`

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonResolver.java
```diff
@@ -87,14 +101,14 @@ public void afterPropertiesSet() {
                 // If this column is an array index, ex. "tweet.hashtags[0]"
                 if (node.isArray()) {
                     // If the JSON node is an array, then add it to our list
-                    addFieldFromJsonArray(columnMetadata.getColumnType(), node, columnMetadata.getArrayNodeIndex());
+                    addFieldFromJsonArray(columnMetadata, node, columnMetadata.getArrayNodeIndex());
                 } else {
                     throw new IllegalStateException(columnMetadata.getColumnName() + " is not an array node");
                 }
             } else {
                 // This column is not an array type
```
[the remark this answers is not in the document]
bradfordb-vmware: Yes, the code now supports the example that you have given. For the given JSON record

```json
{"id": 1, "values": [[1], [1, 1], [1, 2, 1]]}
```

and external table definition

```sql
CREATE EXTERNAL TABLE pxf_json_nested_array(id integer, "values[1]" text[])
LOCATION ('pxf://json-data/nested-arrays.json?profile=file:json&server=local')
```

```console
bboyle=# SELECT * FROM pxf_json_nested_array ;
 id | values[1]
----+-----------
  1 | {1,1}
(1 row)
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/PgUtilities.java
bradfordb-vmware: Running a basic comparison on my local system, it looks like the map lookup based translator takes 8% more time than the old translator code (240ms vs 222ms); the standard deviation was about 3x larger.

I was using a data set that consisted of 22457 arrays that had between 0 and 18 elements with an average of 9 elements per array.

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/JsonExtensionTest.java
```diff
@@ -251,32 +265,240 @@ public void testMultipleFiles() throws Exception {
...
+
+        assertFalse(compareOutput(expectedOutput, actualOutput), "Output did not match expected output");
     }
 
-    @Override
-    public Class<? extends Fragmenter> getFragmenterClass() {
-        return HdfsDataFragmenter.class;
+    /**
+     * Uses the given input directory to run through the PXF unit testing framework. Uses the lines in the file for
+     * output testing.<br>
```
ashuka24, line 295: why is the break needed?

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonResolver.java
```diff
@@ -266,4 +286,21 @@ private static void validateBoolean(JsonNode val) throws BadRecordException {
...
     }
+
+    /**
+     * Format the given JSON array in Postgres array syntax
+     *
+     * @param jsonNode the {@link JsonNode} to serialize in Postgres array syntax
+     * @return a {@link String} containing the array elements in Postgre array syntax
+     * @throws JsonProcessingException
+     */
+    private String addAllFromJsonArray(JsonNode jsonNode) throws JsonProcessingException {
```
ashuka24, line 297: does this also handle nested json arrays? it seems like we do not check for that here?

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/JsonExtensionTest.java
```diff
@@ -76,7 +90,7 @@ public void testCompressedMultilineJsonFile() throws Exception {
         output.add("Fri Jun 07 22:45:02 +0000 2013,343136547123646465,text2,patronusdeadly,,,");
         output.add("Fri Jun 07 22:45:02 +0000 2013,343136547136233472,text3,NoSecrets_Vagas,,,");
 
-        super.assertOutput(new Path(System.getProperty("user.dir") + File.separator
+        assertOutput(new Path(System.getProperty("user.dir") + File.separator
```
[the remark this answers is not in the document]
bradfordb-vmware: The method `assertOutput` had been on `PxfUnit` which this class was previously extending. The only class that extended `PxfUnit` was `JsonExtensionTest` so we deleted/removed `PxfUnit` and moved the methods from there to here.

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/JsonExtensionTest.java
bradfordb-vmware: I don't know if it is needed or not, but this was copied/moved from `PxfUnit`.

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/JsonExtensionTest.java
```diff
@@ -251,32 +265,240 @@ public void testMultipleFiles() throws Exception {
...
+     * @param expectedOutput The expected output
+     * @param actualOutput   The actual output
+     * @return True if no errors, false otherwise.
+     */
+    private boolean compareUnorderedOutput(List<String> expectedOutput, List<String> actualOutput) {
+        return compareOutput(expectedOutput, actualOutput, true);
+    }
+
+    private boolean compareOutput(List<String> expectedOutput, List<String> actualOutput, boolean ignoreOrder) {
+        boolean error = false;
```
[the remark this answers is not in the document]
bradfordb-vmware: Not sure. This code is just being moved from `PxfUnit`. We're removing `PxfUnit` because it was not really needed and was making it difficult to setup for testing with new dependencies.

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/JsonResolverTest.java
```diff
@@ -130,12 +157,24 @@ public void testGetFieldsWithQuotedValues() throws Exception {
...
+                "\"type_int_arr_as_textarray\":[\"1\",\"2\",\"3\"]," +
+                "\"type_bigint_arr_as_textarray\":[null,\"10101010101\"]," +
+                "\"type_smallint_arr_as_textarray\":[\"5\",\"null\"]," +
+                "\"type_float_arr_as_textarray\":[\"1.1\"]," +
+                "\"type_double_arr_as_textarray\":[\"1.1\",\"2.2\"]," +
+                "\"type_string_arr_as_textarray\":[\"hello\",\"world\"]," +
+                "\"type_boolean_arr_as_textarray\":[\"true\",\"false\",null]," +
+                "\"type_string_arr_arr_as_textarray\":[[\"good\",\"morning\"],null,[\"guten\",\"morgen\"]]," +
+                "\"type_object_arr_as_textarray\":[{\"id\":1,\"val\":\"a\"},{\"id\":2,\"val\":\"b\"}]," +
+                "\"type_null_as_textarray\":null," + // quoted null doesn't really make sesnse
```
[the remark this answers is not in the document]
bradfordb-vmware: This is the input JSON for `testGetFieldsWithQuotedValues` which is very similar to `testGetFieldsWithUnquotedValues`.  If we look at the field `type_smallint` for the two test casses

```json
{
    "type_smallint": 13,
}
```

versus

```json
{
    "type_smallint": "26",
}
```

The test is seeing if we can "cast" from strings in JSON to the appropriate GPDB column type. For the case of `type_null_as_textarray`, there really isn't any way to "cast" from a string to an array. We could consider some special handling for the string `"null"` (and the various permutations of capitalization).

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonResolver.java
bradfordb-vmware: Good catch. Let me see if I can easily add that.

bradfordb-vmware: This should now support nested JSON arrays and I've updated the nested array test cases in `JsonResolverTest`.

ashuka24 (approved):

PR #647, opened 2021-06-03
docs - pxf 6->6.newer upgrade

lisakowen:
fill in the template for pxf 6->6.newer upgrade.  will add info specific to 6.1.0 in separate PR.

doc review site link:  https://docs-lisa-pxf-upgradetempl.sc2-04-pcf1-apps.oc.vmware.com/pxf/6-1/using/upgrade_6.html

### DIFF
--- a/docs/content/upgrade_6.html.md.erb
+++ b/docs/content/upgrade_6.html.md.erb
@@ -33,7 +33,7 @@ Perform this procedure before you upgrade to a new version of PXF 6.x:
     gpadmin@gpmaster$ pxf cluster stop
     ```
 
-1. Back up the PXF user configuration files; for example, if `PXF_BASE=/usr/local/pxf-gp6`:
+1. (Optional, Recommended) Back up the PXF user configuration files; for example, if `PXF_BASE=/usr/local/pxf-gp6`:
 
     ``` shell
     gpadmin@gpmaster$ cp -avi /usr/local/pxf-gp6 pxf_base.bak
@@ -69,7 +69,7 @@ After you install the new version of PXF, perform the following procedure:
 
 1. **If you are upgrading <i>from</i> PXF version 6.0.x**:
     - If you previously set the `pxf.connection.timeout` property to change the write/upload timeout, you must now set the `pxf.connection.upload-timeout` property for this purpose.
-    - Existing external tables that access Avro arrays and JSON objects will continue to work as-is. If you want to take advantage of the new Avro array read/write functionality or the new JSON object support, create a new external with the adjusted DDL. If you can access the data with the new external table as you expect, you may choose to drop and recreate the existing external table.
+    - Existing external tables that access Avro arrays and JSON objects will continue to work as-is. If you want to take advantage of the new Avro array read/write functionality or the new JSON object support, create a new external table with the adjusted DDL. If you can access the data with the new external table as you expect, you may choose to drop and recreate the existing external table.
 
 1. Synchronize the PXF configuration from the master host to the standby master and each Greenplum Database segment host. For example:
 
--- a/docs/content/upgrade_6.html.md.erb
+++ b/docs/content/upgrade_6.html.md.erb
@@ -27,12 +27,16 @@ Perform this procedure before you upgrade to a new version of PXF 6.x:
     gpadmin@gpmaster$ pxf version
     ```
 
-1. ANY NEXT STEP HERE??
-
 1. Stop PXF on each Greenplum host as described in [Stopping PXF](cfginitstart_pxf.html#stop_pxf):
 
     ``` shell
     gpadmin@gpmaster$ pxf cluster stop
+    ```
+
+1. Back up the PXF user configuration files; for example, if `PXF_BASE=/usr/local/pxf-gp6`:
+
+    ``` shell
+    gpadmin@gpmaster$ cp -avi /usr/local/pxf-gp6 pxf_base.bak
     ```
 
 
--- a/docs/content/upgrade_6.html.md.erb
+++ b/docs/content/upgrade_6.html.md.erb
@@ -2,18 +2,18 @@
 title: Upgrading from an Earlier PXF 6 Release
 ---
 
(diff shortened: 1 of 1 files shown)


docs/content/upgrade_6.html.md.erb
```diff
@@ -27,14 +27,21 @@ Perform this procedure before you upgrade to a new version of PXF 6:
     gpadmin@gpmaster$ pxf version
     ```
 
-1. next step
+1. ANY NEXT STEP HERE??
```
bradfordb-vmware, line 30: The only step I can think of adding here is advising users to backup their PXF configuration files (e.g., `keytab/`, `conf/`, `servers/`, etc). They should be safe during the upgrade, but a quick `cp -avi $PXF_BASE $PXF_base.bak` is cheap.

Probably would want to do this after running `pxf cluster stop`.

docs/content/upgrade_6.html.md.erb
```diff
@@ -27,14 +27,21 @@ Perform this procedure before you upgrade to a new version of PXF 6:
...
 
-1. Install the new version of PXF, identify and note the new PXF version number, and then continue your PXF upgrade with [Step 2: Complete the Upgrade to a Newere PXF 6](#pxfup).
+    ``` shell
+    gpadmin@gpmaster$ pxf cluster stop
+    ```
+
+
+## <a id="pxfinst"></a>Step 2: Installing the New PXF 6.x
+
+Follow the steps outlined in [Installing PXF 6.x](../release/installing_pxf.html), and identify and note the new PXF version number.
```
bradfordb-vmware, line 41: I think users need to run `pxf cluster register` after upgrading. The installation page includes a _Next Steps_ section which links to _Configuring PXF_ that includes this step but it's not obvious and I only found it because I was looking for it. Should we explicitly point them to the configuration instructions here?

docs/content/upgrade_6.html.md.erb
```diff
@@ -44,15 +51,26 @@ After you install the new version of PXF, perform the following procedure:
...
+1. PXF 6.x includes a new version of the `pxf` extension. You must update the `pxf` extension in every Greenplum database in which you are using PXF. A database superuser or the database owner must run this SQL command in the `psql` subsystem or in an SQL script:
 
-1. next step
+    ``` sql
+    ALTER EXTENSION pxf UPDATE;
+    ```
+
+1. **If you are upgrading <i>from</i> PXF version 6.0.x**:
+    - upgrade step for config property
+    - (tbd?) upgrade step for external tables accessing avro arrays
```
bradfordb-vmware, line 62: If they have existing tables that are reading Avro or JSON files, they should continue to work as-is. If they want to take advantage of the new array support, they can re-create the table with the appropriate columns adjusted. I think a link to the Avor or JSON documentation should sufficient here.

ashuka24 (approved):

docs/content/upgrade_6.html.md.erb
lisakowen: thanks for reviewing, @bradfordb-vmware .  i wasn't sure about this.  i had thought that since pxf 6 only copies the control file over, they would only need to do this once for pxf 6.x, and they would have already ran the command if they have an existing pxf 6 installed (hence they are on this upgrade page).  i can add that step in if this is not the case.

docs/content/upgrade_6.html.md.erb
lisakowen: good suggestion.  i added the copy, but to an explicit file path because the user probably won't have $PXF_BASE set.

docs/content/upgrade_6.html.md.erb
bradfordb-vmware: The control file contains a field called `default_version` that could potentially change  (e.g., `2.0` to `2.1`) between PXF 6 minor versions. We haven't changed this within 6.x but we could and the safest thing might be to just re-run the command.

mperezfuster (approved):

docs/content/upgrade_6.html.md.erb
```diff
@@ -27,14 +27,25 @@ Perform this procedure before you upgrade to a new version of PXF 6:
...
 
-1. next step
+1. Stop PXF on each Greenplum host as described in [Stopping PXF](cfginitstart_pxf.html#stop_pxf):
 
-1. Stop PXF on each Greenplum host as described in [Stopping PXF](cfginitstart_pxf.html#stop_pxf).
+    ``` shell
+    gpadmin@gpmaster$ pxf cluster stop
+    ```
+
+1. Back up the PXF user configuration files; for example, if `PXF_BASE=/usr/local/pxf-gp6`:
```
denalex, line 36: should we mention that this is an optional recommended operation ?

docs/content/upgrade_6.html.md.erb
```diff
@@ -44,15 +55,31 @@ After you install the new version of PXF, perform the following procedure:
...
+
+1. You must update the `pxf` extension in every Greenplum database in which you are using PXF. A database superuser or the database owner must run this SQL command in the `psql` subsystem or in an SQL script:
+
+    ``` sql
+    ALTER EXTENSION pxf UPDATE;
+    ```
+
+1. **If you are upgrading <i>from</i> PXF version 6.0.x**:
+    - If you previously set the `pxf.connection.timeout` property to change the write/upload timeout, you must now set the `pxf.connection.upload-timeout` property for this purpose.
+    - Existing external tables that access Avro arrays and JSON objects will continue to work as-is. If you want to take advantage of the new Avro array read/write functionality or the new JSON object support, create a new external with the adjusted DDL. If you can access the data with the new external table as you expect, you may choose to drop and recreate the existing external table.
```
denalex, line 72:
```suggestion
    - Existing external tables that access Avro arrays and JSON objects will continue to work as-is. If you want to take advantage of the new Avro array read/write functionality or the new JSON object support, create a new external table with the adjusted DDL. If you can access the data with the new external table as you expect, you may choose to drop and recreate the existing external table.
```

denalex (approved):

PR #649, opened 2021-06-04
Docs: Added Oracle JDBC example and moved JDBC examples to separate pages"

mperezfuster:
Added content on how to connect to Oracle via JDBC.
Since the JDBC section was getting too long, split all the examples into different pages and moved them outside the main page.
Preview: https://mireia-pxf-oracle.sc2-04-pcf1-apps.oc.vmware.com/pxf/6-1/using/jdbc_pxf_oracle.html


bradfordb-vmware (approved):
I focused mostly on the new Oracle content, assuming the old content was just being moved. Looks good.

docs/content/jdbc_pxf_oracle.html.md.erb
```diff
@@ -0,0 +1,182 @@
...
+
+    ``` shell
+    $ sqlplus oracleuser
+    ```
+
+4. Create a table named `countries`, insert some data into this table, and commit the transaction:
+
+    ``` sql
+    > CREATE TABLE countries (country_id int, country_name varchar(40), population float);
+    > INSERT INTO countries (country_id, country_ values (3, 'Portugal', 10.28);
```
lisakowen, line 72: i am not familiar with oracle ... is the `country_` correct?

docs/content/jdbc_pxf_oracle.html.md.erb
```diff
@@ -0,0 +1,182 @@
...
+              LOCATION('pxf://oracleuser.countries?PROFILE=jdbc&SERVER=oracle')
+              FORMAT 'CUSTOM' (formatter='pxfwritable_import');
+    ```
+
+2. Display all rows of the `oracle_countries` table:
+
+    ``` sql
+    gpadmin=# SELECT * FROM oracle_countries ;
+    country_id | country_name | population 
+    ------------+--------------+------------
```
lisakowen, line 147: the +'s do not align with the column lines. (same in the write example.)

lisakowen (approved):
looks great, just a few minor comments.

also wondering if we should add an Examples section on the jdbc_pxf.html page which lists the examples with links.  otherwise the user might not know these examples exist if they do not expand the left nav.

docs/content/jdbc_pxf_oracle.html.md.erb
mperezfuster: I don't know what happened there! must be bad copy paste, thanks for noticing!

mperezfuster:
@lisakowen thanks for your suggestion, what about something like this (new commit):
https://mireia-pxf-oracle.sc2-04-pcf1-apps.oc.vmware.com/pxf/6-1/using/jdbc_pxf_oracle.html
We reference the first three examples separately to the name queries one since they're on different sections, let me know what you think.

lisakowen:
makes sense to separate the examples.  up to you, but i was more thinking of adding an "Examples" section above the "About Named Queries" topic, mostly so it will show up in the right nav.

mperezfuster:
oh yes makes sense, what about now?
https://mireia-pxf-oracle.sc2-04-pcf1-apps.oc.vmware.com/pxf/6-1/using/jdbc_pxf.html#examples

lisakowen (approved):
looks great!

denalex (approved):

docs/content/jdbc_pxf_mysql.html.md.erb
```diff
@@ -0,0 +1,186 @@
...
+under the License.
+-->
+
+In this example, you:
+
+- Create a MySQL database and table, and insert data into the table
+- Create a MySQL user and assign all privileges on the table to the user
+- Configure the PXF JDBC connector to access the MySQL database
+- Create a PXF readable external table that references the MySQL table
+- Read the data in the MySQL table
```
ashuka24, line 30:
```suggestion
- Read the data in the MySQL table using PXF
```

ashuka24: I think maybe this change can be added below in lines 32 and 33 as well for clarity that you are not writing and reading from MySQL directly, but through Greenplum?

dyozie (approved):

PR #650, opened 2021-06-07
Pull container images from GCR instead of DockerHub

bradfordb-vmware:
The GP RelEng team is moving all upstream container images from
DockerHub to a public Google Container Registrt (GCR). This commit
updates our CI code to point at the new location.

- update CCP image tag
- change Concourse docker-image to registry-image
- replace DockerHub repository `pivotldata` with `gcr.io/data-gpdb-public-images`
  - matched FROM in Dockerfile with build arg from couldbuild steps
  - identified several unused Dockerfiles for potential deletion


bradfordb-vmware:
Add @greenplum-db/gp-releng to look over and make sure we are point at the correct upstream images.

concourse/docker/centos6-hdp2-secure/Dockerfile
```diff
@@ -1,6 +1,7 @@
-ARG BASE_IMAGE=gpdb-dev:centos6
+# FIXME: this file does not appear to be used anymore; can we delete it (and the containing directory)?
```
denalex, line 1: yes, we should delete if this is not used anymore. there's an off-chance we might need to be able to create the image if we are to do a patch for an old release that supported centos6, so maybe capture the knowledge of this file somewhere.

concourse/docker/pxf-dev-base/cloudbuild.yaml
```diff
@@ -29,7 +29,7 @@ steps:
   id: gpdb5-centos7-test-pxf-image
   args:
   - 'build'
-  - '--build-arg=BASE_IMAGE=${_BASE_IMAGE_REPOSITORY}/centos-gpdb-dev:7-gcc6.2-llvm3.7'
+  - '--build-arg=BASE_IMAGE=${_BASE_IMAGE_REPOSITORY}/gpdb5-centos7-build-test:latest'
```
denalex, line 32: same comment about build/test

concourse/docker/pxf-dev-server/Dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG BASE_IMAGE=pivotaldata/gpdb-pxf-dev:centos7
+ARG BASE_IMAGE=gcr.io/data-gpdb-ud/gpdb-pxf-dev/gpdb6-centos7-test-pxf:latest
```
denalex, line 1: this is probably correct, but not a straight-forward replacement

concourse/pipelines/certification_pipeline.yml
```diff
@@ -92,8 +92,8 @@ resources:
   type: registry-image
   icon: docker
   source:
-    repository: pivotaldata/ccp
-    tag: 7
+    repository: gcr.io/data-gpdb-public-images/ccp
+    tag: latest
```
denalex, line 96: that's probably ok, but opens us up to be dependent on potentially destabilizing changes to the CCP image

concourse/pipelines/pg_regress_pipeline.yml
```diff
@@ -36,22 +36,22 @@ resources:
...
 
 - name: gpdb6-centos7-build
-  type: docker-image
+  type: registry-image
   source:
-    repository: pivotaldata/gpdb7-centos7-build
+    repository: gcr.io/data-gpdb-public-images/gpdb6-centos7-build
     tag: latest
 
 - name: gpdb-pxf-dev-centos7-hdp2-server
```
denalex, line 50: we now violate the invariant of "name=repository-tag" i.e it was
`gpdb-pxf-dev-centos7-hdp2-server = pivotaldata/gpdb-pxf-dev - centos7-hdp2-server` but now is
`gpdb-pxf-dev-centos7-hdp2-server <> gpdb6-centos7-test-pxf-hdp2`
Maybe we should align the names of resources and underlying repo images

concourse/scripts/download_pivnet_cli.bash
```diff
@@ -7,6 +7,11 @@ set -e
 PATH=${PIVNET_CLI_DIR}:${PATH}
 chmod +x "${PIVNET_CLI_DIR}/pivnet"
 
+if ! type jq &> /dev/null; then
```
denalex, line 10: I think as shell coding style we were trying to always use `[[` notation.

concourse/scripts/download_pivnet_cli.bash
```diff
@@ -7,6 +7,11 @@ set -e
...
+	apt-get update --quiet=1
```
denalex, line 11: are we sure we always will be running this script on Ubuntu ? Can `jq` be added to the docker image ?

concourse/scripts/download_pivnet_cli.bash
bradfordb-vmware: The commands `[` and `[[` are used for evaluating an expression and exiting with a status determined by the expression; for example `[[ -f <some-path> ]]` tests if `some-path` is a regular file. Here `type` takes the place of `[[` and it is actually a syntax error to wrap in `[[...]]`.

The equivalent using `[[...]]` would be `[[ -x /usr/bin/jq ]]` which hard codes the expected path of `jq` while `type` looks for it anywhere on `$PATH`.

concourse/docker/centos6-hdp2-secure/Dockerfile
bradfordb-vmware: As far as I can tell, this image was being built by the `docker-images.yaml` pipeline which was removed by commit c7ee7049d2915017209385bd556b54bc23887e9c. This was the commit that replaced the docker-images pipeline with singlecluster pipeline.

concourse/docker/pxf-dev-server/Dockerfile
bradfordb-vmware: The `cloudbuild.yaml` that builds this image is/was always overriding `$BASE_IMAGE` to `gcr.io/data-gpdb-ud/gpdb-pxf-dev/gpdb6-centos7-test-pxf:latest`. We could remove a layer of complexity by either:

1. Hard-code the `FROM` step in the Dockerfile to be `gcr.io/data-gpdb-ud/gpdb-pxf-dev/gpdb6-centos7-test-pxf:latest`
2. Remove the (unused) default value for the `BASE_IMAGE` build arg from the Dockerfile.

I cannot think of a scenario where we would need to dynamically switch the base image, so I would vote for option 1. Do you have any objections to this?

concourse/pipelines/certification_pipeline.yml
bmdoil: Small clarification, the `7` tag was functionally the same as `latest`, new image digests were pushed to dockerhub using the `7` tag. With the changes, all images are pushed with incrementing v1,v2...vN tags with the most recent being `latest`. 

You may consider pinning the resource at a specific tag. Currently, the tags for `gcr.io/data-gpdb-public-images` and `gcr.io/data-gpdb-public-images` can be viewed with the `gcloud container images list-tags [repository]` command using the same service account key used for reading private images.

```
gcloud container images list-tags gcr.io/data-gpdb-public-images/ccp
DIGEST        TAGS       TIMESTAMP
8a563ebaa041  latest,v3  2021-05-20T21:36:41
0320ac9a6359  v2         2021-05-19T18:26:46
a6e30853e541  dev,devv4  2021-05-19T17:24:38
486cc1628fd5  v1         2021-05-19T12:02:14
8b3d295ac6e4             2021-05-12T17:21:25
```

concourse/pipelines/pg_regress_pipeline.yml
```diff
@@ -36,22 +36,22 @@ resources:
...
-  type: docker-image
+  type: registry-image
   source:
-    repository: pivotaldata/gpdb-pxf-dev
-    tag: centos7-hdp2-server
+    repository: gcr.io/data-gpdb-ud/gpdb-pxf-dev/gpdb6-centos7-test-pxf-hdp2
+    tag: latest
 
 # ~/workspace/gp-continuous-integration/secrets/gpdb6-integration-testing.dev.yml
 - name: libquicklz-centos7
```
bmdoil, line 57: Somewhat related, you can drop these resources from the compile job if you'd like, they are contained in `gpdb6-centos7-build`
```
libquicklz-centos7
libquicklz-devel-centos7
libsigar-centos7
```

concourse/scripts/download_pivnet_cli.bash
bmdoil: `jq` can be added to the image.

bmdoil (approved):

concourse/docker/pxf-dev-base/README.md
```diff
@@ -26,7 +26,7 @@ command to build the image:
 
     pushd ~/workspace/pxf/concourse/docker/pxf-dev-base/
     docker build \
-      --build-arg=BASE_IMAGE=pivotaldata/centos-gpdb-dev:7-gcc6.2-llvm3.7 \
+      --build-arg=BASE_IMAGE=gcr.io/data-gpdb-public-images/gpdb5-centos7-build-test:latest \
```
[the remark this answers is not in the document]
bradfordb-vmware: I followed up with the RelEng team and since this is a new image for GPDB5 CI that replaces an old image that was being used for both building and testing, they are providing a single `-build-test` image instead of separate `-build` and `-test` images.

concourse/docker/pxf-dev-server/Dockerfile
denalex: sounds good, thanks

PR #652, opened 2021-06-09
docs - read/write avro array

lisakowen:
docs for https://github.com/greenplum-db/pxf/pull/636.

if i interpreted the PR correctly, pxf now supports reading/writing arrays of avro primitive types.  and also supports mapping some other types to string arrays when the avro schema is provided.

i updated the data type mapping tables.  also updated the example & DDL to use text[] for followers data field.

questions:
1.  read/write mappings - i could use some suggestions on how to present this better.  i didn't want to have a separate line for each "array of primitive type", but maybe that is needed.
2.  i assumed that pxf could support arrays of all of the primitive types (like double) even though the PR did not explicitly call that out.  let me know if this is not correct.

doc review site link:   https://docs-lisa-pxf-avroarray.sc2-04-pcf1-apps.oc.vmware.com/pxf/6-1/using/hdfs_avro.html

### DIFF
--- a/docs/content/hdfs_avro.html.md.erb
+++ b/docs/content/hdfs_avro.html.md.erb
@@ -36,11 +36,13 @@ Apache Avro is a data serialization framework where the data is serialized in a 
 
 ### <a id="profile_hdfsavrodatamap"></a>Data Type Mapping
 
-Avro supports both primitive and complex data types. 
+The Avro specification defines [primitive](https://avro.apache.org/docs/current/spec.html#schema_primitive), [complex](https://avro.apache.org/docs/current/spec.html#schema_complex), and [logical](https://avro.apache.org/docs/current/spec.html#Logical+Types) types.
 
 To represent Avro primitive data types and Avro arrays of primitive types in Greenplum Database, map data values to Greenplum Database columns of the same type. 
 
 Avro supports other complex data types including arrays of non-primitive types, maps, records, enumerations, and fixed types. Map top-level fields of these complex data types to the Greenplum Database `text` type. While PXF does not natively support reading these types, you can create Greenplum Database functions or application code to extract or further process subcomponents of these complex data types.
+
+Avro supports logical data types including include decimal, date, time, and duration types. You must similarly map these data types to the Greenplum Database `text` type.
 
 ### <a id="datatype_map_read "></a>Read Mapping
 
@@ -57,15 +59,17 @@ PXF uses the following data type mapping when reading Avro data:
 | int | int |
 | long | bigint |
 | string | text |
-| Complex type: Array (any dimension) of type boolean, bytes, double, float, int, long, string | array (any dimension) of type boolean, bytea, double, real, bigint, text   |
-| Complex type: Array of other types **and** *Avro schema is provided* | text[]   |
+| Complex type: Array (any dimension) of type: boolean, bytes, double, float, int, long, string | array (any dimension) of type: boolean, bytea, double, real, bigint, text   |
+| Complex type: Array of {union of null with type}: boolean, bytes, double, float, int, long, string | boolean[], bytea[], double[], real[], bigint[], text[]   |
+| Complex type: Array of other types<br>(*Avro schema is provided*) | text[]   |
 | Complex type: Map, Record, or Enum  | text, with delimiters inserted between collection items, mapped key-value pairs, and record data.   |
 | Complex type: Fixed     | bytea (supported for read operations only).   |
 | Union      | Follows the above conventions for primitive or complex data types, depending on the union; must contain 2 elements, one of which must be null.  |
+| Logical type: decimal, date, time, timestamp, duration | text |
 
 ### <a id="datatype_map_Write"></a>Write Mapping
 
-PXF supports writing only Avro primitive types and arrays of Avro primitive types. PXF supports writing other complex types to Avro as string.
+PXF supports writing Avro primitive types and arrays of Avro primitive types. PXF supports writing other complex types to Avro as string.
 
 PXF uses the following data type mapping when writing Avro data:
 
@@ -84,10 +88,10 @@ PXF uses the following data type mapping when writing Avro data:
 | smallint<sup>2</sup> | int |
 | text | string |
 | varchar | string |
-| date, numeric, time, timestamp, timestamptz **and** *no Avro schema is provided* | string |
-| array (any dimension) of type bigint, boolean, bytea, double, int, real, text **and** *Avro schema is provided* | Array (any dimension) of type long, boolean, bytes, double, int, float, string |
-| bigint[], boolean[], bytea[], double[], int[], real[], text[] **and** *no Avro schema is provided* | long[], boolean[], bytes[], double[], int[], float[], string[] |
-| date[], numeric[], time[], timestamp[], timestamptz[] **and** *Avro is schema is provided* | string[] |
+| numeric, date, time, timestamp, timestamptz<br>(*no Avro schema is provided*) | string |
+| array (any dimension) of type: bigint, boolean, bytea, double, int, real, text <br>(*Avro schema is provided*) | Array (any dimension) of type: long, boolean, bytes, double, int, float, string |
+| bigint[], boolean[], bytea[], double[], int[], real[], text[] <br>(*no Avro schema is provided*) | long[], boolean[], bytes[], double[], int[], float[], string[] |
+| numeric[], date[], time[], timestamp[], timestamptz[] <br> (*Avro is schema is provided*) | string[] |
 | enum, record | string |
 
 </br><sup>1</sup>&nbsp;PXF right-pads <code>char[<i>n</i>]</code> types to length <code><i>n</i></code>, if required, with white space.
@@ -328,7 +332,7 @@ PXF returns an error if you provide a schema that does not include a `union` of 
 PXF supports writing only Avro primitive data types and Avro Arrays of the types identified in [Data Type Write Mapping](#datatype_map_Write). PXF does not support writing complex types to Avro:
 
 - When you specify a `SCHEMA` file in the `LOCATION`, the schema must include only primitive data types.
-- When PXF generates the schema, it writes any complex type that you specify in the writable external table column definition to the Avro file as a single Avro `string` type. For example, if you write an array of numerics, PXF converts the array to a `string`, and you must read this data with a Greenplum `text`-type column.
+- When PXF generates the schema, it writes any complex type that you specify in the writable external table column definition to the Avro file as a single Avro `string` type. For example, if you write an array of the Greenplum `numeric` type, PXF converts the array to a `string`, and you must read this data with a Greenplum `text`-type column.
 
 
 ### <a id="topic_avrowrite_example"></a>Example: Writing Avro Data
--- a/docs/content/hdfs_avro.html.md.erb
+++ b/docs/content/hdfs_avro.html.md.erb
@@ -90,7 +90,7 @@ PXF uses the following data type mapping when writing Avro data:
 | varchar | string |
 | numeric, date, time, timestamp, timestamptz<br>(*no Avro schema is provided*) | string |
 | array (any dimension) of type: bigint, boolean, bytea, double, int, real, text <br>(*Avro schema is provided*) | Array (any dimension) of type: long, boolean, bytes, double, int, float, string |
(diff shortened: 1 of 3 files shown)


docs/content/hdfs_avro.html.md.erb
```diff
@@ -57,13 +57,15 @@ PXF uses the following data type mapping when reading Avro data:
 | int | int |
 | long | bigint |
 | string | text |
-| Complex type: Array, Map, Record, or Enum  | text, with delimiters inserted between collection items, mapped key-value pairs, and record data.   |
+| Complex type: Array (any dimension) of type boolean, bytes, double, float, int, long, string | array (any dimension) of type boolean, bytea, double, real, bigint, text   |
+| Complex type: Array of other types **and** *Avro schema is provided* | text[]   |
+| Complex type: Map, Record, or Enum  | text, with delimiters inserted between collection items, mapped key-value pairs, and record data.   |
```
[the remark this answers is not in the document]
bradfordb-vmware: This the table for reading from Avro files, which doesn't have those types. Avro has `date` and `timestamp` [_logical types_](http://avro.apache.org/docs/current/spec.html#Logical+Types) but we'd still be reading them from the Avro file as an `int` or `long`.

One thing that we don't mention here is that we support _both_

- array of primitive (e.g., an array of integers, none of which are null)
- array of union of null with primitive (e.g., an array of integers some of which might be null)

For both cases, we'd read them to the corresponding Greenplum array type (e.g., `int[]`).

docs/content/hdfs_avro.html.md.erb
```diff
@@ -329,10 +325,10 @@ If you do not specify a `SCHEMA` file, PXF generates a schema for the Avro file
...
 
-PXF supports writing only Avro primitive data types. It does not support writing complex types to Avro:
+PXF supports writing only Avro primitive data types and Avro Arrays of the types identified in [Data Type Write Mapping](#datatype_map_Write). PXF does not support writing complex types to Avro:
 
 - When you specify a `SCHEMA` file in the `LOCATION`, the schema must include only primitive data types.
-- When PXF generates the schema, it writes any complex type that you specify in the writable external table column definition to the Avro file as a single Avro `string` type. For example, if you write an array of integers, PXF converts the array to a `string`, and you must read this data with a Greenplum `text`-type column.
+- When PXF generates the schema, it writes any complex type that you specify in the writable external table column definition to the Avro file as a single Avro `string` type. For example, if you write an array of numerics, PXF converts the array to a `string`, and you must read this data with a Greenplum `text`-type column.
```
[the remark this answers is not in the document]
bradfordb-vmware: For the particular case of `numeric`, it will be a string because there isn't a matching Avro type for `numeric`; GPDB `numeric` is a variable storage size while the Avro numeric primitives (`int`, `long`, `float`, and `double`) are fixed (32 or 64-bit) size.

docs/content/hdfs_avro.html.md.erb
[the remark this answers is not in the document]
denalex: I guess my questions is from a customer point of view -- if I have a logical type which is date or timestamp in my Avro file (and an array of those as well), and I want this info in GPDB, what should I do ? Should I map it to int ? What do I do with that int afterwards to arrive at the proper date / timestamp ? Or should I define it as `date` in GPDB and expect PXF to do the conversion ? 

Right now there is no info whatsoever for READ case and for WRITE we are saying" `date, numeric, time, timestamp, timestamptz&nbsp;and&nbsp;no Avro schema is provided | string" so we are storing them as strings that GPDB sends us, not as Avro logical types. Maybe smith to look at for the next enhancement.


.

docs/content/hdfs_avro.html.md.erb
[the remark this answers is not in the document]
denalex: thanks for clarification @bradfordb-vmware , it reads very generically though, so "array of numerics" sounds to me like an Arya of smith with is a number, not quite a `numeric[]`. It also is not obvious that `numeric` is a `complex type`, maybe we need to explain it before giving an example.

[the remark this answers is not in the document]
bradfordb-vmware: I agree 100% and it's unfortunate that the upstream Postgres [docs](https://www.postgresql.org/docs/9.1/datatype-numeric.html) use "numeric types" to refer to types like `smallint`, `integer`, `numeric`, ...

Perhaps we should clearly define on this page that 
- _numeric types_ refers to all types listed in Table 8-2 from the previous link
- `numeric` refers specfically to the user-specified precision GPDB type

I don't know if that helps with or adds to the confusion.

docs/content/hdfs_avro.html.md.erb
```diff
@@ -57,13 +57,15 @@ PXF uses the following data type mapping when reading Avro data:
...
+| Complex type: Array of other types **and** *Avro schema is provided* | text[]   |
+| Complex type: Map, Record, or Enum  | text, with delimiters inserted between collection items, mapped key-value pairs, and record data.   |
 | Complex type: Fixed     | bytea (supported for read operations only).   |
 | Union      | Follows the above conventions for primitive or complex data types, depending on the union; must contain 2 elements, one of which must be null.  |
 
-### <a id="datatype_map_Write "></a>Write Mapping
+### <a id="datatype_map_Write"></a>Write Mapping
 
-PXF supports writing only Avro primitive data types. It does not support writing complex types to Avro.
+PXF supports writing only Avro primitive types and arrays of Avro primitive types. PXF supports writing other complex types to Avro as string.
```
mperezfuster, line 68: maybe "only" is not required anymore?

mperezfuster (approved):

docs/content/hdfs_avro.html.md.erb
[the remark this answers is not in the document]
lisakowen: i added a link to the avro spec for primitive/complex/logical types.  i separately address the logical types (as a group) in the "data type" intro.  also added a table row for these for read.

re-pushed the review site, may need to refresh.

docs/content/hdfs_avro.html.md.erb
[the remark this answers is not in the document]
lisakowen: for this specific issue, i clarify that we are referring to the greenplum `numeric` type.

docs/content/hdfs_avro.html.md.erb
```diff
@@ -57,13 +59,17 @@ PXF uses the following data type mapping when reading Avro data:
...
 | long | bigint |
 | string | text |
-| Complex type: Array, Map, Record, or Enum  | text, with delimiters inserted between collection items, mapped key-value pairs, and record data.   |
+| Complex type: Array (any dimension) of type: boolean, bytes, double, float, int, long, string | array (any dimension) of type: boolean, bytea, double, real, bigint, text   |
+| Complex type: Array of {union of null with type}: boolean, bytes, double, float, int, long, string | boolean[], bytea[], double[], real[], bigint[], text[]   |
+| Complex type: Array of other types<br>(*Avro schema is provided*) | text[]   |
+| Complex type: Map, Record, or Enum  | text, with delimiters inserted between collection items, mapped key-value pairs, and record data.   |
 | Complex type: Fixed     | bytea (supported for read operations only).   |
 | Union      | Follows the above conventions for primitive or complex data types, depending on the union; must contain 2 elements, one of which must be null.  |
+| Logical type: decimal, date, time, timestamp, duration | text |
```
bradfordb-vmware, line 68: This one is not quite correct. When PXF writes GPDB date/time types to Avro, it saves the data in the Avro file with type `string`. To read this back in to GPDB, you would want to have your column be of type `text` as you have here.

However, Avro natively stores

- the `date` logical type as an `int`
- the `time-millis`, `timestamp-millis`, `local-timestampe-millis` (millisecond precision) as an `int`
- the `time-micros`, `timestamp-micros`, `local-timestamp-micros (microsecond precision) as a `long`
- the `duration` logical type is stored as a `fixed` with size 12

To read these fields, the user will need to declare their GPDB columns as `int`, `bigint`, or `bytea` (respectively).

docs/content/hdfs_avro.html.md.erb
```diff
@@ -36,11 +36,13 @@ Apache Avro is a data serialization framework where the data is serialized in a
...
+To represent Avro primitive data types and Avro arrays of primitive types in Greenplum Database, map data values to Greenplum Database columns of the same type. 
 
-Avro supports complex data types including arrays, maps, records, enumerations, and fixed types. Map top-level fields of these complex data types to the Greenplum Database `TEXT` type. While Greenplum Database does not natively support reading these types, you can create Greenplum Database functions or application code to extract or further process subcomponents of these complex data types.
+Avro supports other complex data types including arrays of non-primitive types, maps, records, enumerations, and fixed types. Map top-level fields of these complex data types to the Greenplum Database `text` type. While PXF does not natively support reading these types, you can create Greenplum Database functions or application code to extract or further process subcomponents of these complex data types.
+
+Avro supports logical data types including include decimal, date, time, and duration types. You must similarly map these data types to the Greenplum Database `text` type.
```
dyozie, line 45: including include -> including

dyozie (changes requested):

docs/content/hdfs_avro.html.md.erb
```diff
@@ -82,8 +88,11 @@ PXF uses the following data type mapping when writing Avro data:
 | smallint<sup>2</sup> | int |
 | text | string |
 | varchar | string |
-| array ([]), enum, record | string |
-| date, time, timestamp, timestamptz | string |
+| numeric, date, time, timestamp, timestamptz<br>(*no Avro schema is provided*) | string |
+| array (any dimension) of type: bigint, boolean, bytea, double, int, real, text <br>(*Avro schema is provided*) | Array (any dimension) of type: long, boolean, bytes, double, int, float, string |
+| bigint[], boolean[], bytea[], double[], int[], real[], text[] <br>(*no Avro schema is provided*) | long[], boolean[], bytes[], double[], int[], float[], string[] |
```
ashuka24, line 93: Do we need to specify here that if no avro schema is provided, it only creates a one-dimensional array?

docs/content/hdfs_avro.html.md.erb
lisakowen: are you saying that if the avro file was generated by another source, pxf could read date/timemilli/timemicro/duration as int/int/bigint/bytea.  would decimal be read as text?

in the case where pxf generated the avro file, and those types were written, the type is string already, right?  so it would be a string->text mapping.

docs/content/hdfs_avro.html.md.erb
lisakowen: i was thinking that `[]` would imply a one-dimensional array.  i can also explicitly state that.

docs/content/hdfs_avro.html.md.erb
bradfordb-vmware: If the Avro file was generated by another source, then PXF **would have to** read `date`/`timemilli`/`timemicro`/`duration` as an `int`/`int`/`long`/`bytea`. If you try to read as `text`, it would fail with the following error:

```console
ERROR:  For field value schema requires type TEXT but input record has type INTEGER  (seg0 slice1 127.0.1.1:6000 pid=519527)
```

If PXF wrote the Avro file, then the GPDB column types for date/timestamp/etc would be written as `string` and it would be a `string` to `text` mapping.

lisakowen: thanks, @bradfordb-vmware .  i created separate rows for the (groups of) logical types.

how do we handle the Avro decimal type on read?  do we need to call that type out?

docs/content/hdfs_avro.html.md.erb
```diff
@@ -57,13 +59,20 @@ PXF uses the following data type mapping when reading Avro data:
 | int | int |
 | long | bigint |
 | string | text |
-| Complex type: Array, Map, Record, or Enum  | text, with delimiters inserted between collection items, mapped key-value pairs, and record data.   |
+| Complex type: Array (any dimension) of type: boolean, bytes, double, float, int, long, string | array (any dimension) of type: boolean, bytea, double, real, bigint, text   |
```
bradfordb-vmware, line 62: I don't think we can put `NOT NULL` on columns in external tables, so I think we can combine this row and the following one.

bradfordb-vmware (approved):

PR #653, opened 2021-06-09
docker: add diagram for container images

ashuka24:


denalex (approved):

denalex:
Might be good to mention somewhere the tool you used to create the image and how to go about editing it.

concourse/docker/diagram/Makefile
```diff
@@ -0,0 +1,2 @@
+images.svg: images.dot
+	dot  -Tsvg images.dot -o images.svg
```
bradfordb-vmware, line 2: Building on @denalex's comment, we might want to include a check if `dot` exists and if not print a message to install the GraphViz package.

concourse/docker/diagram/images.dot
```diff
@@ -0,0 +1,168 @@
+digraph G {
```
bradfordb-vmware, line 1: Do we want to give this a more descriptive filename (e.g., `container-build-flow` or `container-build-diagram`)? We might also want to include a comment at the top to remind folks to re-run `make` after editing this file to update the SVG.

concourse/docker/diagram/images.dot
```diff
@@ -0,0 +1,168 @@
...
+        r0[label="Dockerfile"]
+    }
+
+    # UD GCR images
+    subgraph cluster_gcr_data_gpdb_ud {
+        label = "gcr.io/data-gpdb-ud"
+        style=dashed
+        node [shape=box]
+
+        g1[label="gpdb-pxf-dev/gpdb5-centos7-test-pxf:$COMMIT_SHA"]
```
bradfordb-vmware, line 52: I think all of these images have the same prefix (`gpdb-pxf-dev`) so we could move that to the subgraph label.

It might be a little painful, but can we rename the node names to align with the label: for example, change `g11` to `gpdb7_ubuntu1804_test_pxf_commit_sha`. This should help with reading the edges further down.

ashuka24: The `rpmrebuild` image does not have the same prefix, but maybe i can rearrange the subgraphs so that the names aren't as long

bradfordb-vmware (approved):

PR #654, opened 2021-06-09
remove hdp2 docker image

ashuka24:
The HDP2 docker image is not used in any PXF related pipelines. It is only occasionally used for local development with docker. This method of development is rarely used and the scripts it uses and is being used by is not being maintained.


README.md
```diff
@@ -179,14 +179,11 @@ cp ${PXF_HOME}/templates/*-site.xml ${PXF_BASE}/servers/default
...
-The quick and easy is to download the GPDB 6.6 RPM from Github and move it into the `/downloads` folder. Then run `./dev/start.bash` to get a docker image with a running GPDB6, Hadoop cluster and an installed PXF.
-
-If you would like more control over the GPDB installation, you can use the steps below.
+You can use the steps below to spin up a docker container:
 
 ```bash
-# Get the latest centos7 image, works for GPDB5 and GPDB6
-docker pull pivotaldata/gpdb-pxf-dev:centos7
+# Get the latest centos7 image for GPDB6
+docker pull gcr.io/data-gpdb-ud/gpdb-pxf-dev/gpdb6-centos7-test-pxf:latest
```
denalex, line 186: this is not a public image, I'd assume. I am not sure whether the original one is (was), but the idea here is that OSS devs should (ideally) be able to create these images inter own GCR account, that's why it is defined as a variable.

ashuka24:
After discussions with @denalex and @bradfordb-vmware, we are looking to continue maintaining the dev docker image and creating a chore that would help developers keep the image relatively up to date.

PR #660, opened 2021-06-15
Add make targets for TPC-H performance pipelines

ashuka24:
This commit adds make targets for data-gen-1g and
query-execution-parquet-1g which are used by PXF for running the TPC-H
benchmark.


concourse/Makefile
```diff
@@ -242,6 +242,21 @@ singlecluster:
 		-l ~/workspace/gp-continuous-integration/secrets/gpdb_common-ci-secrets.yml \
 		-v pxf-git-branch=master -p pxf-singlecluster
 
+.PHONY: data-gen-1g
+data-gen-1g:
+	make -C ~/workspace/gp-performance-testing \
```
bradfordb-vmware, line 247: Can we test if gp-performance-testing is checked out in workspace and print an error message to clone it if it's not there?

bradfordb-vmware (approved):

PR #670, opened 2021-07-06
Upgrade Spring Boot to 2.5.2, Gradle to 7

denalex:


bradfordb-vmware (approved):

server/gradle.properties
```diff
@@ -19,14 +19,13 @@ version=0.0.0-SNAPSHOT
 apiVersion=0
 license=ASL 2.0
 hadoopVersion=2.10.1
-htraceVersion=4.0.1-incubating
```
ashuka24, line 22: why is this no longer needed?

server/pxf-api/src/test/java/org/greenplum/pxf/api/security/SecureLoginTest.java
```diff
@@ -53,6 +53,7 @@ public static void resetProperties() {
         resetProperty(PROPERTY_KEY_SERVICE_KEYTAB, kerberosKeytab);
         resetProperty(PROPERTY_KEY_KERBEROS_KDC, kdcDefault);
         resetProperty(PROPERTY_KEY_KERBEROS_REALM, realmDefault);
+        UserGroupInformation.reset(); // clean up so that downstream tests in the same JVM are not affected
```
ashuka24, line 56: why is this needed now and not before?

server/build.gradle
```diff
@@ -67,28 +67,130 @@ configure(javaProjects) {
...
+            dependency("org.mortbay.jetty:jetty-util:6.1.26")
+            dependency("org.objenesis:objenesis:2.1")
+            dependency("org.simplify4u:slf4j-mock:2.1.0")
+            dependency("org.threeten:threeten-extra:1.5.0")
+            dependency("org.tukaani:xz:1.8")
+            dependency("org.wildfly.openssl:wildfly-openssl:1.0.7.Final")
+            dependency("org.xerial.snappy:snappy-java:1.0.5")
+
+            // Hadoop dependencies
+            dependencySet(group:"org.apache.hadoop", version:"${hadoopVersion}") {
```
ashuka24, line 111: by providing the hadoop version here like this and having the dependency sets what allows us to remove the versioning in the other build.gradle files?

denalex: sets are just groupings that avoid re-definition of group and version for every module. What allows us to skip versions everywhere else is the fact that we define these inside the `dependencies` DSL provided by  `apply plugin: 'io.spring.dependency-management'` which makes it similar to `<dependencyManagement>` Maven mechanism. We declare the version just once here and when we need the dependency in a subproject, it is already has its version defined from here.

server/gradle.properties
denalex: it was only used in 1 place, so I hardcoded it there, this is a minor component, no reason to have its version extracted like this

server/pxf-api/src/test/java/org/greenplum/pxf/api/security/SecureLoginTest.java
denalex: I don't know, maybe a change in how newer Gradle runs tests ? The test was dirtying the JVM, this now cleans it up for downstream tests not to fail.

ashuka24 (approved):

PR #671, opened 2021-07-14
docs: add table for unsupported types in Hive JDBC driver

bradfordb-vmware:

### DIFF
--- a/docs/content/jdbc_pxf.html.md.erb
+++ b/docs/content/jdbc_pxf.html.md.erb
@@ -53,14 +53,16 @@ The PXF JDBC connector supports the following data types:
 
 Any data type not listed above is not supported by the PXF JDBC connector.
 
-**Note**: PXF includes Hive JDBC driver version 1.1.0 which does **not** support the following data types when reading from or writing to Hive tables.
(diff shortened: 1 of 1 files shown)


mperezfuster:
LGTM, just a couple of minor changes.

docs/content/jdbc_pxf.html.md.erb
```diff
@@ -53,7 +53,7 @@ The PXF JDBC connector supports the following data types:
 
 Any data type not listed above is not supported by the PXF JDBC connector.
 
-**Note**: PXF includes Hive JDBC driver 1.1.0 which does not support the following data types:
+**Note**: PXF includes Hive JDBC driver version 1.1.0 which does **not** support the following data types when reading from or writing to Hive tables.
```
bradfordb-vmware, line 56: I double checked and with the exception of `binary`/`bytea`, this table is only for _writing_ to Hive. Should we add another column for whether it is read or write (or both)?

lisakowen:
@bradfordb-vmware @mperezfuster - can you check out my recent commit to this PR and let me know what you think?

lisakowen:
merging this PR.

PR #678, opened 2021-08-10
docs: add example for ORC array support

ashuka24:
Updated docs for https://github.com/greenplum-db/pxf/pull/675

This PR updates the PXF documentation to include support for the ORC list compound time and provide an example for how to query such columns with PXF.

### DIFF
--- a/docs/content/hdfs_orc.html.md.erb
+++ b/docs/content/hdfs_orc.html.md.erb
@@ -57,6 +57,8 @@ PXF supports only the list ORC compound type, and only for a subset of the ORC s
 | ORC Compound Type | PXF/Greenplum Data Type |
 |-------------------|-------------------------|
 | array\<string> | Text[] |
+| array\<char> | Bpchar[] |
+| array\<varchar> | Varchar[] |
 | array\<binary> | Bytea[] |
 | array\<float> | Real[] |
 | array\<double> | Float8[] |
--- a/docs/content/hdfs_orc.html.md.erb
+++ b/docs/content/hdfs_orc.html.md.erb
@@ -12,7 +12,7 @@ The `hdfs:orc` profile:
 - Supports the compound list type for a subset of ORC scalar types.
 - Does not support the map, union or struct compound types.
 
-The `hdfs:orc` profile currently supports reading scalar data types from ORC files. If the data resides in a Hive table, and you want to read complex types or the Hive table is partitioned, use the [`hive:orc`](hive_pxf.html#hive_orc) profile.
+The `hdfs:orc` profile currently supports reading scalar data types and lists of certain scalar types from ORC files. If the data resides in a Hive table, and you want to read complex types or the Hive table is partitioned, use the [`hive:orc`](hive_pxf.html#hive_orc) profile.
 
 
 ## <a id="prereq"></a>Prerequisites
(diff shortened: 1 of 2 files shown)


docs/content/hdfs_orc.html.md.erb
```diff
@@ -51,6 +52,19 @@ To read ORC scalar data types in Greenplum Database, map ORC data values to Gree
 | Integer | bigint (64 bit) | Bigint |
 | Integer | date | Date |
 
+Lists are the only supported compound type and only for a subset of the ORC scalar types. The supported compound mapping is as follows:
```
mperezfuster, line 55: I am not sure what this means, maybe due to my ignorance about the topic but confused with the "Lists" term at the start of the sentence. 
I might rephrase to something like: 
"The compound list type are the only supported compound type and only for a subset of ORC scalar types. The supported compound mapping is as follows"

docs/content/hdfs_orc.html.md.erb
```diff
@@ -151,12 +168,27 @@ Beijing,Jul,411,11600.67' > /tmp/sampledata.csv
...
+    (5 rows)
     ```
 
+1. The data can be queried on any column including the array column. For example, this query returns the rows where the items sold include boots and/or pants:
+
+    ``` sql
+    testdb=# SELECT * FROM sample_orc WHERE items_sold && '{"boots", "pants"}';
+    ```
+
+    ``` shell
```
mperezfuster, line 187: I know this was the case on the previous version of the doc, but maybe it would look better to have a single code block instead of two consecutive ones? Since it creates a blank gap in between.

mperezfuster (approved):
few minor changes but otherwise looks good! Thanks so much for creating this

lisakowen:
@ashuka24 - i will address @mperezfuster 's comments.

docs/content/hdfs_orc.html.md.erb
lisakowen: i originally split the SQL from the command output because shell formatting of the output looked better than when it was sql-formatted.  i'll check this out again.

docs/content/hdfs_orc.html.md.erb
```diff
@@ -151,12 +168,27 @@ Beijing,Jul,411,11600.67' > /tmp/sampledata.csv
...
+     location  | month | num_orders | total_sales |                  items_sold
+    -----------+-------+------------+-------------+----------------------------------------------
+     Prague    | Jan   |        101 |     4875.33 | {boots,hats}
+     Bangalore | May   |        317 |     8936.99 | {"winter socks","long-sleeved shirts",boots}
+     Beijing   | Jul   |        411 |    11600.67 | {hoodies/sweaters,pants}
+    (3 rows)
+    ```
```
denalex, line 194: should we also provide an example where we pick a single element from the array, like `SELECT * FROM sample_orc WHERE items_sold[0] = 'boots'` ?

denalex (approved):

docs/content/hdfs_orc.html.md.erb
lisakowen: added this example.

docs/content/hdfs_orc.html.md.erb
lisakowen: i reworded that sentence a bit.

PR #680, opened 2021-08-16
fix CURLOPT_RESOLVE optimization

bradfordb-vmware:
You cannot `#ifdef` on `CURLOPT_RESOLVE` since it is an enum member; the
`#ifdef` is always false and the block is not compiled.

Using `CURLOPT_RESOLVE` to provide a custom host name to IP address
resolution for the case of `localhost` allows PXF to pre-populate the
DNS cache and avoid a lookup via `getaddrinfo`.

We have seen some RHEL7 environments where the call to `getaddrinfo`
fails and calls `abort` which causes the GPDB segments to panic.

### DIFF
--- a/external-table/src/libchurl.c
+++ b/external-table/src/libchurl.c
@@ -344,21 +344,19 @@ churl_init(const char *url, CHURL_HEADERS headers)
 	clear_error_buffer(context);
 
 /* Required for resolving localhost on some docker environments that
- * had intermittent networking issues when using pxf on HAWQ
- * However, CURLOPT_RESOLVE is only available in curl versions 7.21 and above */
-#ifdef CURLOPT_RESOLVE
+ * had intermittent networking issues when using pxf on HAWQ.
+ */
 	if (strstr(url, LocalhostIpV4) != NULL)
 	{
 		struct curl_slist *resolve_hosts = NULL;
-		char	   *pxf_host_entry = (char *) palloc0(strlen(PxfServiceAddress) + strlen(LocalhostIpV4Entry) + 1);
+		char	   *pxf_host_entry = (char *) palloc0(LOCAL_HOST_RESOLVE_STRING_MAX_LENGTH);
 
-		strcat(pxf_host_entry, PxfServiceAddress);
-		strcat(pxf_host_entry, LocalhostIpV4Entry);
+		snprintf(pxf_host_entry, LOCAL_HOST_RESOLVE_STRING_MAX_LENGTH, LOCAL_HOST_RESOLVE_STRING_FORMAT, get_pxf_port());
+		elog(DEBUG3, "adding CURLOPT_RESOLVE with entry '%s'", pxf_host_entry);
 		resolve_hosts = curl_slist_append(NULL, pxf_host_entry);
 		set_curl_option(context, CURLOPT_RESOLVE, resolve_hosts);
 		pfree(pxf_host_entry);
 	}
-#endif
 
 	set_curl_option(context, CURLOPT_URL, url);
 	set_curl_option(context, CURLOPT_VERBOSE, (const void *) FALSE);
--- a/external-table/src/libchurl.h
+++ b/external-table/src/libchurl.h
@@ -143,6 +143,11 @@ void		churl_cleanup(CHURL_HANDLE handle, bool after_error);
  */
 void		print_http_headers(CHURL_HEADERS headers);
 
+#define LOCAL_HOST_RESOLVE_STRING_FORMAT "localhost:%d:127.0.0.1"
+/* PORT can be at most five digits giving a total length for the resolve string
+ * of 25 plus one for the trailing '\0'
+ */
+#define LOCAL_HOST_RESOLVE_STRING_MAX_LENGTH 26
 #define LocalhostIpV4Entry ":127.0.0.1"
 #define LocalhostIpV4 "localhost"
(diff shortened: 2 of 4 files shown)


bradfordb-vmware:
This PR is targeting master/PXF 6.x; we will also want to cherry-pick/backport to branch-5.16.

external-table/src/libchurl.c
```diff
@@ -346,14 +346,15 @@ churl_init(const char *url, CHURL_HEADERS headers)
 /* Required for resolving localhost on some docker environments that
  * had intermittent networking issues when using pxf on HAWQ
  * However, CURLOPT_RESOLVE is only available in curl versions 7.21 and above */
-#ifdef CURLOPT_RESOLVE
+#if LIBCURL_VERSION_NUM >= 0x071500
```
denalex, line 349: if we base off the version, then it seems it has been added in 7.21.3 : https://github.com/curl/curl/commit/1b24b89cca3c06e36b69969af8edeeaca659515b

that will make: 
```suggestion
#if LIBCURL_VERSION_NUM >= 0x071503
```

external-table/src/libchurl.c
```diff
@@ -346,14 +346,15 @@ churl_init(const char *url, CHURL_HEADERS headers)
...
 	if (strstr(url, LocalhostIpV4) != NULL)
 	{
 		struct curl_slist *resolve_hosts = NULL;
-		char	   *pxf_host_entry = (char *) palloc0(strlen(PxfServiceAddress) + strlen(LocalhostIpV4Entry) + 1);
+		int	   host_entry_size = snprintf(NULL, 0, "%s:%d%s", LocalhostIpV4, get_pxf_port(), LocalhostIpV4Entry) + 1;
```
denalex, line 353: We could've continued to estimate this as port is max at 5 characters so:
`strlen(LocalhostIpV4Entry)+1+5+strlen(LocalhostIpV4Entry)+1` or really just `9+1+5+10+1=26`
but this is ok as well.

denalex (approved):

external-table/src/libchurl.c
denalex: On the other hand, should we just not compare and have compiler fail if the build version is less than 7.21.3 ? Was it only the case for centos6 or even centos5 at some point ? This current patch is safe and ok for PXF5, but maybe in PXF 6 we should do just that ?

ashuka24 (approved):

hpandeycodeit (approved):

external-table/src/libchurl.c
bradfordb-vmware: Great catch on the version -- I assumed the comment mean 7.21.0 but I should have checked. In GPDB 5, we [vendor cURL 7.54.0](https://github.com/greenplum-db/gpdb/blob/f9cdb519790ce36c32dc8269771880e91cddfdbe/gpAux/releng/make/dependencies/ivy.xml#L27) so we should be able to safely remove this conditional altogether.

fdw/libchurl.c
```diff
@@ -342,21 +342,20 @@ churl_init(const char *url, CHURL_HEADERS headers)
...
+ */
 	if (strstr(url, LocalhostIpV4) != NULL)
 	{
 		struct curl_slist *resolve_hosts = NULL;
-		char	   *pxf_host_entry = (char *) palloc0(strlen(PxfServiceAddress) + strlen(LocalhostIpV4Entry) + 1);
+		char	   *pxf_host_entry = (char *) palloc0(LOCAL_HOST_RESOLVE_STRING_MAX_LENGTH);
 
-		strcat(pxf_host_entry, PxfServiceAddress);
-		strcat(pxf_host_entry, LocalhostIpV4Entry);
+		/* TODO: do not hardcode the port here */
```
ashuka24, line 352: do we need to handle this todo?

external-table/src/libchurl.h
```diff
@@ -143,6 +143,11 @@ void		churl_cleanup(CHURL_HANDLE handle, bool after_error);
  */
 void		print_http_headers(CHURL_HEADERS headers);
 
+#define LOCAL_HOST_RESOLVE_STRING_FORMAT "localhost:%d:127.0.0.1"
```
ashuka24, line 146: do we need to worry about IPv6?

fdw/libchurl.c
bradfordb-vmware: Good question. I think the answer is no.

For the external table version of the PXF extension, we set the PXF port via an environment variable and there is a helper function `get_pxf_port` that we can call to get the port. For the FDW version of the PXF extension, it looks like the port is set as an option on the foreign table (see `fdw/pxf_option.c`) and is stored in a struct `PxfOptions` which isn't being passed in to `churl_init`. This plumbing workin needs to be done at some point but that's a larger refactor. I only touched the fdw code to minimize drift between the two.

external-table/src/libchurl.h
bradfordb-vmware: Looking at the documentation for [`CURLOPT_RESOLVE`](https://curl.se/libcurl/c/CURLOPT_RESOLVE.html), I think we can include both IPv4 and IPv6 by changing this to `"localhost:%d:127.0.0.1,::1"`.

bradfordb-vmware: Closing the loop: we have other IPv6 blockers and we don't have IPv6 testing setup yet. We'll defer on that for now.

PR #681, opened 2021-08-17
Introduced operation retries when GSS connection failures are encountered

denalex:
When Kerberos security is enabled for a Hadoop cluster and especially when PXF is configured to use principals without distinct  hostnames, the Kerberos library in HDFS Namenode can fail to accept a SASL/GSS/Kerberos connection from PXF if it determines that the Kerberos Authenticator it receives from PXF has already been seen by the Namenode within a short window of time. The Namenode deems it a Kerberos replay attack.

The solution implemented here is to retry the operation if an `IOException` with the `GSS initiate failed` message is observed. The number of retries is set to 5 by default, but is configurable by `pxf.sasl.connection.retries` property introduced into the `pxf-site.xml` template.

### DIFF
--- a/server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/FragmenterResource.java
+++ b/server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/FragmenterResource.java
@@ -29,6 +29,7 @@ import org.greenplum.pxf.api.utilities.FragmenterCacheFactory;
 import org.greenplum.pxf.api.utilities.FragmenterFactory;
 import org.greenplum.pxf.api.utilities.FragmentsResponse;
 import org.greenplum.pxf.api.utilities.FragmentsResponseFormatter;
+import org.greenplum.pxf.service.utilities.GSSFailureHandler;
 import org.greenplum.pxf.api.utilities.Utilities;
 import org.greenplum.pxf.service.HttpRequestParser;
 import org.greenplum.pxf.service.RequestParser;
@@ -61,6 +62,7 @@ import static org.greenplum.pxf.api.model.RequestContext.RequestType;
 @Path("/" + Version.PXF_PROTOCOL_VERSION + "/Fragmenter/")
 public class FragmenterResource extends BaseResource {
 
+    private final GSSFailureHandler failureHandler;
     private FragmenterFactory fragmenterFactory;
 
     private FragmenterCacheFactory fragmenterCacheFactory;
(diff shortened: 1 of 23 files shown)


server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/FragmenterResource.java
```diff
@@ -183,7 +187,15 @@ public Response getFragmentsStats(@Context final ServletContext servletContext,
 
     private List<Fragment> getFragments(RequestContext context) throws Exception {
         /* Create a fragmenter instance with API level parameters */
-        List<Fragment> fragments = AnalyzeUtils.getSampleFragments(fragmenterFactory.getPlugin(context).getFragments(), context);
+        Fragmenter fragmenter = fragmenterFactory.getPlugin(context);
+        // We can't support lambdas here because asm version doesn't support it
```
ashuka24, line 191: What is `asm`? And should we write the version number here? that way if it does end up getting supported by a later version we have some better context?

server/pxf-service/src/test/java/org/greenplum/pxf/service/bridge/ReadSamplingBridgeTest.java
```diff
@@ -58,7 +58,8 @@
     /**
      * Writable test object to test ReadSamplingBridge. The object receives a
      * string and returns it in its toString function.
-     */
+
```
ashuka24, line 61: why are we no longer running this test suite?

server/pxf-service/src/templates/user/templates/pxf-site.xml
```diff
@@ -54,4 +54,10 @@
         <description>Specifies whether Predicate Pushdown feature is enabled for Hive profiles.</description>
     </property>
 
+    <property>
+        <name>pxf.sasl.connection.retries</name>
```
lisakowen, line 58: we are adding a new property.  do we want to add this property/description to the main docs, or just mention setting the property in the context of the resolved issue?

server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/FragmenterResource.java
denalex: asm is a library we use, the version is in build.gradle, we use a newer version in PXF6

server/pxf-service/src/templates/user/templates/pxf-site.xml
denalex: yes, we should add it to the main docs as well

server/pxf-service/src/test/java/org/greenplum/pxf/service/bridge/ReadSamplingBridgeTest.java
denalex: the test was `@Ignore`d before this change since we do not support analytics endpoint with GPDB (we did with HAWQ). Now the code would've needed to change to accommodate for handler, so I just commented it all out, did not want to remove the class.

bradfordb-vmware (approved):

ashuka24 (approved):

PR #682, opened 2021-08-17
Fix curl resolve

bradfordb-vmware:
This is a backport of PR #680 to the PXF 5.16 support branch.

### DIFF
--- a/external-table/src/libchurl.h
+++ b/external-table/src/libchurl.h
@@ -143,6 +143,11 @@ void		churl_cleanup(CHURL_HANDLE handle, bool after_error);
  */
 void		print_http_headers(CHURL_HEADERS headers);
 
+#define LOCAL_HOST_RESOLVE_STRING_FORMAT "localhost:%d:127.0.0.1"
+/* PORT can be at most five digits giving a total length for the resolve string
+ * of 25 plus one for the trailing '\0'
+ */
+#define LOCAL_HOST_RESOLVE_STRING_MAX_LENGTH 26
 #define LocalhostIpV4Entry ":127.0.0.1"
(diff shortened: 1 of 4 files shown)


fdw/libchurl.c
```diff
@@ -319,21 +319,20 @@ churl_init(const char *url, CHURL_HEADERS headers)
...
+ */
 	if (strstr(url, LocalhostIpV4) != NULL)
 	{
 		struct curl_slist *resolve_hosts = NULL;
-		char	   *pxf_host_entry = (char *) palloc0(strlen(PxfServiceAddress) + strlen(LocalhostIpV4Entry) + 1);
+		char	   *pxf_host_entry = (char *) palloc0(LOCAL_HOST_RESOLVE_STRING_MAX_LENGTH);
 
-		strcat(pxf_host_entry, PxfServiceAddress);
-		strcat(pxf_host_entry, LocalhostIpV4Entry);
+		/* TODO: do not hardcode the port here */
```
ashuka24, line 329: do we need to handle this todo?

external-table/src/libchurl.h
```diff
@@ -143,6 +143,11 @@ void		churl_cleanup(CHURL_HANDLE handle, bool after_error);
  */
 void		print_http_headers(CHURL_HEADERS headers);
 
+#define LOCAL_HOST_RESOLVE_STRING_FORMAT "localhost:%d:127.0.0.1"
```
ashuka24, line 146: this format is only valid for IPv4. Do we need to worry about IPv6?

fdw/libchurl.c
ashuka24, line 329: do we need to handle this todo?

PR #684, opened 2021-08-18
docs - for new pxf.sasl.connection.retries property

lisakowen:
the pxf.sasl.connection.retries  property is introduced to allow the user to specify the number of times PXF retries a connection request after a connection is rejected and returns a `GSS initiate failed` error.

in this PR:
- describe the new property
- add a step to the 5.x upgrade guide

### DIFF
--- a/docs/content/cfg_server.html.md.erb
+++ b/docs/content/cfg_server.html.md.erb
@@ -98,6 +98,7 @@ You configure properties in the `pxf-site.xml` file for a PXF server when one or
 - You want to enable/disable user impersonation on the remote Hadoop or external database system.
 - You will access a network file system with the server configuration.
 - You will access a remote Hadoop or object store file system with the server configuration, and you want to allow a user to access only a specific directory and subdirectories.
+- You will use a SASL connection to access an external data store with the server configuration.
 
 `pxf-site.xml` includes the following properties:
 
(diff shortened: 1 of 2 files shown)


docs/content/cfg_server.html.md.erb
```diff
@@ -98,6 +98,7 @@ You configure properties in the `pxf-site.xml` file for a PXF server when one or
 - You want to enable/disable user impersonation on the remote Hadoop or external database system.
 - You will access a network file system with the server configuration.
 - You will access a remote Hadoop or object store file system with the server configuration, and you want to allow a user to access only a specific directory and subdirectories.
+- You will use a SASL connection to access an external data store with the server configuration.
```
denalex, line 101: I don't think we need to mention this here, the property should not be touched by customer unless needed (setting to 0 to disable retry logic or increasing to more than 5 if they still face the issue)

docs/content/upgrade_pxf_rpm.html.md.erb
```diff
@@ -107,6 +107,20 @@ After you install the new version of PXF, perform the following procedure:
...
+    <property>
+      <name>pxf.sasl.connection.retries</name>
+      <value>5</value>
+      <description>
+        Specifies the number of retries to perform when a SASL connection is refused by a Namenode due to 'GSS initiate failed' error.
+      </description>
+    </property>
+    ```
+
+    You may choose to update the default value.
```
denalex, line 122: again, this is not strictly necessary as the feature is ON by default and the default value in code is 5.

docs/content/cfg_server.html.md.erb
```diff
@@ -109,6 +110,7 @@ You configure properties in the `pxf-site.xml` file for a PXF server when one or
...
 | pxf.fs.basePath | Identifies the base path or share point on the remote file system. This property is applicable when the server configuration is used with a profile that accesses a file. | None; this property is commented out by default. |
 | pxf.ppd.hive<sup>1</sup> | Specifies whether or not predicate pushdown is enabled for queries on external tables that specify the `Hive`, `HiveRC`, and `HiveORC` profiles. | True; predicate pushdown is enabled. |
+| pxf.sasl.connection.retries | Specifies the maximum number of times that PXF retries a SASL connection request after a refused connection returns a `GSS initiate failed` error. | 5 |
```
denalex, line 113: this is good, we just document that the property exists and what it does.

denalex (approved):

docs/content/upgrade_pxf_rpm.html.md.erb
lisakowen: @denalex - thank you for reviewing.

i rephrased that upgrade step to also state "and if you want to update the default value".  or did you just want me to completely remove that step?

bradfordb-vmware (approved):

docs/content/upgrade_pxf_rpm.html.md.erb
```diff
@@ -107,6 +107,18 @@ After you install the new version of PXF, perform the following procedure:
 
             You may need to add this property block to the `pxf-site.xml` file.
 
+1. **If you are upgrading to PXF version 5.16.3 to resolve the erroneous replay attack issue in a Kerberos-secured environment and you want to change the default value of the new `pxf.sasl.connection.retries` property**, add the following to the `$PXF_CONF/servers/<server-name>/pxf-site.xml` file for your PXF Hadoop server:
```
mperezfuster, line 110: The way this is phrased it made me think that we need to update something on the Hadoop side. Maybe we can say something like: 
"add the following to the `$PXF_CONF/servers/<server-name>/pxf-site.xml` file on your PXF servers". 
Or whatever is more correct, I forget where exactly this needs to be configured for 5.x.

mperezfuster (approved):
Thanks for adding this! Just had a small comment.

ashuka24 (approved):

PR #689, opened 2021-08-24
external-table: add simple cURL debug callback function

bradfordb-vmware:
Enable CURLOPT_VERBOSE if the client or log level is the DEBUG3 level
(or higher); when verbose logging is enabled, libcurl will call the
debug callback given by CURLOPT_DEBUGFUNCTION.

external-table: conditionally enable curl debug output

### DIFF
--- a/external-table/src/libchurl.c
+++ b/external-table/src/libchurl.c
@@ -131,30 +131,6 @@ static char	   *build_header_str(const char *format, const char *key, const char
 static bool	IsValidJson(text *json);
 
 
-/*
- * Debug function - print the http headers
- */
-void
-print_http_headers(CHURL_HEADERS headers)
-{
-	if ((DEBUG2 >= log_min_messages) || (DEBUG2 >= client_min_messages))
-	{
-		churl_settings *settings = (churl_settings *) headers;
-		struct curl_slist *header_cell = settings->headers;
-		char	   *header_data;
-		int			count = 0;
-
-		while (header_cell != NULL)
-		{
-			header_data = header_cell->data;
-			elog(DEBUG2, "churl http header: cell #%d: %s",
-				 count, header_data ? header_data : "NONE");
-			header_cell = header_cell->next;
-			++count;
-		}
-	}
-}
-
 CHURL_HEADERS
 churl_headers_init(void)
 {
@@ -335,6 +311,44 @@ churl_headers_cleanup(CHURL_HEADERS headers)
 	pfree(settings);
 }
 
+/*
+ * debug callback for CURLOPT_DEBUGFUNCTION
+ * logs debug information via postgres' logging mechanism
+ */
+static int
+log_curl_debug(CURL *handle, curl_infotype type, char *data, size_t size, void *userp)
+{
+	static const char prefix_map[CURLINFO_END][13] = {
+		"info: ",
+		"header in: ",
+		"header out: ",
+		"data in: ",
+		"data out: ",
+		"data in: ",
+		"data out: "
+	};
+
+	switch(type) {
+	case CURLINFO_TEXT:
+	case CURLINFO_HEADER_IN:
+	case CURLINFO_HEADER_OUT:
+		ereport(DEBUG3,
+			(errmsg("curl debug - %s%s", prefix_map[type], data)));
+		break;
+	case CURLINFO_DATA_IN:
+	case CURLINFO_DATA_OUT:
+	case CURLINFO_SSL_DATA_IN:
+	case CURLINFO_SSL_DATA_OUT:
+		ereport(DEBUG3,
+			(errmsg("curl debug - %s%zu bytes", prefix_map[type], size)));
+		break;
+	default:
+		break;
+	}
+
+	return 0;
+}
+
 static CHURL_HANDLE
 churl_init(const char *url, CHURL_HEADERS headers)
 {
@@ -358,8 +372,11 @@ churl_init(const char *url, CHURL_HEADERS headers)
 		pfree(pxf_host_entry);
 	}
 
+	long curl_verbose = (DEBUG3 >= log_min_messages) || (DEBUG3 >= client_min_messages);
+
 	set_curl_option(context, CURLOPT_URL, url);
-	set_curl_option(context, CURLOPT_VERBOSE, (const void *) FALSE);
+	set_curl_option(context, CURLOPT_DEBUGFUNCTION, log_curl_debug);
+	set_curl_option(context, CURLOPT_VERBOSE, (const void *) curl_verbose);
 	set_curl_option(context, CURLOPT_ERRORBUFFER, context->curl_error_buffer);
 	set_curl_option(context, CURLOPT_IPRESOLVE, (const void *) CURL_IPRESOLVE_V4);
 	set_curl_option(context, CURLOPT_WRITEFUNCTION, write_callback);
@@ -385,7 +402,6 @@ churl_init_upload(const char *url, CHURL_HEADERS headers)
 	churl_headers_append(headers, "Transfer-Encoding", "chunked");
 	churl_headers_append(headers, "Expect", "100-continue");
 
-	print_http_headers(headers);
 	setup_multi_handle(context);
 	return (CHURL_HANDLE) context;
 }
@@ -397,7 +413,6 @@ churl_init_download(const char *url, CHURL_HEADERS headers)
 
 	context->upload = false;
 
-	print_http_headers(headers);
 	setup_multi_handle(context);
 	return (CHURL_HANDLE) context;
 }
@@ -892,6 +907,11 @@ check_response_status(churl_context *context)
 
 			initStringInfo(&err);
 
+			/* if the request did not complete correctly, show the error
+			 * information. If no detailed error information was written to errbuf
+			 * show the more generic information from curl_easy_strerror instead.
+			 */
+
 			appendStringInfo(&err, "transfer error (%ld): %s",
 							 status, curl_easy_strerror(status));
 
@@ -900,9 +920,22 @@ check_response_status(churl_context *context)
 				appendStringInfo(&err, " from %s", addr);
 				pfree(addr);
 			}
-			elog(ERROR, "%s", err.data);
+
+			size_t len = strlen(context->curl_error_buffer);
+			if (len > 0)
+			{
+				ereport(ERROR,
+					(errmsg("%s", err.data),
+					errdetail("curl error buffer: %s", context->curl_error_buffer)));
+			}
+			else
+			{
+				ereport(ERROR,
+					(errmsg("%s", err.data)));
+			}
 		}
-		elog(DEBUG2, "check_response_status: msg %d done with status OK", i++);
+		ereport(DEBUG2,
+			(errmsg("check_response_status: msg %d done with status OK", i++)));
 	}
 }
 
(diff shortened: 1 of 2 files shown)


external-table/src/libchurl.c
```diff
@@ -359,7 +385,8 @@ churl_init(const char *url, CHURL_HEADERS headers)
 	}
 
 	set_curl_option(context, CURLOPT_URL, url);
-	set_curl_option(context, CURLOPT_VERBOSE, (const void *) FALSE);
+	set_curl_option(context, CURLOPT_DEBUGFUNCTION, showit);
+	set_curl_option(context, CURLOPT_VERBOSE, (const void *) ((DEBUG3 >= log_min_messages) || (DEBUG3 >= client_min_messages) ? 1L : 0L));
```
ashuka24, line 389: just so i'm understanding this correctly, 
now when we say debug3, we will print the verbose curl errors/logs. but if we do any other level, then we will not print anything?

along those same lines, if we err out, will we also print the debug err logs regardless of the debug level?

external-table/src/libchurl.c
```diff
@@ -15,6 +15,8 @@
  * KIND, either express or implied.  See the License for the
  * specific language governing permissions and limitations
  * under the License.
+ *
+ * Portions Copyright (C) 1998 - 2012, Daniel Stenberg, <daniel@haxx.se>, et al.
```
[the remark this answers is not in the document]
bradfordb-vmware: The `showit()` function that this PR adds is copied from the default logging function in the curl code base.

external-table/src/libchurl.c
bradfordb-vmware: This will cause libcurl to call our debug function (`showit`) if either the log or client error level is `DEBUG3`, `DEBUG4`, or `DEBUG5`. If we error out, our debug function will not be called unless the logging level is set to `DEBUG3` or lower.

external-table/src/libchurl.c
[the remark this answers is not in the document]
denalex: This might have OSL implications, will chat offline

external-table/src/libchurl.c
```diff
@@ -335,6 +337,30 @@ churl_headers_cleanup(CHURL_HEADERS headers)
 	pfree(settings);
 }
 
+/*
+ * based on the showit function in lib/sendf.c from curl 7.29
+ * instead of writing to stderr, we use postgres' elog facility
+ */
+static int
+showit(CURL *handle, curl_infotype type, char *data, size_t size, void *userp)
```
denalex, line 345: didn't you mention that curl debug statements show up in pglog anyways, since the stdout (or stderr) is redirected there ? If yes, then why do we need this ?

external-table/src/libchurl.c
```diff
@@ -335,6 +337,30 @@ churl_headers_cleanup(CHURL_HEADERS headers)
...
+{
+	static const char s_infotype[CURLINFO_END][3] = {
+		"* ", "< ", "> ", "{ ", "} ", "{ ", "} "
+	};
+
+	switch(type) {
+	case CURLINFO_TEXT:
+	case CURLINFO_HEADER_OUT:
+	case CURLINFO_HEADER_IN:
+		elog(DEBUG3, "curl debug: %s%s",s_infotype[type], data);
```
denalex, line 355: so will this log headers and some info ? What about errors ? I think we also log headers that we send, I'm concerned that we'll have too much info logged at DEBUG3 level

external-table/src/libchurl.c
denalex: if C does not have boolean, why bother with ternary operator and just not use the result of comparison ?
`(const void *) ((DEBUG3 >= log_min_messages) || (DEBUG3 >= client_min_messages))`

external-table/src/libchurl.c
```diff
@@ -892,8 +919,13 @@ check_response_status(churl_context *context)
 
 			initStringInfo(&err);
 
+			/* if the request did not complete correctly, show the error
+			 * information. if no detailed error information was written to errbuf
+			 * show the more generic information from curl_easy_strerror instead.
```
denalex, line 924: curious why / how these different buffers will have different information

external-table/src/libchurl.c
bradfordb-vmware: If you enable CURLOPT_VERBOSE without providing a callback function (CURLOPT_DEBUGFUNCTION), then curl will print to stderr; however, postmaster captures stderr and writes it to the postgres log at the LOG level with the prefix `3rd party error log:`. I don't think we want curl debug info being logged at the LOG level and it seems counter-intuitive to turn it on by setting client_min_messages or log_min_messages to something like DEBUG3 but the messages are at the LOG level.

Providing a custom debug function allows us to pick which log level we want this information at and customize the prefix.

external-table/src/libchurl.c
bradfordb-vmware: Correct, this will log the headers and other informational text. The PXF extension already includes a method (`print_http_headers`) for logging the headers, which we could potentially remove. I've omitted logging for the other info types (`CURLINFO_DATA_IN`, `CURLINFO_DATA_OUT`) because I don't think we want to dump the data to the log file. We could just log the number of bytes, which might help with finding a byte-offset in a file if we suspect bad data for causing connection issues.

external-table/src/libchurl.c
bradfordb-vmware: My understanding is that `curl_easy_strerror` provides a very generic error message for a curl error code (e.g,. 56 => "Failure when receiving data from the peer") while `context->curl_error_buffer` may contain information specific to the failed operation.

external-table/src/libchurl.c
bradfordb-vmware: The result of the comparison is an integer which is smaller than a pointer; when we omit the ternary operator, I get the following warning:

```
warning: cast to pointer from integer of different size [-Wint-to-pointer-cast]
```

hpandeycodeit (approved):

external-table/src/libchurl.c
bradfordb-vmware: Here are some examples of what the  curl debug messages will look like in the log file:

```
curl debug - info: Added localhost:5888:127.0.0.1 to DNS cache
curl debug - header out: GET /pxf/read?trace=true HTTP/1.1\r
Host: localhost:5888\r
Accept: */*\r
X-GP-FORMAT: TEXT\r
X-GP-OPTIONS-FORMAT: csv\r
curl debug - header in: HTTP/1.1 200 \r
curl debug - data in: 8071 bytes
```

external-table/src/libchurl.c
bradfordb-vmware: I've updated the error logging to keep the original format as the message and to provide the curl error buffer as detail (if the buffer is non-empty):

Here is what the error looks like in `psql` if the error buffer is empty:

```console
ERROR:  transfer error (56): Failure when receiving data from the peer from '127.0.0.1:5888' (libchurl.c:936)  (seg0 slice1 127.0.1.1:6000 pid=660962) (libchurl.c:936)
CONTEXT:  External table netflow_full, line 6861885 of file pxf:///netflow.csv?PROFILE=file:csv&SERVER=alpine
```

and here is what the error looks like if the error buffer is non-empty:

```console
ERROR:  transfer error (56): Failure when receiving data from the peer from '127.0.0.1:5888' (libchurl.c:927)  (seg0 slice1 127.0.1.1:6000 pid=660730) (libchurl.c:927)
DETAIL:  curl error buffer: Recv failure: Software caused connection abort
CONTEXT:  External table netflow_full, line 6805146 of file pxf:///netflow.csv?PROFILE=file:csv&SERVER=alpine
```

To trigger this error, I used [`ss`](https://man7.org/linux/man-pages/man8/ss.8.html) to forcibly close sockets while GPDB was reading from PXF (`sudo ss -K dst 127.0.0.1 dport = 5888`).

external-table/src/libchurl.c
```diff
@@ -335,6 +311,44 @@ churl_headers_cleanup(CHURL_HEADERS headers)
...
+		"data in: ",
+		"data out: "
+	};
+
+	switch(type) {
+	case CURLINFO_TEXT:
+	case CURLINFO_HEADER_IN:
+	case CURLINFO_HEADER_OUT:
+		ereport(DEBUG3,
+			(errmsg("curl debug - %s%s",prefix_map[type], data)));
```
denalex, line 336:
```suggestion
			(errmsg("curl debug - %s%s", prefix_map[type], data)));
```

external-table/src/libchurl.c
```diff
@@ -335,6 +311,44 @@ churl_headers_cleanup(CHURL_HEADERS headers)
...
+		break;
+	case CURLINFO_DATA_IN:
+	case CURLINFO_DATA_OUT:
+	case CURLINFO_SSL_DATA_IN:
+	case CURLINFO_SSL_DATA_OUT:
+		ereport(DEBUG3,
+			(errmsg("curl debug - %s%zu bytes", prefix_map[type], size)));
+		break;
+	default: /* nada */
```
denalex, line 345:
```suggestion
	default:
```

external-table/src/libchurl.c
denalex:
```suggestion
log_curl_debug(CURL *handle, curl_infotype type, char *data, size_t size, void *userp)
```

external-table/src/libchurl.c
```diff
@@ -358,8 +372,11 @@ churl_init(const char *url, CHURL_HEADERS headers)
 		pfree(pxf_host_entry);
 	}
 
+	long curl_verbose = (DEBUG3 >= log_min_messages) || (DEBUG3 >= client_min_messages);
+
 	set_curl_option(context, CURLOPT_URL, url);
-	set_curl_option(context, CURLOPT_VERBOSE, (const void *) FALSE);
+	set_curl_option(context, CURLOPT_DEBUGFUNCTION, showit);
+	set_curl_option(context, CURLOPT_VERBOSE, (const void *) curl_verbose);
```
denalex, line 379: do you still need the cast to `(const void *)` since the variable is now `long` already ?

external-table/src/libchurl.c
denalex: this looks great !

external-table/src/libchurl.c
```diff
@@ -900,9 +920,22 @@ check_response_status(churl_context *context)
...
+					errdetail("curl error buffer: %s", context->curl_error_buffer)));
+			}
+			else
+			{
+				ereport(ERROR,
+					(errmsg("%s", err.data)));
+			}
 		}
-		elog(DEBUG2, "check_response_status: msg %d done with status OK", i++);
+		ereport(DEBUG2,
```
denalex, line 937: what's the difference between `elog` and `ereport` (other than ability to add details section ?) Is that we get context logged as well ? if yes, do we really want context for this line ?

bradfordb-vmware: My understanding is that `elog` is the older API and is defined in terms of `ereport` ([source](https://github.com/greenplum-db/gpdb/blob/ff356e4319324dd777c31efa4e9231005440f962/src/include/utils/elog.h#L248-L254)). Context is printed/included even if we use `elog`.

external-table/src/libchurl.c
bradfordb-vmware: I think so; `set_curl_option` is a PXF helper method that expects the third argument to be a pointer.

ashuka24 (approved):

denalex (approved):

PR #693, opened 2021-08-31
Fixed GSS failure handler retry logic

denalex:


hpandeycodeit (approved):

server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/FragmenterResource.java
```diff
@@ -186,15 +186,11 @@ public Response getFragmentsStats(@Context final ServletContext servletContext,
...
+        // create a fragmenter holder that will get a fragmenter from the factory based on the request context
+        FragmenterHolder holder = new FragmenterHolder(context);
         // We can't support lambdas here because asm version doesn't support it
-        List<Fragment> fragments = failureHandler.execute(fragmenter.getConfiguration(), "get fragments", new Callable<List<Fragment>>() {
-            @Override
-            public List<Fragment> call() throws Exception {
-                return fragmenter.getFragments();
-            }
-        });
+        // use the holder that has both an operation to run  to fetch fragments and a callback to reset the fragmenter before retries
```
ashuka24, line 192: small nitpick
```suggestion
        // use the holder that has both an operation to run to fetch fragments and a callback to reset the fragmenter before retries
```

ashuka24: just to confirm we need to do this so that it doesn't pick up the old fragmenter when doing the retry logic, correct?

server/pxf-service/src/main/java/org/greenplum/pxf/service/rest/FragmenterResource.java
```diff
@@ -237,4 +233,49 @@ private void logFragmentStatistics(Level level, RequestContext context, List<Fra
...
+            this.fragmenter = fragmenterFactory.getPlugin(context);
+        }
+
+        /* The operation that can be retried by the failure handler */
+        @Override
+        public List<Fragment> call() throws Exception {
+            return fragmenter.getFragments();
+        }
+
+        /* The callback that is performaedcan be retried by the failure handler */
```
ashuka24, line 275:
```suggestion
        /* The callback that is performed can be retried by the failure handler */
```

server/pxf-service/src/test/java/org/greenplum/pxf/service/bridge/ReadBridgeTest.java
```diff
@@ -0,0 +1,107 @@
...
+        when(mockAccessor1.getConfiguration()).thenReturn(configuration);
+        when(mockAccessor1.openForRead()).thenThrow(new IOException("GSS initiate failed"));
+        when(mockAccessor2.openForRead()).thenReturn(true);
+
+        bridge = new ReadBridge(context, mockAccessorFactory, mockResolverFactory, handler);
+        boolean result = bridge.beginIteration();
+
+        assertTrue(result);
+        verify(mockAccessorFactory, times(2)).getPlugin(context);
+        verify(mockAccessor1).openForRead(); // first  attempt on accessor #1
```
bradfordb-vmware, line 81: This might be too specific, but would we want to check the order of these calls and/or that there is no other interactions with `mockAccessor1` after the failed call to `openForRead`?

bradfordb-vmware (approved):

lisakowen (approved):

Issue #683, opened 2021-08-18
pxf can not read composite avro file from aws s3 and return  error org.apache.avro.generic.GenericData$EnumSymbol cannot be cast to java.lang.String


se7enshidb:
# avsc meta data 
```
vim avro_schema.avsc 
{
"type" : "record",
  "name" : "example_schema",
  "namespace" : "com.example",
  "fields" : [ {
    "name" : "id",
    "type" : "long",
    "doc" : "Id of the user account"
  }, {
    "name" : "username",
    "type" : "string",
    "doc" : "Name of the user account"
  }, {
    "name" : "followers",
    "type" : {"type": "array", "items": "string"},
    "doc" : "Users followers"
  }, {
    "name": "fmap",
    "type": {"type": "map", "values": "long"}
  }, {
    "name": "relationship",
    "type": {
        "type": "enum",
        "name": "relationshipEnum",
        "symbols": ["MARRIED","LOVE","FRIEND","COLLEAGUE","STRANGER","ENEMY"]
    }
  }, {
    "name": "address",
    "type": {
        "type": "record",
        "name": "addressRecord",
        "fields": [
            {"name":"number", "type":"int"},
            {"name":"street", "type":"string"},
            {"name":"city", "type":"string"}]
    }
  } ],
  "doc:" : "A basic schema for storing messages"
}
```

# avro data 
```
vim pxf_avro.txt 
{"id":1, "username":"john","followers":["kate", "santosh"], "relationship": "FRIEND", "fmap": {"kate":10,"santosh":4}, "address":{"number":1, "street":"renaissance drive", "city":"san jose"}}
{"id":2, "username":"jim","followers":["john", "pam"], "relationship": "COLLEAGUE", "fmap": {"john":3,"pam":3}, "address":{"number":9, "street":"deer creek", "city":"palo alto"}}
```
java -jar ./avro-tools-1.9.1.jar fromjson --schema-file /tmp/avro_schema.avsc /tmp/pxf_avro.txt > /tmp/pxf_avro.avro

# convert
java -jar ./avro-tools.jar fromjson --schema-file ./avro_schema.avsc ./pxf_avro.txt > ./pxf_avro.avro

# create table
```
DROP FOREIGN TABLE s3_pxf_cn_with_avro;
CREATE FOREIGN TABLE s3_pxf_cn_with_avro(
id bigint, username text, followers text, fmap text, relationship text, address text
)
SERVER s3_server_cn
[...]

PR #696, opened 2021-09-01
Enum and bytea types are now handled properly for external tables using FORMAT "TEXT" or FORMAT "CSV"

hpandeycodeit:
Fixed the code to handle Enum and Bytea type casting
Enum and bytea are now handled properly for external tables
using FORMAT "TEXT" or FORMAT "CSV"
This commit fixes issue #683.

### DIFF
--- a/automation/src/test/java/org/greenplum/pxf/automation/features/avro/HdfsReadableAvroTest.java
+++ b/automation/src/test/java/org/greenplum/pxf/automation/features/avro/HdfsReadableAvroTest.java
@@ -217,7 +217,7 @@ public class HdfsReadableAvroTest extends BaseFeature {
         runTincTest("pxf.features.hdfs.readable.avro.complex_types_csv.runTest");
     }
 
-    /**PivotalSshRsa
+    /**
      * Read an Avro file that includes all supported complex types, reading the schema from another file.
      *
      * @throws Exception
--- a/automation/src/test/java/org/greenplum/pxf/automation/features/avro/HdfsReadableAvroTest.java
+++ b/automation/src/test/java/org/greenplum/pxf/automation/features/avro/HdfsReadableAvroTest.java
@@ -197,7 +197,17 @@ public class HdfsReadableAvroTest extends BaseFeature {
         runTincTest("pxf.features.hdfs.readable.avro.complex_types.runTest");
     }
 
-    /**
+    @Test(groups = {"features", "gpdb", "hcfs", "security"})
+    public void avroComplexTextFormat() throws Exception {
+        prepareReadableTable("avrotest_complex_text", AVRO_ALL_TYPES_FIELDS, hdfsPath + avroComplexFileName + SUFFIX_AVRO);
+        exTable.setFormatter(null);
+        exTable.setFormat("TEXT");
+        gpdb.createTableAndVerify(exTable);
+        // Verify results
+        runTincTest("pxf.features.hdfs.readable.avro.complex_types_text.runTest");
+    }
+
+    /**PivotalSshRsa
      * Read an Avro file that includes all supported complex types, reading the schema from another file.
      *
      * @throws Exception
--- a/automation/src/test/java/org/greenplum/pxf/automation/features/avro/HdfsReadableAvroTest.java
+++ b/automation/src/test/java/org/greenplum/pxf/automation/features/avro/HdfsReadableAvroTest.java
@@ -207,6 +207,16 @@ public class HdfsReadableAvroTest extends BaseFeature {
         runTincTest("pxf.features.hdfs.readable.avro.complex_types_text.runTest");
     }
 
+    @Test(groups = {"features", "gpdb", "hcfs", "security"})
+    public void avroComplexCsvFormat() throws Exception {
+        prepareReadableTable("avrotest_complex_csv", AVRO_ALL_TYPES_FIELDS, hdfsPath + avroComplexFileName + SUFFIX_AVRO);
+        exTable.setFormatter(null);
+        exTable.setFormat("CSV");
+        gpdb.createTableAndVerify(exTable);
+        // Verify results
+        runTincTest("pxf.features.hdfs.readable.avro.complex_types_csv.runTest");
+    }
+
     /**PivotalSshRsa
      * Read an Avro file that includes all supported complex types, reading the schema from another file.
      *
--- a/automation/tincrepo/main/pxf/features/hdfs/readable/avro/complex_types_text/expected/query01.ans
+++ b/automation/tincrepo/main/pxf/features/hdfs/readable/avro/complex_types_text/expected/query01.ans
@@ -1,0 +1,68 @@
+-- @description query01 for PXF HDFS Readable Avro supported complex types test cases
+
+GP_IGNORE:-- start_ignore
+GP_IGNORE:-- end_ignore
+
+SELECT type_long, type_string, type_array, type_union, type_enum, type_fixed from avrotest_complex_text ORDER BY type_long;
+GP_IGNORE: type_long | type_string |          type_array           |                            type_union                             | type_enum |  type_fixed
+GP_IGNORE:-----------+-------------+-------------------------------+-------------------------------------------------------------------+-----------+---------------
+1|shivram|[karthik,santosh,girish,arun]||FRIEND|?zst
+2|noa|[jimmy,godon]|this is optional type_unionellaneious information about this user|COLLEAGUE|\020!\003\004
+3|jimmy|[kate,santosh]||FRIEND|\001\002\003{
+4|godon|[caleb,parham]|new user|COLLEAGUE|4\002<\004
+(4 rows)
+
+CREATE view avro_view
+as select type_string, string_to_array(substring(type_map from 2 for (char_length(type_map) - 2)),',')::text[]
+as type_map, string_to_array(substring(type_record from 2 for (char_length(type_record) - 2)),',')::text[] as type_record from avrotest_complex_text;
+CREATE VIEW
+select exists(select type_map from avro_view where type_map <@ '{caleb:3,parham:3}' and type_map @> '{caleb:3,parham:3}');
+ ?column?
+----------
+ t
+(1 row)
+
+select exists(select type_map from avro_view where type_map <@ '{kate:10,santosh:4}' and type_map @> '{kate:10,santosh:4}');
+ ?column?
+----------
+ t
+(1 row)
+
+select exists(select type_map from avro_view where type_map <@ '{godon:1,jimmy:2}' and type_map @> '{godon:1,jimmy:2}');
+ ?column?
+----------
+ t
+(1 row)
+
+select exists(select type_map from avro_view where type_map <@ '{karthik:3,prak:4,girish:3,arun:3}' and type_map @>'{karthik:3,prak:4,girish:3,arun:3}');
+ ?column?
+----------
+ t
+(1 row)
+
+select exists(select type_record from avro_view where type_record <@ '{street:krishna street,number:12,city:chennai}' and type_record @> '{street:krishna street,number:12,city:chennai}');
+ ?column?
+----------
+ t
+(1 row)
+
+select exists(select type_record from avro_view where type_record <@ '{street:melon ct,number:754,city:sunnyvale}'and type_record @> '{street:melon ct,number:754,city:sunnyvale}');
+ ?column?
+----------
+ t
+(1 row)
+
+select exists(select type_record from avro_view where type_record <@ '{street:renaissance drive, number:1,city:san jose}' and type_record @> '{street:renaissance drive, number:1,city:san jose}');
+ ?column?
+----------
+ t
+(1 row)
+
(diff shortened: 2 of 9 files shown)


automation/src/test/java/org/greenplum/pxf/automation/features/avro/HdfsReadableAvroTest.java
```diff
@@ -197,7 +197,17 @@ public void avroComplex() throws Exception {
         runTincTest("pxf.features.hdfs.readable.avro.complex_types.runTest");
     }
 
-    /**
+    @Test(groups = {"features", "gpdb", "hcfs", "security"})
```
ashuka24, line 200: do we need to add a test case for format type CSV?

ashuka24 (commented):
Since this fix is also applicable to fdw, we can probably reword the commit and remove the reference to external tables

server/pxf-service/src/test/java/org/greenplum/pxf/service/BridgeOutputBuilderTest.java
```diff
@@ -144,7 +145,8 @@ public void testCSVSerialization() throws Exception {
                 new OneField(DataType.NUMERIC.getOID(), "0"),
                 new OneField(DataType.TIMESTAMP.getOID(), new Timestamp(0)),
                 new OneField(DataType.DATE.getOID(), new Date(1)),
-                new OneField(DataType.VARCHAR.getOID(), null)
+                new OneField(DataType.VARCHAR.getOID(), null),
+                new OneField(DataType.VARCHAR.getOID(), DataType.VARCHAR)
```
bradfordb-vmware, line 149: This seems a little odd to use the type name as the value as well. Since `col7` is already a `VARCHAR` with non-null data, what does this new column cover?

automation/src/test/java/org/greenplum/pxf/automation/features/avro/HdfsReadableAvroTest.java
denalex: good point, we should

server/pxf-service/src/test/java/org/greenplum/pxf/service/BridgeOutputBuilderTest.java
denalex: we needed some enum, `DataType.VARCHAR` was handy, but probably confusing, +1 for using smth else here

server/pxf-service/src/main/java/org/greenplum/pxf/service/BridgeOutputBuilder.java
```diff
@@ -468,18 +468,19 @@ void fillOneGPDBWritableField(OneField oneField, int colIdx)
     private String fieldListToCSVString(List<OneField> fields) {
         return fields.stream()
                 .map(field -> {
+                    // Check first if the field.val is null then using .toString() is safe in else branches.
                     if (field.val == null)
                         return greenplumCSV.getValueOfNull();
                     else if (field.type == DataType.BYTEA.getOID())
-                        return "\\x" + Hex.encodeHexString((byte[]) field.val);
+                        return "\\\\x" + Hex.encodeHexString((byte[]) field.val);
```
denalex, line 475: interesting, so this change did not cause any other test failures ?

denalex (approved):

server/pxf-service/src/main/java/org/greenplum/pxf/service/BridgeOutputBuilder.java
hpandeycodeit: So this works fine except for FORMAT "CSV" but I have handled that in the new commit.

server/pxf-service/src/test/java/org/greenplum/pxf/service/BridgeOutputBuilderTest.java
hpandeycodeit: Added an enum type for this just to avoid the confusion.

automation/src/test/java/org/greenplum/pxf/automation/features/avro/HdfsReadableAvroTest.java
hpandeycodeit: Added test for FORMAT "CSV"

automation/tincrepo/main/pxf/features/hdfs/readable/avro/complex_types_text/expected/query01.ans
```diff
@@ -0,0 +1,68 @@
+-- @description query01 for PXF HDFS Readable Avro supported complex types test cases
```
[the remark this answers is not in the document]
hpandeycodeit: updated these comments to make them more specific

server/pxf-service/src/main/java/org/greenplum/pxf/service/BridgeOutputBuilder.java
```diff
@@ -76,7 +76,7 @@
     private final GreenplumCSV greenplumCSV;
     private final OutputFormat outputFormat;
     private final List<ColumnDescriptor> columnDescriptors;
-
+    private final String format;
```
ashuka24, line 79: Can we rename this to be a little more specific? Maybe something like `gpdbTableFormat`

server/pxf-service/src/main/java/org/greenplum/pxf/service/BridgeOutputBuilder.java
```diff
@@ -468,18 +469,24 @@ void fillOneGPDBWritableField(OneField oneField, int colIdx)
...
+                    // Check first if the field.val is null then using .toString() is safe in else branches.
                     if (field.val == null)
                         return greenplumCSV.getValueOfNull();
                     else if (field.type == DataType.BYTEA.getOID())
-                        return "\\x" + Hex.encodeHexString((byte[]) field.val);
+                    {
+                        // check for Format Type here. if the Format Type is CSV, we should escape using single \
+                        // for Text or Custom Format types, it should \\ 
+                        String encodedStr = Hex.encodeHexString((byte[]) field.val);
+                        return (format != null && format.equalsIgnoreCase("csv")) ? "\\x" + encodedStr : "\\\\x" + encodedStr;
```
ashuka24, line 480: on what occasion would the format be null?

server/pxf-service/src/main/java/org/greenplum/pxf/service/BridgeOutputBuilder.java
```diff
@@ -468,18 +469,24 @@ void fillOneGPDBWritableField(OneField oneField, int colIdx)
...
                 .map(field -> {
+                    // Check first if the field.val is null then using .toString() is safe in else branches.
                     if (field.val == null)
                         return greenplumCSV.getValueOfNull();
                     else if (field.type == DataType.BYTEA.getOID())
-                        return "\\x" + Hex.encodeHexString((byte[]) field.val);
+                    {
+                        // check for Format Type here. if the Format Type is CSV, we should escape using single \
+                        // for Text or Custom Format types, it should \\ 
+                        String encodedStr = Hex.encodeHexString((byte[]) field.val);
```
ashuka24, line 479: would it be cleaner to have the format be a string? ex: 
```
String hexPrepend = format.equalsIgnoreCase("csv") ? '\\x' : '\\\\x';
return hexPrepend + Hex.encodeHexString((byte[]) field.val); 
```

server/pxf-service/src/test/java/org/greenplum/pxf/service/BridgeOutputBuilderTest.java
ashuka24: It looks like you added the TestEnum `HELLO` to the class but haven't updated the test to use the value?

server/pxf-service/src/main/java/org/greenplum/pxf/service/BridgeOutputBuilder.java
hpandeycodeit: yes looks much cleaner. Updated it.

server/pxf-service/src/main/java/org/greenplum/pxf/service/BridgeOutputBuilder.java
hpandeycodeit: From the unit test, the value of `format` was coming null. But I have set it now in the test class and removed the null check here.

server/pxf-service/src/main/java/org/greenplum/pxf/service/BridgeOutputBuilder.java
hpandeycodeit: done.

automation/src/test/java/org/greenplum/pxf/automation/features/avro/HdfsReadableAvroTest.java
```diff
@@ -197,7 +197,17 @@ public void avroComplex() throws Exception {
...
+    public void avroComplexTextFormat() throws Exception {
+        prepareReadableTable("avrotest_complex_text", AVRO_ALL_TYPES_FIELDS, hdfsPath + avroComplexFileName + SUFFIX_AVRO);
+        exTable.setFormatter(null);
+        exTable.setFormat("TEXT");
+        gpdb.createTableAndVerify(exTable);
+        // Verify results
+        runTincTest("pxf.features.hdfs.readable.avro.complex_types_text.runTest");
+    }
+
+    /**PivotalSshRsa
```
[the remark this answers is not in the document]
hpandeycodeit: fixed now.

server/pxf-service/src/test/java/org/greenplum/pxf/service/BridgeOutputBuilderTest.java
```diff
@@ -113,7 +113,7 @@ public void testFillGPDBWritable() throws Exception {
     @Test
     public void testCSVSerialization() throws Exception {
         RequestContext context = new RequestContext();
-
+        context.setFormat("TEXT");
```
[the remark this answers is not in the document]
hpandeycodeit: Unit test here was not setting `format` earlier and if we don't set it here, it will be passed as `null` in the `BridgeOutputBuilder.java` for this test. 
We have to write a separate unit test case if we want to set it format to `CSV` as well since we cannot set both `TEXT` and `CSV` in same test.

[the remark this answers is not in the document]
ashuka24: If that is the case, then this test should be renamed appropriately. the current test name indicates that it should be a csv test and not a text test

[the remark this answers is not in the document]
hpandeycodeit: SInce this is the CSV test, I have changed the FORMAT to `CSV` and changed the test as well.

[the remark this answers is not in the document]
denalex: I don't think our C extension will sed 'CSV' as the value: https://github.com/greenplum-db/pxf/blob/master/external-table/src/pxfheaders.c#L534-L537 -- CSV is treated as TEXT I believe, you can check it while debugging PXF

automation/src/test/java/org/greenplum/pxf/automation/features/avro/HdfsReadableAvroTest.java
```diff
@@ -197,6 +197,26 @@ public void avroComplex() throws Exception {
...
     }
 
+    @Test(groups = {"features", "gpdb", "hcfs", "security"})
+    public void avroComplexTextFormat() throws Exception {
+        prepareReadableTable("avrotest_complex_text", AVRO_ALL_TYPES_FIELDS, hdfsPath + avroComplexFileName + SUFFIX_AVRO);
+        exTable.setFormatter(null);
+        exTable.setFormat("TEXT");
+        gpdb.createTableAndVerify(exTable);
+        // Verify results
+        runTincTest("pxf.features.hdfs.readable.avro.complex_types_text.runTest");
```
bradfordb-vmware, line 207: I was able to change this to

```java
runTincTest("pxf.features.hdfs.readable.avro.complex_types.runTest");
```

and the test still passes. It looks like the created external tables are not being dropped between tests. Looking in to the code, `gpdb.createTableAndVerify(...)` will first drop the table before (re-)creating it. With that insight, I think we could re-use the table name (`avro_complex`) and the tinc test (`pxf.features.hdfs.readable.avro.complex_types.runTest`) to avoid adding two new tinc test cases.

If we want to keep the two new tinc test cases, can we include something like `\d avro_complex_text` in the SQL to verify the format of the table?

server/pxf-service/src/test/java/org/greenplum/pxf/service/BridgeOutputBuilderTest.java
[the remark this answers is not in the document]
bradfordb-vmware: If the C extension doesn't differentiate between `TEXT` or `CSV`, then that will be problematic. It looks like Postgres has different rules for `bytea` between the two.

bradfordb-vmware (approved):

ashuka24 (approved):

PR #697, opened 2021-09-02
Fixed NullPointerException in GSS failure handling retry logic

denalex:
This is a port of a similar fix done in branch-5.16: https://github.com/greenplum-db/pxf/pull/693

### DIFF
--- a/server/pxf-service/src/main/java/org/greenplum/pxf/service/bridge/ReadBridge.java
+++ b/server/pxf-service/src/main/java/org/greenplum/pxf/service/bridge/ReadBridge.java
@@ -60,7 +60,8 @@ public class ReadBridge extends BaseBridge {
      */
     @Override
     public boolean beginIteration() throws Exception {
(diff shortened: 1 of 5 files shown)


bradfordb-vmware (approved):
Looks good. One thing I wanted to double check my understanding on is why we don't need the `FragmentHolder` from #693. Looking at the code in `FragmentService`, we have the method `getFragmenter(RequestContext)` so in essence, the `FragmentService` takes the place of `FragmentHolder`. Is that correct (to first order approximation)?

denalex:
@bradfordb-vmware - we do not need `FragmenterHolder` here because in PXF6 `Configuration` is available from the `RequestContext` and hence we do not need to obtain an instance of `fragmenter` before calling the `failureHandler` and hence we can fold dynamically getting the fragmenter inside the processing lambda. The lambda will capture the instance of the `FragmenterService` and the "effectively final" local `context` variable which will allow it to dynamically request the new Fragmenter instance every time it is executed. There is also no need for a callback here.

hpandeycodeit (approved):

ashuka24 (approved):

PR #702, opened 2021-09-10
Remove CI logic for adjusting GPHOME on Ubuntu

bradfordb-vmware:
A recent changein GPDB to how greenplum_path.sh is generated caused the
grep for GPHOME to start failing. This code was only being run when the
target OS is Ubuntu. It looks like the binary tarball no longer expects
to be in /usr/local/gpdb. The only combination of CI
script/task/pipeline that is still using a binary tarball of GPDB is the
certification pipeline.

While manually reviewing the CI code to determine which CI tasks would
be impacted by this change, an unused task was identified and removed.


denalex (approved):

concourse/pipelines/certification_pipeline.yml
```diff
@@ -57,7 +57,17 @@ resources:
   type: git
   icon: git
   source:
-    tag_filter: release-*
+    # This pipeline should be using a release tag for running
+    # the certification tests, but the Ubuntu test is red on the latest release
+    # of PXF (currently 6.2.0) because of an upstream change in
+    # greenplum_path.sh. The upstream change only impacts PXF's CI code and
+    # not the product's functionality. Until the next release of PXF is tagged
+    # this pipeline should be pointed a the master branch and have the pxf_src
```
ashuka24, line 65:
```suggestion
    # this pipeline should be pointed at the master branch and have the pxf_src
```

concourse/tasks/test_pxf_multinode.yml
```diff
@@ -1,25 +0,0 @@
-platform: linux
```
ashuka24, line 1: why are we removing this yml file? and if we are not using it, can we also get rid of the test_pxf_multinode.bash file?

bradfordb-vmware: This task yaml is not referenced by any pipeline. The `test_pxf_multinode.bash` script is still being used by `test_pxf_on_ccp.yml`.

ashuka24 (approved):

hpandeycodeit (approved):

PR #703, opened 2021-09-10
Added Support for Avro Logical Types for Readable External Tables

hpandeycodeit:
Added Support for Avro Logical Types for Readable External Tables. Following Avro Logical Types are supported in Readable External Tables: 

```
Decimal
UUID
Date
Time (millisecond precision)
Time (microsecond precision)
Timestamp (millisecond precision)
Timestamp (microsecond precision)
Local timestamp (millisecond precision)
Local timestamp (microsecond precision)
```

This PR contains Implementation changes for the support of above types and  Unit and Automation Tests for these changes. 

https://avro.apache.org/docs/current/spec.html#Logical+Types


automation/src/test/java/org/greenplum/pxf/automation/features/avro/HdfsReadableAvroTest.java
```diff
@@ -217,6 +218,15 @@ public void avroComplexCsvFormat() throws Exception {
         runTincTest("pxf.features.hdfs.readable.avro.complex_types_csv.runTest");
     }
 
+    @Test(groups = {"features", "gpdb", "hcfs", "security"})
+    public void avroLogicalTypesUUID() throws Exception {
+        prepareReadableTable("logical_types_uuid", new String[]{"name uuid", "age int"}, hdfsPath + avroLogicalTypeFileName + SUFFIX_AVRO);
```
denalex, line 223: I think we should have 1 test for `logical_types` (similar to how we have primitive and complex ones). All supported logical types be part of one table / file so that we can verify all of them in 1 query and save on test execution time.

server/pxf-service/src/test/java/org/greenplum/pxf/service/BridgeOutputBuilderTest.java
```diff
@@ -129,6 +129,7 @@ public void testCSVSerialization() throws Exception {
         addColumn(context, 12, DataType.DATE, "col12");
         addColumn(context, 13, DataType.VARCHAR, "col13");
         addColumn(context, 14, DataType.VARCHAR, "col14");
+        addColumn(context, 14, DataType.TEXT, "col15");
```
denalex, line 132:
```suggestion
        addColumn(context, 15, DataType.TEXT, "col15");
```
wonder why nothing failed because of the wrong index here ?

automation/src/test/java/org/greenplum/pxf/automation/features/avro/HdfsReadableAvroTest.java
hpandeycodeit: Makes sense to test all the types (for Readable Tables) in one Test. We can put this on hold till we add all the logical types in the same test.

automation/jsystem.properties
```diff
@@ -1,17 +1,17 @@
-handlers=java.util.logging.FileHandler java.util.logging.ConsoleHandler
-htmlReportDir=log
+java.util.logging.FileHandler.pattern=jsystem%g.log
```
bradfordb-vmware, line 1: I think this diff is the result of using Java 11 for running automation locally (maybe) but I don't think we want to check this in.

server/pxf-api/src/main/java/org/greenplum/pxf/api/io/DataType.java
```diff
@@ -76,6 +82,12 @@
         BYTEAARRAY.typeElem = BYTEA;
         BPCHARARRAY.typeElem = BPCHAR;
         VARCHARARRAY.typeElem = VARCHAR;
+        DATEARRAY.typeElem = DATE;
+        UUIDARRAY.typeElem = TEXT; // since UUID's underlying type is String in AVRO
```
bradfordb-vmware, line 86: I think this should be

```java
UUIDARRAY.typeElem = UUID;
```

Avro does store UUIDs as strings, but in GPDB/Postgres, the elements of a UUID array are UUIDs.

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/avro/AvroTypeConverterTest.java
```diff
@@ -7,7 +7,7 @@
 import java.util.List;
 import java.util.stream.Collectors;
 
-public enum AvroTypeConverter {
+public enum AvroTypeConverterTest {
```
bradfordb-vmware, line 10: I know we needed to change the enum name but I don't think `AvroTypeConverterTest` is a good replacement since it sounds like its a test for the AvroTypeConverter` class but it isn't.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/AvroResolver.java
```diff
@@ -304,6 +330,7 @@ int populateRecord(List<OneField> record, Object fieldValue,
         return ret;
     }
 
+
```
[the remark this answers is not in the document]
hpandeycodeit: and this as well.

ashuka24 (approved):

bradfordb-vmware (approved):

PR #707, opened 2021-09-15
Enabled Kerberos Constrained Delegation impersonation for secure clusters

denalex:
Previously PXF could impersonate end-users to secure Hadoop clusters by requiring that Hadoop user (typically `gpadmin`) that maps to the Kerberos principal that PXF server runs as (typically `gpadmin/_HOST@REALM`) be granted Hadoop proxy privileges to impersonate other users. 

In case where it is not possible or desired to grant these Hadoop proxy privileges, users can now utilize Kerberos Constrained Delegation feature (also known as [S4U2proxy](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-sfu/bde93b0e-f3c9-4ddf-9f44-e1453be7af5a) ) to configure the PXF service principal in Microsoft Active Directory or RedHat IPA Server to be able to act as a delegate for obtaining Kerberos tickets from end users to the HDFS namenode service principal (typically `hdfs/_HOST@REALM`). It is no longer necessary to configure Hadoop proxy privileges since with the constrained delegation PXF will present to Hadoop the Kerberos ticket it obtained from the KDC on behalf of the end-user.

This new feature can now be enabled in PXF on a per-server basis by setting the `pxf.service.kerberos.constrained-delegation` property to `true` in `pxf-site.xml` file of the appropriate PXF configuration server. When this property is set to `true`, it is also required that the server is configured for Kerberos security (property `hadoop.security.authentication` is set to `kerberos` in `core-site.xml` file), otherwise the PXF server will return an error.

The intermediary S4U2 credentials are cached (per user/server in a JVM) to reduce the number of KDC calls. Caching duration is managed by a new property `pxf.service.kerberos.constrained-delegation.credential-cache.expiration` and is set to the value of `1 day` by default.


server/pxf-service/src/main/java/org/greenplum/pxf/service/security/PxfSaslPropertiesResolver.java
```diff
@@ -0,0 +1,105 @@
...
+/**
+ * Customized resolver that is called by Hadoop SASL mechanism when a SASL connection needs to be established.
+ * It obtains and stores an S4U2 proxy credential under "javax.security.sasl.credentials" property name so that
+ * the downstream machinery can perform the Kerberos Constrained Delegation logic. This resolver must be set
+ * on the Hadoop configuration object under "hadoop.security.saslproperties.resolver.class" property name.
+ */
+public class PxfSaslPropertiesResolver extends SaslPropertiesResolver {
+
+    private static final Logger LOG = LoggerFactory.getLogger(PxfSaslPropertiesResolver.class);
+    private final Supplier<GSSManager> gssManagerProvider; // break dependency on static method of GSSManager
```
ashuka24, line 30: is this a todo?

server/pxf-service/src/main/java/org/greenplum/pxf/service/security/PxfSaslPropertiesResolver.java
```diff
@@ -0,0 +1,105 @@
...
+    PxfSaslPropertiesResolver(Supplier<GSSManager> gssManagerProvider) {
+        this.gssManagerProvider = gssManagerProvider;
+    }
+
+    @Override
+    public Map<String, String> getClientProperties(InetAddress serverAddress, int ingressPort) {
+        return getClientProperties(serverAddress);
+    }
+
+    @SuppressWarnings("unchecked")
```
ashuka24, line 52: why is this warning suppression necessary?

server/pxf-service/src/main/java/org/greenplum/pxf/service/security/PxfSaslPropertiesResolver.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    /**
+     * Obtains an S4U2 proxy credential for the user and realm specified by the configuration.
+     * It must be run under from a "doAs" block from a Subject having a service Kerberos ticket.
+     *
+     * @return the proxy credential
+     */
+    private GSSCredential getKerberosProxyCredential() {
+        // TODO: cache credential per user/server and check if credential is near expiration ?
```
ashuka24, line 68: does this need to be handled in this pr?

server/pxf-service/src/main/java/org/greenplum/pxf/service/security/PxfSaslPropertiesResolver.java
denalex: no, this is the explanation why this provider is there (not to use static methods in the main flow). this will be changed in the next commit, though.

server/pxf-service/src/main/java/org/greenplum/pxf/service/security/PxfSaslPropertiesResolver.java
denalex: because MapWrapper<T> is being used as just MapWrapper since there's a mismatch between String and Object types.

server/pxf-service/src/main/java/org/greenplum/pxf/service/security/PxfSaslPropertiesResolver.java
denalex: yes, will be added in the next commit

bradfordb-vmware:
Good catches!

server/pxf-api/src/test/java/org/greenplum/pxf/api/security/SecureLoginTest.java
```diff
@@ -421,9 +421,16 @@ public void testGlobalImpersonationPropertyAbsent() {
         assertTrue(secureLogin.isUserImpersonationEnabled(configuration));
     }
 
+    @Test
+    public void testServerConfigurationImpersonationPropertyJunk() {
+        // true if missing or non-parsable boolean
+        configuration.set("pxf.service.user.impersonation", "junk");
```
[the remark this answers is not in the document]
denalex: correct, this is intentional, we ship the template with default `true`, so if users will go to the extends of changing it, they will have to make sure they specified `false` precisely. Any typo will preserve the "default" setting since they did not express their intention correctly.

automation/tincrepo/main/pxf/proxy/small_data_ipa/expected/query01.ans
```diff
@@ -0,0 +1,126 @@
...
+\set OLD_GP_USER :USER
+DROP ROLE IF EXISTS testuser;
+DROP ROLE
+CREATE ROLE testuser LOGIN RESOURCE QUEUE pg_default;
+CREATE ROLE
+\connect - testuser
+SELECT * FROM pxf_proxy_ipa_small_data_allowed ORDER BY name;
+ name   | num | dub |    longnum     | bool 
+---------+-----+-----+----------------+------
+ row_1   |   1 |   1 |   100000000000 | f
```
bradfordb-vmware, line 21: I don't think we need to change this for this PR, but for connectivity tests do we need to use 100 rows of data? Would using 10 be OK? I prefer reading/reviewing the `*.ans` files to the `*.sql` files because I can see both the query and the expectation (i.e., success/failure). I think if we used 10 rows it would be easier for internalizing the whole query file.

For data formatting tests, using 100 rows makes sense.

bradfordb-vmware (approved):
This looks good. I really appreciate the detailed comments around some of the more complex parts of the implementation. They definitely helped me with my understanding and reviewing the code.

I've made some comments here and there but none of them are show stoppers. I defer to your judgement which (if any) you want to address.

automation/src/main/java/org/greenplum/pxf/automation/components/hdfs/Hdfs.java
```diff
@@ -154,6 +171,26 @@ public void init() throws Exception {
...
+            namenodeSso.setHost(namenodeHost);
+            namenodeSso.setUserName(getSshUserName());
+            namenodeSso.setPrivateKey(getSshPrivateKey());
+            namenodeSso.init();
+
+            namenodePrincipal = config.get("dfs.namenode.kerberos.principal");
+            namenodeKeytab = config.get("dfs.namenode.keytab.file");
+            if (namenodePrincipal != null) {
+                // substitute _HOST portion of the principal with the namenode FQDN
+                namenodePrincipal = namenodePrincipal.replace("_HOST", getHostForConfiguredNameNode1HA());
```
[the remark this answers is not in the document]
denalex: Strictly speaking, you're correct. For `ipa-hadoop` config this will be the same, for any other config (none of them are using this setting yet to use a namenode principal to ssh to the namenode, kinit and do something there) we will have to pick the hostname provided in the SUT file so it has to be the full FQDN of the namenode host that was used to create a KRB principal. That FQDN might be on internal Hadoop network and not be routable via SSH from the place where the test runs (we made sure it works for IPA specifically). But we'll solve this when the need arises. I will make the suggested change.

automation/src/test/java/org/greenplum/pxf/automation/proxy/HiveProxySmokeTest.java
```diff
@@ -45,65 +49,88 @@ public void afterClass() throws Exception {
...
         // Generate Small data, write to HDFS and load to Hive
         Table dataTable = getSmallData();
         String fileName = "hiveSmallData.txt";
-        hdfs.writeTableToFile((hdfs.getWorkingDirectory() + "/" + fileName), dataTable, ",");
-        hive.loadData(hiveTableAllowed, (hdfs.getWorkingDirectory() + "/" + fileName), false);
-        hdfs.writeTableToFile((hdfs.getWorkingDirectory() + "/" + fileName), dataTable, ",");
-        hive.loadData(hiveTableProhibited, (hdfs.getWorkingDirectory() + "/" + fileName), false);
+        hdfsTarget.writeTableToFile((hdfsTarget.getWorkingDirectory() + "/" + fileName), dataTable, ",");
+        hiveTarget.loadData(hiveTableAllowed, (hdfsTarget.getWorkingDirectory() + "/" + fileName), false);
+        hdfsTarget.writeTableToFile((hdfsTarget.getWorkingDirectory() + "/" + fileName), dataTable, ",");
```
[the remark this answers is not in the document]
denalex: Yes, loading data in Hive will move the file in HDFS: https://cwiki.apache.org/confluence/display/hive/gettingstarted 
```
Note that loading data from HDFS will result in moving the file/directory. As a result, the operation is almost instantaneous.
```

automation/src/test/resources/sut/LocalToIPAMultiNodeHadoopHA.xml
```diff
@@ -0,0 +1,98 @@
...
+To use this file for running automation test, set the following line in your jsystem.properties file:
+sutFile=LocalToIPAMultiNodeHadoopHA.xml
+and then run:
+make TEST=HdfsHAFailoverTest
+-->
+<sut>
+    <cluster>
+        <class>org.greenplum.pxf.automation.components.cluster.SingleCluster</class>
+        <host>localhost</host>
+        <hiveBaseHdfsDirectory>/user/hive/warehouse</hiveBaseHdfsDirectory>
```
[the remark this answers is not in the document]
denalex: no, the real value used is `/user/hive/warehouse` for Hive 2 , we dynamically change it in CI from `/hive/warehouse/` to `/user/hive/warehouse` in `Multi*NodesCluster.xml` and if you have a mismatch you get all sorts of Metastore schema errors. That CI substitution (https://github.com/greenplum-db/pxf/blob/master/concourse/scripts/test_pxf_multinode.bash#L416) did cost me a few hours to troubleshoot :)

dev/ipa-cluster.bash
```diff
@@ -0,0 +1,317 @@
...
+
+    if ! type ansible-playbook &>/dev/null; then
+      fail 'ansible-playbook is not found, did you install it ?'
+    fi
+
+    if ! type xmlstarlet &>/dev/null; then
+      fail 'xmlstarlet is not found, did you install it (e.g. "brew install xmlstarlet") ?'
+    fi
+
+    : "${PXF_BASE?"PXF_BASE is required. export PXF_BASE=[YOUR_PXF_CONFIG_LOCATION]"}"
```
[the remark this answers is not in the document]
denalex: the alternative will be:
```
echo "${FOO?is not set}"
-bash: FOO: is not set
```
or in this case:
```
 → ./ipa-cluster.bash --destroy
./ipa-cluster.bash: line 70: TF_VAR_gcp_project: TF_VAR_gcp_project is required. export TF_VAR_gcp_project=[YOUR_GCP_PROJECT_NAME]
```
Is there a way not to print first `TF_VAR_gcp_project:` ? Otherwise
```
./ipa-cluster.bash: line 70: TF_VAR_gcp_project: is required. 
```
looks a bit empty as if the name of smth that was required is missing.
This pattern was copy & paste from somewhere else and is only for dev / CI scripts. In user-visible scripts we use explicit checking with proper messaging. I will leave this as is for now, until we come up with the uniform pattern.

dev/ipa-cluster.bash
```diff
@@ -0,0 +1,317 @@
...
+      // Ephemeral public IP
+    }
+  }
+}
+EOF
+}
+
+# apply terraform template to provision the cluster
+function terraform_apply() {
+    cd "${terraform_dir}" || fail "cannot go to ${terraform_dir}"
```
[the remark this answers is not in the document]
denalex: we remember `$WORKING_DIR` in the very beginning and go there in `fail()` and in successful end. We do not return there after each function as we jump between terraform and ansible directories. We just make sure we return back on error (via `fail`) or at the very end.

automation/tincrepo/main/pxf/proxy/small_data_ipa/expected/query01.ans
denalex: I agree, we call it `small_data` but it can be even smaller :) Can be done in a different PR across multiple test cases.

automation/src/test/java/org/greenplum/pxf/automation/proxy/HiveProxySmokeTest.java
```diff
@@ -45,65 +49,88 @@ public void afterClass() throws Exception {
...
+        // get HDFS and Hive to work against, it is "hdfs" and "hive" defined in SUT file by default,
+        // but subclasses can choose a different HDFS to use, such as HDFS in IPA-based Hadoop cluster
+        Hdfs hdfsTarget = getHdfsTarget();
+        Hive hiveTarget = getHiveTarget();
+
         // Create Hive tables
         hiveTableAllowed = TableFactory.getHiveByRowCommaTable("hive_table_allowed", HIVE_FIELDS);
         hiveTableProhibited = TableFactory.getHiveByRowCommaTable("hive_table_prohibited", HIVE_FIELDS);
 
         // hive.dropTable(hiveTable, false);
```
ashuka24, line 84: can we remove this unused code?

concourse/scripts/install_multinode_hadoop.bash
```diff
@@ -9,21 +9,24 @@ set -o errexit
 # 120-column terminal.
 stty cols 120
 
-cluster_name="$(cat terraform_hadoop/name)"
-metadata_path="terraform_hadoop/metadata"
+if ! type jq &>/dev/null; then
+	yum install -y jq
```
ashuka24, line 13: instead of doing a yum install here, can we just add jq to the base image?

concourse/scripts/install_pxf.bash
```diff
@@ -115,6 +115,7 @@ function create_pxf_installer_scripts() {
 			if [[ -d ~/ipa_env_files ]]; then
 				REALM_3="\$(< ipa_env_files/REALM)"
 				sed -i \
+					-e '/^\[libdefaul.*/a  \\\\tforwardable=true' \
```
ashuka24, line 118: is this supposed to be libdefault?

automation/src/main/java/org/greenplum/pxf/automation/components/hdfs/Hdfs.java
```diff
@@ -154,6 +172,37 @@ public void init() throws Exception {
...
+            namenodeSso.setPrivateKey(getSshPrivateKey());
+            namenodeSso.init();
+
+
+            // source environment file
+            namenodeSso.runCommand("source ~/.bash_profile");
+            namenodePrincipal = config.get("dfs.namenode.kerberos.principal");
+            namenodeKeytab = config.get("dfs.namenode.keytab.file");
+            if (namenodePrincipal != null) {
+                // substitute _HOST portion of the principal with the namenode FQDN, need to get it from the
```
ashuka24, line 196: can we write out `fully qualified domain name` here instead of using FQDN? or is it safe to assume that everyone will know what FQDN stands for?

server/pxf-api/src/main/java/org/apache/hadoop/security/PxfUserGroupInformation.java
```diff
@@ -260,7 +262,7 @@ private static String getOSLoginModuleName() {
      *
      * @return the user's TGT or null if none was found
      */
-    private synchronized KerberosTicket getTGT(Subject subject) {
+    public synchronized KerberosTicket getTGT(Subject subject) {
```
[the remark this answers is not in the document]
denalex: we now need to obtain current TGT in `GSSCredentialProvider` class for caching purposes

concourse/scripts/install_multinode_hadoop.bash
denalex: yes, makes sense, it is currently used from:
```
- task: Generate Multinode Hadoop Cluster
        file: pxf_src/concourse/tasks/install_multinode_hadoop.yml
        image: gpdb[[gp_ver]]-pxf-dev-centos7-image
```
Let's do that in a separate PR

concourse/scripts/install_pxf.bash
denalex: probably, but I think current regex would cover it anyways, @bradfordb-vmware - any feedback on this ?

automation/src/main/java/org/greenplum/pxf/automation/components/hdfs/Hdfs.java
denalex: I think it is safe to assume developers will know (or will google) FQDN acronym

PR #711, opened 2021-10-05
docs - indent all lines in echo commands

lisakowen:
each text line in an echo command should be indented to the same character location as the codeblock backticks.  the current content works, but is probably not markdown-compliant.


mperezfuster (approved):
Thanks for looking into this Lisa!

dyozie (approved):

PR #715, opened 2021-11-09
Update IPA Hadoop cluster for HA

bradfordb-vmware:
The IPA-based Hadoop cluster is updated with high availability using NFS
for the shared storage required by the NameNodes. The official Apache
Hadoop guide [0] was used to update the Hadoop configuration for HA.

The Terraform module is updated to provision an instance of Google
Filestore to provide a managed NFS server. Additionally, the number of
NameNode related resources (instances, TLS private keys, and self-signed
certifcates) are pinned to two (2) since Hadoop HA requires exactly two
NameNodes.

The Ansible inventory `namenode` group is split into two (one for the
active and one for the standby NameNode) in order to accomdate the
different steps for starting the active and standby NameNode processes.

This commit also fixes a bug in how the truststore for the Hadoop nodes
was being generated. For some reason, the PKCS12 file generated by the
OpenSSL CLI was not being properly read by Java so `keytool` is used to
generate a compatible PKCS12 file.

[0]: https://hadoop.apache.org/docs/r2.10.1/hadoop-project-dist/hadoop-hdfs/HDFSHighAvailabilityWithNFS.html


concourse/scripts/install_multinode_hadoop.bash
```diff
@@ -9,21 +9,24 @@ set -o errexit
 # 120-column terminal.
 stty cols 120
 
+if ! type keytool &>/dev/null; then
+	yum install -y java-1.8.0-openjdk
```
denalex, line 13: do we need to install Java, should it be already present ? Also what if we need Java 11 ?

denalex (approved):

concourse/terraform/ipa-multinode-hadoop/main.tf
```diff
@@ -51,11 +51,31 @@ resource "google_compute_instance" "ipa" {
...
+}
+
 data "google_compute_default_service_account" "default" {
 }
 
 resource "google_compute_instance" "namenode" {
-  count         = var.hdfs_namenode_count
+  # a maximum of two NameNodes may be configured per nameservice
+  # https://hadoop.apache.org/docs/r2.10.1/hadoop-project-dist/hadoop-hdfs/HDFSHighAvailabilityWithNFS.html
+  count         = 2
```
ashuka24, line 78: is there ever a case where we do not want both name nodes to be configured?

bradfordb-vmware: Good question. One of the challenges is that configuration of the Hadoop cluster varies significantly if we want HA or not -- if we are setting up HA, we have to set several configurations related to the name service. We could introduce a config parameter (e.g., `$SETUP_HADOOP_HA`) that we check in both Terraform and Ansible and take the appropriate steps. However, @denalex and I were discussing this earlier and we think we might not need it because:

1. If constrained delegation works for HA, it should be working for non-HA.
2. It should be possible to configure PXF to use this HA Hadoop cluster as though it was a non-HA cluster.

For now, pinning to exactly two nodes seems like a reasonable choice for a CI cluster. It's possible I am missing something though, so let me know if the above doesn't make sense.

PR #716, opened 2021-11-11
Decouple Constrained Delegation from impersonation

denalex:
Now Kerberos Constrained delegation also works when there is no impersonation, but a `pxf.service.user.name` property is provided in `pxf-site.xml`. In this case PXF requests that service user's TGS when connecting to Hadoop. If the service user is not defined and there is no impersonation, then KCD is not even attempted, since presenting login user (PXF service principal) identity is sufficient.


automation/src/test/java/org/greenplum/pxf/automation/proxy/HdfsProxySmokeTest.java
```diff
@@ -61,45 +68,35 @@ protected void prepareData() throws Exception {
...
+        // server with impersonation
+        createPxfTable(serverName, "_small_data_allowed", locationAllowed);
+        // server without impersonation but with a service user
+        createPxfTable(serverName + "-no-impersonation", "_small_data_allowed_no_impersonation", locationAllowed);
+        // server without impersonation and without a service user
+        createPxfTable(serverName + "-no-impersonation-no-svcuser", "_small_data_allowed_no_impersonation_no_svcuser", locationAllowed);
+
+    }
+
+    private void createPxfTable(String serverName, String tableSuffix, String location) throws Exception {
```
bradfordb-vmware, line 90: Maybe rename to `createReadablePxfTable()`

bradfordb-vmware (approved):

PR #718, opened 2021-11-15
Docs: specified examples for PXF_BASE path with and without relocation

mperezfuster:
It was brought to Docs attention that the section https://gpdb.docs.pivotal.io/pxf/6-2/using/jdbc_pxf_oracle.html could do with further clarification for locating $PXF_BASE for both cases, when it was been relocated and when it has not.
Adding an extra option to step 1 when configuring JDBC connectors for MySQL, Oracle and Trino, so we distinguish the path to specify when PXF_BASE has been relocated and when it has not.
Please let me know if you think this explanation is necessary or if it should be done differently. Thanks!
Preview: https://mireia-pxf-jdbc.sc2-04-pcf1-apps.oc.vmware.com/pxf/6-2/using/jdbc_pxf.html

### DIFF
--- a/docs/content/jdbc_pxf_oracle.html.md.erb
+++ b/docs/content/jdbc_pxf_oracle.html.md.erb
@@ -82,9 +82,17 @@ This procedure will typically be performed by the Greenplum Database administrat
 
 1. Download the Oracle JDBC driver and place it under `$PXF_BASE/lib` of your Greenplum Database master host. If you [relocated $PXF_BASE](about_pxf_dir.html#movebase), make sure you use the updated location. You can download a Oracle JDBC driver from your preferred download location. The following example places a driver downloaded from Oracle webiste under `$PXF_BASE/lib` of the Greenplum Database master:
 
-    ```shell
-    $ scp ojdbc10.jar gpadmin@gpmaster:/usr/local/pxf-gp<version>/lib/
-    ```
+    1. If you did not relocate `$PXF_BASE`, run the following:
+
+        ```shell
+        $ scp ojdbc10.jar gpadmin@gpmaster:/usr/local/pxf-gp<version>/lib/
+        ```
+
+    2. If you relocated `$PXF_BASE`, run the following:
+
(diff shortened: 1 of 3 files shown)


docs/content/jdbc_pxf_oracle.html.md.erb
```diff
@@ -82,9 +82,17 @@ This procedure will typically be performed by the Greenplum Database administrat
...
+    1. If you did not relocate `$PXF_BASE`, run the following:
+
+        ```shell
+        $ scp ojdbc10.jar gpadmin@gpmaster:/usr/local/pxf-gp<version>/lib/
+        ```
+
+    2. If you relocated `$PXF_BASE`, run the following:
+
+        ```shell
+        $ scp ojdbc10.jar gpadmin@gpmaster:$PXF_BASE/lib/
```
denalex, line 94: this assumes the command is run from another machine and PXF_BASE might not be set there, so if the command is just copied and pasted, it will evaluate `$PXF_BASE` to nothing and the file will be placed to `/lib/`, which is incorrect. Maybe indicate somehow that the actual value has to be instead there `gpadmin@gpmaster:<$PXF_BASE>/lib/` ?

mperezfuster: good point, let me make a few changes to it and see if it makes more sense.

lisakowen (approved):
looks good.  if $PXF_BASE is set, we instructed the user to add the setting to .bashrc, which would be picked up by the scp command.

the only other thing that popped in my head was that we could link to the "relocating PXF_BASE" topic in the "if you did not relocate ..." step.  not sure if that would overcomplicate things... ?

mperezfuster:
Thanks @lisakowen. Do you mean to link it under the "if you relocated ..." step instead? I am not sure why someone not relocating PXF_BASE would be interested in accessing that section. Also it is mentioned in the paragraph before splitting into the two options.

dyozie (approved):

PR #720, opened 2021-11-23
Redirect PXF stdout and stderr to files in PXF_LOGDIR

bradfordb-vmware:
Most of log messages generated by PXF are captured by log4j and sent to
the configured log file. However, log messages from start-up before
log4j is fully configured are sent to stdout/stderr. Additionally, some
debug output (e.g., `-Dsun.security.krb5.debug=true`) are only sent to
stdout/stderr. Redirecting stdout/stderr of the JVM process will allow
for easier access to this debug info.

Additionally, this comment moves the default value for the log level
from pxf-log4j2.xml to pxf-application.properties. The `bundle` lookup
will throw an error if the key is not found in the file. To supress this
output, the key is uncommented in the properties file. Removing it from
the XML file prevents the default value from being specified in two
locations.

### DIFF
--- a/server/pxf-service/src/scripts/pxf
+++ b/server/pxf-service/src/scripts/pxf
@@ -410,7 +410,7 @@ function do_start()
     echoGreen "   PXF_JVM_OPTS=${PXF_JVM_OPTS}"
     [[ $PXF_DEBUG == true ]] && echoGreen "   LOADER_PATH=${LOADER_PATH}"
 
-    "$javaexe" "${arguments[@]}" >> /dev/null 2>&1 &
+    "$javaexe" "${arguments[@]}" 1>>"${PXF_LOGDIR}/${identity}.out" 2>&1 &
     pid=$!
     disown $pid
     echo "$pid" > "$pid_file"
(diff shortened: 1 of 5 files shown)


server/pxf-service/src/templates/conf/pxf-log4j2.xml
```diff
@@ -25,7 +24,7 @@
         <AppenderRef ref="RollingFile"/>
     </Root>
 
-    <Logger name="org.greenplum.pxf" level="${env:PXF_LOG_LEVEL:-${sys:PXF_LOG_LEVEL:-info}}"/>
+    <Logger name="org.greenplum.pxf" level="${env:PXF_LOG_LEVEL:-${bundle:pxf-application:pxf.log.level}}"/>
```
denalex, line 27: I thought `env:` prefix refers to local variables declared in this file and `sys:` refers to system variables in which case we should use `sys:` prefix to refer to `PXF_LOG_LEVEL` variable that can be set in the shell before starting PXF, since the local variable definition is now removed.

ashuka24 (approved):
Nit: There's a spelling mistake in the commit message: `supress` --> `suppress`

server/pxf-service/src/scripts/pxf
```diff
@@ -410,7 +410,7 @@ function do_start()
     echoGreen "   PXF_JVM_OPTS=${PXF_JVM_OPTS}"
     [[ $PXF_DEBUG == true ]] && echoGreen "   LOADER_PATH=${LOADER_PATH}"
 
-    "$javaexe" "${arguments[@]}" >> /dev/null 2>&1 &
+    "$javaexe" "${arguments[@]}" 1>"${PXF_LOGDIR}/${identity}.out.log" 2>"${PXF_LOGDIR}/${identity}.err.log" &
```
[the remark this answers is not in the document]
bradfordb-vmware: This is capturing items printed to stdout (e.g., `System.out.println(...)`) so there really isn't anything we can control with respect to its format.

server/pxf-service/src/templates/conf/pxf-log4j2.xml
bradfordb-vmware: The `env` lookup is provided by [EnvironmentLookup](https://logging.apache.org/log4j/2.x/log4j-core/apidocs/org/apache/logging/log4j/core/lookup/EnvironmentLookup.html) which looks up keys from environment variables.; the `sys` lookup is provided by [SystemPropertiesLookup](https://logging.apache.org/log4j/2.x/log4j-core/apidocs/org/apache/logging/log4j/core/lookup/SystemPropertiesLookup.html) which looks up keys from system properties.

hpandeycodeit (approved):

denalex (approved):

PR #721, opened 2021-12-08
set default MySQL fetchSize to Integer.MIN_VALUE

denalex:
MySQL JDBC driver brings the complete ResultSet in PXF memory before returning it to PXF which causes PXF OutOfMemory errors when retrieving large datasets. The solution is to either use streaming mode by setting the `fetchSize` to `Integer.MIN_VALUE` (-2147483648) or use `useCursorFetch` property along with desired `fetchSize` to enable cursor fetch, which results in setting up temporary table on the server. Testing showed that using streaming is more performant and hence preferable. To allow customers use PXF with MySQL out of the box without additional tweaking, the default `fetchSize` for MySQL JDBC driver is now set to Integer.MIN_VALUE.


denalex:
NOTE: unit test for this is not done, since the logic relies on the name of the JDBC driver and the initialization logic tries to load the driver, so it cannot be mocked and we do not provide MySQL driver during test execution. The logic has been tested manually.

bradfordb-vmware (approved):

lisakowen (approved):

ashuka24 (approved):

hpandeycodeit (approved):

PR #726, opened 2021-12-14
pxf-hive: Catch TTransportException when working with metastore client

ashuka24:
Catching TTransportExceptions in the HiveMetastoreClientCompatibility1xx
allow PXF to retry connecting to the Hive metastore if necessary. The
default number of retries is 1 and can be set in the hive-site.xml using
`hive.metastore.failure.retries`.

### DIFF
--- a/server/pxf-hive/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClientCompatibility1xx.java
+++ b/server/pxf-hive/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClientCompatibility1xx.java
@@ -63,7 +63,7 @@ public class HiveMetaStoreClientCompatibility1xx extends HiveMetaStoreClient imp
                 throw ex;
             } catch (TTransportException transportException) {
                 /*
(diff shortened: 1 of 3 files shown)


hpandeycodeit (approved):

server/pxf-hive/src/main/java/org/apache/hadoop/hive/metastore/HiveMetaStoreClientCompatibility1xx.java
```diff
@@ -60,6 +61,12 @@ public Table getTable(String dbname, String name) throws TException {
             } catch (MetaException | NoSuchObjectException ex) {
                 LOG.debug("Original exception not re-thrown", e);
                 throw ex;
+            } catch (TTransportException transportException) {
+                /*
+                RetryingMetaStoreClient.java already contains logic to retry connecting to the metastore.
```
denalex, line 66:
```suggestion
                Propagate a TTransportException to allow RetryingMetaStoreClient that proxies this class to retry connecting to the metastore.
```

bradfordb-vmware (approved):

denalex (approved):

PR #727, opened 2021-12-14
Bump log4j2 version to 2.16.0

denalex:
A new CVE was raised related to Log4J2:  http://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-45046
Mitigation is to upgrade to 2.16.0 or remove the JndiLookup class from the log4j-core JAR.

Also, since at this moment PXF uses `org.slf4j:slf4j-api:1.7.31` and hence from the description in https://lists.apache.org/thread/d6v4r6nosxysyq9rvnr779336yf0woz4 can continue using the artifact `log4j-slf4j-impl-2.16.0.jar` and does not need to switch to`log4j-slf4j18-impl-2.16.0.jar`

### DIFF
--- a/server/build.gradle
+++ b/server/build.gradle
@@ -94,14 +94,14 @@ configure(javaProjects) {
             dependency("org.apache.htrace:htrace-core:3.1.0-incubating")
             dependency("org.apache.htrace:htrace-core4:4.0.1-incubating")
 
-            // --- bump log4j2 to 2.15.0 for CVE-2021-44228 fix, revert once
-            // org.springframework.boot:spring-boot-starter-log4j2 is upgraded to bundle log4j2:2.15.0+
-            dependencySet(group:"org.apache.logging.log4j", version:"2.15.0") {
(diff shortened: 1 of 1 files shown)


ashuka24 (approved):

bradfordb-vmware:
> Also, since at this moment PXF uses `org.slf4j:slf4j-api:1.7.31` and hence from the description in https://lists.apache.org/thread/d6v4r6nosxysyq9rvnr779336yf0woz4 can continue using the artifact **`log4j-slf4j-impl-2.16.0.jar`** and does not need to switch to`log4j-slf4j18-impl-2.16.0.jar`

Should that be `log4j-slf4j-impl-1.7.31.jar`?

bradfordb-vmware (approved):

denalex:
@bradfordb-vmware - no, `log4j-slf4j-impl` is part of Log4J project, so we use:
```
BOOT-INF/lib/log4j-slf4j-impl-2.16.0.jar
```

PR #737, opened 2022-01-18
Make IPA cluster HA and add automation test with queries before and after failover

denalex:
This is a PR to add additional testing for the W.I.P. feature of Kerberos Constrained Delegation. It introduces:
- scripts and tools to make the IPA cluster in GCP being Hadoop High-Availability (collaboration with @bradfordb-vmware)
- development scripts / Readme for testing with it from a local workstation
- automation test that runs queries before and after Hadoop HA failover to prove that caching Kerberos S4U2 credentials works in Hadoop HA environment even when the active namenode switches back and forth within the HA cluster.


automation/src/main/java/org/greenplum/pxf/automation/components/common/BaseSystemObject.java
```diff
@@ -23,4 +23,20 @@ public BaseSystemObject(boolean silentReport) {
 			report.setSilent(silentReport);
 		}
 	}
+
+	protected String replaceHome(String value) {
+		if (value == null) {
+			return null;
+		}
+		String home = System.getenv("HOME");
+		return value.replace("${home}", home).replace("${HOME}", home);
```
ashuka24, line 32: why do we have to replace for both lowercase and uppercase instances here?

automation/src/main/java/org/greenplum/pxf/automation/components/common/ShellSystemObject.java
```diff
@@ -107,15 +108,18 @@ public void init() throws Exception {
         /**
          * Direct to Private Key instead of Password based connection.
          */
-        File privateKey = new File(System.getProperty("user.home")
-                + "/.ssh/id_rsa");
-        connection.setPrivateKey(privateKey);
+        String privateKeyFileName = privateKey;
+        if (privateKeyFileName == null) {
+            privateKeyFileName = System.getProperty("user.home") + "/.ssh/id_rsa";
```
ashuka24, line 113: is it possible for this property to return null if it's not set? and if so, would it cause issues below when we try to create the file?

concourse/scripts/test_pxf_multinode.bash
```diff
@@ -332,9 +332,16 @@ function setup_pxf_kerberos_on_cluster() {
 			fi
 		"
 
+		# sync up PXF server configuration
 		ssh gpadmin@mdw "PXF_BASE=${BASE_DIR} ${PXF_HOME}/bin/pxf cluster sync"
+
+		# add configuration information to the SUT file for the automation suite
```
ashuka24, line 338: what is the SUT file?

automation/src/main/java/org/greenplum/pxf/automation/components/common/BaseSystemObject.java
denalex: out of overabundance of caution, basically allowing endusers (test developers) to make a "mistake" by specifying a lower-cased version

automation/src/main/java/org/greenplum/pxf/automation/components/common/ShellSystemObject.java
denalex: I am not sure, it will NPE at the worst case, this was just porting the existing logic, so it was presumed it would not be null before this.

concourse/scripts/test_pxf_multinode.bash
denalex: "System Under Test", the file that is used by automation to describe the system, e.g. `MultiHadoopMultiNodesCluster.xml` in case of this test in CI.

automation/src/test/java/org/greenplum/pxf/automation/features/hdfsha/HdfsHAFailoverTest.java
```diff
@@ -0,0 +1,95 @@
...
+        String locationTestUser = prepareData(TEST_USER);
+
+        // create PXF external table for no impersonation, no service user case (normal Kerberos)
+        createReadablePxfTable("hdfs-ipa-no-impersonation-no-svcuser", locationAdminUser);
+
+        // create PXF external table for impersonation based on Constrained Delegation
+        createReadablePxfTable("hdfs-ipa", locationTestUser);
+
+        // run tinc to read PXF data, it will issue 2 queries per PXF server to cache the tokens / use them
+        runTincTest("pxf.features.hdfsha.step_1_pre_failover.runTest");
```
bradfordb-vmware, line 45: Looking at the tinc test, there is no difference in the queries being run for steps 1, 2, or three. Could we just re-run the same tinc test at each step (e.g. `runTincTest("pxf.features.hdfsha.runTest");`)? My comment is less about the file duplication and more about indicating that the queries being run are not different.

bradfordb-vmware (approved):
Looks good. I made a few comments where I thought things could maybe be made a little bit clearer but I defer to your judgement.

automation/Makefile
```diff
@@ -74,8 +74,10 @@ endif
 
 ifeq "$(PXF_TEST_DEBUG)" "true"
 	MAVEN_DEBUG_OPTS+= -DPXF_TEST_DEBUG=$(PXF_TEST_DEBUG)
-	# Uncomment below line to remote debug pxf automation test
-	# MAVEN_DEBUG_OPTS+= -Dmaven.surefire.debug
+	# set PXF_TEST_DEBUG_ATTACH to true to remotely debug pxf automation test
+	ifeq "$(PXF_TEST_DEBUG_ATTACH)" "true"
+		MAVEN_DEBUG_OPTS+= -Dmaven.surefire.debug
```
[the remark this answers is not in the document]
denalex: Yes. PXF_TEST_DEBUG basically means "do not remove HDFS directories after the test so we can inspect them". To attach the debugger you were supposed to go into this file and uncomment the MAVEN_DEBUG_OPTS and manage not to checkin that change. I made it more flexible by driving via an environment variable that is easier to change back and forth between runs.

automation/src/main/java/org/greenplum/pxf/automation/components/common/BaseSystemObject.java
denalex: a user can place either `${HOME}` or `${home}` token in the property value and this will resolve it, either one option will match or none

automation/src/main/java/org/greenplum/pxf/automation/components/common/BaseSystemObject.java
```diff
@@ -23,4 +23,20 @@ public BaseSystemObject(boolean silentReport) {
...
+	}
+
+	protected String replaceUser(String value) {
+		if (value == null) {
+			return null;
+		}
+		String user = System.getenv("USER");
```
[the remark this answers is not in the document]
denalex: same answer :)

automation/src/main/java/org/greenplum/pxf/automation/components/common/ShellSystemObject.java
```diff
@@ -27,6 +27,7 @@
     private String hostName = "";
     private String userName;
     private String password;
+    private String privateKey;
```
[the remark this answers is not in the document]
denalex: no, we don't have any dev docs on the design of the automation framework, it's by inspecting code that we learn of details. would be nice to have one, I agree.

automation/src/main/java/org/greenplum/pxf/automation/components/common/ShellSystemObject.java
```diff
@@ -107,15 +108,18 @@ public void init() throws Exception {
         /**
          * Direct to Private Key instead of Password based connection.
          */
-        File privateKey = new File(System.getProperty("user.home")
-                + "/.ssh/id_rsa");
-        connection.setPrivateKey(privateKey);
+        String privateKeyFileName = privateKey;
+        if (privateKeyFileName == null) {
+            privateKeyFileName = System.getProperty("user.home") + "/.ssh/id_rsa";
+        }
```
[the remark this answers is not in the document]
denalex: yes, we can do that, I can make a change in a subsequent PR

automation/src/test/java/org/greenplum/pxf/automation/features/hdfsha/HdfsHAFailoverTest.java
denalex: As a matter of practice we try not to run the same query more than once to preserve its output and diffs (if any). We do not want the subsequent run overriding the previous one's results. Also, the queries technically do not have to be the same for this scenario, they just are because we do not really care about the actual query here.

automation/src/main/java/org/greenplum/pxf/automation/components/common/BaseSystemObject.java
bradfordb-vmware: I was reading this code backwards -- I thought we were performing the replacement on the value returned by `System.getenv()`.

PR #738, opened 2022-01-19
Fix: For reading the records correctly from a MultiLine JSON file 

hpandeycodeit:
There is an issue currently where in large Multiline JSON files (>128MB) with special characters, the bytes are not counted correctly. This could result in duplicate records in the GPDB.
    
This PR fixes the above issue.


server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
```diff
@@ -91,7 +90,7 @@ public JsonRecordReader(JobConf conf, FileSplit split) throws IOException {
             }
             is = fileIn;
         }
-        parser = new PartitionedJsonParser(is);
+        parser = new PartitionedJsonParser(fileIn);
```
denalex, line 93: what about the codec case from above, the codec was used to create the InputStream and the value is saved in `is` variable, but it seems not to be used anymore ?

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParser.java
```diff
@@ -42,26 +42,22 @@
...
-		this.inputStreamReader = new InputStreamReader(is, StandardCharsets.UTF_8);
+ 		this.fileIn =   fileIn;
 	}
 
 	private boolean scanToFirstBeginObject() throws IOException {
 		// seek until we hit the first begin-object
 		char prev = ' ';
 		int i;
-		while ((i = inputStreamReader.read()) != EOF) {
+		while ((i = fileIn.read()) != EOF) {
```
denalex, line 58: this method now returns a byte, not a character, since we switched from `InputStreamReader` to a subclass of `InputStream`. Not sure if the rest of the logic will be ok with that. `(char)` cast will only work for ASCII. Also, reading bytes and representing them as `int` is not very efficient.

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParserOffsetTest.java
```diff
@@ -19,37 +19,51 @@
...
-		assertEquals(54, parser.getBytesRead());
-		assertEquals(28, parser.getBytesRead() - result.length());
-		jsonInputStream.close();
-	}
-
-	public InputStream createFromString(String s) {
-		return new ByteArrayInputStream(s.getBytes());
-	}
+
+    private static String jsonFile = "src/test/resources/parser-tests/offset/input.json";
```
denalex, line 39: I think there is a way to refer to test resources without `src/test/resources/` path, smth like:
```
File f = new File(this.getClass().getClassLoader().getResource("example.txt").toURI());
```

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParserSeekTest.java
```diff
@@ -105,4 +116,18 @@ public void seekToStart(InputStream jsonInputStream) throws IOException {
...
+        context.setConfiguration(new Configuration());
+        jobConf = new JobConf(context.getConfiguration(), PartitionedJsonParserSeekTest.class);
+    }
+
+    public void getDataInputStream(File file) throws IOException {
+        Path path = new Path(file.getPath());
+        fs = path.getFileSystem(jobConf);
+        fileIn = fs.open(path);
+        fileIn.seek(0);
+    }
```
denalex, line 132: not sure what this test is doing, why we seek to 0 each time and how is it different from NoSeek test.

server/pxf-json/src/test/resources/parser-tests/offset/input.json
```diff
@@ -0,0 +1,13 @@
...
+  },
+  {
+    "color": "red",
+    "v": "vv"
+  }
+]
+
+
+
+
```
denalex, line 13: We need tests that work with UTF-8 multibyte chars to verify the issue. Also tests that introduce split sizes that fall in the middle of the current test file to make sure the logic works across splits for the same file (as well as across file boundaries, which maybe current Seek test is doing ?)

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParser.java
```diff
@@ -42,26 +42,22 @@
...
 	private boolean endOfStream = false;
+	private FSDataInputStream fileIn;
 
-	public PartitionedJsonParser(InputStream is) {
+	public PartitionedJsonParser(FSDataInputStream fileIn  ) {
 		this.lexer = new JsonLexer();
-
-		// You need to wrap the InputStream with an InputStreamReader, so that it can encode the incoming byte stream as
-		// UTF-8 characters
-		this.inputStreamReader = new InputStreamReader(is, StandardCharsets.UTF_8);
```
bradfordb-vmware, line 55: This breaks the use-case where `identifier` (i.e., `jsonMemberName` in `JsonRecordReader`) contains non-ASCII characters. Consider JSON data that looks like:

```json
[
  {
    "cüsötmerstätüs": "...",
    ...
  },
  {
    "cüsötmerstätüs": "...",
    ...
  },
  ...
]
```

with an external table 

```sql
CREATE EXTERNAL TABLE pxf_multiline_json (<columns>)
LOCATION ('pxf://data.json?profile=hdfs:json&identifier=cüsötmerstätüs') ON ALL
FORMAT 'CUSTOM' ( FORMATTER='pxfwritable_import' )
ENCODING 'UTF8';
```

Running `SELECT count(*) FROM pxf_multiline_json` will return a count of zero while the master branch returns a non-zero count.

Since JSON identifiers can contain UTF-8 characters, I think we need to keep the `InputStreamReader`.

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParser.java
```diff
@@ -42,26 +42,22 @@
...
 	private static final int EOF = -1;
-	private final InputStreamReader inputStreamReader;
-	private final JsonLexer lexer;
-	private long bytesRead = 0;
+ 	private final JsonLexer lexer;
 	private boolean endOfStream = false;
+	private FSDataInputStream fileIn;
 
-	public PartitionedJsonParser(InputStream is) {
+	public PartitionedJsonParser(FSDataInputStream fileIn  ) {
```
ashuka24, line 49: nitpick: 
```suggestion
	public PartitionedJsonParser(FSDataInputStream fileIn) {
```

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParser.java
```diff
@@ -42,26 +42,22 @@
...
 		this.lexer = new JsonLexer();
-
-		// You need to wrap the InputStream with an InputStreamReader, so that it can encode the incoming byte stream as
-		// UTF-8 characters
-		this.inputStreamReader = new InputStreamReader(is, StandardCharsets.UTF_8);
+ 		this.fileIn =   fileIn;
```
ashuka24, line 51: nitpick: 
```suggestion
 		this.fileIn = fileIn;
```

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParserOffsetTest.java
```diff
@@ -19,37 +19,51 @@
...
+    private JobConf jobConf;
+    private FileSystem fs;
+
+    @Test
+    public void testOffset() throws IOException {
+
+        PartitionedJsonParser parser = new PartitionedJsonParser(fileIn);
+        String result = parser.nextObjectContainingMember("color");
+        assertNotNull(result);
+        assertEquals(38, fileIn.getPos());
```
ashuka24, line 50: seems a little strange to me that we are not checking bytes read by calling the underlying datatype's `getPos()` function instead of just calling `parser.getBytesRead`. can we keep `getBytesRead` as wrapper over this function?

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParser.java
```diff
@@ -39,162 +39,165 @@
...
-		return bytesRead;
-	}
-
-	/**
-	 * @return Returns true if the end of the stream has been reached and false otherwise.
-	 */
-	public boolean isEndOfStream() {
-		return endOfStream;
-	}
+    private static final char BACKSLASH = '\\';
```
bradfordb-vmware, line 42: It looks like this file was re-indenting using spaces instead of tabs which makes it hard to view what functionally was changed. Can we leave this file indented with tabs so that the PR highlights the code changes?

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParserNoSeekTest.java
```diff
@@ -24,11 +24,7 @@
 import org.apache.commons.logging.LogFactory;
 import org.junit.jupiter.api.Test;
 
-import java.io.File;
-import java.io.FileInputStream;
-import java.io.FilenameFilter;
-import java.io.IOException;
-import java.io.InputStream;
+import java.io.*;
```
bradfordb-vmware, line 27: We want to avoid wildcard imports. You might need to configure your IDE to not use them; I don't think IntelliJ allows you to disable them completely but you can configure it to have really high number.

bradfordb-vmware (changes requested):

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParser.java
hpandeycodeit: There was some issue with the line endings which is fixed so you should see only the changes related to this PR.

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParserNoSeekTest.java
hpandeycodeit: fixed this as well.

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParser.java
```diff
@@ -61,9 +61,13 @@ private boolean scanToFirstBeginObject() throws IOException {
 		int i;
 		while ((i = inputStreamReader.read()) != EOF) {
 			char c = (char) i;
-			bytesRead++;
+			bytesRead += Character.toString(c).getBytes(StandardCharsets.UTF_8).length;
```
bradfordb-vmware, line 64: I wonder if it might be more performant to maintain a `CharBuffer` that we append characters to and then at the end of the function, call `charBuffer.toString.getBytes(StandardCharsets.UTF_8).length`.

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParser.java
```diff
@@ -174,6 +177,7 @@ public String nextObjectContainingMember(String memberName) throws IOException {
 					objectCount--;
 					if (objectCount < 0) {
 						// we're done! we reached an "}" which is at the same level as the member we found
+						bytesRead += currentObject.toString().getBytes(StandardCharsets.UTF_8).length;
```
bradfordb-vmware, line 180: We're calling `currentObject.toString()` twice here; I'd extract to a local variable.

```java
final String result = currentObject.toString()
bytes += result.getBytes(StandardCharsets.UTF_8).length;
return result
```

server/pxf-json/src/test/resources/parser-tests/offset/input.json
```diff
@@ -0,0 +1,8 @@
+[
+{"cüsötmerstätüs":"välid","name": "äää", "year": "2022", "address": "söme city", "zip": "95051"},
+{"cüsötmerstätüs":"invälid","name": "yī", "year": "2020", "address": "anöther city", "zip": "12345"}
+]
+
```
ashuka24, line 5: nit pick: remove extra lines

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParserOffsetTest.java
```diff
@@ -19,37 +19,148 @@
...
+
+        PxfInputFormat pxfInputFormat = new PxfInputFormat();
+        PxfInputFormat.setInputPaths(jobConf, path);
+    }
+
+    public LongWritable createKey() {
+        return new LongWritable();
+    }
+
+
```
ashuka24, line 162: nitpick: remove extra line

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParser.java
```diff
@@ -112,7 +116,6 @@ public String nextObjectContainingMember(String memberName) throws IOException {
 
 		while ((i = inputStreamReader.read()) != EOF) {
 			char c = (char) i;
-			bytesRead++;
```
bradfordb-vmware, line 115: I think we are still (potentially) under counting. There are some places where we call `currentObject.setLength(0)` and/or `currentObject.delete(...)` where we are not including those characters' bytes in `bytesRead`.

I wonder if it might be simpler to add an additional `CharBuffer` that we append every character read (`c`)  to. When the function exits, we can call `charBuffer.toString.getBytes(StandardCharsets.UTF_8).length`

As an example of where I think we might be miss counting is the following JSON where `memberName` is `id`:

```json
[
  {
    "id": ...,
  },
  {
    "skip": ...,
  },
  {
    "id": ...,
  }
]
```

Here the second object does not contain the `memberName` we are looking for, so we read past it to the third object but we only account for the third object's bytes.

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParser.java
```diff
@@ -61,9 +61,13 @@ private boolean scanToFirstBeginObject() throws IOException {
 		int i;
 		while ((i = inputStreamReader.read()) != EOF) {
 			char c = (char) i;
-			bytesRead++;
+			bytesRead += Character.toString(c).getBytes(StandardCharsets.UTF_8).length;
 			if (c == START_BRACE && prev != BACKSLASH) {
 				lexer.setState(JsonLexer.JsonLexerState.BEGIN_OBJECT);
+				//in case of START_BRACE "{" , subtract one byte as we will be counting this
```
ashuka24, line 67: why do we not have to worry about the ending brace?

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParserOffsetTest.java
```diff
@@ -19,37 +19,148 @@
...
+        assertEquals(data, data1);
+    }
+
+    @Test
+    public InputStream createFromString(String s) {
+        return new ByteArrayInputStream(s.getBytes());
+    }
+
+    @BeforeEach
```
bradfordb-vmware, line 143: Style-nit: Can we move the setup method to the beginning of the class?

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParserOffsetTest.java
```diff
@@ -19,37 +19,148 @@
...
+        jobConf.set(RECORD_MEMBER_IDENTIFIER, "cüsötmerstätüs");
+        file = new File(this.getClass().getClassLoader().getResource("parser-tests/offset/input.json").toURI());
+        context.setDataSource(file.getPath());
+        path = new Path(file.getPath());
+
+        PxfInputFormat pxfInputFormat = new PxfInputFormat();
+        PxfInputFormat.setInputPaths(jobConf, path);
+    }
+
+    public LongWritable createKey() {
```
bradfordb-vmware, line 158: Does this need to be public?

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParserOffsetTest.java
```diff
@@ -19,37 +19,148 @@
...
+        return new LongWritable();
+    }
+
+
+    public Text createValue() {
```
bradfordb-vmware, line 163: Does this need to be public?

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParserOffsetTest.java
```diff
@@ -19,37 +19,148 @@
...
+    /**
+     *  Here the record overlaps between Split-1 and Split-2.
+     *  reader will start reading the first record in the
+     *  middle of the split. It won't find the start { of that record but will
+     *  read till the end. It will successfully return the second record from the Split-2
+     */
+    public void testInBetweenSplits() throws IOException {
+
+        long start = 32;
+        fileSplit = new FileSplit(path, start, 100l, hosts);
```
bradfordb-vmware, line 86: For long literals, we should use `L` because it is really hard to tell (depending on the font) if that is a lowercase L (`l`) or a one (`1`).

server/pxf-json/src/test/resources/parser-tests/offset/input.json
bradfordb-vmware: This test set includes two-byte UTF-8 code points (umlauts) but we should include wider code points (three and four bytes are valid). As an example 'hello' in Chinese is `你好` which is two code points and is encoded using 6 total bytes. An example code point that is encoded with four bytes is `🦈` (shark emoji).

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParserOffsetTest.java
```diff
@@ -19,37 +19,148 @@
...
 import java.io.IOException;
 import java.io.InputStream;
+import java.net.URISyntaxException;
+import java.nio.charset.StandardCharsets;
 
-import org.junit.jupiter.api.Test;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
 
 public class PartitionedJsonParserOffsetTest {
```
denalex, line 45:
```suggestion
public class PartitionedJsonParserTest {
```

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParserOffsetTest.java
```diff
@@ -19,37 +19,148 @@
...
+        key = createKey();
+        data = createValue();
+        jsonRecordReader.next(key, data);
+        assertEquals(107, data.toString().getBytes(StandardCharsets.UTF_8).length);
+
+        // since the FileSplit starts at 32, which is the middle of the first record.
+        // so the reader reads the first record till it finds the end } and then starts reading the next record in the split
+        // it discards the previous read data but keeps track of the bytes read.
+
+        assertEquals(184, jsonRecordReader.getPos() - start);
```
denalex, line 97: this is testing `jsonRecordReader` in the unit test for the `PartitionedJsonParser`. Maybe create a separate utnit test class for testing the reader specifically. Parser test needs to assert the results of parsing

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParserOffsetTest.java
```diff
@@ -19,37 +19,148 @@
...
+    }
+
+    public LongWritable createKey() {
+        return new LongWritable();
+    }
+
+
+    public Text createValue() {
+        return new Text();
+    }
```
denalex, line 165: maybe we should have a set of tests using codec to exercise that flow.

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParser.java
hpandeycodeit: I tried with CharBuffer but it needs converting every character to a UTF. I have used StringBuilder Object Instead.

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParser.java
```diff
@@ -61,9 +61,13 @@ private boolean scanToFirstBeginObject() throws IOException {
...
+				// when we read the bytes from the currentObject string.
+				bytesRead--;
```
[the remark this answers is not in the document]
hpandeycodeit: Changed the logic so don't have to check for `{`

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParser.java
hpandeycodeit: Changed the logic so don't have to check for `{`

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParserOffsetTest.java
hpandeycodeit: Created a separate test class for `jsonRecordReader`

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParserOffsetTest.java
hpandeycodeit: Added the test for codec as well.

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParser.java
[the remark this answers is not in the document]
hpandeycodeit: I need some clarification on this. How this will help with optimization? In both the cases it will go through the same logic.

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParser.java
```diff
@@ -46,6 +46,7 @@
 	private final JsonLexer lexer;
 	private long bytesRead = 0;
 	private boolean endOfStream = false;
+	private StringBuilder charObject;
```
bradfordb-vmware, line 49: I think the name `charObject` is a little too generic -- maybe something like `uncountedCharsReadFromStream`? I don't know if that's the best name either but it should reflect that purpose of this builder -- buffering characters read from the stream that haven't been added to `bytesRead`.

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParser.java
```diff
@@ -112,11 +115,11 @@ public String nextObjectContainingMember(String memberName) throws IOException {
 
 		while ((i = inputStreamReader.read()) != EOF) {
 			char c = (char) i;
-			bytesRead++;
 
 			lexer.lex(c);
 
 			currentObject.append(c);
+			charObject.append(c);
```
bradfordb-vmware, line 122: Minor nit: can we move this up to right after the cast to `char` so that it matches the loop in `scanToFirstBeginObject`?

bradfordb-vmware (approved):

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParser.java
```diff
@@ -112,7 +115,7 @@ public String nextObjectContainingMember(String memberName) throws IOException {
 
 		while ((i = inputStreamReader.read()) != EOF) {
 			char c = (char) i;
-			bytesRead++;
+			uncountedCharsReadFromStream.append(c);
```
denalex, line 118: if we are reading a large file and the object we need is way at the end (or not even present) the `StringBuffer` will grow too large. Should we keep a long counter of characters read and update `bytesRead` and reset the `StringBuffer` every N characters (say 8096) ?

server/pxf-json/src/test/resources/parser-tests/offset/input.json
```diff
@@ -0,0 +1,7 @@
+[
+{"cüsötmerstätüs":"välid","name": "äää", "year": "2022", "address": "söme city", "zip": "95051"},
```
denalex, line 2:
```suggestion
{"cüstömerstätüs":"välid","name": "äää", "year": "2022", "address": "söme city", "zip": "95051"},
```
unless intentional. Here and everywhere else, seems like a typo

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParserTest.java
```diff
@@ -19,37 +19,37 @@
...
+    public void testOffset() throws IOException, URISyntaxException {
+       File file = new File(this.getClass().getClassLoader().getResource("parser-tests/offset/input.json").toURI());
+
+        InputStream jsonInputStream = new FileInputStream(file);
+
+        PartitionedJsonParser parser = new PartitionedJsonParser(jsonInputStream);
+        String result = parser.nextObjectContainingMember("cüsötmerstätüs");
+        assertNotNull(result);
+        assertEquals(107, parser.getBytesRead());
+        assertEquals(11, parser.getBytesRead() - result.length());
```
denalex, line 46: numbers here are try to visualize / verify. Maybe put a comment what we expect to see in the input file at these byte offsets. Also, the math subtracts number of characters `result.length()` from number of bytes read `parser.getBytesRead()`. I don't think this is correct, I think you need to do `result.getBytes(UTF_8).length` for proper accounting. If I am right, then how is that the test does not fail ?

server/pxf-json/src/test/resources/parser-tests/offset/input.json
```diff
@@ -0,0 +1,7 @@
...
+{"cüsötmerstätüs":"invälid","name": "yī", "year": "2020", "address": "anöther city", "zip": "12345"},
+{"cüsötmerstätüs":"invälid","name": "₡¥", "year": "2022", "address": "𑀓exas", "zip": "12345"},
+{"cüsötmerstätüs":"välid","name": "हिमांशु", "year": "1234", "address": "Same ₡i𑀓y", "zip": "00010"},
+{"cüsötmerstätüs":"välid","name": "你好", "year": "2033", "address": "0𑀓a", "zip": "19348"}
+]
```
denalex, line 7: I think we should try to test more files with different content, some I can think of:
* empty files
* empty collection files (contains only `[]` or `{}`)
* files that have no elements matching the keyword
* files that have some records matching the keyword intermixed with the ones that don't
* records where the keyword is not at the top level, asserting that it did not match

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParserTest.java
```diff
@@ -19,37 +19,37 @@
...
 
-import org.junit.jupiter.api.Test;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+public class PartitionedJsonParserTest {
+
+    @Test
+    public void testOffset() throws IOException, URISyntaxException {
+       File file = new File(this.getClass().getClassLoader().getResource("parser-tests/offset/input.json").toURI());
```
denalex, line 38: nit: alignment

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/parser/JsonRecordReaderTest.java
```diff
@@ -0,0 +1,140 @@
...
+        key = createKey();
+        data = createValue();
+        jsonRecordReader.next(key, data);
+        assertEquals(107, data.toString().getBytes(StandardCharsets.UTF_8).length);
+
+        // since the FileSplit starts at 32, which is the middle of the first record.
+        // so the reader reads the first record till it finds the end } and then starts reading the next record in the split
+        // it discards the previous read data but keeps track of the bytes read.
+
+        assertEquals(184, jsonRecordReader.getPos() - start);
```
denalex, line 97: it would be nice to assert that the record it read is expected (i.e. look inside the record or just compare JSON strings)

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/parser/JsonRecordReaderTest.java
```diff
@@ -0,0 +1,140 @@
...
+    public void testRecordSizeSmallerThanSplit() throws IOException {
+
+        fileSplit = new FileSplit(path, 0, 1000, hosts);
+        jsonRecordReader = new JsonRecordReader(jobConf, fileSplit);
+        key = createKey();
+        data = createValue();
+        while(jsonRecordReader.next(key, data))
+        {
+            assertNotNull(data);
+        }
```
denalex, line 130: we need to assert more, this should be similar to `testWithCodec`

server/pxf-json/src/test/resources/parser-tests/offset/input.json
hpandeycodeit: Added these additional tests and fixed other comments as well.

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
```diff
@@ -80,7 +80,7 @@ public JsonRecordReader(JobConf conf, FileSplit split) throws IOException {
 
         // openForWrite the file and seek to the start of the split
         FileSystem fs = file.getFileSystem(conf);
-        FSDataInputStream fileIn = fs.open(split.getPath());
+        fileIn = fs.open(split.getPath());
```
denalex, line 83: it does not seem we still need to have `fileIn` to be a class member. I think having it as local variable, as before, is sufficient as we are not accessing it from any other method.

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParser.java
```diff
@@ -46,24 +46,27 @@
...
+	private StringBuilder uncountedCharsReadFromStream;
 
 	public PartitionedJsonParser(InputStream is) {
 		this.lexer = new JsonLexer();
 
 		// You need to wrap the InputStream with an InputStreamReader, so that it can encode the incoming byte stream as
 		// UTF-8 characters
 		this.inputStreamReader = new InputStreamReader(is, StandardCharsets.UTF_8);
+
+		this.uncountedCharsReadFromStream = new StringBuilder(8192);
```
denalex, line 58: let's extract `8192` as a constant

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParser.java
```diff
@@ -46,24 +46,27 @@
 	private final JsonLexer lexer;
 	private long bytesRead = 0;
 	private boolean endOfStream = false;
+	private StringBuilder uncountedCharsReadFromStream;
```
denalex, line 49:
```suggestion
	private final StringBuilder uncountedCharsReadFromStream;
```

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParser.java
```diff
@@ -197,4 +200,21 @@ public long getBytesRead() {
 	public boolean isEndOfStream() {
 		return endOfStream;
 	}
+
+	private int readNextChar() throws IOException {
+		int i = inputStreamReader.read();
```
denalex, line 205: this read character can be EOF here, which is not even a character by itself, it is value -1 that signifies "nothing more to read", so we need to check for this and not append / count it.

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParser.java
```diff
@@ -197,4 +200,21 @@ public long getBytesRead() {
...
+		uncountedCharsReadFromStream.append((char) i);
+		if (uncountedCharsReadFromStream.length() > 8096) {
```
denalex, line 207: I think you meant `8192`, so let's use the constant here. And that's too late, this will cause StringBuilder to expand capacity, we need to prevent it going over the target capacity.

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/parser/JsonRecordReaderTest.java
```diff
@@ -0,0 +1,355 @@
...
+        data = createValue();
+        int recordCount = 0;
+        if (jsonRecordReader.next(key, data)) {
+            assertNotNull(data);
+            recordCount++;
+        }
+
+        assertEquals(0, recordCount);
+        assertEquals(12, jsonRecordReader.getPos());
+        assertEquals("", data.toString());
```
bradfordb-vmware, line 193:
```suggestion
        assertFalse(jsonRecordReader.next(key, value));
        assertEquals(12, jsonRecordReader.getPos());
```

Since `jsonRecordReader.next(key, value)` is only supposed to return `true` if a key/value was read, I don't know if we need to assert anything about `data`.

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/parser/JsonRecordReaderTest.java
```diff
@@ -0,0 +1,355 @@
...
+        path = new Path(file.getPath());
+        fileSplit = new FileSplit(path, 0, 10, hosts);
+        jsonRecordReader = new JsonRecordReader(jobConf, fileSplit);
+
+        assertEquals(0, jsonRecordReader.getPos());
+
+        key = createKey();
+        data = createValue();
+        int recordCount = 0;
+        if (jsonRecordReader.next(key, data)) {
```
bradfordb-vmware, line 211: Same comment as the previous two tests.

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/parser/JsonRecordReaderTest.java
```diff
@@ -0,0 +1,355 @@
...
+        key = createKey();
+        data = createValue();
+        int recordCount = 0;
+        while (jsonRecordReader.next(key, data)) {
+            assertNotNull(data);
+            recordCount++;
+        }
+
+        //cüstömerstätüs identifier will retrieve 4 records
+        assertEquals(4, recordCount);
```
bradfordb-vmware, line 241: Do we expect that all of the bytes of `mixed_input.json` will have been read at this point? Can we assert the value of `jsonRecordReader.getPos()`?

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/parser/JsonRecordReaderTest.java
```diff
@@ -0,0 +1,355 @@
...
+        jsonRecordReader = new JsonRecordReader(jobConf, fileSplit);
+        key = createKey();
+        data = createValue();
+        recordCount = 0;
+        while (jsonRecordReader.next(key, data)) {
+            assertNotNull(data);
+            recordCount++;
+        }
+
+        assertEquals(1, recordCount);
```
bradfordb-vmware, line 255: Do we expect that all of the bytes of `mixed_input.json` will have been read at this point? Can we assert the value of `jsonRecordReader.getPos()`?

server/pxf-json/src/test/resources/parser-tests/offset/complex_input.json
```diff
@@ -0,0 +1,16 @@
+[{
```
bradfordb-vmware, line 1: I don't think this file is valid JSON -- when I try to parse it with `jq`, I get an error:

```console
$ <complex_input.json jq
parse error: Objects must consist of key:value pairs at line 16, column 1
```

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/parser/JsonRecordReaderTest.java
```diff
@@ -0,0 +1,355 @@
...
+            recordCount++;
+        }
+
+        //cüstömerstätüs identifier will retrieve 4 records
+        assertEquals(2, recordCount);
+    }
+
+    @Test
+    public void testSplitBeforeMemberName() throws URISyntaxException, IOException {
+        // Split After { before MEMBER_NAME or After RECORD_MEMBER_IDENTIFIER will cause the record to skip
```
bradfordb-vmware, line 285: I had to read and re-read this description to understand the test and I am still not sure that I do:

I think we are testing the case where a file split starts after the JSON `BEGIN_OBJECT` (i.e., `{`) but before the identifier we are looking for (e.g., `{<split-start>"<identifier>"...`). In this case, we expect the record reader to skip over this object entirely.

Is that a correct understanding of this test?

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/parser/JsonRecordReaderTest.java
```diff
@@ -0,0 +1,355 @@
...
+
+        key = createKey();
+        data = createValue();
+        int recordCount = 0;
+        while (jsonRecordReader.next(key, data)) {
+            assertNotNull(data);
+            recordCount++;
+        }
+
+        assertEquals(1, recordCount);
```
bradfordb-vmware, line 303: Here I think it would be really helpful to assert on what we expect data to be. Since `data` is expected to be a valid JSON object, it might even make sense to convert it to an object and assert the field values:

```java
JsonNode jsonNode = objectMapper.readTree(data);
String color = jsonNode.get("color").asText();
assertEquals("red", color);
...
```

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/parser/JsonRecordReaderTest.java
```diff
@@ -0,0 +1,355 @@
...
+        //cüstömerstätüs identifier will retrieve 4 records
+        assertEquals(2, recordCount);
+    }
+
+    @Test
+    public void testSplitBeforeMemberName() throws URISyntaxException, IOException {
+        // Split After { before MEMBER_NAME or After RECORD_MEMBER_IDENTIFIER will cause the record to skip
+        //Here the split start after the <SEEK> keyword in the file ( Line # 3 ) and right before the identifier.
+        jobConf.set(RECORD_MEMBER_IDENTIFIER, "name");
+        file = new File(this.getClass().getClassLoader().getResource("parser-tests/seek/seek-into-mid-object-1/input.json").toURI());
```
bradfordb-vmware, line 288: I know this PR doesn't add this JSON file, but it is not valid JSON

```console
$ <seek/seek-into-mid-object-1/input.json jq
parse error: Invalid numeric literal at line 3, column 11
```

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/parser/JsonRecordReaderTest.java
hpandeycodeit: Yes that is correct.

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/parser/JsonRecordReaderTest.java
hpandeycodeit: This is due to the `<SEEK>` in these files where the other tests are reading this keyword as an offset.

server/pxf-json/src/test/resources/parser-tests/offset/complex_input.json
hpandeycodeit: Fixed this.

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/parser/JsonRecordReaderTest.java
hpandeycodeit: Added assertions on some of these data values and fixed other comments as well.

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParser.java
```diff
@@ -42,28 +42,32 @@
 	private static final char BACKSLASH = '\\';
 	private static final char START_BRACE = '{';
 	private static final int EOF = -1;
+	private static final long CHARS_READ_LIMIT = 8192L;
```
ashuka24, line 45: Where did this number come from? Is it a standard char limit value?

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParser.java
```diff
@@ -42,28 +42,32 @@
...
 		// seek until we hit the first begin-object
 		char prev = ' ';
 		int i;
-		while ((i = inputStreamReader.read()) != EOF) {
+		while ((i = readNextChar()) != EOF) {
 			char c = (char) i;
-			bytesRead++;
 			if (c == START_BRACE && prev != BACKSLASH) {
 				lexer.setState(JsonLexer.JsonLexerState.BEGIN_OBJECT);
+
```
ashuka24, line 70: nit: unnecessary new line

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/parser/JsonRecordReaderTest.java
```diff
@@ -0,0 +1,376 @@
...
+        // the first record will end at 107 bytes and this record will be ignored since the split started after {
+        long start = 32;
+        fileSplit = new FileSplit(path, start, 100, hosts);
+        jsonRecordReader = new JsonRecordReader(jobConf, fileSplit);
+
+        assertEquals(32, jsonRecordReader.getPos());
+
+        key = createKey();
+        data = createValue();
+        jsonRecordReader.next(key, data); // why it's reading beyond split boundaries?? Have more clarification around the boundary
```
ashuka24, line 90: Does this comment need to be addressed?

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/parser/JsonRecordReaderTest.java
```diff
@@ -0,0 +1,376 @@
...
+        jsonRecordReader.next(key, data);
+        // reads the full 1 record here
+        assertEquals(105, data.toString().getBytes(StandardCharsets.UTF_8).length);
+
+        assertEquals("{\"cüstömerstätüs\":\"välid\",\"name\": \"äää\", \"year\": \"2022\", \"address\": \"söme city\", \"zip\": \"95051\"}", data.toString());
+    }
+
+    @Test
+    /**
+     * The Split size is large only single spilt will be able to read all the records.
```
ashuka24, line 130:
```suggestion
     * The Split size is large so a single split will be able to read all the records.

File: server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParser.java (hunk 48 above)
Comment (hpandeycodeit) [reply server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParser.java:45]: Not standard. It is `8K ( 1024*8)` chars count limit that we thought would be good to reset the buffer in case of very large object or zipped file to avoid buffer overflow.

File: server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/parser/JsonRecordReaderTest.java (hunk 50 above)
Comment (hpandeycodeit) [reply server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/parser/JsonRecordReaderTest.java]: This was already addressed in the comments already. Missed to remove it. Thanks! 

Fixed rest of the comments as well.

File: server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParser.java [hunk 52]
```diff
@@ -42,26 +42,29 @@
...
+	private final StringBuilder uncountedCharsReadFromStream;
 
 	public PartitionedJsonParser(InputStream is) {
 		this.lexer = new JsonLexer();
 
 		// You need to wrap the InputStream with an InputStreamReader, so that it can encode the incoming byte stream as
 		// UTF-8 characters
 		this.inputStreamReader = new InputStreamReader(is, StandardCharsets.UTF_8);
+
+		this.uncountedCharsReadFromStream = new StringBuilder();
```
denalex, line 59:
```suggestion
		this.uncountedCharsReadFromStream = new StringBuilder(CHARS_READ_LIMIT);
```

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParser.java
```diff
@@ -197,4 +200,24 @@ public long getBytesRead() {
 	public boolean isEndOfStream() {
 		return endOfStream;
 	}
+
+	private int readNextChar() throws IOException {
+		int i = inputStreamReader.read();
+
+		if (i != -1) {
```
denalex, line 207:
```suggestion
		if (i != EOF) {
```

denalex (approved):

PR #741, opened 2022-01-24
Improved performance of iterating over a list of fragments

denalex:
A `LinkedList` of fragments was accessed using index-based access which resulted in suboptimal performance when the list contained 10M+ objects. The operational complexity was O(N^2) and now it is O(N). The execution time went down from 15 mins to 1.3 seconds for 10M fragments.

### DIFF
--- a/server/pxf-service/src/main/java/org/greenplum/pxf/service/FragmenterService.java
+++ b/server/pxf-service/src/main/java/org/greenplum/pxf/service/FragmenterService.java
@@ -168,14 +168,18 @@ public class FragmenterService {
      * @return the filtered list of fragments for the given segment
      */
     private List<Fragment> filterFragments(List<Fragment> fragments, int segmentId, int totalSegments, int gpSessionId, int gpCommandCount) {
-        int shift = gpSessionId % totalSegments;
-        int fragmentCount = fragments.size();
+        // use artificial shift so that queries with low fragment count are not always executed by the low-numbered segments
+        // this helps to spread out the workload across the cluster to different PXF JVMs
+        int shift = gpSessionId % totalSegments + gpCommandCount;
 
-        List<Fragment> filteredFragments = new ArrayList<>((int) Math.ceil(fragmentCount / totalSegments));
-        for (int i = 0; i < fragmentCount; i++) {
-            if (segmentId == ((i + shift + gpCommandCount) % totalSegments)) {
-                filteredFragments.add(fragments.get(i));
+        List<Fragment> filteredFragments = new ArrayList<>((int) Math.ceil(fragments.size() / totalSegments));
+        // do not use "for loop with index" to iterate over fragments as the incoming fragment list is a LinkedList
+        int i = 0;
+        for (Fragment fragment : fragments) {
+            if (segmentId == ((i + shift) % totalSegments)) {
+                filteredFragments.add(fragment);
             }
+            i++;
         }
         return filteredFragments;
     }
--- a/server/pxf-service/src/main/java/org/greenplum/pxf/service/FragmenterService.java
+++ b/server/pxf-service/src/main/java/org/greenplum/pxf/service/FragmenterService.java
@@ -168,18 +168,23 @@ public class FragmenterService {
      * @return the filtered list of fragments for the given segment
      */
     private List<Fragment> filterFragments(List<Fragment> fragments, int segmentId, int totalSegments, int gpSessionId, int gpCommandCount) {
-        // use artificial shift so that queries with low fragment count are not always executed by the low-numbered segments
-        // this helps to spread out the workload across the cluster to different PXF JVMs
-        int shift = gpSessionId % totalSegments + gpCommandCount;
+        /*
+        We use a mod function inside the loop to distribute fragments across all N segments for processing
+        in a round-robin fashion, where each segment will be allocated to work on every N-th fragment.
+
+        We use an artificially shifted index so that queries with low fragment count are not always executed by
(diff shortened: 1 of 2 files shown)


server/pxf-service/src/test/java/org/greenplum/pxf/service/FragmenterServiceTest.java
```diff
@@ -431,4 +433,42 @@ public void testBeginIterationGSSFailureAfterMaxRetries() throws Throwable {
...
+        long start = System.currentTimeMillis();
+        List<Fragment> response = fragmenterService.getFragmentsForSegment(context1);
+        long end = System.currentTimeMillis();
+
+        verify(fragmenter1, times(1)).getFragments();
+
+        assertTrue(response instanceof ArrayList);
+        assertEquals(100000, response.size());
+        assertEquals("foo.bar", response.get(0).getSourceName());
+        assertTrue(end-start < 10000L); // should be less than 10 secs (8x margin), not minutes
```
ashuka24, line 471: how did we decide on this value?

denalex: purely using intuition and estimation that probably no real system that will run a test suite will be 8x slower than MacBookPro and 10 seconds looks round enough. 1 second is too low, 5 seconds is a bit risky. What other value would you suggest ?

server/pxf-service/src/main/java/org/greenplum/pxf/service/FragmenterService.java
```diff
@@ -168,14 +168,18 @@ public FragmenterService(FragmenterCacheFactory fragmenterCacheFactory,
      * @return the filtered list of fragments for the given segment
      */
     private List<Fragment> filterFragments(List<Fragment> fragments, int segmentId, int totalSegments, int gpSessionId, int gpCommandCount) {
-        int shift = gpSessionId % totalSegments;
-        int fragmentCount = fragments.size();
+        // use artificial shift so that queries with low fragment count are not always executed by the low-numbered segments
+        // this helps to spread out the workload across the cluster to different PXF JVMs
+        int shift = gpSessionId % totalSegments + gpCommandCount;
```
bradfordb-vmware, line 173: Are we adding `gpCommandCount` for the case where multiple commands are executed in the same session?

server/pxf-service/src/main/java/org/greenplum/pxf/service/FragmenterService.java
```diff
@@ -168,14 +168,18 @@ public FragmenterService(FragmenterCacheFactory fragmenterCacheFactory,
...
 
-        List<Fragment> filteredFragments = new ArrayList<>((int) Math.ceil(fragmentCount / totalSegments));
-        for (int i = 0; i < fragmentCount; i++) {
-            if (segmentId == ((i + shift + gpCommandCount) % totalSegments)) {
-                filteredFragments.add(fragments.get(i));
+        List<Fragment> filteredFragments = new ArrayList<>((int) Math.ceil(fragments.size() / totalSegments));
+        // do not use "for loop with index" to iterate over fragments as the incoming fragment list is a LinkedList
```
bradfordb-vmware, line 176: Just checking my understanding here: The issue isn't the "for loop with index" itself but calling `fragments.get(i)` inside the loop since that method is `O(n)` which makes the "for loop with index" `O(n^2)`.

bradfordb-vmware (approved):

server/pxf-service/src/main/java/org/greenplum/pxf/service/FragmenterService.java
denalex: yes, correct. this change is just moving that constant out of the loop, so no overall logic change to the hashing / mod function.

server/pxf-service/src/main/java/org/greenplum/pxf/service/FragmenterService.java
denalex: yes, correct. I guess my default non explicitly stated assumption is that "for loop with index" is usually used when we want to use the loop counter for index-based access into a collection, otherwise we tend to use an "iterator" loop if we are not directly interested in the index itself. Here we are interested in the index, but for the mod function only.

server/pxf-service/src/test/java/org/greenplum/pxf/service/FragmenterServiceTest.java
```diff
@@ -431,4 +433,42 @@ public void testBeginIterationGSSFailureAfterMaxRetries() throws Throwable {
...
+
+    // ----- TESTS for performance of list traversal -----
+    @Test
+    public void testListTraversalPerformance() throws Throwable {
+
+        // This test makes sure we iterate properly (using an iterator, not the index-based for loop) over a LinkedList
+        // that is returned by a fragmenter when building a list of fragment for a segment.
+        // Tested on MacBookPro, the timings are as follows:
+        // 10M fragments - from  15 mins to 1.3 secs
+        //  1M fragments - from 8.2 secs to 1.3 secs
```
hpandeycodeit, line 445: Is the run time same for 10M and 1M?

hpandeycodeit (approved):

server/pxf-service/src/test/java/org/greenplum/pxf/service/FragmenterServiceTest.java
denalex: yes, pretty much the same within a margin of error with precision of a tenth of a second

PR #742, opened 2022-01-26
Use yum instead of rpm to install GPDB from an RPM package

bradfordb-vmware:
If the runtime container is missing any dependencies required by the
GPDB package, then the rpm command will fail with `error: Failed
dependencies:`. If yum is used to install the package, then missing
dependencies will be resolved and installed.

### DIFF
--- a/concourse/scripts/build.bash
+++ b/concourse/scripts/build.bash
@@ -16,7 +16,7 @@ function install_gpdb() {
     if command -v rpm; then
         pkg_file=$(find "${GPDB_PKG_DIR}" -name "greenplum-db-${GPDB_VERSION}-rhel*-x86_64.rpm")
         echo "Installing RPM ${pkg_file}..."
-        rpm --quiet -ivh "${pkg_file}" >/dev/null
+        yum -y -d1 "${pkg_file}"
     elif command -v apt-get; then
         # apt-get wants a full path
         pkg_file=$(find "${PWD}/${GPDB_PKG_DIR}" -name "greenplum-db-${GPDB_VERSION}-ubuntu18.04-amd64.deb")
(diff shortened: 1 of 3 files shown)


concourse/scripts/build.bash
```diff
@@ -16,7 +16,7 @@ function install_gpdb() {
     if command -v rpm; then
         pkg_file=$(find "${GPDB_PKG_DIR}" -name "greenplum-db-${GPDB_VERSION}-rhel*-x86_64.rpm")
         echo "Installing RPM ${pkg_file}..."
-        rpm --quiet -ivh "${pkg_file}" >/dev/null
+        yum -y -d1 install "${pkg_file}"
```
denalex, line 19: how much of a concern is the fact that we are assuming all dependencies are accounted for in the current docker images and hence `rpm` install was sufficient and would fail if the assumption was not true (as is the case right now). That would've caught missing dependency and we would put it back in the docker (at some level) and document for customer's use. Right now this will not happen as the dependency will be auto-installed. In general, we prefer not to install dips during test phase.

bradfordb-vmware: The source of truth for GPDB's runtime dependencies should be the [RPM spec](https://github.com/greenplum-db/greenplum-database-release/blob/main/ci/concourse/scripts/greenplum-db-7.spec) and not the Dockerfile that builds the test image. Having the packages pre-installed in the test container is an optimization.

I believe the official GPDB documentation recommends using `yum` over `rpm` for GPDB6.

denalex: yes, true, so in the spirit of maintaining the optimization, failing RPM install will signal the need to adjust the optimized docker test image so we avoid extra deps installs during tests.

PR #749, opened 2022-02-07
Tests for Hive with Kerberos Constrained delegation

denalex:
PR from a sub-feature to a feature branch. Introduces `HiveProxyIPASmokeTest` to test impersonated access provided by Kerberos Constrained delegation feature. The pattern is similar to that of  `HdfsProxyIPASmokeTest`.


concourse/scripts/install_multinode_hadoop.bash
```diff
@@ -65,19 +65,23 @@ popd || exit 1
...
 # Hadoop cluster has been successfully configured; now create environment files
 # for downstream CI tasks to configure and run automation tests
-hadoop_namenode="ccp-${cluster_name}-nn01"
+hadoop_namenode_1="ccp-${cluster_name}-nn01"
+hadoop_namenode_2="ccp-${cluster_name}-nn02"
 ipa_server="ccp-${cluster_name}-ipa"
 mkdir -p ipa_env_files
 jq <"${metadata_path}" -r ".etc_hosts" >ipa_env_files/etc_hostfile
-echo "${hadoop_namenode}" >ipa_env_files/name
+echo "${hadoop_namenode_1}" >ipa_env_files/name
```
bradfordb-vmware, line 73: I think when we were originally iterating on this, we tried to match the IPA cluster's environment files to the DataProce clusters' environment files -- thinking we might be able to swap one for the other. I don't think we are still planning to do that, so I wonder if it might be OK to change this to

```sh
echo "${hadoop_namenode_1}" >ipa_env_files/nn01
```

to match the new `nn02` file.

concourse/scripts/install_multinode_hadoop.bash
```diff
@@ -65,19 +65,23 @@ popd || exit 1
...
-scp "${hadoop_namenode}:\$HADOOP_PREFIX/etc/hadoop/*-site.xml" ipa_env_files/conf/
+scp "${hadoop_namenode_1}:\$HADOOP_PREFIX/etc/hadoop/*-site.xml" ipa_env_files/conf/
+scp "${hadoop_namenode_2}:\$HIVE_HOME/conf/hive-site.xml" ipa_env_files/conf/
 
 cp ~/.ssh/"${cluster_name}" ipa_env_files/google_compute_engine
 cp ~/.ssh/"${cluster_name}".pub ipa_env_files/google_compute_engine.pub
 
 gcp_project="$(jq <"${metadata_path}" -r ".project")"
 domain_name="c.${gcp_project}.internal"
+echo "${domain_name}" >ipa_env_files/DOMAIN
```
bradfordb-vmware, line 84: I think most the other filenames are in lowercase. I think `DOMAIN` is in uppercase because of the Kerberos case-sensitivity. Would we want to leave this as lowercase?

bradfordb-vmware (approved):

PR #750, opened 2022-02-07
Add RHEL8 image for running rpmrebuild to add OSL to RPMs

bradfordb-vmware:

### DIFF
--- a/concourse/docker/rpmrebuild/cloudbuild.yaml
+++ b/concourse/docker/rpmrebuild/cloudbuild.yaml
@@ -15,7 +15,19 @@ steps:
     - 'concourse/docker/rpmrebuild/centos'
   waitFor: ['-']
 
+# Builds the rpmrebuild-rhel8 image
+- name: 'gcr.io/cloud-builders/docker'
+  id: rpmrebuild-rhel8-image
+  args:
+    - 'build'
+    - '--build-arg=BASE_IMAGE=${_PRIVATE_BASE_IMAGE_REPOSITORY}/gpdb6-rhel8-test:latest'
+    - '--tag=gcr.io/$PROJECT_ID/rpmrebuild-rhel8:latest'
+    - '-f'
+    - 'concourse/docker/rpmrebuild/rhel/Dockerfile'
+    - 'concourse/docker/rpmrebuild/rhel'
+  waitFor: ['-']
+
 # Push images from Cloud Build to Container Registry
 images:
   - 'gcr.io/$PROJECT_ID/rpmrebuild-centos7:latest'
-
+  - 'gcr.io/$PROJECT_ID/rpmrebuild-rhel8:latest'
--- a/concourse/docker/rpmrebuild/rhel/Dockerfile
+++ b/concourse/docker/rpmrebuild/rhel/Dockerfile
@@ -1,0 +1,16 @@
+ARG BASE_IMAGE=gcr.io/data-gpdb-private-images/gpdb6-rhel8-test:latest
+
+FROM centos:7 as downloader
+
+# download source RPM for rpmrebuild from EPEL repository for EL7
+RUN yum install -y epel-release && yumdownloader --source rpmrebuild
+
+FROM ${BASE_IMAGE}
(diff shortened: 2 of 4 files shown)


concourse/docker/rpmrebuild/rhel/Dockerfile
```diff
@@ -0,0 +1,16 @@
+ARG BASE_IMAGE=gcr.io/data-gpdb-private-images/gpdb6-rhel8-test:latest
```
denalex, line 1: can this use `${_PRIVATE_BASE_IMAGE_REPOSITORY}` and not hardcode the private image repo ?

concourse/docker/rpmrebuild/cloudbuild.yaml
```diff
@@ -15,7 +15,19 @@ steps:
     - 'concourse/docker/rpmrebuild/centos'
   waitFor: ['-']
 
+# Builds the rpmrebuild-rhel8 image
+- name: 'gcr.io/cloud-builders/docker'
+  id: rpmrebuild-rhel8-image
+  args:
+    - 'build'
+    - '--build-arg=BASE_IMAGE=${_PRIVATE_BASE_IMAGE_REPOSITORY}/gpdb6-rhel8-test:latest'
```
denalex, line 23: would this cloud build be triggered when any changes are made to `gpdb6-rhel8-test:latest` ?

concourse/docker/rpmrebuild/rhel/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+
+FROM centos:7 as downloader
+
+# download source RPM for rpmrebuild from EPEL repository for EL7
+RUN yum install -y epel-release && yumdownloader --source rpmrebuild
+
+FROM ${BASE_IMAGE}
+
+COPY --from=downloader /rpmrebuild-*.src.rpm /
```
denalex, line 10: we are assuming the sources for `rpmrebuild` are the same between `Centos7` and `Rhel8`, is this safe ? Will it always be the case ?

concourse/docker/rpmrebuild/rhel/Dockerfile
bradfordb-vmware: We could do something like

```dockerfile
ARG PRIVATE_BASE_IMAGE_REPOSITORY

FROM gcr.io/${PRIVATE_BASE_IMAGE_REPOSITORY}/gpdb6-rhel8-test:latest
```

but that assumes the container registry is in Google Cloud and Docker Hub. The most generic would be to just leave `BASE_IMAGE` unspecified but our existing Dockerfiles do provide default values ([pxf-dev-base/gpdb6/centos7/Dockerfile](https://github.com/greenplum-db/pxf/blob/master/concourse/docker/pxf-dev-base/gpdb6/centos7/Dockerfile#L1-L3) has

```dockerfile
ARG BASE_IMAGE=gcr.io/data-gpdb-public-images/gpdb6-centos7-test:latest

FROM ${BASE_IMAGE}
```

We felt like including the default value was consistent with the existing Dockerfiles.

concourse/docker/rpmrebuild/cloudbuild.yaml
bradfordb-vmware: Currently, no. We'd need to add a job to the Cloud Build pipeline to watch for changes on this upstream image and then "manually" trigger a Cloud Build job.

concourse/docker/rpmrebuild/rhel/Dockerfile
bradfordb-vmware: The source RPM contains the upstream source tarball (from the rpmrebuild repo) that is distribution independent. The source RPM also includes a spec-file for building from the upstream source tarball on RHEL/Fedora. We could modify the `yumdownloader` command a few lines up to request a specific version

```sh
yumdownloader --source rpmrebuild 2.11-3.el7
```

so that  we always get the same version.

denalex (approved):

PR #752, opened 2022-02-07
Add support for GPDB6 on RHEL 8

ashuka24:
This PR adds the infrastructure needed to build and test PXF for GPDB6 on RHEL 8.


concourse/pipelines/templates/build_pipeline-tpl.yml
```diff
@@ -479,6 +528,14 @@ resources:
...
     json_key: ((pxf-storage-service-account-key))
     versioned_file: ((pxf-build-bucket-prefix))/snapshots/pxf-gp[[gp_ver]]-ubuntu18.04.tar.gz
+
+- name: pxf_gp[[gp_ver]]_tarball_rhel8
+  type: gcs
+  icon: google-drive
+  source:
+    bucket: data-gpdb-ud-pxf-build
+    json_key: ((pxf-storage-service-account-key))
+    versioned_file: ((pxf-build-bucket-prefix))/snapshots/pxf-gp[[gp_ver]]-rhel8.04.tar.gz
```
denalex, line 538:
```suggestion
    versioned_file: ((pxf-build-bucket-prefix))/snapshots/pxf-gp[[gp_ver]]-el8.tar.gz
```
this would seem to fit the pattern, is this how it is produced ?

package/pxf-gp6.spec
```diff
@@ -1,6 +1,9 @@
 # Disable repacking of jars, since it takes forever
 %define __jar_repack %{nil}
 
+# Disable build_id links for ELF files on RHEL 8
+%define _build_id_links none
```
denalex, line 5: I assume:
- we do not need build-ids when our RPM installs and RHEL8 will be ok with it
- presence of this definition is still ok for RHEL7 as the spec is used for that as well

denalex (approved):

concourse/pipelines/templates/build_pipeline-tpl.yml
ashuka24: It is pushed up to GCP in a put step and the param for the put does correctly match the `.el8` convention. I will update it accordingly

package/pxf-gp6.spec
bradfordb-vmware: It looks like `build-ids` are use when a package provides binaries and debug info in separate packages (https://fedoraproject.org/wiki/Summer_Coding_2010_ideas_-_Universal_Build-ID#Build-ID_background).

This is defining a macro, for for the RHEL7 RPM point of view its just an unused macro. We could make it conditional on the platform:

```
%if 0%{?rhel} == 8
%define _build_id_links none
%endif
```

but I don't know if that's any clearer.

denalex: thanks, we can leave it as is, no benefit in conditional

concourse/pipelines/templates/dev_build_pipeline-tpl.yml
```diff
@@ -448,6 +477,56 @@ jobs:
...
+    file: pxf_src/concourse/tasks/test.yml
+    image: gpdb[[gp_ver]]-pxf-dev-rhel8-image
+    params:
+      ACCESS_KEY_ID: ((tf-machine-access-key-id))
+      GP_VER: [[gp_ver]]
+      GROUP: gpdb,proxy,profile
+      SECRET_ACCESS_KEY: ((tf-machine-secret-access-key))
+{% set gp_ver = None %}
+
+
```
hpandeycodeit, line 529: nit: remove extra line

concourse/scripts/build.bash
```diff
@@ -62,6 +61,9 @@ function package_pxf() {
 }
 
 install_gpdb
+# installation of GPDB from RPM/DEB doesn't insure that the installation location will match the version
```
hpandeycodeit, line 64:
```suggestion
# installation of GPDB from RPM/DEB doesn't ensure that the installation location will match the version
```

PR #753, opened 2022-02-09
ci: fix up pivnet artifacts

ashuka24:
Existing Pivnet artifacts do not contain a RHEL 8 RPM, so continue using the existing RHEL 8 RPM until we can start automatically pulling down a newer RPM.

### DIFF
--- a/concourse/scripts/get_product_files.bash
+++ b/concourse/scripts/get_product_files.bash
@@ -24,6 +24,7 @@ gpdb_version=$(
 )
 echo -e "Latest - ${VERSIONS_BEFORE_LATEST} GPDB version found:\n${GPDB_VERSION}X:\t${gpdb_version}"
 
+# get the string of product names and turn it into list with correct gpdb_version found above
 IFS=, read -ra product_files_GPDB_VERSION <<<"${LIST_OF_PRODUCTS}"
 for file in "${product_files_GPDB_VERSION[@]}"; do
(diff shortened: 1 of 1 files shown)


concourse/scripts/get_product_files.bash
```diff
@@ -56,9 +63,19 @@ for ((i = 0; i < ${#product_files[@]}; i++)); do
...
 		fi
 	fi
-	rm -f "${product_dirs[$i]}"/*.{rpm,deb}
+	# if the version doesn't exist in either i-1, i-th bucket or the SHAs don't match, download it from pivnet.
 	id=$(jq <<<"${product_files_json}" -r --arg object_key "${file}" '.[] | select(.aws_object_key == $object_key).id')
-	echo "Downloading ${file} with id ${id} to ${product_dirs[$i]}..."
+
+	if [[ -z "${id}" ]]; then
+		echo "Did not find '${file}' in product files for GPDB '${gpdb_version}'"
+		if [[ $file =~ ^.*rhel8.*$ ]]; then
```
bradfordb-vmware, line 71: Based on this [StackOverflow answer](https://stackoverflow.com/a/35924143)

> While not always necessary, it is advisable to store the regex in a variable first, because Bash has trouble with regex literals containing `\`

hpandeycodeit (approved):

bradfordb-vmware (approved):

PR #754, opened 2022-02-11
Add scripts for modifying PXF extension to support gpupgrade

bradfordb-vmware:
Includes a basic automation test for the scripts. The test makes sure
that PXF can still work after running the prepare script and after the
rollback script.

### DIFF
--- a/automation/src/test/java/org/greenplum/pxf/automation/features/gpupgrade/GpupgradeTest.java
+++ b/automation/src/test/java/org/greenplum/pxf/automation/features/gpupgrade/GpupgradeTest.java
@@ -1,0 +1,62 @@
+package org.greenplum.pxf.automation.features.gpupgrade;
+
+import org.greenplum.pxf.automation.BaseFunctionality;
+import org.greenplum.pxf.automation.structures.tables.basic.Table;
+import org.greenplum.pxf.automation.structures.tables.pxf.ReadableExternalTable;
+import org.greenplum.pxf.automation.structures.tables.utils.TableFactory;
+import org.testng.annotations.AfterMethod;
+import org.testng.annotations.Test;
+
+public class GpupgradeTest extends BaseFunctionality {
+
+    public static final String[] FIELDS = {
+            "name    text",
+            "num     integer",
+            "dub     double precision",
+            "longNum bigint",
+            "bool    boolean"
+    };
+
+    private ReadableExternalTable externalTable;
+
+    @Override
+    protected void beforeMethod() throws Exception {
+        super.beforeMethod();
+        String location = prepareData();
+        createReadablePxfTable("default", location);
+    }
+
+    @Override
+    protected void afterMethod() throws Exception {
+        gpdb.dropTable(externalTable, true);
+        super.afterMethod();
+    }
+
+    @Test(groups = {"features", "gpdb"})
+    public void testGpdbUpgradeScenario() throws Exception {
+        runTincTest("pxf.features.gpupgrade.step_1_before_running_pxf_pre_gpupgrade.runTest");
+
+        cluster.runCommand("pxf-pre-gpupgrade");
+        runTincTest("pxf.features.gpupgrade.step_2_after_running_pxf_pre_gpupgrade.runTest");
+
+        cluster.runCommand("pxf-post-gpupgrade");
+        runTincTest("pxf.features.gpupgrade.step_3_after_running_pxf_post_gpupgrade.runTest");
+    }
+
+    private String prepareData() throws Exception {
+        Table smallData = getSmallData("", 10);
+        String location = hdfs.getWorkingDirectory() + "/gpupgrade-test-data.txt";
+        hdfs.writeTableToFile(location, smallData, ",");
+
+        return location;
+    }
+
+    private void createReadablePxfTable(String serverName, String location) throws Exception {
+        externalTable = TableFactory.getPxfReadableTextTable("pxf_gpupgrade_test", FIELDS, location, ",");
+        externalTable.setHost(pxfHost);
+        externalTable.setPort(pxfPort);
+        externalTable.setServer("SERVER=" + serverName);
+        gpdb.createTableAndVerify(externalTable);
+    }
+
+}
--- a/automation/tincrepo/main/pxf/features/gpupgrade/step_1_before_running_pxf_pre_gpupgrade/runTest.py
+++ b/automation/tincrepo/main/pxf/features/gpupgrade/step_1_before_running_pxf_pre_gpupgrade/runTest.py
@@ -1,0 +1,12 @@
+from mpp.models import SQLTestCase
+from mpp.models import SQLConcurrencyTestCase
+
+class PxfGpupgradeStep1BeforeRunningPxfPreGpupgrade(SQLConcurrencyTestCase):
+    """
+    @db_name pxfautomation
+    @concurrency 1
+    @gpdiff True
+    """
+    sql_dir = 'sql'
+    ans_dir = 'expected'
+    out_dir = 'output'
--- a/automation/tincrepo/main/pxf/features/gpupgrade/step_2_after_running_pxf_pre_gpupgrade/runTest.py
+++ b/automation/tincrepo/main/pxf/features/gpupgrade/step_2_after_running_pxf_pre_gpupgrade/runTest.py
@@ -1,0 +1,12 @@
+from mpp.models import SQLTestCase
+from mpp.models import SQLConcurrencyTestCase
+
+class PxfGpupgradeStep2AfterRunningPxfPreGpupgrade(SQLConcurrencyTestCase):
+    """
+    @db_name pxfautomation
+    @concurrency 1
+    @gpdiff True
+    """
+    sql_dir = 'sql'
+    ans_dir = 'expected'
+    out_dir = 'output'
--- a/automation/tincrepo/main/pxf/features/gpupgrade/step_3_after_running_pxf_post_gpupgrade/runTest.py
+++ b/automation/tincrepo/main/pxf/features/gpupgrade/step_3_after_running_pxf_post_gpupgrade/runTest.py
@@ -1,0 +1,12 @@
+from mpp.models import SQLTestCase
+from mpp.models import SQLConcurrencyTestCase
+
+class PxfGpupgradeStep3AfterRunningPxfPostGpupgrade(SQLConcurrencyTestCase):
+    """
+    @db_name pxfautomation
+    @concurrency 1
+    @gpdiff True
+    """
+    sql_dir = 'sql'
+    ans_dir = 'expected'
+    out_dir = 'output'
--- a/external-table/Makefile
+++ b/external-table/Makefile
@@ -27,6 +27,8 @@ stage: pxf.so
 	install -c -m 644 pxf--1.0.sql build/stage/
 	install -c -m 644 pxf--2.0.sql build/stage/
 	install -c -m 644 pxf--1.0--2.0.sql build/stage/
+	@echo "gpdb.version=$(GP_VERSION)" > build/stage/metadata
+	@echo "gpdb.major-version=$(GP_MAJORVERSION)" > build/stage/metadata
 	@echo $(GP_MAJORVERSION) > build/metadata/gp_major_version
 	@echo $(BLD_ARCH) | sed 's/_/-/' > build/metadata/build_arch
 
--- a/server/build.gradle
+++ b/server/build.gradle
@@ -263,6 +263,8 @@ task stage(type: Copy) {
         include 'pxf'
         include 'kill-pxf.sh'
         include 'merge-pxf-config.sh'
+        include 'pxf-pre-gpupgrade'
+        include 'pxf-post-gpupgrade'
         into 'bin'
         filter { line -> line.replaceAll('_PXF_VERSION_', "${version}") }
     }
--- a/server/pxf-service/src/scripts/pxf-post-gpupgrade
+++ b/server/pxf-service/src/scripts/pxf-post-gpupgrade
@@ -1,0 +1,118 @@
+#!/usr/bin/env bash
+
+set -o errexit
+
+# Follow symlinks to find the real script
+cd "$(dirname "$0")" || exit 1
+script_file=$(pwd)/$(basename "$0")
+while [[ -L "$script_file" ]]; do
+	script_file=$(readlink "$script_file")
+	cd "$(dirname "$script_file")" || exit 1
+	script_file=$(pwd)/$(basename "$script_file")
+done
+parent_script_dir="$( (cd "$(dirname "${script_file}")/.." && pwd -P))"
+
+# establish PXF_HOME and global vars used by all commands
+export PXF_HOME=${parent_script_dir}
+
+# Path to PXF_BASE directories, defaults to PXF_HOME
+export PXF_BASE=${PXF_BASE:=$PXF_HOME}
+
+# Path to Log directory
+export PXF_LOGDIR=${PXF_LOGDIR:=${PXF_BASE}/logs}
+
+: "${GPHOME:?GPHOME must be set before running this script}"
+# shellcheck source=/dev/null
+. "${GPHOME}/greenplum_path.sh"
+
+# create a log file with a timestamp in the name
+# example: $PXF_LOGDIR/pxf-post-gpupgrade.20220302135812.log
+log_file="${PXF_LOGDIR}/$(basename "${0}").$(date +%Y%m%d%H%M%S).log"
+touch "${log_file}"
+
+export PGHOST="${PGHOST:-localhost}"
+export PGPORT=${PGPORT:-5432}
+export PGDATABASE="${PGDATABASE:-gpadmin}"
+export PGUSER=${PGUSER:-gpadmin}
+export PGPASSWORD=${PGPASSWORD:-}
+
+cat <<EOF >>"${log_file}"
+Postgres Connection Parameters
+    \$PGHOST='${PGHOST}'
+    \$PGPORT='${PGPORT}'
+    \$PGDATABASE='${PGDATABASE}'
+    \$PGUSER='${PGUSER}'
+    \$PGPASSWORD='${PGPASSWORD//?/*}'
+EOF
+
+pxf_gpdb_major_version="$(awk 'BEGIN { FS = "=" } /gpdb.major-version/{ print $2 }' "${PXF_HOME}/gpextable/metadata")"
+gp_version="$(psql --no-align --tuples-only --command 'SELECT substring(version(), $$.*Greenplum Database (.*) build.*$$)')"
+
+echo "PXF compiled against GPDB major version '${pxf_gpdb_major_version}'" >>"${log_file}"
+echo "Running GPDB cluster is version '${gp_version}'" >>"${log_file}"
+
+if [[ "${pxf_gpdb_major_version}" != "${gp_version%%.*}" ]]; then
+	echo "ERROR: This version of PXF only works with GPDB ${pxf_gpdb_major_version}+ but the targeted GPDB cluster is ${gp_version}" | tee -a "${log_file}"
+	exit 1
+fi
+
+if [[ "${gp_version}" = 5* ]]; then
+	master_data_dir_query="SELECT fselocation FROM pg_catalog.pg_filespace_entry WHERE fsedbid = 1 AND fsefsoid = (SELECT oid FROM pg_catalog.pg_filespace WHERE fsname = 'pg_system')"
+else
+	master_data_dir_query="SELECT datadir FROM pg_catalog.gp_segment_configuration WHERE dbid = 1"
+fi
+export MASTER_DATA_DIRECTORY="${MASTER_DATA_DIRECTORY:-$(psql --no-align --tuples-only --command "${master_data_dir_query}")}"
+echo "GPDB master data directory is '${MASTER_DATA_DIRECTORY}'" >>"${log_file}"
+
+PXF_HOME_REGEX="(.*:)*\/gpextable.*"
+dynamic_library_path="$(gpconfig --show dynamic_library_path | grep 'Master.*value:' | sed -e 's/Master.*value: \(.*\)/\1/')"
+
+if [[ ! "${dynamic_library_path}" =~ $PXF_HOME_REGEX ]]; then
+	echo "GUC 'dynamic_library_path=${dynamic_library_path}' does not contain \$PXF_HOME/gpextable" >>"${log_file}"
+	echo "Skipping removing it from 'dynamic_library_path'" >>"${log_file}"
+else
+	echo "Removing '${PXF_HOME}/gpextable' from 'dynamic_library_path=${dynamic_library_path}'" >>"${log_file}"
+	new_dynamic_library_path="$(echo -n "${dynamic_library_path}" | perl -n -e 'print join(":", grep(!/.*gpextable.*/, split(":", $_)))')"
+	# for GPDB 5, must `\`-escape dollar-sign ($) in the value passed to gpconfig
+	if [[ "${gp_version}" = 5* ]]; then
+		new_dynamic_library_path="${new_dynamic_library_path//$/\\$}"
+	fi
+	{
+		echo "New value for 'dynamic_library_path' is '${new_dynamic_library_path}'"
+		gpconfig --change dynamic_library_path --value "${new_dynamic_library_path}"
+		gpstop -u
+	} &>>"${log_file}"
+fi
+
+echo "Updating PXF function definitions" >>"${log_file}"
+psql --no-align --tuples-only --command "SELECT datname FROM pg_catalog.pg_database WHERE datname != 'template0';" | while read -r dbname; do
+	echo -n "checking if database '${dbname}' has PXF extension installed... " >>"${log_file}"
+	if ! psql --dbname="${dbname}" --no-align --tuples-only --command "SELECT extname FROM pg_catalog.pg_extension WHERE extname = 'pxf'" | grep . &>/dev/null; then
+		echo "skipping database '${dbname}'" >>"${log_file}"
+		continue
+	fi
+	echo "updating PXF extension UDFs in database '${dbname}'" >>"${log_file}"
+	psql --dbname="${dbname}" --set ON_ERROR_STOP=on &>>"${log_file}" <<-END_OF_SQL
+		CREATE OR REPLACE FUNCTION pg_catalog.pxf_write() RETURNS integer
+		AS '${PXF_HOME}/gpextable/pxf', 'pxfprotocol_export'
+		LANGUAGE C STABLE;
+
+		CREATE OR REPLACE FUNCTION pg_catalog.pxf_read() RETURNS integer
+		AS '${PXF_HOME}/gpextable/pxf', 'pxfprotocol_import'
+		LANGUAGE C STABLE;
+
+		CREATE OR REPLACE FUNCTION pg_catalog.pxf_validate() RETURNS void
+		AS '${PXF_HOME}/gpextable/pxf', 'pxfprotocol_validate_urls'
+		LANGUAGE C STABLE;
+
+		CREATE OR REPLACE FUNCTION pg_catalog.pxfwritable_import() RETURNS record
+		AS '${PXF_HOME}/gpextable/pxf', 'gpdbwritableformatter_import'
+		LANGUAGE C STABLE;
+
+		CREATE OR REPLACE FUNCTION pg_catalog.pxfwritable_export(record) RETURNS bytea
+		AS '${PXF_HOME}/gpextable/pxf', 'gpdbwritableformatter_export'
+		LANGUAGE C STABLE;
+	END_OF_SQL
+done
+
+echo "Success" | tee -a "${log_file}"
--- a/automation/tincrepo/main/pxf/features/gpupgrade/step_1_before_running_pxf_pre_gpupgrade/expected/query01.ans
+++ b/automation/tincrepo/main/pxf/features/gpupgrade/step_1_before_running_pxf_pre_gpupgrade/expected/query01.ans
@@ -1,0 +1,45 @@
+-- start_ignore
+-- end_ignore
+-- @description query01 for PXF gpupgrade test on small data
+-- start_matchsubs
+--
+-- m{.*/usr/local/pxf-(dev|gp\d).*}
+-- s{/usr/local/pxf-(dev|gp\d)}{\$PXF_HOME}
+--
+-- end_matchsubs
+SELECT * FROM pxf_gpupgrade_test;
+  name  | num | dub |    longnum    | bool
+--------+-----+-----+---------------+------
+ row_1  |   1 |   1 |  100000000000 | f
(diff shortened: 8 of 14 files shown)


server/pxf-service/src/scripts/prepare-for-gpupgrade.sh
```diff
@@ -0,0 +1,73 @@
...
+SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)
+PXF_HOME=$(cd -- "$(dirname -- "${SCRIPT_DIR}")" &>/dev/null && pwd)
+
+function echoYellow() { echo $'\e[0;33m'"$1"$'\e[0m'; }
+function echoGreen() { echo $'\e[0;32m'"$1"$'\e[0m'; }
+
+: "${GPHOME:?must be set when running this script; source greenplum_path.sh}"
+: "${MASTER_DATA_DIRECTORY:? must be set when running this script}"
+
+kPXF_HOME_REGEX="(.*:)*${PXF_HOME}/gpextable:.*"
```
denalex, line 30: do we care for the exact value of PXF_HOME ? What if it is a symlink ? maybe checking for presence of `/gpexttable` only will be sufficient ?

server/pxf-service/src/scripts/prepare-for-gpupgrade.sh
```diff
@@ -0,0 +1,73 @@
...
+
+dynamic_library_path="$(gpconfig --show dynamic_library_path | grep 'Master.*value:' | sed -e 's/Master.*value: \(.*\)/\1/')"
```
denalex, line 32: are we sure `gpconfig` will be on the PATH ?

server/pxf-service/src/scripts/prepare-for-gpupgrade.sh
```diff
@@ -0,0 +1,73 @@
...
+if [[ $dynamic_library_path =~ $kPXF_HOME_REGEX ]]; then
+	echoYellow "dynamic_library_path already contains \$PXF_HOME/gpextable"
+	echoYellow "Skipping adding '${PXF_HOME}/gpextable' to 'dynamic_library_path'"
+else
+	echoGreen "Adding '${PXF_HOME}/gpextable' to 'dynamic_library_path'"
+	new_dynamic_library_path="${PXF_HOME}/gpextable:${dynamic_library_path}"
+	echoGreen "New value for 'dynamic_library_path' is '${new_dynamic_library_path}'"
+	gpconfig --change dynamic_library_path --value "${new_dynamic_library_path}" &>/dev/null
```
denalex, line 40: wonder if we need to have a log file for these operations to log all detailed messages and also outputs from gp* commands

server/pxf-service/src/scripts/prepare-for-gpupgrade.sh
```diff
@@ -0,0 +1,73 @@
...
+		LANGUAGE C STABLE;
+
+		CREATE OR REPLACE FUNCTION pg_catalog.pxfwritable_import() RETURNS record
+		AS 'pxf', 'gpdbwritableformatter_import'
+		LANGUAGE C STABLE;
+
+		CREATE OR REPLACE FUNCTION pg_catalog.pxfwritable_export(record) RETURNS bytea
+		AS 'pxf', 'gpdbwritableformatter_export'
+		LANGUAGE C STABLE;
+	END_OF_SQL
```
denalex, line 72: would that be better to externalize this section into a file ?

server/pxf-service/src/scripts/prepare-for-gpupgrade.sh
```diff
@@ -0,0 +1,73 @@
...
+done
```
denalex, line 73: Would be good to print out a success message to the user.

server/pxf-service/src/scripts/rollback-gpupgrade-changes.sh
```diff
@@ -0,0 +1,72 @@
...
+
+kPXF_HOME_REGEX="(.*:)*${PXF_HOME}/gpextable:.*"
+
+dynamic_library_path="$(gpconfig --show dynamic_library_path | grep 'Master.*value:' | sed -e 's/Master.*value: \(.*\)/\1/')"
+if [[ ! $dynamic_library_path =~ $kPXF_HOME_REGEX ]]; then
+	echoYellow "dynamic_library_path does not contain \$PXF_HOME/gpextable"
+	echoYellow "Skipping removing it from 'dynamic_library_path'"
+else
+	echoGreen "Removing '${PXF_HOME}/gpextable' from 'dynamic_library_path'"
+	new_dynamic_library_path="$(echo -n "${dynamic_library_path}" | awk -v RS=: -v ORS=: '/pxf-gp[0-9]/ { next } { print }' | sed 's/:*$//')"
```
denalex, line 38: there is no guarantee PXF is installed in a directory containing `pxf-gp*`

server/pxf-service/src/scripts/prepare-for-gpupgrade.sh
```diff
@@ -0,0 +1,73 @@
...
+else
+	echoGreen "Adding '${PXF_HOME}/gpextable' to 'dynamic_library_path'"
+	new_dynamic_library_path="${PXF_HOME}/gpextable:${dynamic_library_path}"
+	echoGreen "New value for 'dynamic_library_path' is '${new_dynamic_library_path}'"
+	gpconfig --change dynamic_library_path --value "${new_dynamic_library_path}" &>/dev/null
+	gpstop -u &>/dev/null
+fi
+
+echoGreen "Updating PXF function definitions"
+psql --no-align --tuples-only --command "SELECT datname FROM pg_catalog.pg_database WHERE datname != 'template0' AND datname != 'template1';" postgres | while read -r dbname; do
```
denalex, line 45: what if someone managed to install the extension into `template0` or `template1` ?

automation/src/test/java/org/greenplum/pxf/automation/features/gpupgrade/GpupgradeTest.java
```diff
@@ -0,0 +1,62 @@
...
+    @Override
+    public void runAfterMethod() throws Exception {
+        gpdb.dropTable(externalTable, true);
+    }
+
+    @Test(groups = {"features", "gpdb"})
+    public void testGpdbUpgradeScenario() throws Exception {
+        runTincTest("pxf.features.gpupgrade.step_1_pre_prepare.runTest");
+
+        cluster.runCommand("source $GPHOME/greenplum_path.sh");
```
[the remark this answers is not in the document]
bradfordb-vmware: That is my understanding -- `cluster.runCommand()` runs the given command in a long-lived `bash` session and subsequent commands are run in the same shell.

concourse/scripts/test.bash
```diff
@@ -93,6 +93,7 @@ function run_pxf_automation() {
 		export GPHD_ROOT=${GPHD_ROOT}
 		export PXF_HOME=${PXF_HOME}
 		export PGPORT=${PGPORT}
+		export MASTER_DATA_DIRECTORY=\${HOME}/data/master/gpseg-1
```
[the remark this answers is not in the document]
bradfordb-vmware: In order for `gpstop -u` to be run, we either need to have the environment variable `MASTER_DATA_DIRECTORY` set or pass it explicitly to `gpstop -u -d <master-data-directory>`.

I can see if there is a way to determine it automatically.

server/build.gradle
```diff
@@ -260,6 +260,12 @@ task stage(type: Copy) {
         into 'bin'
         filter { line -> line.replaceAll('_PXF_VERSION_', "${version}") }
     }
+    from('pxf-service/src/scripts/') {
+        include 'prepare-for-gpupgrade.sh'
+        include 'rollback-gpupgrade-changes.sh'
```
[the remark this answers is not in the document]
bradfordb-vmware: I think that is a good idea; let me check with the `gpupgrade` team since this is part of our "API" with them.

server/pxf-service/src/scripts/prepare-for-gpupgrade.sh
bradfordb-vmware: I was actually thinking about that when writing this script. Would we want to just redirect STDOUT/STDERR to `$PXF_LOG_DIR/<script-name>.log`?

server/pxf-service/src/scripts/prepare-for-gpupgrade.sh
bradfordb-vmware: That can be done. I also wanted to see if we could combine the different SQL queries (listing DBs, checking for the extension, recreating the UDFs) into a single PL/pgSQL block.

server/pxf-service/src/scripts/rollback-gpupgrade-changes.sh
bradfordb-vmware: True - I'll update this to look for the value of `$PXF_HOME` instead of the hard-coded `pxf-gp*`.

server/pxf-service/src/scripts/prepare-for-gpupgrade.sh
bradfordb-vmware: That's a good question that I might need some help with. Including `template1` won't be a problem, but I wasn't able to connect to `template0`. Let me follow-up with the server team.

server/build.gradle
```diff
@@ -260,6 +260,12 @@ task stage(type: Copy) {
...
+        into 'sbin'
```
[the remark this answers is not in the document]
bradfordb-vmware: The reason I didn't want to put the scripts in `$PXF_HOME/bin` is that we document/recommend users add `$PXF_HOME/bin` to their `$PATH`. That would mean that these scripts are also always on the `$PATH` even though they may only bee needed once. Additionally, the scripts are named very generically (e.g., `before-gpupgrade.sh`) and could conflict with other extensions' scripts.

I am not aware of any issue with adding a new `sbin` subdir to `$PXF_HOME` but I might be missing some context.

server/pxf-service/src/scripts/prepare-for-gpupgrade.sh
```diff
@@ -0,0 +1,73 @@
...
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+set -o errexit
+
+SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)
+PXF_HOME=$(cd -- "$(dirname -- "${SCRIPT_DIR}")" &>/dev/null && pwd)
```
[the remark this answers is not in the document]
bradfordb-vmware: My understanding is that it will work as long as the last component is not a symlink; I can reuse the logic from the `pxf` script since that handles that case as well.

server/pxf-service/src/scripts/prepare-for-gpupgrade.sh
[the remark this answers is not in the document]
bradfordb-vmware: The `k` prefix is to indicate that this variable is a constant and not a dynamic value. This was a recommendation I read while reading up on using Bash's regex match.

server/pxf-service/src/scripts/prepare-for-gpupgrade.sh
bradfordb-vmware: In general no. We need users to have source `$GPHOME/greenplum_path.sh`. Since we are checking if `$GPHOME` is set, we could just include an explicit `. $GPHOME/greenplum_path.sh` at the beginning of the script. We would be able to test for this by removing the explicit `source $GPHOME/greenplum_path.sh` in the automation test case.

server/pxf-service/src/scripts/prepare-for-gpupgrade.sh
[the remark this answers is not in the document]
denalex: seems like this is more of a coding style issue and its the first time I even see it, so it is not very conventional either. IMO it distracts the reader (looks like a typo), rather than helps, I'd say we do not use it, it does not bring any value

kalensk:
For the BASH files the various psql commands might benefit from adding `-v ON_ERROR_STOP=1 ` to stop if there are any errors, etc. Some addition [fun reading](https://www.manniwood.com/postgresql_and_bash_stuff/) perhaps.

server/pxf-service/src/scripts/pxf-before-gpupgrade
```diff
@@ -0,0 +1,91 @@
...
+export PXF_BASE=${PXF_BASE:=$PXF_HOME}
+
+# Path to Log directory
+export PXF_LOGDIR=${PXF_LOGDIR:=${PXF_BASE}/logs}
+
+: "${GPHOME:?must be set when running this script; source greenplum_path.sh}"
+# shellcheck source=/dev/null
+. "${GPHOME}/greenplum_path.sh"
+
+export MASTER_DATA_DIRECTORY="${MASTER_DATA_DIRECTORY:-$(psql --no-align --tuples-only --command 'SELECT datadir FROM pg_catalog.gp_segment_configuration WHERE dbid = 1')}"
```
bhuvnesh2703, line 30: would it also run on gpdb5? (datadir column doesn't exist in gp_segment_configuration)

server/pxf-service/src/scripts/pxf-before-gpupgrade
```diff
@@ -0,0 +1,91 @@
...
+
+dynamic_library_path="$(gpconfig --show dynamic_library_path | grep 'Master.*value:' | sed -e 's/Master.*value: \(.*\)/\1/')"
+pxf_gpdb_major_version="$(awk 'BEGIN { FS = "=" } /gpdb.major-version/{ print $2 }' "${PXF_HOME}/gpextable/metadata")"
+gp_version="$(psql --no-align --tuples-only --command 'SELECT substring(version(), $$.*Greenplum Database (.*) build.*$$)')"
+
+if [[ "${pxf_gpdb_major_version}" != "${gp_version%%.*}" ]]; then
+	echo "This version of PXF only works with GPDB ${pxf_gpdb_major_version}+ but the targeted GPDB cluster is ${gp_version}"
+fi
+
+if [[ $dynamic_library_path =~ $PXF_HOME_REGEX ]]; then
```
bhuvnesh2703, line 44: should it be in double quotes to handle spaces etc, "$dynamic_library_path"

server/pxf-service/src/scripts/pxf-before-gpupgrade
```diff
@@ -0,0 +1,91 @@
...
+	echo "GUC 'dynamic_library_path' already contains \$PXF_HOME/gpextable" >"${log_file}"
+	echo "Skipping adding '${PXF_HOME}/gpextable' to 'dynamic_library_path'" >"${log_file}"
```
bhuvnesh2703, line 46: is it overwriting the log_file content.

bradfordb-vmware: Good catch. I've updated the logging to use appending.

server/pxf-service/src/scripts/pxf-after-gpupgrade
```diff
@@ -0,0 +1,101 @@
...
+
+# Path to PXF_BASE directories, defaults to PXF_HOME
+export PXF_BASE=${PXF_BASE:=$PXF_HOME}
+
+# Path to Log directory
+export PXF_LOGDIR=${PXF_LOGDIR:=${PXF_BASE}/logs}
+
+: "${GPHOME:?must be set when running this script; source greenplum_path.sh}"
+# shellcheck source=/dev/null
+. "${GPHOME}/greenplum_path.sh"
```
denalex, line 26: should we check if the file exists first in case user provided incorrect GPHOME value ? Or the error in that case will be descriptive enough ?

server/pxf-service/src/scripts/pxf-after-gpupgrade
```diff
@@ -0,0 +1,101 @@
...
+
+log_file="${PXF_LOGDIR}/$(basename "${0}").$(date +%s).log"
```
denalex, line 28: might be helpful to comment with an example of what the expected log file name will be

server/pxf-service/src/scripts/pxf-after-gpupgrade
```diff
@@ -0,0 +1,101 @@
...
+touch "${log_file}"
+
+export PGHOST="${PGHOST:-localhost}"
+export PGPORT=${PGPORT:-5432}
+export PGDATABASE="${PGDATABASE:-gpadmin}"
+export PGUSER=${PGUSER:-gpadmin}
+export PGPASSWORD=${PGPASSWORD:-}
```
denalex, line 35: should we log all this values into a log file for troubleshooting ?

server/pxf-service/src/scripts/pxf-after-gpupgrade
```diff
@@ -0,0 +1,101 @@
...
+if [[ "${pxf_gpdb_major_version}" != "${gp_version%%.*}" ]]; then
+	echo "ERROR: This version of PXF only works with GPDB ${pxf_gpdb_major_version}+ but the targeted GPDB cluster is ${gp_version}"
+	exit 1
+fi
+
+case "${gp_version}" in
+5*) master_data_dir_query="SELECT fselocation FROM pg_catalog.pg_filespace_entry WHERE fsedbid = 1 AND fsefsoid = (SELECT oid FROM pg_catalog.pg_filespace WHERE fsname = 'pg_system')" ;;
+*) master_data_dir_query="SELECT datadir FROM pg_catalog.gp_segment_configuration WHERE dbid = 1" ;;
+esac
+export MASTER_DATA_DIRECTORY="${MASTER_DATA_DIRECTORY:-$(psql --no-align --tuples-only --command "${master_data_dir_query}")}"
```
denalex, line 49: should we log this value into a log file for troubleshooting ?

server/pxf-service/src/scripts/pxf-after-gpupgrade
```diff
@@ -0,0 +1,101 @@
...
+
+PXF_HOME_REGEX="(.*:)*${PXF_HOME}/gpextable:.*"
+dynamic_library_path="$(gpconfig --show dynamic_library_path | grep 'Master.*value:' | sed -e 's/Master.*value: \(.*\)/\1/')"
+
+if [[ ! "${dynamic_library_path}" =~ $PXF_HOME_REGEX ]]; then
+	echo "GUC 'dynamic_library_path' does not contain \$PXF_HOME/gpextable" >>"${log_file}"
```
denalex, line 55:
```suggestion
	echo "GUC 'dynamic_library_path'=${dynamic_library_path} does not contain \$PXF_HOME/gpextable" >>"${log_file}"
```

server/pxf-service/src/scripts/pxf-after-gpupgrade
```diff
@@ -0,0 +1,101 @@
...
+	echo "Skipping removing it from 'dynamic_library_path'" >>"${log_file}"
+else
+	echo "Removing '${PXF_HOME}/gpextable' from 'dynamic_library_path'" >>"${log_file}"
```
denalex, line 58:
```suggestion
	echo "Removing '${PXF_HOME}/gpextable' from 'dynamic_library_path'=${dynamic_library_path}" >>"${log_file}"
```

server/pxf-service/src/scripts/pxf-after-gpupgrade
```diff
@@ -0,0 +1,101 @@
...
+	new_dynamic_library_path="$(echo -n "${dynamic_library_path}" | perl -n -e 'print join(":", grep(!/.*gpextable.*/, split(":", $_)))')"
```
denalex, line 59: are we guaranteed `perl` is present on the system and has the sufficient version ?

denalex, line 59: this only looks for presence of `gpextable`, which is ok, so maybe relax PXF_HOME_REGEX to look only for this as well and not include `PXF_HOME` there to avoid potential symlink problems

server/pxf-service/src/scripts/pxf-after-gpupgrade
```diff
@@ -0,0 +1,101 @@
...
+	echo "New value for 'dynamic_library_path' is '${new_dynamic_library_path}'" >>"${log_file}"
+	# for GPDB 5, must `\`-escape dollar-sign ($) in the value passed to gpconfig
+	case "${gp_version}" in
```
denalex, line 62: why `case` with 1 option when `if` would do ?

server/pxf-service/src/scripts/pxf-after-gpupgrade
```diff
@@ -0,0 +1,101 @@
...
+PXF_HOME_REGEX="(.*:)*${PXF_HOME}/gpextable:.*"
+dynamic_library_path="$(gpconfig --show dynamic_library_path | grep 'Master.*value:' | sed -e 's/Master.*value: \(.*\)/\1/')"
+
+if [[ ! "${dynamic_library_path}" =~ $PXF_HOME_REGEX ]]; then
+	echo "GUC 'dynamic_library_path' does not contain \$PXF_HOME/gpextable" >>"${log_file}"
+	echo "Skipping removing it from 'dynamic_library_path'" >>"${log_file}"
+else
+	echo "Removing '${PXF_HOME}/gpextable' from 'dynamic_library_path'" >>"${log_file}"
+	new_dynamic_library_path="$(echo -n "${dynamic_library_path}" | perl -n -e 'print join(":", grep(!/.*gpextable.*/, split(":", $_)))')"
+	echo "New value for 'dynamic_library_path' is '${new_dynamic_library_path}'" >>"${log_file}"
```
denalex, line 60: should this be logged after `case` since the value can change ? Also, I'd add in the beginning "Updating .. to .." to indicate we're about to make a change

server/pxf-service/src/scripts/pxf-after-gpupgrade
```diff
@@ -0,0 +1,101 @@
...
+	# for GPDB 5, must `\`-escape dollar-sign ($) in the value passed to gpconfig
+	case "${gp_version}" in
+	5*) new_dynamic_library_path="${new_dynamic_library_path//$/\\$}" ;;
+	*) ;;
+	esac
+	gpconfig --change dynamic_library_path --value "${new_dynamic_library_path}" &>"${log_file}"
```
denalex, line 66: this will append and not overwrite the log file, right ?

server/pxf-service/src/scripts/pxf-after-gpupgrade
```diff
@@ -0,0 +1,101 @@
...
+		AS '${PXF_HOME}/gpextable/pxf', 'gpdbwritableformatter_import'
+		LANGUAGE C STABLE;
+
+		CREATE OR REPLACE FUNCTION pg_catalog.pxfwritable_export(record) RETURNS bytea
+		AS '${PXF_HOME}/gpextable/pxf', 'gpdbwritableformatter_export'
+		LANGUAGE C STABLE;
+	END_OF_SQL
+done
+
+echo "Success"
```
denalex, line 101:
```suggestion
echo "Success" | tee "${log_file}"
```

server/pxf-service/src/scripts/pxf-before-gpupgrade
```diff
@@ -0,0 +1,101 @@
...
+
+log_file="${PXF_LOGDIR}/$(basename "${0}").$(date +%s).log"
+touch "${log_file}"
+
+export PGHOST="${PGHOST:-localhost}"
+export PGPORT=${PGPORT:-5432}
+export PGDATABASE="${PGDATABASE:-gpadmin}"
+export PGUSER=${PGUSER:-gpadmin}
+export PGPASSWORD=${PGPASSWORD:-}
+
```
denalex, line 36: most of the comments to the file above apply here as well

server/pxf-service/src/scripts/pxf-after-gpupgrade
```diff
@@ -0,0 +1,101 @@
...
+# establish PXF_HOME and global vars used by all commands
+export PXF_HOME=${parent_script_dir}
+
+# Path to PXF_BASE directories, defaults to PXF_HOME
+export PXF_BASE=${PXF_BASE:=$PXF_HOME}
+
+# Path to Log directory
+export PXF_LOGDIR=${PXF_LOGDIR:=${PXF_BASE}/logs}
+
+: "${GPHOME:?must be set when running this script; source greenplum_path.sh}"
```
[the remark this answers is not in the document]
bradfordb-vmware: I should update this error message -- we just need `$GPHOME` to be set and then this script will source `greenplum_path.sh` so that we can be sure that Greenplum-specific Python environment variables are set since we rely on `gpconfig`.

server/pxf-service/src/scripts/pxf-after-gpupgrade
```diff
@@ -0,0 +1,101 @@
...
+# shellcheck source=/dev/null
```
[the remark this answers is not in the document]
bradfordb-vmware: [`shellcheck`](https://github.com/koalaman/shellcheck) will try to follow sourced files and since it can't find/follow this, it will normally print a warning. This silences that one warning for the next line.

server/pxf-service/src/scripts/pxf-after-gpupgrade
bradfordb-vmware: Perl is a required dependency of both GPDB 5 and 6.

server/pxf-service/src/scripts/pxf-after-gpupgrade
bradfordb-vmware: Should I update all of the redirects to use `<cmd> | tee "${log_file}"` so that the user can see script making progress? Or do you think it's better to have a few high-level `echo <msg> | tee "${log_file}"`?

denalex: not all please, ideal is no more than 5 statements (if any) should be printed, Success / Failure should be the last one

kalensk:
Let's use the terms pre/post for the script names and throughout as those are the "official" terms that gpupgrade uses throughout its docs and such.

server/pxf-service/src/scripts/pxf-post-gpupgrade
```diff
@@ -0,0 +1,118 @@
...
+export PXF_HOME=${parent_script_dir}
+
+# Path to PXF_BASE directories, defaults to PXF_HOME
+export PXF_BASE=${PXF_BASE:=$PXF_HOME}
+
+# Path to Log directory
+export PXF_LOGDIR=${PXF_LOGDIR:=${PXF_BASE}/logs}
+
+: "${GPHOME:?GPHOME must be set before running this script}"
+# shellcheck source=/dev/null
```
ashuka24, line 25: is this still needed here?

external-table/Makefile
```diff
@@ -27,6 +27,8 @@ stage: pxf.so
 	install -c -m 644 pxf--1.0.sql build/stage/
 	install -c -m 644 pxf--2.0.sql build/stage/
 	install -c -m 644 pxf--1.0--2.0.sql build/stage/
+	@echo "gpdb.version=$(GP_VERSION)" > build/stage/metadata
```
ashuka24, line 30: I don't see where this is being used. Do we still need this in the metadata?

automation/tincrepo/main/pxf/features/gpupgrade/step_2_after_running_pxf_pre_gpupgrade/runTest.py
```diff
@@ -0,0 +1,12 @@
+from mpp.models import SQLTestCase
+from mpp.models import SQLConcurrencyTestCase
+
+class PxfGpupgradeStep2AfterRunningPxfBeforeGpupgrade(SQLConcurrencyTestCase):
```
denalex, line 4: nit: 
```suggestion
class PxfGpupgradeStep2AfterRunningPxfPreGpupgrade(SQLConcurrencyTestCase):
```

automation/tincrepo/main/pxf/features/gpupgrade/step_1_before_running_pxf_pre_gpupgrade/runTest.py
```diff
@@ -0,0 +1,12 @@
+from mpp.models import SQLTestCase
+from mpp.models import SQLConcurrencyTestCase
+
+class PxfGpupgradeStep1BeforeRunningPxfBeforeGpupgrade(SQLConcurrencyTestCase):
```
denalex, line 4: nit: 
```suggestion
class PxfGpupgradeStep1BeforeRunningPxfPreGpupgrade(SQLConcurrencyTestCase):
```

automation/tincrepo/main/pxf/features/gpupgrade/step_3_after_running_pxf_post_gpupgrade/runTest.py
```diff
@@ -0,0 +1,12 @@
+from mpp.models import SQLTestCase
+from mpp.models import SQLConcurrencyTestCase
+
+class PxfGpupgradeStep3AfterRunningPxfAfterGpupgrade(SQLConcurrencyTestCase):
```
denalex, line 4: nit: 
```suggestion
class PxfGpupgradeStep3AfterRunningPxfPostGpupgrade(SQLConcurrencyTestCase):
```

server/pxf-service/src/scripts/pxf-post-gpupgrade
```diff
@@ -0,0 +1,118 @@
...
+. "${GPHOME}/greenplum_path.sh"
+
+# create a log file with a Unix epoch timestamp in the name
+# example: $PXF_LOGDIR/pxf-post-gpupgrade.1645739333.log
```
denalex, line 29: nit: would it be better to have user readable date to a second precision like `20220302165512` (4:55:12 pm on Mar 2, 2022) ? only if there is a handy format to `date` command that achieves this

denalex (approved):

external-table/Makefile
```diff
@@ -27,6 +27,8 @@ stage: pxf.so
...
+	@echo "gpdb.major-version=$(GP_MAJORVERSION)" > build/stage/metadata
```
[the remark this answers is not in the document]
bradfordb-vmware: The file `build/metadata/gp_major_version` isn't included in the RPM package and is only used during build time. We now need this information at runtime. Additionally, `build/metadata/gp_major_version` contains just the major version while this new file is a properties file (i.e., `key=value`) so that it is not opaque.

server/pxf-service/src/scripts/pxf-post-gpupgrade
bradfordb-vmware: Yes. The scripts are source `greenplum_path.sh` and this comment is a directive to `shellcheck` to no follow this path and try to also check `greenplum_path.sh`. Since `bash` sees it as a comment, it has no runtime impact.

external-table/Makefile
bradfordb-vmware: We are not currently using it but including it for future use and to potentially help with debugging any reported issues.

ashuka24 (approved):

PR #756, opened 2022-02-15
Fixed HiveDataFragmenter not closing connections to Hive Metastore

denalex:


server/pxf-hive/src/main/java/org/greenplum/pxf/plugins/hive/HiveDataFragmenter.java
```diff
@@ -94,10 +93,11 @@
 
     private static final TreeTraverser TRAVERSER = new TreeTraverser();
 
-    private IMetaStoreClient client;
-    private final HiveClientWrapper hiveClientWrapper;
+    protected final Logger LOG = LoggerFactory.getLogger(getClass());
```
bradfordb-vmware, line 96: Is there a reason we removed the `static` keyword here? I was under the impression that static loggers are generally better.

**Edit:** It looks like we made the `Logger` non-static so that we could use `getClass()`. Doing a quick grep of the rest of the code, it looks like there is a mix of static and non-static loggers being used. Should we prefer one over the other?

server/pxf-hive/src/main/java/org/greenplum/pxf/plugins/hive/HiveDataFragmenter.java
```diff
@@ -127,15 +126,34 @@ public void afterPropertiesSet() {
     @Override
     public List<Fragment> getFragments() throws Exception {
         Metadata.Item tblDesc = hiveClientWrapper.extractTableFromName(context.getDataSource());
-        fetchTableMetaData(tblDesc);
+        IMetaStoreClient client = hiveClientWrapper.initHiveClient(this.context, configuration);
```
bradfordb-vmware, line 129: This might be over-engineering, but would we want to consider wrapping the client in `AutoClosable` so that we can use a `try-with-resource` here (and in `getFragmentStats()`)?

server/pxf-hive/src/test/java/org/greenplum/pxf/plugins/hive/HiveDataFragmenterTest.java
```diff
@@ -20,46 +20,89 @@
...
+        context.setDataSource("sometable");
         context.setServerName("default");
         context.setUser("dummy");
         context.setConfiguration(configuration);
     }
 
     @Test
-    public void constructorCantAccessMetaStore() {
-        when(hiveClientWrapper.initHiveClient(context, configuration)).thenThrow(new RuntimeException("Failed connecting to Hive MetaStore service: which way to albuquerque"));
+    public void failsToInitHiveClient() {
```
bradfordb-vmware, line 67: I am not 100% sure I understand what behavior this test is covering. It looks like it its verifying that when we call a function that raises an error, we don't catch or modify that exception. Is that correct?

bradfordb-vmware (approved):

server/pxf-hive/src/main/java/org/greenplum/pxf/plugins/hive/HiveDataFragmenter.java
denalex: We use `private static` when it is just a single class since we don't usually need to create an instance of a logger for each instance of the class. We use `protected non-static` in the base class of the inheritance hierarchy to provide a logger to all subclasses while preserving the fact that the logger should be named with the classname of the actual subclass. Yes, there's an overhead there since we now create an instance of the logger for each instance of the class, but that overhead is considered negligible.

server/pxf-hive/src/main/java/org/greenplum/pxf/plugins/hive/HiveDataFragmenter.java
denalex: thought about it, but `IMetaStoreClient` does not implement `AutoCloseable` or `Closeable` despite having a `close()` method. But we can definitely wrap it, if that's a pattern we want to follow. Let me try doing this.

server/pxf-hive/src/test/java/org/greenplum/pxf/plugins/hive/HiveDataFragmenterTest.java
denalex: this was mostly port of an existing test. now it basically verifies that if we can't init HiveClient we will throw the exception that caused it from the `getFragments()` method.

server/pxf-hive/src/main/java/org/greenplum/pxf/plugins/hive/HiveDataFragmenter.java
```diff
@@ -127,15 +126,18 @@ public void afterPropertiesSet() {
     @Override
     public List<Fragment> getFragments() throws Exception {
         Metadata.Item tblDesc = hiveClientWrapper.extractTableFromName(context.getDataSource());
-        fetchTableMetaData(tblDesc);
+
+        try (HiveClientWrapper.MetaStoreClientHolder clientHolder = hiveClientWrapper.initHiveClient(context, configuration)) {
```
ashuka24, line 130: why is it ok to now init the hive client when getting the fragments instead of during property setting?

denalex: The only purpose of this Fragmenter is to get fragments or stats when requested, those are the only 2 public methods. `afterPropertySet` is an "extended constructor" and creating `HiveClient` there was premature, since we only expect 1 call per Fragmenter instance lifetime and even not that, only 1 Fragmenter instance was actually called per query, others would get the values from the cache, so we were creating clients, not using them and not closing them. Not we obtain the client only when needed and close it immediately when the function is over.

ashuka24 (approved):

PR #764, opened 2022-02-23
Add job to the certification pipeline for RHEL8

ashuka24:


denalex (approved):

concourse/pipelines/certification_pipeline.yml
```diff
@@ -230,6 +255,38 @@ jobs:
       GP_VER: 6
       PXF_CERTIFICATION_FOLDER: data-gpdb-ud-pxf-build/prod/certifications
 
+## ---------- RHEL 8 Swimlane ----------
+- name: Certify GPDB-6 with PXF-GP6 on RHEL8
+  plan:
+    - in_parallel:
```
bradfordb-vmware, line 261: Nit: YAML array indentation. I generally prefer (I think we are mostly consistent) to not indent arrays:

```yaml
plan:
- in_parallel:
    steps:
    - get:
    - get:
- task:
- task:
```

Additionally, I think that we should be using  `in_parallel_config` going forward instead of passing the array of steps directly

```yaml
in_parallel:
  steps:
  - get:
  - get
```

instead of

```yaml
in_parallel:
- get:
- get:
```

bradfordb-vmware (approved):

PR #768, opened 2022-02-24
Add support for RHEL8 in the perf pipeline

ashuka24:
Refactor the perf pipeline to allow for perf testing on RHEL 8

### DIFF
--- a/concourse/Makefile
+++ b/concourse/Makefile
@@ -9,10 +9,11 @@ BUILD_PIPELINE_NAME         ?= pxf-build
 DEV_BUILD_PIPELINE_NAME     ?= dev:$(USER)-$(BRANCH)
 PR_BUILD_PIPELINE_NAME      ?= pxf_pr_pipeline
 CERTIFICATION_PIPELINE_NAME ?= pxf-certification
-PERF_PIPELINE_NAME          ?= pxf_perf-$(SCALE)g
+PERF_PIPELINE_NAME          ?= pxf_perf-$(SCALE)g-RHEL$(REDHAT_MAJOR_VERSION)
 PIVNET_PIPELINE_NAME        ?= pivnet_artifacts
 NUM_GPDB5_VERSIONS          ?= 10
 NUM_GPDB6_VERSIONS          ?=  9
+REDHAT_MAJOR_VERSION        ?= 7
 FLY_CMD                      = /usr/local/bin/fly
 TEMPLATE_CMD                 = ./template_tool
 FLY_OPTION_NON-INTERACTIVE   =
@@ -248,10 +249,13 @@ perf:
 		echo 'Specify the SCALE for the test (i.e make SCALE=10 perf). Allowed SCALE values are 10, 50, 500'; \
 		exit 1; \
 	fi
+	@PIPELINE_FILE=$$(mktemp) && \
+	$(TEMPLATE_CMD) --template perf_pipeline-tpl.yml --vars \
+		redhat_major_version=$(REDHAT_MAJOR_VERSION) >"$${PIPELINE_FILE}" && \
 	$(FLY_CMD) --target=$(CONCOURSE) \
 		set-pipeline \
 		--check-creds \
-		--config $(HOME)/workspace/pxf/concourse/pipelines/perf_pipeline.yml \
+		--config "$${PIPELINE_FILE}" \
 		--load-vars-from=$(HOME)/workspace/pxf/concourse/settings/perf-settings-$(SCALE)g.yml \
 		--load-vars-from=$(SECRETS_HOME)/gpdb_common-ci-secrets.yml \
 		--load-vars-from=$(SECRETS_HOME)/pxf-secrets.yml \
--- a/concourse/README.md
+++ b/concourse/README.md
@@ -71,6 +71,10 @@ make SCALE=50 -C "${HOME}/workspace/pxf/concourse" perf
 make SCALE=500 -C "${HOME}/workspace/pxf/concourse" perf
 ```
 
+By default, these pipelines run perf on RHEL7.
+If you would like to run pipelines using RHEL8, please include `REDHAT_MAJOR_VERSION=8` to the command.
+Ex: `make SCALE=10 REDHAT_MAJOR_VERSION=8 -C "${HOME}/workspace/pxf/concourse" perf`
+
 # Deploy development PXF pipelines
 
 The dev pipeline is an abbreviated version of the `pxf-build` pipeline.
--- a/concourse/pipelines/templates/perf_pipeline-tpl.yml
+++ b/concourse/pipelines/templates/perf_pipeline-tpl.yml
@@ -105,15 +105,16 @@ resources:
     branch: {{pxf-git-branch}}
     uri: {{pxf-git-remote}}
 
-- name: gpdb6_rhel7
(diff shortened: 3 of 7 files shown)


concourse/pipelines/templates/perf_pipeline-tpl.yml
```diff
@@ -121,6 +122,18 @@ resources:
     tag: latest
     username: _json_key
     password: ((pxf-cloudbuild-service-account-key))
+{% endif %}
+
+{% if redhat_major_version == 8 %}
```
bradfordb-vmware, line 127:
```suggestion
{% elif redhat_major_version == 8 %}
```

[Jinja Documentation - If](https://jinja.palletsprojects.com/en/3.0.x/templates/#if)

concourse/scripts/install_pxf.bash
```diff
@@ -13,6 +15,8 @@ else
 	PXF_HOME=${GPHOME}/pxf
 fi
 PXF_VERSION=${PXF_VERSION:=6}
+# for centos7, there exists a ~centos and in rhel8, there exists a ~rhel
+OS_VERSION=$( echo $PLATFORM | tr -cd '[:alpha:].-')
```
bradfordb-vmware, line 19: Looking at how this variable is used, I think a better name might `DEFAULT_OS_USER`. I think what through me for a loop was that `~centos` is a short-hand for a path (`${HOME}/centos`) and I was expecting this to be looking for files.

concourse/scripts/install_pxf.bash
```diff
@@ -13,6 +15,8 @@ else
 	PXF_HOME=${GPHOME}/pxf
 fi
 PXF_VERSION=${PXF_VERSION:=6}
+# for centos7, there exists a ~centos and in rhel8, there exists a ~rhel
+DEFAULT_OS_USER=$( echo $PLATFORM | tr -cd '[:alpha:].-')
```
bradfordb-vmware, line 19: I would probably write this as either

```sh
DEFAULT_OS_USER="${PLATFORM//[[:digit:]]/}"
```

which does the same thing as `tr` but without spawning a subshell.

bradfordb-vmware (approved):

concourse/scripts/install_pxf.bash
```diff
@@ -131,8 +135,8 @@ function create_pxf_installer_scripts() {
 		      sed -i -e "s|gpadmin/_HOST@EXAMPLE.COM|gpadmin@${REALM}|g" ${BASE_DIR}/servers/default/pxf-site.xml
 		      gpscp -f ~gpadmin/hostfile_all -v -r -u gpadmin ~/dataproc_env_files/pxf.service.keytab =:${BASE_DIR}/keytabs/
 		    fi
-		    gpscp -f ~gpadmin/hostfile_all -v -r -u centos /tmp/krb5.conf =:/tmp/krb5.conf
-		    gpssh -f ~gpadmin/hostfile_all -v -u centos -s -e 'sudo mv /tmp/krb5.conf /etc/krb5.conf'
+		    gpscp -f ~gpadmin/hostfile_all -v -r -u ${DEFAULT_OS_USER} /tmp/krb5.conf =:/tmp/krb5.conf
```
hpandeycodeit, line 138: `DEFAULT_OS_USER` is the OS or the User?

concourse/README.md
```diff
@@ -71,6 +71,10 @@ make SCALE=50 -C "${HOME}/workspace/pxf/concourse" perf
 make SCALE=500 -C "${HOME}/workspace/pxf/concourse" perf
 ```
 
+These pipelines default running perf on RHEL7.
```
hpandeycodeit, line 74:
```suggestion
These pipelines by default running perf on RHEL7.
```

concourse/scripts/install_pxf.bash
bradfordb-vmware: Each of the Google Cloud VMs have two users by default: 1) an OS-specific user with `sudo`-privileges, and 2) `gpadmin` who does not have `sudo`-privileges. This variable holds the name of the OS-specific user.

concourse/pipelines/templates/perf_pipeline-tpl.yml
```diff
@@ -181,22 +192,33 @@ resources:
 
 jobs:
 
-- name: Build PXF-GP6 on RHEL7
+- name: Build PXF-GP6 on RHEL[[redhat_major_version]]
```
denalex, line 195: we are somewhat inconsistent about using  `Centos7` and `RHEL7` which prevents us to use loops, especially for names of docker images. Does not have to be solved here, just noticing. Or maybe we can call the resource `gpdb6-pxf-dev-rhel7-image` and it will actually pull `gpdb6-pxf-dev-centos7-image` ?

concourse/scripts/install_pxf.bash
```diff
@@ -2,6 +2,8 @@
 
 set -euxo pipefail
 
+: "${PLATFORM:?PLATFORM must be set}"
```
denalex, line 5: it is not immediately clear what values this variable might take. Maybe document it ? The best would be to auto-discover it, though.

concourse/scripts/install_pxf.bash
```diff
@@ -13,6 +15,8 @@ else
 	PXF_HOME=${GPHOME}/pxf
 fi
 PXF_VERSION=${PXF_VERSION:=6}
+# for centos7, there exists a ~centos and in rhel8, there exists a ~rhel
+DEFAULT_OS_USER="${PLATFORM//[[:digit:].-]/}"
```
denalex, line 19: what about `ubuntu` or `oel` ?

concourse/Makefile
```diff
@@ -9,10 +9,11 @@ BUILD_PIPELINE_NAME         ?= pxf-build
 DEV_BUILD_PIPELINE_NAME     ?= dev:$(USER)-$(BRANCH)
 PR_BUILD_PIPELINE_NAME      ?= pxf_pr_pipeline
 CERTIFICATION_PIPELINE_NAME ?= pxf-certification
-PERF_PIPELINE_NAME          ?= pxf_perf-$(SCALE)g
+PERF_PIPELINE_NAME          ?= pxf_perf-$(SCALE)g-RHEL$(REDHAT_MAJOR_VERSION)
 PIVNET_PIPELINE_NAME        ?= pivnet_artifacts
 NUM_GPDB5_VERSIONS          ?= 10
 NUM_GPDB6_VERSIONS          ?=  9
+REDHAT_MAJOR_VERSION        ?= 7
```
denalex, line 16: does this need to go above line 12 since you want it to take default value if not set before using it there ?

denalex (approved):

concourse/Makefile
ashuka24: It didn't seem to be an issue when I was testing. I think because of where the variable is actually used, it does still properly evaluate to the default value if the `REDHAT_MAJOR_VERSION` is not provided in the command line

denalex: I don't get it :) It is used in line 12 but takes a default value only in line 16, how would that work ?

ashuka24: `?=` in a Makefile means to lazy set a variable, so the value of line 12 is evaluated when it is actually accessed not when it's declared. https://stackoverflow.com/questions/448910/what-is-the-difference-between-the-gnu-makefile-variable-assignments-a

PR #772, opened 2022-03-01
Added separate SUT file for IPA test cluster environment

denalex:


automation/src/test/resources/sut/MultiHadoopIPAMultiNodesCluster.xml
```diff
@@ -0,0 +1,139 @@
+<sut>
```
bradfordb-vmware, line 1: Minor nit: It looks like this file has executable permission (`-rwxr-xr-x`) which it probably inherited when copying `MultiHadoopMultiNodesCluster.xml`. Do these files need to be executable?

bradfordb-vmware (approved):
Would it be possible to add some context for this change to either the commit message and/or where we have the conditional check in `MultiServerTest.java` and `text_pxf_multinode.bash`?

hpandeycodeit (approved):

PR #777, opened 2022-03-11
Update build and dev pipelines to include a mirrorless cluster test

bradfordb-vmware:
* Modify the PXF-GP6-SECURE-MULTI-NO-IMPERS pipeline job to use a
mirrorless CCP cluster
  * CCP configuration was taken from [PR 13116][] in greenplum-db/gpdb
* Update PXF CLI tests to work with clusters that do not include a
standby master host

[PR 13116]: https://github.com/greenplum-db/gpdb/pull/13116

### DIFF
--- a/concourse/pipelines/templates/build_pipeline-tpl.yml
+++ b/concourse/pipelines/templates/build_pipeline-tpl.yml
@@ -1378,12 +1378,13 @@ jobs:
           terraform_source: ccp_src/google/
           vars:
             PLATFORM: centos7
+            WITH_MIRRORS: false
             number_of_nodes: ((number_of_gpdb_nodes))
             extra_nodes: 1
             segments_per_host: 4
             instance_type: n1-standard-4
             ccp_reap_minutes: 120
-            standby_master: true
+            standby_master: false
       - task: Generate Greenplum Cluster
         input_mapping:
           gpdb_rpm: gpdb_package
@@ -1402,6 +1403,13 @@ jobs:
       - in_parallel:
         - task: Initialize Greenplum
           file: ccp_src/ci/tasks/gpinitsystem.yml
+          params:
+            POSTGRES_CONF_ADDONS:
+            - wal_level=minimal
+            - max_wal_senders=0
+            - gp_dispatch_keepalives_idle=30
+            - gp_dispatch_keepalives_interval=10
+            - gp_dispatch_keepalives_count=4
         - task: Install Hadoop
           file: pxf_src/concourse/tasks/install_hadoop.yml
           image: gpdb[[gp_ver]]-pxf-dev-centos7-image
--- a/concourse/scripts/cli/common.sh
+++ b/concourse/scripts/cli/common.sh
@@ -30,6 +30,9 @@ err_cnt=0
 test_cnt=0
 failed_tests_cnt=0
 
+declare -a all_cluster_hosts
+mapfile -t all_cluster_hosts <hostfile_all
+
 run_test() {
 	local usage='test <func> <message>'
 	local func=${1:?${usage}} message="$((++test_cnt))) ${2:?${usage}}"
@@ -135,3 +138,30 @@ list_remote_pxf_running_pid() {
   ssh "${1}" "ps -aef | grep pxf | grep -v grep | tr -s ' ' | cut -d ' ' -f 2"
 }
 
+has_standby_master() {
+  grep -q smdw hostfile_all
+}
+
+get_cluster_description() {
+  local cluster_description="master host"
+  if has_standby_master; then
+    cluster_description+=", standby master host,"
+  fi
+  local num_segment_hosts
+  num_segment_hosts="$(grep -c -P 'sdw\d+' hostfile_all)"
+  cluster_description+=" and ${num_segment_hosts} segment hosts..."
+
+  echo "${cluster_description}"
+}
+
+get_cluster_sync_description() {
+  local cluster_sync_description="master host to "
+  if has_standby_master; then
+    cluster_sync_description+="standby master host and "
+  fi
+  local num_segment_hosts
+  num_segment_hosts="$(grep -c -P 'sdw\d+' hostfile_all)"
+  cluster_sync_description+="${num_segment_hosts} segment hosts..."
+
+  echo "${cluster_sync_description}"
+}
--- a/concourse/scripts/cli/test_reset_init.sh
+++ b/concourse/scripts/cli/test_reset_init.sh
@@ -6,6 +6,9 @@ source "${dir}/common.sh"
 
 expected_extension_file="${GPHOME}/share/postgresql/extension/pxf.control"
 
+cluster_description="$(get_cluster_description)"
+num_hosts="${#all_cluster_hosts[@]}"
+
 list_extension_files() {
 	local usage='list_extension_files <host>' host=${1:?${usage}}
 	ssh "${host}" "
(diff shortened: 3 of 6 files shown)


concourse/scripts/cli/test_reset_init.sh
```diff
@@ -55,20 +58,22 @@ successful_init_message=\
...
+  done <hostfile_all
   # when : "pxf cluster init" command is run
   local result="$(pxf cluster init)"
   # then : it succeeds and prints the expected message
   assert_equals "${successful_init_message}" "${result}" "pxf cluster init (no extension) should succeed"
   #      : AND the extension file is copied to GPHOME
-  assert_equals "${expected_extension_file}" "$(list_extension_files ${host})" "extension file should exist on host ${host}"
+  while read -r host; do
+    assert_equals "${expected_extension_file}" "$(list_extension_files ${host})" "extension file should exist on host ${host}"
+  done <hostfile_all
```
ashuka24, line 76: some of these while loops you use `hostfile_all` and some of then use `hostfile_init`. I'm a little confused what the difference between these files are and how do we determine when one should be used over the other?

bradfordb-vmware: The file `hostfile_all` includes all hosts in the cluster (e.g., mdw, smdw, sdw1, ..., sdwn) while `hostfile_init` does not include `mdw`.

That said, `hostfile_all` and `hostfile_init` are not the best names and I am open to any suggestions for better names.

ashuka24: Are there requirements to the file convention? (aka does it have to start with `hostfile_`? I think `hostfile_all` is ok for what it is. maybe we could rename `hostfile_init` to `hostfile_segments` or something that's more descriptive of the contents?

concourse/pipelines/templates/build_pipeline-tpl.yml
```diff
@@ -1402,6 +1403,13 @@ jobs:
       - in_parallel:
         - task: Initialize Greenplum
           file: ccp_src/ci/tasks/gpinitsystem.yml
+          params:
+            POSTGRES_CONF_ADDONS:
+            - wal_level=minimal
+            - max_wal_senders=0
+            - gp_dispatch_keepalives_idle=30
+            - gp_dispatch_keepalives_interval=10
+            - gp_dispatch_keepalives_count=4
```
denalex, line 1412: is this really needed for mirrorless ? If yes, maybe give feedback to CCP to pass only 1 parameter to the job and have the job setup this variables, as CCP users should not care about these internals.

concourse/scripts/cli/common.sh
```diff
@@ -102,19 +102,19 @@ assertion_error() {
 }
 
 remove_remote_file() {
-  ssh "${1}" "rm -rf "${2}""
+  ssh -n "${1}" "rm -rf "${2}""
```
denalex, line 105: wonder why `-n` flag is needed ? Seems it prohibits reading from `stdin` which is recommended for ssh tasks running in the background.

concourse/scripts/cli/test_reset_init.sh
```diff
@@ -55,20 +58,22 @@ successful_init_message=\
 *
 *****************************************************************************
 
-Initializing PXF on master host, standby master host, and 2 segment hosts...
-PXF initialized successfully on 4 out of 4 hosts"
+Initializing PXF on ${cluster_description}
```
denalex, line 61: smart idea, but it introduces a pattern of dynamically generating an expected outcome based on the runtime state, which is prone to garbage-in / garbage-out problem. The concern is purely theoretical, though, I think this change is good overall.

denalex (approved):

concourse/pipelines/templates/build_pipeline-tpl.yml
bradfordb-vmware: This was the recommended setting for mirrorless CCP given in this [PR on greenplum-db/gpdb](https://github.com/greenplum-db/gpdb/pull/13116)

@edespino Would it be possible to move these PGOPTIONS into CCP?

concourse/scripts/cli/common.sh
bradfordb-vmware: This is needed because this function ends up being called inside a `while`-loop that is reading from `stdin`. We could avoid the `while`-loop reading from `stdin` by reading the files into an array variable and then looping over that variable instead.

denalex: yes, that would be way preferable instead of leaking the concern of "no stdin" into functions residing in different files

concourse/scripts/cli/test_stop_start_restart_status.sh
```diff
@@ -345,6 +345,8 @@ test_start_succeeds_one_running() {
   assert_not_empty "${sdw1_pid}" "PXF should be running on host sdw1"
   #      : AND PXF is not running on master, standby master or segment host 2
   for host in "${hosts_all[@]}"; do
+    [[ "${host}" == "sdw1" ]] && continue
+
```
denalex, line 349: I think you do not need to quote when using `[[ ]]` tests

concourse/scripts/cli/test_stop_start_restart_status.sh
```diff
@@ -295,22 +302,22 @@ run_test test_restart_succeeds_none_running "pxf cluster restart (none running)
...
-PXF restarted successfully on 4 out of 4 hosts"
+"Restarting PXF on ${cluster_description}
+PXF restarted successfully on ${num_hosts} out of ${num_hosts} hosts"
 test_restart_succeeds_one_running() {
   # given: PXF is running on segment host 1
   local sdw1_pid="$(list_remote_pxf_running_pid sdw1)"
   assert_not_empty "${sdw1_pid}" "PXF should be running on host sdw1"
   #      : AND PXF is not running on master, standby master or segment host 2
-  for host in {s,}mdw sdw2; do
+  for host in "${all_cluster_hosts[@]}"; do
```
ashuka24, line 312: looks like this original check was skipping over `sdw1` but we aren't doing that anymore. i'm a little concerned that this test did not fail?

concourse/scripts/cli/test_stop_start_restart_status.sh
```diff
@@ -392,22 +407,22 @@ run_test test_status_succeeds_one_running "pxf cluster status (one running) shou
...
-PXF stopped successfully on 4 out of 4 hosts"
+"Stopping PXF on ${cluster_description}
+PXF stopped successfully on ${num_hosts} out of ${num_hosts} hosts"
 test_stop_succeeds_one_running() {
   # given: PXF is running on segment host 1
   local sdw1_pid="$(list_remote_pxf_running_pid sdw1)"
   assert_not_empty "${sdw1_pid}" "PXF should be running on host sdw1"
   #      : AND PXF is not running on master, standby master or segment host 2
-  for host in {s,}mdw sdw2; do
+  for host in "${all_cluster_hosts[@]}"; do
```
ashuka24, line 417: same here, this test looks like it should be running on `sdw1`...

bradfordb-vmware:
> what was the purpose of changing the order ? The overall logic is still the same, right ?

Yes. The original test checked that PXF was running on `sdw1` and then ran `pxf stop` on `mdw`, `smdw`, and `sdw2` (arranging) before then running `pxf cluster restart` (acting) and then asserting the expected output message.

When adjusting the `for`-loop, I accidentally included `sdw1` so that PXF was not running on any of the hosts before executing `pxf cluster restart`. As @ashuka24 pointed out, the test was still passing even though the setup wasn't exactly the same. Strictly speaking I did not need to change the order, just omit `sdw1` from the loop that executes `pxf stop` but I wanted to make sure the test would fail if the scenario isn't correct.

ashuka24 (approved):

PR #781, opened 2022-03-22
Fix: In case of UnsupportedOperationException, Add an error message.

hpandeycodeit:
There are places in the code where we are throwing `UnsupportedOperationException` without any error message  and the empty error message isn't very useful to the user. For eg: `ERROR:  PXF server error : No message available`

This PR adds error messages to the places where we are throwing `UnsupportedOperationException`

### DIFF
--- a/server/pxf-api/src/main/java/org/greenplum/pxf/api/examples/DemoAccessor.java
+++ b/server/pxf-api/src/main/java/org/greenplum/pxf/api/examples/DemoAccessor.java
@@ -34,7 +34,7 @@ public class DemoAccessor extends BasePlugin implements Accessor {
     private int rowNumber;
     private int fragmentNumber;
     private static final int NUM_ROWS = 2;
-    private static final String UNSUPPORTED_ERR_MESSAGE = "Write operation is not supported";
+    private static final String UNSUPPORTED_ERR_MESSAGE = "Demo accessor does not support write operation";
 
     @Override
     public boolean openForRead() {
--- a/server/pxf-api/src/main/java/org/greenplum/pxf/api/examples/DemoAccessor.java
+++ b/server/pxf-api/src/main/java/org/greenplum/pxf/api/examples/DemoAccessor.java
@@ -34,6 +34,7 @@ public class DemoAccessor extends BasePlugin implements Accessor {
     private int rowNumber;
     private int fragmentNumber;
     private static final int NUM_ROWS = 2;
+    private static final String UNSUPPORTED_ERR_MESSAGE = "Write operation is not supported";
 
     @Override
     public boolean openForRead() {
@@ -91,7 +92,7 @@ public class DemoAccessor extends BasePlugin implements Accessor {
      */
     @Override
     public boolean openForWrite() throws Exception {
-        throw new UnsupportedOperationException();
+        throw new UnsupportedOperationException(UNSUPPORTED_ERR_MESSAGE);
     }
 
     /**
@@ -103,7 +104,7 @@ public class DemoAccessor extends BasePlugin implements Accessor {
      */
     @Override
     public boolean writeNextObject(OneRow onerow) throws Exception {
-        throw new UnsupportedOperationException();
+        throw new UnsupportedOperationException(UNSUPPORTED_ERR_MESSAGE);
     }
 
     /**
@@ -113,6 +114,6 @@ public class DemoAccessor extends BasePlugin implements Accessor {
      */
     @Override
     public void closeForWrite() throws Exception {
-        throw new UnsupportedOperationException();
+        throw new UnsupportedOperationException(UNSUPPORTED_ERR_MESSAGE);
     }
 }
--- a/server/pxf-api/src/main/java/org/greenplum/pxf/api/examples/DemoResolver.java
+++ b/server/pxf-api/src/main/java/org/greenplum/pxf/api/examples/DemoResolver.java
@@ -65,6 +65,6 @@ public class DemoResolver extends BasePlugin implements Resolver {
      */
     @Override
     public OneRow setFields(List<OneField> record) throws Exception {
-        throw new UnsupportedOperationException("Current operation is not supported");
+        throw new UnsupportedOperationException("Demo resolver does not support write operation");
     }
 }
--- a/server/pxf-api/src/main/java/org/greenplum/pxf/api/examples/DemoResolver.java
+++ b/server/pxf-api/src/main/java/org/greenplum/pxf/api/examples/DemoResolver.java
@@ -65,6 +65,6 @@ public class DemoResolver extends BasePlugin implements Resolver {
      */
     @Override
     public OneRow setFields(List<OneField> record) throws Exception {
-        throw new UnsupportedOperationException();
+        throw new UnsupportedOperationException("Current operation is not supported");
     }
 }
--- a/server/pxf-api/src/main/java/org/greenplum/pxf/api/model/BaseFragmenter.java
+++ b/server/pxf-api/src/main/java/org/greenplum/pxf/api/model/BaseFragmenter.java
@@ -14,6 +14,8 @@ public class BaseFragmenter extends BasePlugin implements Fragmenter {
 
     @Override
     public FragmentStats getFragmentStats() throws Exception {
-        throw new UnsupportedOperationException("Current fragmenter does not support getFragmentStats");
+
+        String profile = context.getProfile();
+        throw new UnsupportedOperationException(String.format("Profile '%s' does not support statistics for fragments", profile));
     }
 }
--- a/server/pxf-api/src/main/java/org/greenplum/pxf/api/model/BaseFragmenter.java
+++ b/server/pxf-api/src/main/java/org/greenplum/pxf/api/model/BaseFragmenter.java
@@ -14,6 +14,6 @@ public class BaseFragmenter extends BasePlugin implements Fragmenter {
 
     @Override
     public FragmentStats getFragmentStats() throws Exception {
-        throw new UnsupportedOperationException("Current operation is not supported");
+        throw new UnsupportedOperationException("Current fragmenter does not support getFragmentStats");
     }
 }
--- a/server/pxf-api/src/main/java/org/greenplum/pxf/api/model/BaseFragmenter.java
+++ b/server/pxf-api/src/main/java/org/greenplum/pxf/api/model/BaseFragmenter.java
@@ -14,6 +14,6 @@ public class BaseFragmenter extends BasePlugin implements Fragmenter {
 
     @Override
     public FragmentStats getFragmentStats() throws Exception {
-        throw new UnsupportedOperationException("Operation getFragmentStats is not supported");
+        throw new UnsupportedOperationException("Current operation is not supported");
     }
 }
--- a/server/pxf-diagnostic/src/main/java/org/greenplum/pxf/diagnostic/UserDataVerifyAccessor.java
+++ b/server/pxf-diagnostic/src/main/java/org/greenplum/pxf/diagnostic/UserDataVerifyAccessor.java
@@ -18,6 +18,7 @@ public class UserDataVerifyAccessor extends BasePlugin implements Accessor {
 
     private int counter = 0;
     private char firstColumn = 'A';
+    private static final String UNSUPPORTED_ERR_MESSAGE = "Write operation is not supported";
 
     @Override
     public boolean openForRead() {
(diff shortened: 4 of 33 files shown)


server/pxf-service/src/main/java/org/greenplum/pxf/service/controller/PxfErrorReporter.java
```diff
@@ -40,6 +40,9 @@ protected T invokeWithErrorHandling(ThrowingSupplier<T, Exception> action) {
             // let PxfRuntimeException and Error propagate themselves
             log.error(e.getMessage(), e);
             throw e;
+        } catch (UnsupportedOperationException e) {
+            log.error(e.getStackTrace()[0].getMethodName()+" is not implemented" , e);
+            throw new PxfRuntimeException(e.getStackTrace()[0].getMethodName()+" is not implemented", e );
```
denalex, line 45: nit: formatting
```suggestion
            String errorMessage = e.getStackTrace()[0].getMethodName() + " is not implemented";
            log.error(errorMessage, e);
            throw new PxfRuntimeException(errorMessage, e );
```

denalex: `e.getStackTrace()[0]` might be null (in some special cases) so we'll need a null check

denalex: Do we need unit test ? Also in the PR description can you please describe what problem does this solve, what is the output before and after the change ? I am not 100% sure we need to do exactly this rather than going into our code and adding descriptions to places where we throw `UnsupportedOperationException`

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/QuotedLineBreakAccessor.java
```diff
@@ -35,12 +35,16 @@
...
+
+    private static final String UNSUPPORTED_ERR_MESSAGE = "Write operation is not supported";
+
     private boolean fileAsRow;
     private boolean firstLine, lastLine;
     private int skipHeaderCount;
     BufferedReader reader;
     Queue<String> lineQueue;
 
+
```
ashuka24, line 47: nit: extra new line not needed

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/QuotedLineBreakAccessor.java
```diff
@@ -35,12 +35,16 @@
  * multi-line records, that are read from a single source (non-parallel).
  */
 public class QuotedLineBreakAccessor extends HdfsAtomicDataAccessor {
+
+    private static final String UNSUPPORTED_ERR_MESSAGE = "Write operation is not supported";
```
ashuka24, line 39: When is this accessor used? Could we have a more specific error message here?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedAccessor.java
```diff
@@ -57,6 +57,7 @@
     private static final TreeTraverser TRAVERSER = new TreeTraverser();
 
     static final String MAP_BY_POSITION_OPTION = "MAP_BY_POSITION";
+    private static final String UNSUPPORTED_ERR_MESSAGE = "Write operation is not supported";
```
ashuka24, line 60:
```suggestion
    private static final String UNSUPPORTED_ERR_MESSAGE = "Write operation is not supported for vectorized ORC profile";
```
or maybe something along these lines?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedResolver.java
```diff
@@ -127,6 +127,8 @@
 
     private List<List<OneField>> cachedBatch;
 
+    private static final String UNSUPPORTED_ERR_MESSAGE = "Current operation is not supported";
```
ashuka24, line 130: maybe this would be better? or maybe even specify read/write
```suggestion
    private static final String UNSUPPORTED_ERR_MESSAGE = "Vectorized ORC operations are not supported";
```

server/pxf-hive/src/main/java/org/greenplum/pxf/plugins/hive/HiveAccessor.java
```diff
@@ -96,6 +96,7 @@
     private static final Logger LOG = LoggerFactory.getLogger(HiveAccessor.class);
     private static final String PXF_PPD_HIVE = "pxf.ppd.hive";
     private static final String HIVE_DEFAULT_PARTITION = "__HIVE_DEFAULT_PARTITION__";
+    private static final String UNSUPPORTED_ERR_MESSAGE = "Write operation is not supported";
```
ashuka24, line 99:
```suggestion
    private static final String UNSUPPORTED_ERR_MESSAGE = "Write operation is not supported for Hive";
```

ashuka24 (commented):
Some of the error messages could be more specific instead of just having `Current operation is not supported`. I think it would be worth adding more specific error messages.

server/pxf-api/src/main/java/org/greenplum/pxf/api/examples/DemoAccessor.java
```diff
@@ -34,6 +34,7 @@ public class DemoAccessor extends BasePlugin implements Accessor {
     private int rowNumber;
     private int fragmentNumber;
     private static final int NUM_ROWS = 2;
+    private static final String UNSUPPORTED_ERR_MESSAGE = "Write operation is not supported";
```
bradfordb-vmware, line 37: Do we want to be a little more specific here and include the name of the accessor e.g., `Demo accessor does not support write operations`

server/pxf-api/src/main/java/org/greenplum/pxf/api/examples/DemoResolver.java
```diff
@@ -65,6 +65,6 @@ public List<OneField> getFields(OneRow row) {
      */
     @Override
     public OneRow setFields(List<OneField> record) throws Exception {
-        throw new UnsupportedOperationException();
+        throw new UnsupportedOperationException("Current operation is not supported");
```
bradfordb-vmware, line 68: This message is too vague -- we should say what the operation is (write or read).

server/pxf-api/src/main/java/org/greenplum/pxf/api/model/BaseFragmenter.java
```diff
@@ -14,6 +14,6 @@ public List<Fragment> getFragments() throws Exception {
 
     @Override
     public FragmentStats getFragmentStats() throws Exception {
-        throw new UnsupportedOperationException("Operation getFragmentStats is not supported");
+        throw new UnsupportedOperationException("Current operation is not supported");
```
bradfordb-vmware, line 17: I don't think `getFragmentStats` is actually used anywhere but I think the message should indicate that the current fragmenter does not support get fragment stats.

server/pxf-diagnostic/src/main/java/org/greenplum/pxf/diagnostic/UserDataVerifyAccessor.java
```diff
@@ -18,6 +18,7 @@ public class UserDataVerifyAccessor extends BasePlugin implements Accessor {
 
     private int counter = 0;
     private char firstColumn = 'A';
+    private static final String UNSUPPORTED_ERR_MESSAGE = "Write operation is not supported";
```
bradfordb-vmware, line 21: We should include either the profile name or the accessor name.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedResolver.java
hpandeycodeit: I am leaving the ORC related changes for now as it is since we are working on orc write and this will be changed.

hpandeycodeit:
Addressed the PR feedback

server/pxf-hive/src/main/java/org/greenplum/pxf/plugins/hive/HiveAccessor.java
```diff
@@ -96,6 +96,7 @@ public class HiveAccessor extends HdfsSplittableDataAccessor {
     private static final Logger LOG = LoggerFactory.getLogger(HiveAccessor.class);
     private static final String PXF_PPD_HIVE = "pxf.ppd.hive";
     private static final String HIVE_DEFAULT_PARTITION = "__HIVE_DEFAULT_PARTITION__";
+    private static final String UNSUPPORTED_ERR_MESSAGE = "Write operation is not supported for Hive";
```
bradfordb-vmware, line 99:
```suggestion
    private static final String UNSUPPORTED_ERR_MESSAGE = "Hive accessor does not support write operation.";
```

We should try and follow that same pattern for all error messages of the same kind.

server/pxf-hive/src/main/java/org/greenplum/pxf/plugins/hive/HiveResolver.java
```diff
@@ -150,7 +150,7 @@ public List<OneField> getFields(OneRow onerow) throws Exception {
      */
     @Override
     public OneRow setFields(List<OneField> record) {
-        throw new UnsupportedOperationException();
+        throw new UnsupportedOperationException("Current operation is not supported");
```
bradfordb-vmware, line 153: The method `setFields` is called in the write path so this should be
```suggestion
        throw new UnsupportedOperationException("Hive resolver does not support write operation.");
```

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonAccessor.java
```diff
@@ -43,6 +43,7 @@ public class JsonAccessor extends HdfsSplittableDataAccessor {
 
     public static final String IDENTIFIER_PARAM = "IDENTIFIER";
     public static final String RECORD_MAX_LENGTH_PARAM = "MAXLENGTH";
+    private static final String UNSUPPORTED_ERR_MESSAGE = "Write for JSON is not supported";
```
bradfordb-vmware, line 46:
```suggestion
    private static final String UNSUPPORTED_ERR_MESSAGE = "JSON accessor does not support write operation.";
```

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonResolver.java
```diff
@@ -123,7 +123,7 @@ public List<OneField> getFields(OneRow row) throws Exception {
      */
     @Override
     public OneRow setFields(List<OneField> record) throws UnsupportedOperationException {
-        throw new UnsupportedOperationException();
+        throw new UnsupportedOperationException("Current operation is not supported");
```
bradfordb-vmware, line 126:
```suggestion
        throw new UnsupportedOperationException("JSON resolver does not support write operation.");
```

server/pxf-s3/src/main/java/org/greenplum/pxf/plugins/s3/S3SelectAccessor.java
```diff
@@ -46,6 +46,7 @@ public class S3SelectAccessor extends BasePlugin implements Accessor {
     public static final String FILE_HEADER_INFO_IGNORE = "IGNORE";
     public static final String FILE_HEADER_INFO_USE = "USE";
     public static final String JSON_TYPE = "JSON-TYPE";
+    private static final String UNSUPPORTED_ERR_MESSAGE = "S3 Select does not support writing";
```
bradfordb-vmware, line 49:
```suggestion
    private static final String UNSUPPORTED_ERR_MESSAGE = "S3 Select accessor does not support write operation.";
```

server/pxf-service/src/main/java/org/greenplum/pxf/service/bridge/ReadBridge.java
```diff
@@ -150,7 +150,7 @@ protected boolean isDataException(IOException ex) {
      */
     @Override
     public boolean setNext(DataInputStream inputStream) {
-        throw new UnsupportedOperationException("setNext is not implemented");
+        throw new UnsupportedOperationException("Current operation is not supported");
```
bradfordb-vmware, line 153: The method `setNext` is called from `WriteServiceImpl` -- I can't imagine how the 'ReadBridge` would ever end up being used in the write path but I think we need a better message here. We should probably also add some debug logging message here to help determine how the code ended up here.

hpandeycodeit: `ReadBridge` implements abstract class `BaseBridge` which implements interface `Bridge` which has this method. That's why `setNext` is present in `ReadBridge` and yeah it will never get used in this class.  I have modified the message here.

ashuka24 (approved):

bradfordb-vmware (approved):

PR #788, opened 2022-04-08
Replace prefix macro with environment variable in scriptlets

bradfordb-vmware:
Since RPM macros are expanded at RPM build time, scriptlets should use
the install prefix environment variable instead.

### DIFF
--- a/package/pxf-gp5.spec
+++ b/package/pxf-gp5.spec
@@ -35,8 +35,8 @@ fi
 %__cp -R %{_sourcedir}/* %{buildroot}/%{prefix}
 
 %post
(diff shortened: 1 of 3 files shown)


bradfordb-vmware:
I tested this change manually by install the RPMs with following command

```sh
rpm -i <path-to-rpm> --prefix /opt/foo\ bar
```

and ensuring there was no warning/error output from the post-install scriptlet. I also verified the content of the extension control file.

denalex (approved):

hpandeycodeit (approved):

ashuka24 (approved):

PR #789, opened 2022-04-08
Upgrade to Springboot 2.5.12

ashuka24:
This commit updates the Springboot version to 2.5.12 to mitigate any
 vulnerabilities that come from the following CVE: https://github.com/advisories/GHSA-36p3-wjmg-h94x

Springboot 2.5.12 updates Tomcat from v9.0.43 to v9.0.60. This included
a change that causes the stream to close immediately instead of cleanly
if an error occurs after the response is committed. This means that the
chunk that marks the end of the response body is no longer sent [1]. Due
to the change in error handling during streaming, the
HdfsReadableTextTest#errorInTheMiddleOfStream integration test must be
updated to reflect the new behavior.

Springboot 2.5.12 comes with Log4j v2.17.2 [2] so the patch introduced
in commit https://github.com/greenplum-db/pxf/commit/09e944d564594430469ec9d253f07b738dbfb00c (PR https://github.com/greenplum-db/pxf/pull/735) is no longer necessary.

[1] https://github.com/apache/tomcat/commit/cf9abfc1decae26e61eefd89a54317ae8696be7b
[2] https://docs.spring.io/spring-boot/docs/2.5.12/reference/html/dependency-versions.html


bradfordb-vmware (approved):

server/gradle.properties
```diff
@@ -25,7 +25,7 @@ hbaseVersion=1.3.2
 junitVersion=4.11
 parquetVersion=1.11.1
 awsJavaSdk=1.11.801
-springBootVersion=2.4.3
+springBootVersion=2.5.12
```
denalex, line 28: does this version include fixes for Log4J2 and Postgres CVEs ? If yes, we then can revert the patches to `build.gradle` we had to make for those CVEs.

denalex (approved):
Can we please put a detailed description in the PR of what has changed with Spring Boot (with a link to Spring PR/commit if possible) and how the error handling during streaming is now different so that we had to update the integration test.

hpandeycodeit (approved):

PR #794, opened 2022-04-15
Fix hive orc vectorize repeating

bradfordb-vmware:


automation/src/main/java/org/greenplum/pxf/automation/structures/tables/utils/TableFactory.java
```diff
@@ -129,7 +129,8 @@ public static ReadableExternalTable getPxfHiveVectorizedOrcReadableTable(String
                 fields, hiveTable.getName(), "CUSTOM");
 
         if (useProfile) {
-            exTable.setProfile("HiveVectorizedORC");
+            exTable.setProfile("hive:orc");
```
ashuka24, line 132: i see that there are other locations that use `HiveVectorizedORC` instead of `hive:orc` with `VECTORIZE=true`. For example in `HiveVectorizedOrcTest.java`. Do we need to update the profile there as well?

automation/src/test/java/org/greenplum/pxf/automation/features/hive/HiveOrcTest.java
```diff
@@ -106,6 +122,36 @@ private void preparePxfHiveSmallData() throws Exception {
...
 
+    private void prepareOrcDataWithRepeatingData() throws Exception {
+        String dataFileName = "hive_repeating_data.csv";
+        String[] hiveTableColumns = new String[] {
+                "col1 INT",
+                "col2 VARCHAR(16)",
+                "col3 CHAR(1)",
+                "col4 STRING",
+                "col5 DATE",
+                "col6 TINYINT"
```
denalex, line 133: ideally we should test handling of repeated values for all supported column types, not only the ones affected by this fix. Can we re-use table definition that has all types and populate it with repeating data ?

automation/src/test/resources/data/hive/hive_repeating_data.csv
```diff
@@ -0,0 +1,10 @@
+0,col2 value,3,col4 value,2022-04-14,64
+1,col2 value,3,col4 value,2022-04-14,64
+2,col2 value,3,col4 value,2022-04-14,64
+3,col2 value,3,col4 value,2022-04-14,64
+4,col2 value,3,col4 value,2022-04-14,64
+5,col2 value,3,col4 value,2022-04-14,64
+6,col2 value,3,col4 value,2022-04-14,64
+7,col2 value,3,col4 value,2022-04-14,64
+8,col2 value,3,col4 value,2022-04-14,64
+9,col2 value,3,col4 value,2022-04-14,64
```
denalex, line 10: I think we also should test repeated nulls for all supported types, perhaps as a separate query / test case

automation/src/test/java/org/greenplum/pxf/automation/features/hive/HiveOrcTest.java
```diff
@@ -74,6 +77,19 @@ void prepareHiveCollection() throws Exception {
...
 
+    @Override
+    protected void afterClass() throws Exception {
+        if (hiveRepeatingOrcTable != null) {
+            hive.dropTable(hiveRepeatingOrcTable, true);
+        }
+
+        if (hiveRepeatingCsvTable != null) {
+            hive.dropTable(hiveRepeatingCsvTable, true);
+        }
```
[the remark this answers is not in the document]
bradfordb-vmware: The `createTableAndVerify` helper method takes care of dropping the table (if it exists) at the beginning of the test so we can just omit this then.

automation/src/main/java/org/greenplum/pxf/automation/structures/tables/utils/TableFactory.java
bradfordb-vmware: I've reverted this change and added tests for both `HiveVectorizedORC` and `hive:orc` w/ `VECTORIZE=true` profiles.

automation/src/test/java/org/greenplum/pxf/automation/features/hive/HiveOrcTest.java
bradfordb-vmware: `HiveORCVectorizedResolver` doesn't support `TIMESTAMP` so I wasn't able to re-use the `hive_types` table definition but I've expanded the table definition to include all columns from `hive_types` except timestamp.

automation/src/test/resources/data/hive/hive_repeating_data.csv
bradfordb-vmware: Add a second file where every column is `null` repeating.

denalex (approved):

PR #801, opened 2022-04-27
ci: include PXF major version in cloud storage path

bradfordb-vmware:
The build pipelines for PXF 5.16 and 6 use the same path for storing
snapshot tarballs. The pipelines can distinguish artifacts produced by
the different pipelines by correlating the files generation with the CI
job that produced the artifact. CI jobs outside of the build pipelines
cannot distinguish between snapshot builds of PXF 5.16 and 6 and will
randomly get one or the other depending on which ran last.

This commit moves PXF snapshot tarballs to be stored in a location that
includes the PXF major version. This will allow downstream consumers of
snapshot builds to select which PXF major version to pull.

### DIFF
--- a/concourse/pipelines/templates/build_pipeline-tpl.yml
+++ b/concourse/pipelines/templates/build_pipeline-tpl.yml
@@ -534,7 +534,7 @@ resources:
   source:
     bucket: data-gpdb-ud-pxf-build
     json_key: ((pxf-storage-service-account-key))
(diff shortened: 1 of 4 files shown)


kalensk (approved):
LGTM. Not necessarily needed, but if there are objects populated in those locations this gpupgrade PR https://github.com/greenplum-db/gpupgrade/pull/663 and pipeline can test it out before merging.

denalex (approved):

ashuka24 (approved):

PR #814, opened 2022-06-03
Fix pxf extension sigsegv

bradfordb-vmware:


bradfordb-vmware:
Example usage: Base table DDL is

```sql
CREATE READABLE EXTERNAL TABLE pxf_file_avro_r(col1 int, col2 text, col3 text)
LOCATION ('pxf://avro-data/20220523?profile=file:avro&server=alpine')
```

Query fails without crashing the Greenplum Database cluster, but nothing is added to the error table

```sql
SELECT * FROM pxf_file_avro_r ;
-- ERROR:  column 1 has length that exceeds tuple length  (seg2 slice1 127.0.1.1:6002 pid=1179056)
-- CONTEXT:  External table pxf_file_avro_r
SELECT * FROM gp_read_error_log('pxf_file_avro_r');
--  cmdtime | relname | filename | linenum | bytenum | errmsg | rawdata | rawbytes
-- ---------+---------+----------+---------+---------+--------+---------+----------
-- (0 rows)
```

Update the DDL to include a formatter option to generate a core dump on error

```sql
CREATE READABLE EXTERNAL TABLE pxf_file_avro_r(col1 int, col2 text, col3 text)
LOCATION ('pxf://avro-data/20220523?profile=file:avro&server=alpine')
FORMAT 'CUSTOM' (FORMATTER='pxfwritable_import', dump_core_on_error=1);
```

```sql
SELECT * FROM pxf_file_avro_r ;
-- ERROR:  column 1 has length that exceeds tuple length  (seg2 slice1 127.0.1.1:6002 pid=1181100)
-- CONTEXT:  External table pxf_file_avro_r

SELECT * FROM gp_read_error_log('pxf_file_avro_r');
--  cmdtime | relname | filename | linenum | bytenum | errmsg | rawdata | rawbytes
-- ---------+---------+----------+---------+---------+--------+---------+----------
-- (0 rows)
```

```console
$ coredumpctl list
...
Fri 2022-06-10 11:34:40 PDT  1181100  1001  1002   6 present   /usr/local/greenplum-db-6/bin/postgres
```

Note that the PID (1181100) shown in the output of `coredumpctl list` matches with the PID from the error message in the PSQL prompt.
Using `gdb` to debug this core and running `backtrace` shows that we can debug inside of `gpdbwritableformatter_import`.

```console
(gdb) backtrace
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007f41cf210537 in __GI_abort () at abort.c:79
#2  0x000055d4c87334e2 in errfinish (dummy=<optimized out>) at elog.c:763
#3  0x00007f41bb449130 in gpdbwritableformatter_import (fcinfo=<optimized out>) at src/gpdbwritableformatter.c:883
```

external-table/src/gpdbwritableformatter.c
```diff
@@ -663,6 +669,23 @@ gpdbwritableformatter_export(PG_FUNCTION_ARGS)
 	PG_RETURN_BYTEA_P(myData->export_format_tuple->data);
 }
 
+static void
+parse_params(FunctionCallInfo fcinfo, format_t *myData) {
```
bradfordb-vmware, line 673: Should we be strict here in parsing extra options to the formatter? If we encounter anything other than `dump_core_on_error`, should we report an error? This might help catch user typos (e.g., `dump_core_on_eror`).

external-table/src/gpdbwritableformatter.c
```diff
@@ -716,6 +740,9 @@ gpdbwritableformatter_import(PG_FUNCTION_ARGS)
 		myData->outlen = palloc(sizeof(int) * ncolumns);
 		myData->typioparams = (Oid *) palloc(ncolumns * sizeof(Oid));
 		myData->io_functions = palloc(sizeof(FmgrInfo) * ncolumns);
+		myData->formatter_error_level = ERROR;
```
bradfordb-vmware, line 743: We should mirror this in the formatter export function.

external-table/src/gpdbwritableformatter.c
```diff
@@ -80,6 +80,19 @@ typedef struct
...
+	 * PANIC leve; by default log at the ERROR level.
+	 */
+	int		   formatter_error_level;
+
+	/* When always_verify_table_definition='1', the import formatter function
+	 * will verify the column type information inside every tuples' GPDB
+	 * writable record. If there is a mis-match, then an error will be
+	 * reported.
+	 */
+	bool	    always_verify_table_definition;
```
denalex, line 95: The name is misleading at first reading. Without knowing what it does, one can assume that we want to  "always verify table definition" which for me means when a table is defined (`create table ...`) we want to verify it. And we want to do it `always` as opposed to `sometimes` ?  I think we need a better name here.

external-table/src/gpdbwritableformatter.c
```diff
@@ -80,6 +80,19 @@ typedef struct
 	 */
 	FmgrInfo   *io_functions;
 	Oid		   *typioparams;
+
+	/* When dump_core_on_error='1' and a malformed tuple is encounted, the
+	 * query will fail and the PXF formatter will log an error message at the
+	 * PANIC leve; by default log at the ERROR level.
+	 */
```
denalex, line 87: I'd start the comment by describing the meaning of the variable first and then give examples on when it will take diff values. Also `dump_core_on_error` is mentioned here, but it is unclear what it refers to.

external-table/src/gpdbwritableformatter.c
```diff
@@ -427,6 +440,26 @@ verifyExternalTableDefinition(int16 ncolumns_remote, AttrNumber nvalidcolumns, A
...
+		const char *key = FORMATTER_GET_NTH_ARG_KEY(fcinfo, i + 1);
+		const char *val = FORMATTER_GET_NTH_ARG_VAL(fcinfo, i + 1);
+
+		if (strcmp("dump_core_on_error", key) == 0 && strcmp("1", val) == 0)
+		{
+			myData->formatter_error_level = PANIC;
+		}
+		if (strcmp("always_verify_table_definition", key) == 0 && strcmp("1", val) == 0) {
+			myData->always_verify_table_definition = TRUE;
+		}
```
denalex, line 459: should we error if user provides any values other that "0" and "1" for both of these ?

external-table/src/gpdbwritableformatter.c
```diff
@@ -427,6 +440,26 @@ verifyExternalTableDefinition(int16 ncolumns_remote, AttrNumber nvalidcolumns, A
...
+parse_params(FunctionCallInfo fcinfo, format_t *myData) {
+	int nargs = FORMATTER_GET_NUM_ARGS(fcinfo);
+	ereport(DEBUG1, (errmsg("number of formatter args: %d", nargs)));
+
+	for (int i = 0; i < nargs; i++)
+	{
+		const char *key = FORMATTER_GET_NTH_ARG_KEY(fcinfo, i + 1);
+		const char *val = FORMATTER_GET_NTH_ARG_VAL(fcinfo, i + 1);
+
+		if (strcmp("dump_core_on_error", key) == 0 && strcmp("1", val) == 0)
```
denalex, line 453: since we are not actively dumping core on error and it comes as a side effect of PANIC log, should we name the parameter as such, like `error_log_level` and take the actual value like DEBUGX, ERROR, PANIC, etc ?

external-table/src/gpdbwritableformatter.c
```diff
@@ -427,6 +440,26 @@ verifyExternalTableDefinition(int16 ncolumns_remote, AttrNumber nvalidcolumns, A
...
 }
 
+static void
+parse_params(FunctionCallInfo fcinfo, format_t *myData) {
+	int nargs = FORMATTER_GET_NUM_ARGS(fcinfo);
+	ereport(DEBUG1, (errmsg("number of formatter args: %d", nargs)));
+
+	for (int i = 0; i < nargs; i++)
+	{
+		const char *key = FORMATTER_GET_NTH_ARG_KEY(fcinfo, i + 1);
```
denalex, line 450: are they `1` based or is it because formatter name comes first ? what if user puts these params first ?

external-table/src/gpdbwritableformatter.c
```diff
@@ -816,7 +859,7 @@ gpdbwritableformatter_import(PG_FUNCTION_ARGS)
 	ncolumns_remote = readInt2FromBuffer(data_buf, &bufidx);
 
 	/* Verify once on the first row */
-	if (FIRST_LINE_NUM == myData->lineno++)
+	if (FIRST_LINE_NUM == myData->lineno++ || myData->always_verify_table_definition)
```
denalex, line 862: why would we want to always verify this ?

external-table/src/gpdbwritableformatter.c
```diff
@@ -788,6 +830,7 @@ gpdbwritableformatter_import(PG_FUNCTION_ARGS)
 		}
 		FORMATTER_RETURN_NOTIFICATION(fcinfo, FMT_NEED_MORE_DATA);
 	}
+	tupleEndIdx = data_cur + tuplelen;
```
denalex, line 833: can we put a comment here, smth like `calculate the index of last byte of this tuple in the buffer`

external-table/src/gpdbwritableformatter.c
```diff
@@ -851,6 +894,10 @@ gpdbwritableformatter_import(PG_FUNCTION_ARGS)
 			{
 				bufidx = INTALIGN(bufidx);
 				myData->outlen[i] = readIntFromBuffer(data_buf, &bufidx);
+				if (myData->outlen[i] > tuplelen)
```
denalex, line 897: can we be more precise, since we already have advanced in the buffer when we read the values of proceeding columns ? `tuplelen` is the max, but we really should be comparing to `tuplelen - bufidx`

server/pxf-api/src/main/java/org/greenplum/pxf/api/io/GPDBWritable.java
```diff
@@ -471,7 +471,8 @@ public void write(DataOutput out) throws IOException {
                     /* For text format, add 4byte length header. string is already '\0' terminated */
                     default: {
                         byte[] data = ((String) colValue[i]).getBytes(databaseEncoding);
-                        out.writeInt(colLength[i]);
+                        boolean isPoisonPill = ((String) colValue[i]).startsWith("poison pill");
+                        out.writeInt(isPoisonPill ? Integer.MAX_VALUE : colLength[i]);
```
denalex, line 475: I don't think we should have this in production code

bradfordb-vmware: This change is part of the `DO NOT MERGE` commit and is only here to allow for "testing" the changes in the C extension.

I'm trying to see if there is a way for us to have a test class like this that we load during automation tests to verify that malformed data doesn't crash the cluster.

external-table/src/gpdbwritableformatter.c
bradfordb-vmware: I have seen examples where the data buffer is malformed/corrupted in either: 1) the column values, or; 2) the column type info. If we don't call `verifyExternalTableDefinition()` then we won't detect any errors that occur in the column type info. In all the examples I have seen though, if the column type info is corrupted then the column values are also corrupted. This is adding an extra (optional) check of the column type info. This matches what we do on the Java side -- for every received tuple, we extract the column type info from the GPDB writable structure. On the Java side, we have to do this because we don't have the equivalent of `tupdesc` that we can use.

external-table/src/gpdbwritableformatter.c
bradfordb-vmware: That's easy enough to add and my original draft include a `TODO` note about whether we should be strict in checking the passed in options. I'll add that.

external-table/src/gpdbwritableformatter.c
bradfordb-vmware: We could do that. My hesitation with doing that is that it feels like we are exposing an implementation detail through the formatter option. Since the option is a user-facing configuration, I thought it would be better to use something that describes the outcome -- should a core dump be generated when an error is encountered? That this is achieved by changing the level we log at is (in my opinion) an implementation detail.

external-table/src/gpdbwritableformatter.c
bradfordb-vmware: It looks like the formatter macros `FORMATTER_GET_NTH_ARG_KEY` and `FORMATTER_GET_NTH_ARG_VALUE` are `1`-based: [formatter.h](https://github.com/greenplum-db/gpdb/blob/22aa86d3f2a7a0536b4ed621ce6e406ff7158eb2/src/include/access/formatter.h#L86-L87)

external-table/src/gpdbwritableformatter.c
bradfordb-vmware: What this boolean is configuring, is whether for every tuple/GPDB writable structure we receive from PXF should we check that the table's column definitions match with the column type info in the GPDB wrtitable structure or not. What about a boolean `verify_column_types_every_row`? Or we could do an enum `verify_column_types` which takes values of `first` (default) or `all`.

external-table/src/gpdbwritableformatter.c
```diff
@@ -815,11 +819,7 @@ gpdbwritableformatter_import(PG_FUNCTION_ARGS)
 
 	ncolumns_remote = readInt2FromBuffer(data_buf, &bufidx);
 
-	/* Verify once on the first row */
-	if (FIRST_LINE_NUM == myData->lineno++)
-		verifyExternalTableDefinition(ncolumns_remote, nvalidcolumns, ncolumns, tupdesc, data_buf, &bufidx);
-	else /* Skip the columns' enum types */
-		bufidx += ncolumns_remote;
+	verifyExternalTableDefinition(ncolumns_remote, nvalidcolumns, ncolumns, tupdesc, data_buf, &bufidx);
```
ashwinstar, line 822: I am guessing the `bufidx` calculation is done correctly in `verifyExternalTableDefinition()` irrespective of the first row or not, as the previous code had logic here to calculate ` bufidx` for all other rows.

external-table/src/gpdbwritableformatter.c
```diff
@@ -789,6 +790,9 @@ gpdbwritableformatter_import(PG_FUNCTION_ARGS)
 		FORMATTER_RETURN_NOTIFICATION(fcinfo, FMT_NEED_MORE_DATA);
 	}
 
+	/* calculate the index of last byte of this tuple in data_buf */
+	tupleEndIdx = data_cur + tuplelen;
```
ashwinstar, line 794: Any reason to introduce the variable, like are the values changing for `data_cur` and `tuplelen` changing before `tupleEndIdx` is used next time? Would be good to state that reason why this has to be done here.

ashwinstar (approved):
:+1:

external-table/src/gpdbwritableformatter.c
bradfordb-vmware: `data_cur` and `tuplelen` are constant at this point. We introduced the variable `tupleEndIndex` because we reference this value inside of a `for`-loop so we wanted to pre-compute it.

external-table/src/gpdbwritableformatter.c
bradfordb-vmware: Yep, `bufIdx` is incremented inside of `verifyExternalTableDefinition`.

external-table/src/gpdbwritableformatter.c
denalex: Does `tuplelen` also takes into account the 4 bytes that store tuple length ? Or rather, in gpdbwritable format, when the tuple length is reported as 100 in the header, does it mean that is is 4 bytes for length and 96 bytes for everything else ?

bradfordb-vmware: Yes, `tuplelen` includes includes four bytes for itself; this is what things look like after we get `tuplelen` by calling `readIntFromBuffer`

```
            tuplelen
              v
         +---------+--------------------+
data_buf | 4 bytes | tuplelen - 4 bytes |
         +---------+--------------------+
         ^         ^                    ^
         |         |> buf_idx           |
         |> data_cur                    |> tupleEndIdx = data_cur + tuplelen
```

denalex:
The big comment in PR describing DDL is no longer applicable. I think we need to remove it and add a description that can be used in the documentation.

bradfordb-vmware:
I have re-based the PR in order to squash the numerous fix-up commits and to re-word the commit message, which is copied here:

Add data buffer boundary checks to PXF extension

If a GPDBWritable message is malformed, it can cause PXF's custom formatter to attempt a read outside of the valid boundaries of the passed in data buffer. This results in a SIGSEGV that panics the segment and leads to a failover.

This commit adds some basic boundary checks before reading values out of the received data and reports an error if the length would exceed the valid range of indices for the passed in data buffer. Additionally, this commit updates the formatter code to check that the received column types are consistent with the table's column descriptors for every received tuple. Previously, this check was only done for the first received tuple. Finally, this commit also checks that the actual length of every received string matches the length given in the GPDBWritable message.

Error are logged at the FATAL level, which will keep the backend around for a short amount of time so that someone can capture a core dump with either gcore or gdp. The amount of time to keep the process around is configured via gp_debug_linger.

PR #818, opened 2022-06-23
Add support for writing ORC primitive types

ashuka24:
This PR adds support for writing ORC files for the following types: 
- boolean
- bytea
- int
- bigint
- smallint
- real
- float
- double
- text
- bpchar
- varchar
- date
- time
- timestamp
- timestamp with timezone
- numeric
- uuid


server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedAccessor.java
```diff
@@ -144,18 +167,58 @@ public void closeForRead() throws IOException {
...
+        // get user-specified compression or use a default one (ZLIB) if not explicitly provided
+        String compressionCodec = context.getOption("COMPRESSION_CODEC", (String) OrcConf.COMPRESS.getDefaultValue());
+        CompressionKind compressionKind = CompressionKind.valueOf(compressionCodec.toUpperCase());
+        LOG.debug("Using compression: {}", compressionKind);
+        orcWriterOptions.compress(compressionKind);
+
+        // check whether to write timestamps in UTC or local timezone, this will be stored in the stripe / file footer and
+        // is important for file readers as Hive 3.1+ will read the value and perform time shifts if necessary
+        String writerTimeZone = getWriterTimezone(WRITE_TIMEZONE_ORC_PROPERTY_NAME, WRITE_TIMEZONE_PROPERTY_NAME);
+        if (writerTimeZone.equalsIgnoreCase(TIMEZONE_UTC_VALUE)) {
```
hpandeycodeit, line 192: what happens when `getWriterTimezone` returns `local` instead of `UTC` ?

automation/src/test/java/org/greenplum/pxf/automation/features/orc/OrcWriteTest.java
```diff
@@ -0,0 +1,310 @@
...
+
+import java.io.File;
+import java.util.ArrayList;
+import java.util.Arrays;
+
+public class OrcWriteTest extends BaseFeature {
+
+    private static final String[] ORC_PRIMITIVE_TABLE_COLUMNS = {
+            "id     integer"      ,
+            "col00  boolean"      , // DataType.BOOLEAN
```
bradfordb-vmware, line 24: I think some/a lot of the other automation tests uses columns names that indicate what the column type is, e.g., `tmst` for a timestamp column. After reviewing the expected output in the answer files, I have a preference for following that convention since it helps with understanding what the columns should look like.

automation/src/test/java/org/greenplum/pxf/automation/features/orc/OrcWriteTest.java
```diff
@@ -0,0 +1,310 @@
...
+
+    @Override
+    public void beforeMethod() throws Exception {
+        filesToDelete = new ArrayList<>();
+        publicStage = "/tmp/publicstage/pxf/";
+    }
+
+    @Test(groups = {"features", "gpdb", "security", "hcfs"})
+    public void orcWritePrimitives() throws Exception {
+        gpdbTable = "orc_primitive_types";
```
bradfordb-vmware, line 134: I would call this `gpdbTableNamePrefix` since the code will append either `_readable` or `_writable` to this value before creating the table.

automation/src/test/java/org/greenplum/pxf/automation/features/orc/OrcWriteTest.java
```diff
@@ -0,0 +1,310 @@
...
+                .append("col13, ") // numeric
+                .append("cast(col14 as string) as col14 ") // uuid cast as string
+                .append("FROM " + hiveTable.getFullName());
+
+        hive.runQuery("DROP TABLE IF EXISTS orc_primitive_types_hive_ctas");
+        hive.runQuery(ctasHiveQuery.toString());
+
+        // use the Hive JDBC profile to avoid using the PXF ORC reader implementation
+        String jdbcUrl = HIVE_JDBC_URL_PREFIX + hive.getHost() + ":10000/default";
+        ExternalTable exHiveOrcTable = TableFactory.getPxfJdbcReadableTable(
```
bradfordb-vmware, line 175: Since there is a Hive ORC profile, I think a clearer name would be `extHiveJdbcTable`.

bradfordb-vmware (commented):
This is an initial set of comments for this PR -- I still need to review `ORCVectorizedResolverWriteTest`.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedResolver.java
```diff
@@ -166,7 +177,7 @@ public List<List<OneField>> getFieldsForBatch(OneRow batch) {
                         .getNullResultSet(columnDescriptor.columnTypeCode(), batchSize);
             } else {
                 TypeDescription orcColumn = positionalAccess
-                        ? columnIndex < readSchema.getChildren().size() ? readSchema.getChildren().get(columnIndex) : null
+                        ? columnIndex < orcSchema.getChildren().size() ? orcSchema.getChildren().get(columnIndex) : null
```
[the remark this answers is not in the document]
denalex: this is part of the `read` flow. We check for schema presence in request in `ensureReadFunctionsAreInitialized` function so here it is assumed to be not null.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedResolver.java
```diff
@@ -221,90 +281,122 @@ public OneRow setFields(List<OneField> record) {
...
+            throw new PxfRuntimeException("No schema detected in request context");
+        }
+
+        OrcFile.WriterOptions writerOptions = (OrcFile.WriterOptions) context.getMetadata();
+        orcSchema = writerOptions.getSchema();
+        List<TypeDescription> columnTypeDescriptions = orcSchema.getChildren();
+        int schemaSize = columnTypeDescriptions.size();
+        writeFunctions = new TriConsumer[schemaSize];
+
+        for (int i=0; i < schemaSize; i++) {
```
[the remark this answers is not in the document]
denalex: Yes, this is a good point, we have been bitten by this before. Given a list, iterate using an iterator (for-each loop), not index-based access. I will make the change.

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/orc/ORCSchemaBuilderTest.java
```diff
@@ -0,0 +1,205 @@
...
+    @Test
+    public void testAllSupportedTypes() {
+        columnDescriptors = buildAllTypes();
+        assertEquals(ALL_TYPES_SCHEMA, builder.buildSchema(columnDescriptors).toString());
+    }
+
+    @Test
+    public void testBpcharMaxLength() {
+        columnDescriptors.add(new ColumnDescriptor("col0", DataType.BPCHAR.getOID(), 0, "", new Integer[]{}));
+        assertEquals("struct<col0:char(256)>", builder.buildSchema(columnDescriptors).toString());
```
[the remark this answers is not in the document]
denalex: Yes, default `maxLength` for ORC "string" types is 256 and it will be used if Greenplum schema definition lacks explicit limit. We could use a different default, but none will be correct if GP can take an unlimited string. What else can we do ?

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/orc/ORCSchemaBuilderTest.java
```diff
@@ -0,0 +1,205 @@
...
+        columnDescriptors.add(new ColumnDescriptor("col0", DataType.NUMERIC.getOID(), 0, "", new Integer[]{}));
+        assertEquals("struct<col0:decimal(38,10)>", builder.buildSchema(columnDescriptors).toString());
+
+        columnDescriptors.clear();
+        columnDescriptors.add(new ColumnDescriptor("col0", DataType.NUMERIC.getOID(), 0, "", new Integer[]{20,5}));
+        assertEquals("struct<col0:decimal(20,5)>", builder.buildSchema(columnDescriptors).toString());
+
+        // precision is null, scale is ignored
+        columnDescriptors.clear();
+        columnDescriptors.add(new ColumnDescriptor("col0", DataType.NUMERIC.getOID(), 0, "", new Integer[]{null,5}));
```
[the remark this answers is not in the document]
denalex: strictly speaking yes, should be an error. Our current code does not even look into whether scale was provided if precision is missing or passed as null. we can add extra check, even though this will likely never happen

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedAccessorTest.java
```diff
@@ -119,22 +129,139 @@ public void testReadUnorderedSubsetOfOrcTypesFile() throws IOException {
...
+
+        assertTrue(result);
+        verify(mockWriter).addRowBatch(mockBatch);
+        // verify we do not reset the batch in accessor, it should be done in resolver, where it is created
+        verify(mockBatch, never()).reset();
+    }
+    @Test
+    public void testCloseForWrite() throws IOException {
+        accessor.getWriterState().setFileWriter(null);
+        accessor.closeForWrite();
```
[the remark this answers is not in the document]
denalex: added a comment

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCSchemaBuilder.java
```diff
@@ -0,0 +1,142 @@
...
+import org.greenplum.pxf.api.utilities.ColumnDescriptor;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.List;
+
+/**
+ * Class for building an ORC schema from a Greenplum table definition.
+ */
+public class ORCSchemaBuilder {
```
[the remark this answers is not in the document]
denalex: Yes, it can be made with static methods. However, using this pattern in Spring application is usually frowned upon, but on the other hand we are not injecting it anywhere. Also, for testing it would be easier to mock it, but the current code that consumes it does not store it, so we are not mocking it either. While we can totally turn it into a static utility-like class, we can as well let it stay as it is.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedAccessor.java
```diff
@@ -252,4 +315,40 @@ private TypeDescription buildReadSchema(TypeDescription originalSchema) {
...
+     * Determines the timezone that ORC writer will store to the stripe footer as the writer timezone. Timestamps will
+     * be interpreted as instants in the writer timezone. The method searches for the value by looking up configuration
+     * properties with the names proved. If the first property is not defined, then the next property is looked up.
+     * Only "local" and "UTC" are considered to be valid values, case-insensitive.
+     * @param propertyNames names of the configuration properties
+     * @return name of the writer timezone specified in the configuration, or UTC if none are found
+     */
+    private String getWriterTimezone(String ...propertyNames) {
+        String timezone = null;
+        for (String propertyName : propertyNames) {
```
hpandeycodeit, line 338: Do we need this `for loop` since `propertyNames` here is a list two params `pxf.write.timezone` and `pxf.write.timezone.orc` ?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedResolver.java
```diff
@@ -198,6 +209,55 @@ public List<List<OneField>> getFieldsForBatch(OneRow batch) {
...
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public OneRow setFieldsForBatch(List<List<OneField>> records) {
+        if (CollectionUtils.isEmpty(records)) {
+            return null; // this will end bridge iterations
+        }
+        // make sure provided record set can fit into a single batch, we do not want to produce multiple batches here
+        if (records.size() > getBatchSize()) {
```
hpandeycodeit, line 226: In case of records.size()> 1024, the user has to reduce the number of records ?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedResolver.java
```diff
@@ -221,90 +281,122 @@ public OneRow setFields(List<OneField> record) {
...
+     * Ensures that functions used in write use case are initialized. If not initialized, this method will
+     * initialize the functions by iterating over the ORC schema and getting a corresponding function for each column.
+     */
+    @SuppressWarnings("unchecked")
+    private void ensureWriteFunctionsAreInitialized() {
+        if (writeFunctions != null) {
+            return;
+        }
+        if (context.getMetadata() == null || !(context.getMetadata() instanceof OrcFile.WriterOptions)) {
+            throw new PxfRuntimeException("No schema detected in request context");
```
hpandeycodeit, line 385:
```suggestion
            throw new PxfRuntimeException("No ORC schema detected in request context");
```

same at line 287

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedAccessor.java
denalex: this is a more generic implementation that allows for arbitrary chains of properties.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedResolver.java
denalex: not the user, this is catching code bug that can be introduced later. The caller of this method is supposed to inquire ahead of time what the batch size is and provide a list of this size to us here. If the list is larger, we are erring out, as we do not want to do streaming at this level.

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/orc/ORCSchemaBuilderTest.java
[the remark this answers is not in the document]
bradfordb-vmware: ORC has three string types: `string`, `char`, and `varchar`. I think only `char` and `varchar` have specified maximum lengths ([ORC Specification v1](https://orc.apache.org/specification/ORCv1/)), while string does not. Since `text` and `varchar` (w/o length specifier) are equivalent in Postgres, I think it would make sense to map them to the same ORC type, namely `string`.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCSchemaBuilder.java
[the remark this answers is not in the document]
denalex: decided to make it a static utility class

hpandeycodeit (approved):

bradfordb-vmware (approved):

PR #823, opened 2022-06-30
Add unsupported exception in case of hive write

hpandeycodeit:
`hive:text` does not support write. If we test if currently for write, It throws `NullPointerException` 

```

CREATE WRITABLE EXTERNAL TABLE hive_curl_partial_test_w(col1 text, col2 date) LOCATION ('pxf://curl_partial_test?profile=hive:text') FORMAT 'CUSTOM' (FORMATTER='pxfwritable_export');

INSERT INTO hive_curl_partial_test_w (col1, col2) VALUES ('gp-row1', '2021-12-01');

```

pxf-servive.logs: 

```
Caused by: java.lang.NullPointerException
        at org.greenplum.pxf.plugins.hive.HiveAccessor.afterPropertiesSet(HiveAccessor.java:241) ~[pxf-hive-6.3.1-SNAPSHOT.jar!/:?]
        ... 78 more
```

This PR changes the exception it to `UnsupportedOperationException`

### DIFF
--- a/server/pxf-hive/src/main/java/org/greenplum/pxf/plugins/hive/HiveAccessor.java
+++ b/server/pxf-hive/src/main/java/org/greenplum/pxf/plugins/hive/HiveAccessor.java
@@ -214,6 +214,10 @@ public class HiveAccessor extends HdfsSplittableDataAccessor {
     public void afterPropertiesSet() {
         super.afterPropertiesSet();
 
+        // HiveAccessor requires fragment metadata which is available for read operations but not write operations
+        if(context.getRequestType() == RequestContext.RequestType.WRITE_BRIDGE)
+            throw new UnsupportedOperationException(UNSUPPORTED_ERR_MESSAGE);
+
         // determine if predicate pushdown is allowed by configuration
         isPredicatePushdownAllowed = configuration.get(PXF_PPD_HIVE, "true").equalsIgnoreCase("true");
         if (isPredicatePushdownAllowed) {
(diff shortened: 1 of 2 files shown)


server/pxf-hive/src/main/java/org/greenplum/pxf/plugins/hive/HiveAccessor.java
```diff
@@ -236,13 +237,17 @@ public void afterPropertiesSet() {
         Properties properties;
         try {
             HiveFragmentMetadata metadata = context.getFragmentMetadata();
+            if(metadata == null && context.getRequestType().equals(RequestContext.RequestType.WRITE_BRIDGE))
```
bradfordb-vmware, line 240: I wonder if we should move this check closer to the start of the function and include some comments about why this accessor needs the check.

```java
    super.afterPropertiesSet();

    /*
     * HiveAccessor requires fragment metadata which is available for read operations but not write operations
     */
    if (contextgetRequestType() == RequestContext.RequestType.WRITE_BRIDGE) {
        throw new UnsupportedOperationException(UNSUPPORTED_ERR_MESSAGE);
    }
    ...
}
```

server/pxf-hive/src/main/java/org/greenplum/pxf/plugins/hive/HiveAccessor.java
```diff
@@ -213,6 +214,10 @@ public HiveAccessor() {
     public void afterPropertiesSet() {
         super.afterPropertiesSet();
 
+        // HiveAccessor requires fragment metadata which is available for read operations but not write operations
+        if(context.getRequestType() == RequestContext.RequestType.WRITE_BRIDGE)
```
denalex, line 218:
```suggestion
        if (context.getRequestType() == RequestContext.RequestType.WRITE_BRIDGE)
```

denalex (approved):

bradfordb-vmware (approved):

PR #831, opened 2022-07-20
Added check if the PXF PID exists before trying a start/restart in PXF bash script.  

hpandeycodeit:
This is related to github issue https://github.com/greenplum-db/sre-test/issues/274

In case the PXF pid doesn't exists, the PXF specific error to be thrown and not the OS one. 

Currently if the pid doesn't exists, the command `ps -p ""` runs which returns `error: list of process IDs must follow -p` which is not very useful.

### DIFF
--- a/server/pxf-service/src/scripts/pxf
+++ b/server/pxf-service/src/scripts/pxf
@@ -159,6 +159,7 @@ function checkWebapp()
 # determines whether the application is running
 function isRunning()
 {
+    [[ -z "$1" ]] && { return 1; }
     ps -o pid,command -p "$1" | grep 'pxf-app' &> /dev/null
 }
 
--- a/server/pxf-service/src/scripts/pxf
+++ b/server/pxf-service/src/scripts/pxf
@@ -89,6 +89,15 @@ function validate_user()
 
 function checkPermissions()
 {
+    # TODO
+    # Currently, in case of any error if PXF doesn't come up, the empty PID file created here remains.
+    # If the pxf is restarted again after the failure, this empty PID file will cause issues. For e.g.
+    # isRunning function will try to ps on the empty PID which will fail with the error "error: list of process IDs must follow -p"
+    # To avoid that we can either
+    # 1. Check if the PID exists in the isRunning function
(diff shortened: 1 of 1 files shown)


server/pxf-service/src/scripts/pxf
```diff
@@ -159,6 +159,7 @@ function checkWebapp()
 # determines whether the application is running
 function isRunning()
 {
+    [[ -z "$1" ]] && { return 1; }
```
denalex, line 162: It seems this will return an error code if the parameter is not defined / empty, which in itself might be ok, but in the overall flow will likely stop further checking. If the pid file got deleted or corrupted, will this result in status api returning with "pxf down" even if the process is up and running ? If yes, is it what we want ?

hpandeycodeit:
We need more discussion around this as to find a better way to handle the stale PID. For now we can put this PR on hold. 

```
# Currently, in case of any error if PXF doesn't come up, the empty PID file created here remains.
# If the pxf is restarted again after the failure, this empty PID file will cause issues. For e.g.
# isRunning function will try to ps on the empty PID which will fail with the error "error: list of process IDs must follow -p"
# To avoid that we can either
# 1. Check if the PID exists in the isRunning function
# 2. Removes the PID file which is created here after checking for write permissions
# 3. Touch a file with another name to check for the write permissions in the directory
# 4. In case PXF doesn't come up for any reason, clean the stale PID
```

bradfordb-vmware:
@hpandeycodeit I've added a check that the PID file has a [non-zero length](https://github.com/greenplum-db/pxf/blob/remove-actuator-shutdown/server/pxf-service/src/scripts/pxf#L472-L476) in the branch for improving the CLI stability. Is it OK if we close this PR in favor of #881?

hpandeycodeit:
> @hpandeycodeit I've added a check that the PID file has a [non-zero length](https://github.com/greenplum-db/pxf/blob/remove-actuator-shutdown/server/pxf-service/src/scripts/pxf#L472-L476) in the branch for improving the CLI stability. Is it OK if we close this PR in favor of #881?

Sure. I will close it out.

PR #836, opened 2022-07-26
Add write support for one-dimensional ORC arrays

ashuka24:
This PR adds support for writing one dimensional arrays of the following primitive types to ORC files: 
- boolean
- bytea as BINARY
- int, bigint, smallint
- real, float, double
- text, bpchar, varchar
- date, timestamp, timestamp with timezone
- numeric
- time, uuid as STRING


server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/orc/OrcUtilitiesTest.java
```diff
@@ -0,0 +1,183 @@
...
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.assertArrayEquals;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertIterableEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+public class OrcUtilitiesTest {
+    private RequestContext context;
+    private String orcDirectory;
```
bradfordb-vmware, line 22: I don't think these are used/needed for this test case.

automation/src/test/java/org/greenplum/pxf/automation/features/orc/OrcWriteTest.java
```diff
@@ -85,6 +86,48 @@ public class OrcWriteTest extends BaseFeature {
             "c_numeric numeric"        , // DataType.NUMERIC
             "c_uuid text"                // DataType.UUID
     };
+
+    private static final String[] ORC_PRIMITIVE_ARRAYS_TABLE_COLUMNS = {
+            "id     integer"                ,
```
bradfordb-vmware, line 91: Style nit - can we add some spaces to align the column type definitions?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedMappingFunctions.java
```diff
@@ -480,6 +520,39 @@ private static void initWriteFunctionsMap() {
         });
     }
 
+    private static void initWriteListFunctionsMap() {
+        // the array write functions rely on the primitive write functions so they must be initialized first
+        // it is assumed that all arrays are one dimensional
```
denalex, line 525: can this be metadata-driven ?

```
for (TypeDescription.Category category : ARRAY_CATEGORIES) {
  writeListFunctionsMap.put(category, getListWriteFunction(category);
}
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedMappingFunctions.java
```diff
@@ -510,6 +583,54 @@ private static TriConsumer<ColumnVector, Integer, Object> getTimestampInLocalWri
         };
     }
 
+    /**
+     * A special function that writes lists to ORC file
```
denalex, line 587: need to add a javadoc for the function parameter

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedMappingFunctions.java
```diff
@@ -510,6 +583,54 @@ private static TriConsumer<ColumnVector, Integer, Object> getTimestampInLocalWri
...
+     */
+    // PXF currently only has the capability to infer an ORC schema from the GPDB schema, so all arrays will be one-dimensional.
+    private static TriConsumer<ColumnVector, Integer, Object> getListWriteFunction(TypeDescription.Category underlyingChildCategory) {
+        final TriConsumer<ColumnVector, Integer, Object> childWriteFunction = writeFunctionsMap.get(underlyingChildCategory);
+
+        return (columnVector, row, val) -> {
+            ListColumnVector listColumnVector = (ListColumnVector) columnVector;
+
+            List<Object> data = orcUtilities.parsePostgresArray((String) val, underlyingChildCategory);
+
```
denalex, line 598: do we want to explicitly check that `columnVector` is a `ListColumnVector` and that `val` is a `String` and produce helpful error messages instead of relying on class-cast exceptions ? Arguably, now the mismatch should not happen since we are auto-generating the schema, but if we accept the schema, then it can be mismatched. Or at least maybe put a TODO here to implement it later.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedMappingFunctions.java
```diff
@@ -510,6 +583,54 @@ private static TriConsumer<ColumnVector, Integer, Object> getTimestampInLocalWri
...
+            int length = data.size();
+            // the childCount refers to "the number of children slots used"
+            int offset = listColumnVector.childCount;
+
```
denalex, line 602: do we need to check that the `data` array is not null and not empty before trying to adjust the `offsets` and `lengths` vectors ?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedMappingFunctions.java
```diff
@@ -510,6 +583,54 @@ private static TriConsumer<ColumnVector, Integer, Object> getTimestampInLocalWri
...
+            // set the offset, length and new childCount for this row
+            listColumnVector.offsets[row] = offset;
+            listColumnVector.lengths[row] = length;
+            listColumnVector.childCount += length;
```
denalex, line 606: what if some child elements are NULL, do they still take a slot even if it will never be accessed since the isNull will be set to true ?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedMappingFunctions.java
```diff
@@ -510,6 +583,54 @@ private static TriConsumer<ColumnVector, Integer, Object> getTimestampInLocalWri
...
+            listColumnVector.child.ensureSize(listColumnVector.childCount, true);
```
denalex, line 607: this will increase size by the number of children in a row and will internally do a new array allocation and System.arraycopy() for both offsets and lengths arrays. This can kill performance when the initial capacity of 1024 is exhausted. 

An optimization would be to parse multiple (or all) records at once, calculate total number of children and ensure capacity once (or minimize the number of times it is called). We will need to prove the performance problem / impact first though by testing with rows of 10/100/1000 elements in array (so that we run out of slots after processing 100/10/1 rows). 

Another optimization is to ensure excess capacity in anticipation of more data (say ensure 2x capacity). Even smarter will be to ensure excess capacity as a function of the current rowId -- see how many slots we used for `rowId` rows, assume the same distribution and ensure we have enough for `batchSize` rows.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedMappingFunctions.java
```diff
@@ -510,6 +583,54 @@ private static TriConsumer<ColumnVector, Integer, Object> getTimestampInLocalWri
...
+
+            // add the data to the child columnvector
+            ColumnVector childColumnVector = listColumnVector.child;
+            int childIndex;
+            int childCounter = 0;
+            for (Object rowElem : data) {
```
denalex, line 613:
```suggestion
            for (Object rowElement : data) {
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/OrcUtilities.java
```diff
@@ -0,0 +1,103 @@
...
+import org.greenplum.pxf.plugins.hdfs.utilities.PgUtilities;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+
+import java.util.ArrayList;
+import java.util.List;
+
+@Component
+public final class OrcUtilities {
```
denalex, line 15: need class description javadoc

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/OrcUtilities.java
```diff
@@ -0,0 +1,103 @@
...
+import org.greenplum.pxf.api.error.PxfRuntimeException;
+import org.greenplum.pxf.plugins.hdfs.utilities.PgUtilities;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+
+import java.util.ArrayList;
+import java.util.List;
+
+@Component
```
denalex, line 14: Does it need to be Spring-managed ? We are doing `new OrcUtilities(pgUtilities)` where we need it.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/OrcUtilities.java
```diff
@@ -0,0 +1,103 @@
...
+     * Parse a Postgres external format into a given ORC schema
+     *
+     * Re-used from GPHDFS
+     * https://github.com/greenplum-db/gpdb/blob/3b0bfdc169fab7f686276be7eccb024a5e29543c/gpAux/extensions/gphdfs/src/java/1.2/com/emc/greenplum/gpdb/hadoop/formathandler/util/FormatHandlerUtil.java
+     * @param val Postgres external format (the output of function named by typoutput in pg_type) or `null` if null value
+     * @param underlyingChildCategory Underlying type for ORC list. This functions assumes the ORC list is one-dimensional
+     * @return
+     */
+    public List<Object> parsePostgresArray (String val, TypeDescription.Category underlyingChildCategory) {
+        LOG.debug("child type={}, value={}, isTopLevel={}", underlyingChildCategory, val);
```
denalex, line 38: there should be 3 params to correspond to 3 `{}` in the log message

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/OrcUtilities.java
```diff
@@ -0,0 +1,103 @@
...
+
```
denalex, line 39: this log might be very verbose, should it be trace ?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/OrcUtilities.java
```diff
@@ -0,0 +1,103 @@
...
+        if (val == null) {
+            return null;
+        }
+
+        List<Object> data = new ArrayList<>();
```
denalex, line 44: we should allocate this ArrayList with the target capacity after the split has been done and we know how many elements are there.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/OrcUtilities.java
```diff
@@ -0,0 +1,103 @@
...
+        this.pgUtilities = pgUtilities;
+    }
+
+    /**
+     * Parse a Postgres external format into a given ORC schema
+     *
+     * Re-used from GPHDFS
+     * https://github.com/greenplum-db/gpdb/blob/3b0bfdc169fab7f686276be7eccb024a5e29543c/gpAux/extensions/gphdfs/src/java/1.2/com/emc/greenplum/gpdb/hadoop/formathandler/util/FormatHandlerUtil.java
+     * @param val Postgres external format (the output of function named by typoutput in pg_type) or `null` if null value
+     * @param underlyingChildCategory Underlying type for ORC list. This functions assumes the ORC list is one-dimensional
```
denalex, line 34:
```suggestion
     * @param underlyingChildCategory Underlying ORC type for ORC list. This functions assumes the ORC list is one-dimensional
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/OrcUtilities.java
```diff
@@ -0,0 +1,103 @@
...
+        }
+        return data;
+    }
+
+    /**
+     * Parse a string based off the ORC type.
+     *
+     * @param val String representation of value
+     * @param valType ORC type
+     * @return an java representation of the value for the given valType
```
denalex, line 62:
```suggestion
     * @return a Java representation of the value for the given valType
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/OrcUtilities.java
```diff
@@ -0,0 +1,103 @@
...
+     */
+    private Object decodeString(String val, TypeDescription.Category valType) {
```
denalex, line 64:
```suggestion
    private Object decodeString(String val, TypeDescription.Category orcType) {
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/OrcUtilities.java
```diff
@@ -0,0 +1,103 @@
...
+            case DATE:
+            case TIMESTAMP:
+            case TIMESTAMP_INSTANT:
+            case DECIMAL:
+                return val;
+            default:
+                throw new PxfRuntimeException(String.format("type: %s is not supported", valType));
+        }
+    }
+
```
denalex, line 95: let's add a javadoc

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/OrcUtilities.java
```diff
@@ -0,0 +1,103 @@
...
+    protected String createErrorHintFromValue(String val) {
+        if (StringUtils.startsWith(val, "{")) {
+            return "Value is a multi-dimensional array, PXF does not currently support multi-dimensional arrays for writing ORC files.";
```
denalex, line 98:
```suggestion
            return "Value is a multi-dimensional array, PXF does not support multi-dimensional arrays for writing ORC files.";
```

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/orc/OrcUtilitiesTest.java
```diff
@@ -0,0 +1,173 @@
...
+    public void testParsePostgresArrayValidBooleanArray() {
+        String value = "{t,f,t}";
+
+        List<Object> result = orcUtilities.parsePostgresArray(value, TypeDescription.Category.BOOLEAN);
+        assertEquals(Arrays.asList(true, false, true), result);
+    }
+
+    @Test
+    public void testParsePostgresArrayInValidBooleanArrayPXFGeneratedSchema() {
+        // this situation should never happen as the toString method of booleans (boolout) should not return a string in this format
```
denalex, line 113: what is `(boolout)` ?

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/orc/OrcUtilitiesTest.java
```diff
@@ -0,0 +1,173 @@
...
+
+    @Test
+    public void testParsePostgresArrayMultiDimensionalArrayInt() {
+        // test the underlying decode string: we expect it to fail and the failure to be caught by parsePostgresArray
+        // as we currently do not support writing multi-dimensional arrays
+        String value = "{{1,2},{3,4}}";
+
+        Exception exception = assertThrows(PxfRuntimeException.class, () -> orcUtilities.parsePostgresArray(value, TypeDescription.Category.INT));
+        assertEquals("Error parsing array element: {1,2} was not of expected type INT", exception.getMessage());
+        assertEquals("Value is a multi-dimensional array, PXF does not currently support multi-dimensional arrays for writing ORC files.", ((PxfRuntimeException) exception).getHint());
```
denalex, line 137: there's a comprehension mismatch for this error message / hint. We are reporting that an array element is not of an expected primitive type, but the hint says `Value is a multi-dimensional array`. I read this as `value` being shown in the error message, which is not a multi-dim array. Maybe we should say `column value` to set proper context ?

automation/src/test/java/org/greenplum/pxf/automation/features/orc/OrcWriteTest.java
```diff
@@ -19,72 +20,114 @@
...
+            "varchar_arr      varchar(7)[]" , // DataType.VARCHARARRAY
+            "date_arr         date[]"       , // DataType.DATEARRAY
+            "time_arr         time[]"       , // DataType.TIMEARRAY
+            "timestamp_arr    timestamp[]"  , // DataType.TIMESTAMPARRAY
+            "timestamptz_arr  timestamptz[]", // DataType.TIMESTAMP_WITH_TIME_ZONE
+            "numeric_arr      numeric[]"    , // DataType.NUMERICARRAY
+            "uuid_arr         uuid[]"         // DataType.UUIDARRAY
+    };
+
+    private static final String[] ORC_PRIMITIVE_ARRAYS_TABLE_COLUMNS_READ = {
```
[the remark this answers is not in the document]
ashuka24: there's comments on line 123 and 12, but `TIME` and `UUID` do not exist in the ORC schema so those are stored as text type. We could make the read and write schemas the same but it would require some internal modification of these types. This currently matches what we do for the primitive type, where the read and write schemas are different.

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/orc/OrcUtilitiesTest.java
ashuka24: This refers to the GPDB boolean output function.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedMappingFunctions.java
ashuka24: because the caller of this function (getColumnWriter) has a check for `TypeDescription.Category.LIST`, I do not think it is possible to have something besides a listcolumnvector here. but i will add a todo

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedMappingFunctions.java
ashuka24: When a ListColumnVector is initialized, the `offsets` and `lengths` vectors are prepopulated with zeros. When there is an empty data array, the length will be null and we do not need to be worried there. in the case of offset, this value marks the starting point of the array, so having an offset value with a length of zero is ok.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedMappingFunctions.java
```diff
@@ -158,8 +164,34 @@ public static String serializeListRow(ListColumnVector columnVector, int row, in
...
+                    } else {
+                        pgArrayBuilder.addElement(buf -> columnVector.child.stringifyValue(buf, childRow));
+                    }
+                    break;
+                case TIMESTAMP:
+                    TimestampColumnVector childVector = (TimestampColumnVector) columnVector.child;
+                    String val = null;
+
+                    if (childVector.noNulls || !childVector.isNull[childRow]) {
+                        DateTimeFormatter formatter = oid == DataType.TIMESTAMP_WITH_TIMEZONE_ARRAY.getOID()
```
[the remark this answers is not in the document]
ashuka24: at this point, we do not have access to the orc schema. however, the oid value that is passed in here is take directly from the orc schema when we first initialize this function (see ORCVectorizedResolver.java:368)

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedMappingFunctions.java
ashuka24: The offsets and lengths here are dependent on the the array data itself. These will also have values for every row. If the entire array is null, the the length will be zero. 

If the array element is null, then the child's start and length vectors at that index will be skipped over. 

For example, if we had 3 rows with values
```
id |  array
---------
 1 | {hello, world}
 2 | NULL
 3 | {test, NULL, string}
```

We would have something like the following 
```
id | lcv.isNull | lcv.offsets | lcv.lengths ... child.isNull | child.start | child.length | child.vector
--------------------------------------------------------------------------------------------------------
 1 | false      |           0 |           2 ...        false |           0 |            5 |        hello
 2 | true       |           2 |           0 ...        false |           5 |            6 |        world
 3 | false      |           2 |           3 ...        false |          11 |            4 |         test  
                                                        true |           0 |            0 |         NULL
                                                       false |          15 |            6 |       string   
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedMappingFunctions.java
```diff
@@ -399,7 +431,12 @@ public static TriConsumer<ColumnVector, Integer, Object> getColumnWriter(TypeDes
         // if timestamps need to be written in local timezone (and not in UTC), get a special function not in the map
         if (columnTypeCategory.equals(TypeDescription.Category.TIMESTAMP) && !timestampsInUTC) {
             writeFunction = timestampInLocalWriteFunction;
-        } else {
+        } else if (columnTypeCategory.equals(TypeDescription.Category.LIST)) {
+            // pass along the underlying category of the list
+            TypeDescription childTypeDescription = typeDescription.getChildren().get(0);
+            writeFunction = writeListFunctionsMap.get(childTypeDescription.getCategory());
```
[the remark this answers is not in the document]
ashuka24: because we use the write function of the primitive type, which handles the difference between UTC and local timezone, we do not have to worry about that here for lists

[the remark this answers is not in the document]
denalex: even the function for the primitive timestamp is determined here, and cannot be simply retrieved from the functionsMap

external-table/src/pxfheaders.c
```diff
@@ -295,11 +299,17 @@ add_tuple_desc_httpheader(CHURL_HEADERS headers, Relation rel)
...
 				case BITOID:
+				case 1561: // bit array oid
 				case TIMESTAMPOID:
+				case 1115: // timestamp array oid
 				case TIMESTAMPTZOID:
+				case 1185: // timestamp with timezone array oid
 				case TIMEOID:
+				case 1183: //time array oid
 				case TIMETZOID:
+				case 1270: //time with timezone array oid
```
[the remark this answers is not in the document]
ashuka24: i don't believe we do currently, but for consistency with the primitive cases, I have preemptively added all array oids for the listed primitive types in this function

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedMappingFunctions.java
```diff
@@ -158,8 +164,34 @@ public static String serializeListRow(ListColumnVector columnVector, int row, in
                         pgArrayBuilder.addElement(((BytesColumnVector) columnVector.child).toString(childRow));
                     }
                     break;
+                case LONG:
+                    // DateColumnVector extends LongColumnVector but does not override the stringifyValue function to print the
+                    // date in human-readable format (i.e. yyyy-MM-dd) so do that here
+                    if (columnVector.child instanceof DateColumnVector) {
+                        DateColumnVector childVector = (DateColumnVector) columnVector.child;
```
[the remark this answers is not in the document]
ashuka24: For all of these cases, the function called while adding the row to the array handles the repeated value case.
See: `getByteBuffer`, `stringifyDateColumnVectorValue`, `stringifyTimestampColumnVectorValue`, `stringifyValue`, etc

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedMappingFunctions.java
```diff
@@ -511,6 +615,76 @@ private static TriConsumer<ColumnVector, Integer, Object> getTimestampInLocalWri
...
+            listColumnVector.offsets[row] = offset;
+            listColumnVector.lengths[row] = length;
+            listColumnVector.childCount += length;
+            // if the number of children slots needed is greater than the current size
+            if (listColumnVector.childCount > listColumnVector.child.isNull.length) {
+                // reallocate to ensure that the columnvector can hold the data
+                // use the average row size to calculate what the new size should be
+                // (childCount/row) * (batchsize - row) => assume average size and # of rows
+                float avgRow = listColumnVector.childCount / (row+1);
+                listColumnVector.child.ensureSize(Math.round(avgRow) * (VectorizedRowBatch.DEFAULT_SIZE - (row+1)),   true);
```
denalex, line 660: shouldn't we give the overall new size required, not the increment ?
```suggestion
                listColumnVector.child.ensureSize(oldChildSize + Math.ceil(avgRow) * (listColumnVector.length - (row+1)),   true);
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedMappingFunctions.java
```diff
@@ -511,6 +615,76 @@ private static TriConsumer<ColumnVector, Integer, Object> getTimestampInLocalWri
...
+            listColumnVector.offsets[row] = offset;
+            listColumnVector.lengths[row] = length;
+            listColumnVector.childCount += length;
+            // if the number of children slots needed is greater than the current size
+            if (listColumnVector.childCount > listColumnVector.child.isNull.length) {
+                // reallocate to ensure that the columnvector can hold the data
+                // use the average row size to calculate what the new size should be
+                double avgRow = (double) listColumnVector.childCount / (row + 1);
+                int batchSize = listColumnVector.isNull.length;
+                listColumnVector.child.ensureSize((int) Math.ceil(avgRow * batchSize), true);
```
bradfordb-vmware, line 660: I was thinking we might want to include a debug log message here that includes the values of `childCount`, `row`, `batchSize`, and `avgRow * batchSize`.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedMappingFunctions.java
```diff
@@ -511,6 +615,76 @@ private static TriConsumer<ColumnVector, Integer, Object> getTimestampInLocalWri
...
+
+            // set the offset, length and new childCount for this row
+            listColumnVector.offsets[row] = offset;
+            listColumnVector.lengths[row] = length;
+            listColumnVector.childCount += length;
+            // if the number of children slots needed is greater than the current size
+            if (listColumnVector.childCount > listColumnVector.child.isNull.length) {
+                // reallocate to ensure that the columnvector can hold the data
+                // use the average row size to calculate what the new size should be
+                double avgRow = (double) listColumnVector.childCount / (row + 1);
```
denalex, line 658:
```suggestion
                double averageChildrenPerRow = (double) listColumnVector.childCount / (row + 1);
```

denalex (approved):

bradfordb-vmware (approved):

PR #838, opened 2022-07-29
Upgrade org.xerial.snappy:snappy-java to 1.1.8.4

bradfordb-vmware:
Support for Apple Silicon was added in 1.1.8.2

### DIFF
--- a/server/build.gradle
+++ b/server/build.gradle
@@ -112,7 +112,7 @@ configure(javaProjects) {
             dependency("org.threeten:threeten-extra:1.5.0")
             dependency("org.tukaani:xz:1.8")
             dependency("org.wildfly.openssl:wildfly-openssl:1.0.7.Final")
-            dependency("org.xerial.snappy:snappy-java:1.0.5")
+            dependency("org.xerial.snappy:snappy-java:1.1.8.4")
 
             // Hadoop dependencies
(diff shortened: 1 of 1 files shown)


ashuka24 (approved):
there was a PR for aws and now this one for snappy that provides support for Apple Silicon. Are there any other dependencies that we need to update that we could include in this PR?

server/build.gradle
```diff
@@ -112,7 +112,7 @@ configure(javaProjects) {
             dependency("org.threeten:threeten-extra:1.5.0")
             dependency("org.tukaani:xz:1.8")
             dependency("org.wildfly.openssl:wildfly-openssl:1.0.7.Final")
-            dependency("org.xerial.snappy:snappy-java:1.0.5")
+            dependency("org.xerial.snappy:snappy-java:1.1.8.4")
```
denalex, line 115: a question to ask is : "If we use this newer library, compress the data with it and write the file to Hadoop, can a downstream consumer using default (presumably older) version included in Hadoop stack to properly decompress it ?"

bradfordb-vmware: I'm pretty sure the answer is yes. I used PXF (with this PR's change included) to write Parquet data with snappy compression to HDFS. I then used Hive 1.2.1000 (HDP, which includes snappy 1.0.5) to check if I could read the data and I was able to.

denalex (approved):
thank you for testing the change

PR #840, opened 2022-08-03
Pass JDBC connection pool parameters from LOCATION URI part

x4m:


reshke:
Hi x4mmm, hope you are doing well.
Please provide here more detailed changes description:)

x4m:
@reshke I hope you are doing well too :)
Thanks for the comment, I've added clarification of what is going on.
Have a nice day.

denalex:
Can you please explain in more details why you would want to set connection pool properties in the table URI ? Currently PXF uses JDBC connection pool per configuration server, so all the external tables that refer to that server will use the same connection pool (when user impersonation is not used). Connection pool properties can be defined in `$PXF_BASE/servers/<your-server>/jdbc-site.xml` using `jdbc.pool.property.` prefix.

Allowing users to set pool properties at the table level will conflict with sharing of the pool between tables as each table can override the pool settings. We can run into problems when queries for these tables arrive to PXF at the same time and try to set conflicting values for the shared pool.

x4m:
Hi @denalex ! Thanks for looking into this PR!

Currently users have to have filesystem access on the server to configure connection pool. Is is very inconvinient for cloud deployments. When connection pool is configured from URI it's much more usable.

I understand that user might be unaware that poolMaximumSize=1 and poolMaximumSize=2 will produce two different pools. Quite similar problem already persists with POOL_SIZE setting which configures thread pool.

Yet I think that providing slightly inaccurate way to specify connection pool properties is OK. And much better then making this task solvable only by system administrator. As one another option we can provide pool name and compare pool descriptors by pool name instead of poolConfig.

What do you think?

Have a nice day :)

x4m:
@denalex kind reminder :) Does it all make sense from your point of view?

PR #842, opened 2022-08-05
Add support for using a PreparedStatement when reading

bradfordb-vmware:

### DIFF
--- a/server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcAccessor.java
+++ b/server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcAccessor.java
@@ -22,9 +22,11 @@ package org.greenplum.pxf.plugins.jdbc;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang.StringUtils;
 import org.greenplum.pxf.api.OneRow;
+import org.greenplum.pxf.api.error.PxfRuntimeException;
 import org.greenplum.pxf.api.model.Accessor;
 import org.greenplum.pxf.api.model.ConfigurationFactory;
 import org.greenplum.pxf.api.security.SecureLogin;
+import org.greenplum.pxf.api.utilities.Utilities;
 import org.greenplum.pxf.plugins.jdbc.utils.ConnectionManager;
 import org.greenplum.pxf.plugins.jdbc.writercallable.WriterCallable;
 import org.greenplum.pxf.plugins.jdbc.writercallable.WriterCallableFactory;
@@ -57,6 +59,8 @@ import java.util.concurrent.Future;
 public class JdbcAccessor extends JdbcBasePlugin implements Accessor {
 
     private static final Logger LOG = LoggerFactory.getLogger(JdbcAccessor.class);
+
+    private static final String JDBC_READ_PREPARED_STATEMENT_PROPERTY_NAME = "jdbc.read.prepared-statement";
 
     private Statement statementRead = null;
     private ResultSet resultSetRead = null;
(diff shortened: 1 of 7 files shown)


server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcAccessor.java
```diff
@@ -344,4 +354,16 @@ private String getQueryText() {
         return queryText;
     }
 
+    private boolean parseJdbcUsePreparedStatementProperty() {
```
denalex, line 357: we just created a very similar function: https://github.com/greenplum-db/pxf/blob/master/server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedAccessor.java#L328 Should we generalize it in Utilities class ?

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcAccessor.java
```diff
@@ -112,14 +115,21 @@ public boolean openForRead() throws SQLException, SQLTimeoutException {
         LOG.trace("Select query: {}", queryRead);
 
         // Execute queries
-        statementRead = connection.createStatement();
+        boolean usePreparedStatement = parseJdbcUsePreparedStatementProperty();
+        statementRead = usePreparedStatement ?
+                connection.prepareStatement(queryRead) :
+                connection.createStatement();
```
denalex, line 121: can we `LOG.debug` the fact that we determined we need to use a Prepared statement ? No log is needed for the default case. This is to troubleshoot if things are not working, so we can look at the log file and make sure the config property has been taken into account.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcAccessor.java
```diff
@@ -58,6 +59,8 @@ public class JdbcAccessor extends JdbcBasePlugin implements Accessor {
 
     private static final Logger LOG = LoggerFactory.getLogger(JdbcAccessor.class);
 
+    private static final String JDBC_READ_PREPARED_STATEMENT_PROPERTY_NAME = "pxf.jdbc.read.prepared-statement";
```
denalex, line 62: we should add this property to `jdbc-site.xml` with the value of false and document there its purpose and when to change the value.

denalex, line 62: all our jdbc properties start with `jdbc.`, so maybe drop `pxf.` prefix ?

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcAccessor.java
```diff
@@ -112,14 +115,21 @@ public boolean openForRead() throws SQLException, SQLTimeoutException {
         LOG.trace("Select query: {}", queryRead);
 
         // Execute queries
-        statementRead = connection.createStatement();
+        boolean usePreparedStatement = parseJdbcUsePreparedStatementProperty();
```
denalex, line 118: lets put a comment here why we even bother with Prepared Statement (and not providing bind params).

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcAccessor.java
```diff
@@ -344,4 +354,16 @@ private String getQueryText() {
...
+        if (jdbcPreparedStmtReadStr.equalsIgnoreCase("false")) {
+            return false;
+        } else if (jdbcPreparedStmtReadStr.equalsIgnoreCase("true")) {
+            return true;
+        } else {
+            throw new PxfRuntimeException(String.format(
+                    "Property %s has invalid value %s", JDBC_READ_PREPARED_STATEMENT_PROPERTY_NAME, jdbcPreparedStmtReadStr));
+        }
+    }
+
```
denalex, line 368: Let's add a unit test. We can mock ConnectionManager to give us mock connection and then verify which method was called to obtain a statement based on the value of the config property. We can also return a mock for a statement and verify which method was called on a statement mock.

denalex (approved):

PR #857, opened 2022-08-31
Changes for supporting PXF for GP7.

hpandeycodeit:
`external-table` code changes to make it compatible/working with GP7. 
Changes in GP7 related to header files or other code changes that are referenced in PXF code needed to be addressed/updated accordingly. 

There are still some changes in the form of `TODOs` or `FIXME` that still needs to be addressed or investigated. 

Major changes includes: 

- The `fileam.h` is no longer available in GPDB and its renamed to `extaccess.h` and location is changed to `gpcontrib/gp_exttable_fdw/extaccess.h' 
- `ProjectionInfo` struct no longer contains `pi_varNumbers , pi_targetlist or pi_numSimpleVars` and this is still needs to be investigated  on what replaces these member variables in the new Postgres code. 

Co-authored-by: Soumyadeep Chakraborty <soumyadeep2007@gmail.com>


Makefile
```diff
@@ -54,6 +54,7 @@ it:
 
 install:
 	make -C $(SOURCE_EXTENSION_DIR) install
+	make -C external-table install
```
denalex, line 57: it seems we now always want to build `external-table` module, so its handling should be added to other targets that now only deal with `$(SOURCE_EXTENSION_DIR)`. And I would recommend renaming and updating the logic so that instead of `$(SOURCE_EXTENSION_DIR)` and `TARGET_EXTENSION_DIR` we use `FDW_SUPPORT` to optionally build/clean/install the FDW module

external-table/src/libchurl.c
```diff
@@ -851,10 +851,11 @@ fill_internal_buffer(churl_context *context, int want)
 			pg_usleep(100);
 		else if (-1 == select(maxfd + 1, &fdread, &fdwrite, &fdexcep, &timeout))
 		{
-			if (errno == EINTR || errno == EAGAIN)
+			int save_errno = errno;
+			if (save_errno == EINTR || save_errno == EAGAIN)
 				continue;
 			elog(ERROR, "internal error: select failed on curl_multi_fdset (maxfd %d) (%d - %s)",
-				 maxfd, errno, strerror(errno));
+				 maxfd, save_errno, strerror(save_errno));
```
denalex, line 858: what is the point of this change, which error does it fix ?

external-table/src/pxfheaders.c
```diff
@@ -77,11 +78,20 @@ build_http_headers(PxfInputData *input)
 		/* Parse fmtOptString here */
 		if (fmttype_is_text(exttbl->fmtcode) || fmttype_is_csv(exttbl->fmtcode))
 		{
+#if PG_VERSION_NUM <= 90400
 			copyFmtOpts = parseCopyFormatString(rel, exttbl->fmtopts, exttbl->fmtcode);
+#else
+			copyFmtOpts = exttbl->options;
```
denalex, line 84: so `exttbl->options` now only contains parsed format options and not any other options so we don't need to parse it ourselves now?

external-table/src/pxfheaders.c
```diff
@@ -77,11 +78,20 @@ build_http_headers(PxfInputData *input)
...
+#endif
 		}
 
+#if PG_VERSION_NUM <= 90400
 		/* pass external table's encoding to copy's options */
 		copyFmtOpts = appendCopyEncodingOption(copyFmtOpts, exttbl->encoding);
+#else
+        //copyFmtOpts = exttbl->options;
+        copyFmtOpts =  lappend(copyFmtOpts, makeDefElem("encoding", (Node *)makeString((char *)pg_encoding_to_char(exttbl->encoding)), -1));
```
denalex, line 93: it seems that function `appendCopyEncodingOption` did the very same thing (except -1 last parameter). Why are we no longer using it ?

external-table/src/pxfuriparser.h
```diff
@@ -21,6 +21,7 @@
 #define _PXFURIPARSER_H_
 
 #include "postgres.h"
+#include "nodes/pg_list.h"
```
denalex, line 24: why is this needed now ?

external-table/src/pxfuriparser.c
```diff
@@ -19,6 +19,7 @@
 
 #include "pxfuriparser.h"
 #include "pxfutils.h"
+#include "pxfbridge.h"
```
denalex, line 22: why is this needed now ?

external-table/src/libchurl.c
hpandeycodeit: To prevent this error during compilation: 

```
src/libchurl.c:859:13: error: called object type 'int' is not a function or function pointer
                                                         maxfd, errno, strerror(errno));
```

https://github.com/greenplum-db/gpdb/blob/master/src/include/utils/elog.h 

```
/*
 * Provide a way to prevent "errno" from being accidentally used inside an
 * elog() or ereport() invocation.  Since we know that some operating systems
 * define errno as something involving a function call, we'll put a local
 * variable of the same name as that function in the local scope to force a
 * compile error.  On platforms that don't define errno in that way, nothing
 * happens, so we get no warning ... but we can live with that as long as it
 * happens on some popular platforms.
 */
#if defined(errno) && defined(__linux__)
#define pg_prevent_errno_in_scope() int __errno_location pg_attribute_unused()
#elif defined(errno) && (defined(__darwin__) || defined(__freebsd__))
#define pg_prevent_errno_in_scope() int __error pg_attribute_unused()
#else
#define pg_prevent_errno_in_scope()
#endif
```

external-table/src/pxfheaders.c
hpandeycodeit: This function `parseCopyFormatString` doesn't exists anymore in the GP7. Not sure if the `exttbl->options` are parsed. I will investigate more into this.

external-table/src/pxfheaders.c
hpandeycodeit: Using `appendCopyEncodingOption` here is causing errors like these. Not sure why. : 

```
Undefined symbols for architecture x86_64:
  "_appendCopyEncodingOption", referenced from:
      _build_http_headers in pxfheaders.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

For GP7, `makeDefElem` function has changed to accept one extra parameter  of type `int`.


`appendCopyEncodingOption` still exists in GP7 but due to the error above, I avoided it.

external-table/src/pxfuriparser.h
hpandeycodeit: without the above import, it is unable to find `List` 

```
src/pxfuriparser.h:55:2: error: unknown type name 'List'
        List       *options;            /* list of OptionData       */
        ^
```

external-table/src/pxfuriparser.c
hpandeycodeit: Certain variables and functions are unable to get resolved. Like: 

```
src/pxfuriparser.c:297:3: error: use of undeclared identifier 'Value'
                Value      *key = makeString(asc_toupper(data->key, strlen(data->key)));
                ^
src/pxfuriparser.c:297:13: error: use of undeclared identifier 'key'
                Value      *key = makeString(asc_toupper(data->key, strlen(data->key)));
                            ^
src/pxfuriparser.c:297:19: error: implicit declaration of function 'makeString' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
                Value      *key = makeString(asc_toupper(data->key, strlen(data->key)));
```

However, I don't need to import `pxfbridge.h` here in this file. I will import the required headers into the `pxfuriparser.h` 

```
#if PG_VERSION_NUM >= 90600
#include "lib/stringinfo.h"
#include "nodes/value.h"
#endif
```

hpandeycodeit:
@denalex This [commit](https://github.com/greenplum-db/gpdb/commit/19cd1cf4b68faff2e29bc2fa884c480e4644cdb4#diff-7193f2087643abaf45931b1955c918109a17c8b55138900fb097bf99f290f533)  changed TupleDesc struct member `FormData_pg_attribute attrs` from pointer to non pointer.  File `tupdesc.h`

```
typedef struct TupleDescData
{
	int			natts;			/* number of attributes in the tuple */
	Oid			tdtypeid;		/* composite type ID for tuple type */
	int32		tdtypmod;		/* typmod for tuple type */
	int			tdrefcount;		/* reference count, or -1 if not counting */
	TupleConstr *constr;		/* constraints, or NULL if none */
	/* attrs[N] is the description of Attribute Number N+1 */
	FormData_pg_attribute attrs[FLEXIBLE_ARRAY_MEMBER];
}			TupleDescData;
typedef struct TupleDescData *TupleDesc;
```

Before this commit 

```
typedef struct tupleDesc
{
	int			natts;			/* number of attributes in the tuple */
	Form_pg_attribute *attrs;
	/* attrs[N] is a pointer to the description of Attribute Number N+1 */
	TupleConstr *constr;		/* constraints, or NULL if none */
	Oid			tdtypeid;		/* composite type ID for tuple type */
	int32		tdtypmod;		/* typmod for tuple type */
	bool		tdhasoid;		/* tuple has oid attribute in its header */
	int			tdrefcount;		/* reference count, or -1 if not counting */
}	*TupleDesc;
```

external-table/src/pxfheaders.c
```diff
@@ -375,25 +391,40 @@ add_projection_desc_httpheader(CHURL_HEADERS headers,
...
 	 * we use expression_tree_walker to access attrno information
 	 * we do it through a helper function add_attnums_from_targetList
 	 */
-	if (projInfo->pi_targetlist)
+	/* FIXME: commenting this out for compilation success */
+	// pi_targetlist is not available anymore in the postgters code
+	// https://doxygen.postgresql.org/structProjectionInfo.html
+	//if (projInfo->pi_targetlist)
+
+	if (targetList)
```
soumyadeep2007, line 427: Now, unfortunately, we can't rely on `varNumbers` (which stored simple vars) anymore as it is a piece of info no longer stored. `simpleVar` has been renamed to `safeVar` (please see ExecBuildProjectionInfo). The notion of whether a var is safe is lost and must be rediscovered below using the expression steps built conditionally in `ExecBuildProjectionInfo()`.

Dirty patch:

```diff
diff --git a/external-table/src/pxfheaders.c b/external-table/src/pxfheaders.c
index 3972d2af..e74a14bb 100644
--- a/external-table/src/pxfheaders.c
+++ b/external-table/src/pxfheaders.c
@@ -404,7 +404,7 @@ add_projection_desc_httpheader(CHURL_HEADERS headers,
     List *targetList = projInfo->pi_targetlist;
 #else
        int                             *varNumbers = NULL; //TODO Need to find a proper value here
-       List *targetList = (List *)projInfo->pi_state.resultslot;
+       List *targetList = (List *) projInfo->pi_state.expr;
 #endif
 
        Bitmapset               *attrs_used;
@@ -428,8 +428,21 @@ add_projection_desc_httpheader(CHURL_HEADERS headers,
        {
 #else
        numSimpleVars = 0;
+       int numNonSimpleVars = 0;
 
-       if (!varNumbers)
+       for (int i = 0; i < projInfo->pi_state->steps_len; i++)
+       {
+               ExprEvalStep *step = &projInfo->pi_state->steps[i];
+               if (step->opcode == EEOP_ASSIGN_INNER_VAR ||
+                       step->opcode == EEOP_ASSIGN_OUTER_VAR ||
+                       step->opcode == EEOP_ASSIGN_SCAN_VAR)
+                       numSimpleVars++;
+               else if (scratch.opcode = EEOP_ASSIGN_TMP_MAKE_RO ||
+                                scratch.opcode == EEOP_ASSIGN_TMP)
+                       numNonSimpleVars++;
+       }
+
+       if (numNonSimpleVars > 0)
        {
                /*
                 * When there are not just simple Vars we need to

```

This function is becoming unreadable with too many ifdefs. I suggest creating 2 functions : `add_projection_desc_httpheader_pg12` and `add_projection_desc_httpheader_pg9.4` which you can call from `add_projection_desc_httpheader` and then we can see later if we can consolidate into 1 function.

external-table/src/libchurl.c
```diff
@@ -395,7 +395,7 @@ churl_init_upload(const char *url, CHURL_HEADERS headers)
 
 	context->upload = true;
 
-	set_curl_option(context, CURLOPT_POST, (const void *) TRUE);
+	set_curl_option(context, CURLOPT_POST, (const void *) true);
```
ashuka24, line 398: why was this change necessary?

external-table/src/pxfuriparser.c
```diff
@@ -291,10 +290,13 @@ GPHDUri_verify_no_duplicate_options(GPHDUri *uri)
 	foreach(option, uri->options)
 	{
 		OptionData *data = (OptionData *) lfirst(option);
-#if PG_VERSION_NUM >= 90400
-		Value	   *key = makeString(asc_toupper(data->key, strlen(data->key)));
+
+// For GP 5.x latest ( 5.29.8) the server_version_num is 80323.
```
ashuka24, line 294: nit
```suggestion
// For GP 5.x latest (5.29.8) the server_version_num is 80323.

File: external-table/src/pxfuriparser.c [hunk 10]
```diff
@@ -291,10 +290,13 @@ GPHDUri_verify_no_duplicate_options(GPHDUri *uri)
...
+// For 6.x or later str_toupper doesn't work as it has different signature ( one extra parameter) so the asc_toupper will work for 6.x or 7.x
```
ashuka24, line 295: why is asc_toupper an acceptable replacement. are there caveats that we may need to be aware of? 

nit: 
```suggestion
// For 6.x or later str_toupper doesn't work as it has different signature (one extra parameter) so the asc_toupper will work for 6.x or 7.x

File: external-table/src/pxfutils.c [hunk 11]
```diff
@@ -22,11 +22,12 @@ normalize_key_name(const char *key)
 				(errcode(ERRCODE_INTERNAL_ERROR),
 				 errmsg("internal error in pxfutils.c:normalize_key_name. Parameter key is null or empty.")));
 	}
-
-#if PG_VERSION_NUM >= 90400
-	return psprintf("X-GP-OPTIONS-%s", asc_toupper(pstrdup(key), strlen(key)));
+// For GP 5.x latest ( 5.29.8) the server_version_num is 80323.
```
ashuka24, line 25: i see this block of code here in pxfutils.c but also above in pxfuriparser.c. why is it in both locations? can we remove one of the instances?

ashuka24 (commented):
just something to be aware of: spacing across all the files seems to be a little all over the place

external-table/src/pxfheaders.c
```diff
@@ -19,22 +19,31 @@
...
 #include "cdb/cdbvars.h"
 
 /* helper function declarations */
 static void add_alignment_size_httpheader(CHURL_HEADERS headers);
 static void add_tuple_desc_httpheader(CHURL_HEADERS headers, Relation rel);
 static void add_location_options_httpheader(CHURL_HEADERS headers, GPHDUri *gphduri);
 static char *get_format_name(char fmtcode);
-static void add_projection_desc_httpheader(CHURL_HEADERS headers, ProjectionInfo *projInfo, List *qualsAttributes, Relation rel);
+
+#if PG_VERSION_NUM >= 120000
```
bradfordb-vmware, line 41: Do we need to give this function different names depending on the Postgres version? This means that we have to `#if` at the callsite. Couldn't we use just one function declaration and then use `#if` when defining the function?

external-table/src/pxfheaders.c
```diff
@@ -492,13 +504,170 @@ add_projection_desc_httpheader(CHURL_HEADERS headers,
...
+  if (number == 0)
+    return;
+
+  attrs_used = NULL;
+
+  /* Convert the number of projection columns to a string */
+  pg_ltoa(number, long_number);
+  churl_headers_append(headers, "X-GP-ATTRS-PROJ", long_number);
+
+    for (i = 0;varNumbers && i < numSimpleVars ; i++)
```
bradfordb-vmware, line 624: It looks like `varNumbers` is an array and therefore will always evaluate as `true`.

external-table/src/gpdbwritableformatter.c
```diff
@@ -624,29 +662,33 @@ gpdbwritableformatter_export(PG_FUNCTION_ARGS)
...
 			appendInt1ToBuffer(myData->export_format_tuple,
-							   getJavaEnumOrdinal(tupdesc->attrs[i]->atttypid));
+							   getJavaEnumOrdinal(attr->atttypid));
 		}
 	}
 
 	/* Write Nullness */
 	nullBit = boolArrayToByteArray(myData->nulls, ncolumns, nvalidcolumns, &nullBitLen, tupdesc);
-	appendBinaryStringInfo(myData->export_format_tuple, nullBit, nullBitLen);
+	appendBinaryStringInfo(myData->export_format_tuple, (char *) nullBit, nullBitLen);
```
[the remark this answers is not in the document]
hpandeycodeit: In 7.x the signature has changed a bit which expects a `char` in the second argument  : 

https://github.com/greenplum-db/gpdb/blob/10c40bc0e39cd5d4bb4492a98301fe9acdfb754a/src/common/stringinfo.c#L243

```
extern void appendBinaryStringInfo(StringInfo str, const char *data, int datalen);
```

In 6.x

```
void
appendBinaryStringInfo(StringInfo str, const void *data, int datalen)
{
```

external-table/src/gpdbwritableformatter.c
```diff
@@ -658,7 +700,7 @@ gpdbwritableformatter_export(PG_FUNCTION_ARGS)
 	/* End padding */
 	appendStringInfoFill(myData->export_format_tuple, endpadding, '\0');
 
-	Insist(myData->export_format_tuple->len == datlen + VARHDRSZ);
+	Assert(myData->export_format_tuple->len == datlen + VARHDRSZ);
```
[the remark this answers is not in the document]
hpandeycodeit: I have seen both `Insist` and `Assert` used in the gpdb code. However for this file, If we compile with GP7, it fails 

```
src/gpdbwritableformatter.c:702:2: error: implicit declaration of function 'Insist' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
        Insist(myData->export_format_tuple->len == datlen + VARHDRSZ);
```

external-table/src/libchurl.c
hpandeycodeit: with `TRUE` the compiler is unable to resolve it with PG12/GP7: 

```
src/libchurl.c:398:56: error: use of undeclared identifier 'TRUE'
        set_curl_option(context, CURLOPT_POST, (const void *) TRUE);
```

external-table/src/pxfutils.c
hpandeycodeit: I think it's okay to leave it in respective files since it is not exactly the same.

external-table/src/pxfuriparser.c
hpandeycodeit: I will look into it if we can still use `str_toupper` but to answer your question, the change was made to resolve any compilation issues.

external-table/src/gpdbwritableformatter.c
```diff
@@ -46,11 +46,15 @@
 #include "utils/lsyscache.h"
 
 #include <unistd.h>
+#if PG_VERSION_NUM >= 120000
```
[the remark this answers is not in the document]
hpandeycodeit: yes

external-table/src/gpdbwritableformatter.c
```diff
@@ -932,3 +998,17 @@ gpdbwritableformatter_import(PG_FUNCTION_ARGS)
 	FORMATTER_SET_TUPLE(fcinfo, tuple);
 	FORMATTER_RETURN_TUPLE(tuple);
 }
+
+static inline Form_pg_attribute
+getAttributeFromTupleDesc(TupleDesc tupdesc, int index)
+{
+    Form_pg_attribute attr;
+    #if PG_VERSION_NUM >= 120000
+         attr = &tupdesc->attrs[index];
```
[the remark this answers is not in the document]
hpandeycodeit: what do you mean @yimingli-vmware ?

Makefile
```diff
@@ -9,11 +9,6 @@ FDW_SUPPORT = $(shell $(PG_CONFIG) --version | egrep "PostgreSQL 12")
 
 SOURCE_EXTENSION_DIR = external-table
```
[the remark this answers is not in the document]
bradfordb-vmware: I think it makes sense to get rid of `SOURCE_EXTENSION_DIR`.. Going forward, I think we will be building `external-table`

Makefile
```diff
@@ -28,7 +23,9 @@ external-table:
 	make -C external-table
 
 fdw:
```
bradfordb-vmware, line 25: I think we need to add `fdw` to the `all` target, otherwise the FDW doesn't get built unless explicitly asked for by running `make fdw`.

automation/tincrepo/main/pxf/features/general/alter/csv/expected/query01.ans
```diff
@@ -12,6 +12,9 @@
 -- m/ERROR/
 -- s/ERROR:  missing data for column "bool"/ERROR:  invalid input syntax for integer: "1.0"/g
 --
+-- m/ERROR/
+-- s/ERROR:  invalid input syntax for integer: "1.0"/ERROR:  invalid input syntax for type integer: "1.0"/g
```
denalex, line 16: @hpandeycodeit - this type of match/sub has been added to global init now, I believe

external-table/src/gpdbwritableformatter.c
```diff
@@ -963,3 +1029,17 @@ gpdbwritableformatter_import(PG_FUNCTION_ARGS)
 	FORMATTER_RETURN_TUPLE(tuple);
 }
 
+static inline Form_pg_attribute
+getAttributeFromTupleDesc(TupleDesc tupdesc, int index)
```
denalex, line 1033: I wonder if this can be a macro instead of a function ?

external-table/src/gpdbwritableformatter.c
```diff
@@ -806,15 +768,15 @@ gpdbwritableformatter_import(PG_FUNCTION_ARGS)
 
 		for (i = 0; i < ncolumns; i++)
 		{
-                Form_pg_attribute attr = getAttributeFromTupleDesc(tupdesc,i);
+            Form_pg_attribute attr = getAttributeFromTupleDesc(tupdesc,i);
```
ashuka24, line 771: nit: spacing here seems a little odd

external-table/src/gpdbwritableformatter.c
```diff
@@ -481,17 +495,20 @@ gpdbwritableformatter_export(PG_FUNCTION_ARGS)
...
 	/*
 	 * Initialize the context structure
 	 */
 	if (myData == NULL)
 	{
-		if (FORMATTER_GET_EXTENCODING(fcinfo) != PG_UTF8)
-			ereport(ERROR, (errcode(ERRCODE_EXTERNAL_ROUTINE_EXCEPTION),
-							errmsg("gpdbwritable formatter can only export UTF8 formatted data. Define the external table with ENCODING UTF8")));
+
+        reportErrorIfNonUTF8Encoding(fcinfo);
```
bradfordb-vmware, line 511: Would it be cleaner if this function was turned into one that returned a boolean?

```suggestion
if (!isUtf8Encoding(fcinfo))
{
    ereport(ERROR, (errcode(ERRCODE_EXTERNAL_ROUTINE_EXCEPTION),
        errmsg(FORMATTER_ENCODING_ERR_MSG, "import")));
}
```

external-table/src/gpdbwritableformatter.c
```diff
@@ -963,3 +991,41 @@ gpdbwritableformatter_import(PG_FUNCTION_ARGS)
...
+{
+// In GP7 FORMATTER_GET_EXTENCODING(fcinfo) gets the database encoding which may not match the table encoding
+// and thus results in exception here. So getting the table encoding from the ExtTableEntry
+#if PG_VERSION_NUM < 120000
+    if (FORMATTER_GET_EXTENCODING(fcinfo) != PG_UTF8)
+    {
+             ereport(ERROR, (errcode(ERRCODE_EXTERNAL_ROUTINE_EXCEPTION),
+                        errmsg(FORMATTER_ENCODING_ERR_MSG, "import")));
+    }
+    #else
```
bradfordb-vmware, line 1018: Your spacing of `#else` doesn't match with the spacing for `#if`. We should be consistent.

external-table/src/pxfheaders.c
```diff
@@ -492,6 +509,163 @@ add_projection_desc_httpheader(CHURL_HEADERS headers,
...
+  initStringInfo(&formatter);
+  numTargetList = 0;
+
+   List *targetList = (List *) projInfo->pi_state.expr;
+     int  numSimpleVars = 0;
+
+  int numNonSimpleVars = 0;
+
+    // TODO Re-evaluate this logic and may be we don't need numNonSimpleVars
+   for (int i = 0; i < projInfo->pi_state.steps_len; i++)
```
bradfordb-vmware, line 554: You're re-declaring `int i` -- do we need `int i` at the top of function?

external-table/src/pxfheaders.c
```diff
@@ -492,6 +509,163 @@ add_projection_desc_httpheader(CHURL_HEADERS headers,
...
+     {
+         ExprEvalStep *step = &projInfo->pi_state.steps[i];
+         ExprEvalOp opcode = ExecEvalStepOp(&projInfo->pi_state, step);
+             if ( opcode == EEOP_ASSIGN_INNER_VAR ||
+                     opcode == EEOP_ASSIGN_OUTER_VAR ||
+                     opcode == EEOP_ASSIGN_SCAN_VAR)
+                     numSimpleVars++;
+             else if (opcode == EEOP_ASSIGN_TMP_MAKE_RO ||
+                        opcode == EEOP_ASSIGN_TMP)
+                     numNonSimpleVars++;
```
bradfordb-vmware: It's a little hard to see what's being done in each branch of the `if`-statement because of the line-breaks in the condition so I would suggest we use `{}` even though its not strictly needed.
```suggestion
        if (opcode == EEOP_ASSIGN_INNER_VAR ||
            opcode == EEOP_ASSIGN_OUTER_VAR ||
            opcode == EEOP_ASSIGN_SCAN_VAR)
        {
            numSimpleVars++;
        }
        else if (opcode == EEOP_ASSIGN_TMP_MAKE_RO ||
            opcode == EEOP_ASSIGN_TMP)
        {
            numNonSimpleVars++;
        }
```

external-table/src/pxfuriparser.c
bradfordb-vmware: The function `asc_toupper` is an ASCII-only function and is available in GPDB 5. Since this is being used to verify the options provided in the PXF location URL, we might be able to say only ASCII characters may be used. But since we accept a server name which is used to find a directory in the file system this might be overly restrictive.

Since we've been using `asc_toupper` with GPDB6 all along, I don't think this is a change in behavior.

external-table/src/pxfutils.c
```diff
@@ -22,11 +22,12 @@ normalize_key_name(const char *key)
...
+// For 6.x or later str_toupper doesn't work as it has different signature ( one extra parameter) so the asc_toupper will work for 6.x or 7.x
+#if PG_VERSION_NUM <= 80323
+    return psprintf("X-GP-OPTIONS-%s", str_toupper(pstrdup(key), strlen(key)));
```
bradfordb-vmware: Here we are capitalizing the characters in the HTTP header name so I'd just use `asc_toupper` regardless of the `PG_VERSION_NUM`.

yimingli-vmware (approved):

automation/tinc/main/tinctest/lib/global_init_file
```diff
@@ -69,4 +69,8 @@ s/, record/, line/
 m/invalid input syntax for type/
 s/invalid input syntax for type/invalid input syntax for/
 
+# alter table syntax is different in GP6 and GP7 for external table.
+m/ALTER/
+s/ALTER EXTERNAL TABLE/ALTER FOREIGN TABLE/g
```
denalex, line 74: a few lines above we mapped foreign -> external , here we're going the other direction. Do we want to be consistent ?

external-table/src/gpdbwritableformatter.c
```diff
@@ -46,11 +46,16 @@
...
 #include <unistd.h>
+#if PG_VERSION_NUM >= 120000
+#include "access/external.h"
+#endif
 
 PG_FUNCTION_INFO_V1(gpdbwritableformatter_export);
 PG_FUNCTION_INFO_V1(gpdbwritableformatter_import);
 Datum		gpdbwritableformatter_import(PG_FUNCTION_ARGS);
 Datum		gpdbwritableformatter_export(PG_FUNCTION_ARGS);
+static Form_pg_attribute getAttributeFromTupleDesc(TupleDesc, int);
```
denalex, line 57: can we make this function being a macro instead ?

external-table/src/gpdbwritableformatter.c
```diff
@@ -481,17 +496,23 @@ gpdbwritableformatter_export(PG_FUNCTION_ARGS)
...
 
 	/*
 	 * Initialize the context structure
 	 */
 	if (myData == NULL)
 	{
-		if (FORMATTER_GET_EXTENCODING(fcinfo) != PG_UTF8)
-			ereport(ERROR, (errcode(ERRCODE_EXTERNAL_ROUTINE_EXCEPTION),
-							errmsg("gpdbwritable formatter can only export UTF8 formatted data. Define the external table with ENCODING UTF8")));
+		if(!isUtf8Encoding(fcinfo))
```
denalex, line 511:
```suggestion
		if (!isUtf8Encoding(fcinfo))
```

Makefile
```diff
@@ -1,37 +1,61 @@
...
+    FDW_PACKAGE := FALSE
+else ifeq ($(GP_MAJORVERSION), 6)
+    FDW_BUILD := TRUE
+    FDW_PACKAGE := FALSE
+else ifeq ($(GP_MAJORVERSION), 7)
+    FDW_BUILD := TRUE
+    FDW_PACKAGE := TRUE
+else
+    $(error GP_MAJORVERSION must be 5, 6, or 7)
+endif
```
denalex, line 27: for the record, I prefer the original logic better

Makefile
```diff
@@ -1,37 +1,61 @@
...
 external-table:
 	make -C external-table
 
 fdw:
-ifneq ($(FDW_SUPPORT),)
+ifeq ($(FDW_BUILD),TRUE)
 	make -C fdw
 else
-	@echo "The variable FDW_SUPPORT was empty; skipping building the fdw/ directory"
+	@echo "Skipping building FDW extension because FDW_BUILD was not set"
```
denalex, line 58: I prefer the original error message

Makefile
```diff
@@ -42,15 +66,16 @@ server:
...
 	make -C server clean
 
 test:
-ifneq ($(FDW_SUPPORT),)
-	make -C fdw test
+ifeq ($(FDW_BUILD),TRUE)
+	make -C fdw installcheck
 else
-	@echo "The variable FDW_SUPPORT was empty; skipping fdw tests"
+	@echo "Skipping testing FDW extension because FDW_BUILD was not set"
```
denalex, line 78: I prefer the original error message

external-table/src/gpdbwritableformatter.c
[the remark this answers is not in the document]
denalex: so which should it be ? If it needs to be Insist, is there a way to make it work ?

external-table/src/gpdbwritableformatter.c
```diff
@@ -721,17 +748,22 @@ gpdbwritableformatter_import(PG_FUNCTION_ARGS)
...
 
 	/*
 	 * Initialize the context structure
 	 */
 	if (myData == NULL)
 	{
-		if (FORMATTER_GET_EXTENCODING(fcinfo) != PG_UTF8)
-			ereport(ERROR, (errcode(ERRCODE_EXTERNAL_ROUTINE_EXCEPTION),
-							errmsg("gpdbwritable formatter can only import UTF8 formatted data. Define the external table with ENCODING UTF8")));
+		if(!isUtf8Encoding(fcinfo))
```
denalex, line 762:
```suggestion
		if (!isUtf8Encoding(fcinfo))
```

external-table/src/gpdbwritableformatter.c
```diff
@@ -963,3 +999,38 @@ gpdbwritableformatter_import(PG_FUNCTION_ARGS)
...
+    return attr;
+
+}
+
+static bool isUtf8Encoding(FunctionCallInfo fcinfo)
+{
+// In GP7 FORMATTER_GET_EXTENCODING(fcinfo) gets the database encoding which may not match the table encoding
+// and thus results in exception here. So getting the table encoding from the ExtTableEntry
+#if PG_VERSION_NUM < 120000
+    if(FORMATTER_GET_EXTENCODING(fcinfo) != PG_UTF8)
```
denalex, line 1021:
```suggestion
    if (FORMATTER_GET_EXTENCODING(fcinfo) != PG_UTF8)
```

external-table/src/gpdbwritableformatter.c
```diff
@@ -963,3 +999,38 @@ gpdbwritableformatter_import(PG_FUNCTION_ARGS)
...
+        return false;
+    return true;
+#else
+    Relation rel = FORMATTER_GET_RELATION(fcinfo);
+    if(rel == NULL)
```
denalex, line 1026:
```suggestion
    if (rel == NULL)
```

external-table/src/pxfheaders.c
```diff
@@ -19,14 +19,17 @@
...
+#include "commands/defrem.h"
+#if PG_VERSION_NUM >= 120000
+#include "access/external.h"
+#include "extension/gp_exttable_fdw/extaccess.h"
+#include "executor/execExpr.h"
+#else
 #include "access/fileam.h"
 #include "catalog/pg_exttable.h"
-#include "commands/defrem.h"
-#if PG_VERSION_NUM >= 90400
```
denalex, line 25: why is this check no longer relevant ?

external-table/src/pxfheaders.c
```diff
@@ -74,14 +77,22 @@ build_http_headers(PxfInputData *input)
 
 		churl_headers_append(headers, "X-GP-FORMAT", format);
 
-		/* Parse fmtOptString here */
-		if (fmttype_is_text(exttbl->fmtcode) || fmttype_is_csv(exttbl->fmtcode))
-		{
-			copyFmtOpts = parseCopyFormatString(rel, exttbl->fmtopts, exttbl->fmtcode);
-		}
+    /* Parse fmtOptString here */
```
denalex, line 80: review formatting for this file, need a lot of corrections

external-table/src/pxfheaders.c
```diff
@@ -74,14 +77,22 @@ build_http_headers(PxfInputData *input)
...
+    if (fmttype_is_text(exttbl->fmtcode) || fmttype_is_csv(exttbl->fmtcode))
+    {
+#if PG_VERSION_NUM >= 120000
+		copyFmtOpts = exttbl->options;
+#else
+		copyFmtOpts = parseCopyFormatString(rel, exttbl->fmtopts, exttbl->fmtcode);
```
denalex, line 86: isn't this function defined only for `if PG_VERSION_NUM < 90400` ?

external-table/src/pxfheaders.c
```diff
@@ -74,14 +77,22 @@ build_http_headers(PxfInputData *input)
...
+#endif
+    }
 
-		/* pass external table's encoding to copy's options */
+#if PG_VERSION_NUM >= 120000
+    /* pass external table's encoding to copy's options */
+		copyFmtOpts = lappend(copyFmtOpts, makeDefElem("encoding", (Node *)makeString((char *)pg_encoding_to_char(exttbl->encoding)), -1));
+#else
```
denalex, line 93: same question here. Should we conditionally define the function here instead of inlining its logic ?

external-table/src/pxfheaders.c
```diff
@@ -228,10 +239,14 @@ add_tuple_desc_httpheader(CHURL_HEADERS headers, Relation rel)
...
-	{
-		FormData_pg_attribute *attribute = tuple->attrs[i];
+  /* Iterate attributes */
+  for (i = 0, attrIx = 0; i < tuple->natts; ++i)
+  {
+        #if PG_VERSION_NUM >= 120000
+            FormData_pg_attribute *attribute = &tuple->attrs[i];
+        #else
+            FormData_pg_attribute *attribute = tuple->attrs[i];
+        #endif
```
denalex, line 249: we had a function for this in another file. Better use a common macro.

external-table/src/pxfheaders.c
```diff
@@ -492,6 +509,159 @@ add_projection_desc_httpheader(CHURL_HEADERS headers,
...
+ * of dropped columns that precede it. For example,
+ *
+ *  ---------------------------------------------
+ * |  col1  |  col2 (dropped)  |  col3  |  col4  |
+ *  ---------------------------------------------
+ *
+ * Let's assume that col1 and col4 are projected, the reported projected
+ * indices will be 0, 2. This is because we use 0-based indexing and because
+ * col2 was dropped, the indices for col3 and col4 get shifted by -1.
+ */
```
denalex, line 527: no reason to repeat identical comment

external-table/src/pxfuriparser.c
```diff
@@ -291,11 +290,8 @@ GPHDUri_verify_no_duplicate_options(GPHDUri *uri)
 	foreach(option, uri->options)
 	{
 		OptionData *data = (OptionData *) lfirst(option);
-#if PG_VERSION_NUM >= 90400
+
 		Value	   *key = makeString(asc_toupper(data->key, strlen(data->key)));
```
denalex, line 294: is `asc_toupper` available in PG_VERSION_NUM < 90400 ?

server/pxf-service/src/test/java/org/greenplum/pxf/service/PxfMetricsIT.java
```diff
@@ -27,6 +29,8 @@
 
 @SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT, classes = PxfServiceApplication.class)
 @AutoConfigureMetrics
+// on MacOS with Intel chip (and JDK 8) this test crashes the JVM. Commenting out for now to not run on Mac.
```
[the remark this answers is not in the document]
denalex: no

Makefile
```diff
@@ -1,37 +1,54 @@
...
+include $(PGXS)
+
+FDW_BUILD   := TRUE
+FDW_PACKAGE := TRUE
+ifeq ($(shell test $(GP_MAJORVERSION) -lt 6; echo $$?),0)
+	FDW_BUILD := FALSE
+endif
+ifeq ($(shell test $(GP_MAJORVERSION) -lt 7; echo $$?),0)
+	FDW_PACKAGE := FALSE
+endif
```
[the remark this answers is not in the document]
hpandeycodeit: As per discussion we refactored the code and changed the messaging.

external-table/src/gpdbwritableformatter.c
hpandeycodeit: As discussed, `INSIST` is not used anywhere else apart from `regcomp.c` in GPDB. So it looks okay to change it to `Assert`

external-table/src/pxfuriparser.c
hpandeycodeit: yes.

external-table/src/gpdbwritableformatter.c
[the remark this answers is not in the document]
denalex: Insist is explained here: https://github.com/greenplum-db/gpdb/blob/6X_STABLE/src/include/utils/elog.h#L109

Makefile
```diff
@@ -88,17 +117,25 @@ tar: stage
 	tar -czf build/dist/$(shell ls build/stage).tar.gz -C build/stage $(shell ls build/stage)
 
 rpm:
-	make -C $(SOURCE_EXTENSION_DIR) stage
+	make -C external-table stage
```
bradfordb-vmware, line 120: The first few lines of the of `stage`, `rpm`, and `deb` targets are the same. Would it simplify things if we made `rpm` and `deb` depend on the `stage` target`?

external-table/src/libchurl.c
denalex: `TRUE` used to be defined a 1 in `c.h`: https://github.com/greenplum-db/gpdb/blob/6X_STABLE/src/include/c.h#L292 but is no longer defined in GP7. The API for set_curl_option(context, CURLOPT_POST, ...) says we need to pass `1` as a parameter : https://linux.die.net/man/3/curl_easy_setopt , so I think `set_curl_option(context, CURLOPT_POST, (const void *) true);` will do the same

external-table/src/pxfheaders.c
denalex: we now include both `utils/timestamp.h` and `nodes/makefuncs.h` and find stuff we need in either of these places depending on GP version.

external-table/src/pxfheaders.c
denalex: GP7 (or rather `gp_exttable_fdw`) parses all external table options (copy or not) and makes them available in `exttbl->options` so there is no need to do extra parsing when run with GP7

Makefile
denalex: this is a good idea, let's do it in a different PR though

external-table/src/gpdbwritableformatter.c
```diff
@@ -142,6 +147,12 @@ appendStringInfoFill(StringInfo str, int occurrences, char ch)
 		} \
 	} while (0)
 
+#if PG_VERSION_NUM < 90400
+// Copied from tupdesc.h (6.x), since this is not present in GPDB 5
+/* Accessor for the i'th attribute of tupdesc. */
+#define TupleDescAttr(tupdesc, i) ((tupdesc)->attrs[(i)])
```
bradfordb-vmware, line 153: Nice!

external-table/src/pxfheaders.c
```diff
@@ -459,32 +546,34 @@ add_projection_desc_httpheader(CHURL_HEADERS headers,
 						 attrNumber + 1 - FirstLowInvalidHeapAttributeNumber);
 	}
 
+	// STEP 4: add projection headers for attributes of relation that are not dropped and were selected in steps before
```
ashuka24, line 549: not sure if this is the best rewording, but it is personally easier for me to parse the following: 
```suggestion 
	// STEP 4: for attributes in the relation that are not dropped, add projection headers for those selected in steps 0 - 3 above

File: external-table/src/gpdbwritableformatter.c [hunk 49]
```diff
@@ -481,17 +500,23 @@ gpdbwritableformatter_export(PG_FUNCTION_ARGS)
...
 	 * Initialize the context structure
 	 */
 	if (myData == NULL)
 	{
 		if (FORMATTER_GET_EXTENCODING(fcinfo) != PG_UTF8)
-			ereport(ERROR, (errcode(ERRCODE_EXTERNAL_ROUTINE_EXCEPTION),
-							errmsg("gpdbwritable formatter can only export UTF8 formatted data. Define the external table with ENCODING UTF8")));
+		{
+		     ereport(ERROR, (errcode(ERRCODE_EXTERNAL_ROUTINE_EXCEPTION),
+		            errmsg(FORMATTER_ENCODING_ERR_MSG, "export", "export")));
```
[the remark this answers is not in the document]
hpandeycodeit: not  sure if I understand this.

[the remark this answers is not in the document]
denalex: @hpandeycodeit - try using use positional parameter access `%1$s` and only 1 parameter:
`#define FORMATTER_ENCODING_ERR_MSG "pxfwritable_%1$s formatter can only %1$s UTF8 formatted data. Define the external table with ENCODING UTF8"`
`errmsg(FORMATTER_ENCODING_ERR_MSG, "export")));` and the same for import

external-table/src/gpdbwritableformatter.c
```diff
@@ -681,7 +712,12 @@ gpdbwritableformatter_export(PG_FUNCTION_ARGS)
 	/* End padding */
 	appendStringInfoFill(myData->export_format_tuple, endpadding, '\0');
 
-	Insist(myData->export_format_tuple->len == datlen + VARHDRSZ);
+	if (myData->export_format_tuple->len != datlen + VARHDRSZ)
```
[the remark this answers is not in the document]
denalex: we still do -- see how original condition was `...->len == datlen...` that had to be true, otherwise an error was thrown. Now we negated the condition here: `...->len != datlen...` and report an error if this is the case. 
So the behavior is equivalent. Using `Assert` was not correct as in production `Assert` is not effective (removed by compiler).

fdw/Makefile
```diff
@@ -1,36 +1,29 @@
...
 .PHONY: stage
 stage: pxf_fdw.so
-	mkdir -p build/metadata
 	mkdir -p build/stage
 	install -c -m 755 pxf_fdw.so build/stage/pxf_fdw.so
 	install -c -m 644 pxf_fdw.control build/stage/
 	install -c -m 644 pxf_fdw--1.0.sql build/stage/
-	@echo $(GP_MAJORVERSION) > build/metadata/gp_major_version
-	@echo $(BLD_ARCH) | sed 's/_/-/' > build/metadata/build_arch
+	@echo "gpdb.version=$(GP_VERSION)" > build/stage/metadata
```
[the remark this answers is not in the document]
denalex: they both come from GP-provided makefile that we include and assume it is consistent.

denalex (approved):

PR #858, opened 2022-09-01
Add JsonProtocolHandler to use HdfsFileFragmenter for multi-line JSON

ashuka24:
This commit adds a JsonProtocolHandler which will use the HdfsFileFragmenter if we detect that the multi-line JSON profile is being used.

### DIFF
--- a/server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonProtocolHandler.java
+++ b/server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonProtocolHandler.java
@@ -35,14 +35,7 @@ public class JsonProtocolHandler implements ProtocolHandler {
         return context.getResolver();
     }
 
-
-    private boolean useMultilineJson(RequestContext context) {
-        String identifier = context.getOption("identifier");
-
-        if (isNotEmpty(identifier)) {
-            return true;
-        }
-
-        return false;
+    public boolean useMultilineJson(RequestContext context) {
+        return isNotEmpty(context.getOption("identifier"));
     }
 }
--- a/server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonProtocolHandler.java
+++ b/server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonProtocolHandler.java
@@ -2,9 +2,10 @@ package org.greenplum.pxf.plugins.json;
 
 import org.greenplum.pxf.api.model.ProtocolHandler;
 import org.greenplum.pxf.api.model.RequestContext;
-import org.greenplum.pxf.plugins.hdfs.HcfsFragmentMetadata;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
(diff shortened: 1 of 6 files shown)


server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonProtocolHandler.java
```diff
@@ -0,0 +1,48 @@
...
+
+
+    private boolean useMultilineJson(RequestContext context) {
+        String identifier = context.getOption("identifier");
+
+        if (isNotEmpty(identifier)) {
+            return true;
+        }
+
+        return false;
```
denalex, line 46:
```suggestion
        return isNotEmpty(context.getOption("identifier"));
```
we can make this a 1-liner or even embed into the calling function

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonProtocolHandler.java
```diff
@@ -0,0 +1,48 @@
...
+    @Override
+    public String getAccessorClassName(RequestContext context) {
+        return context.getAccessor();
+    }
+
+    @Override
+    public String getResolverClassName(RequestContext context) {
+        return context.getResolver();
+    }
+
```
denalex, line 37: nit: extra empty line

server/pxf-service/src/main/resources/pxf-profiles-default.xml
```diff
@@ -723,6 +723,7 @@ under the License.
             <accessor>org.greenplum.pxf.plugins.json.JsonAccessor</accessor>
             <resolver>org.greenplum.pxf.plugins.json.JsonResolver</resolver>
         </plugins>
+        <handler>org.greenplum.pxf.plugins.json.JsonProtocolHandler</handler>
```
denalex, line 726: would the same need to be done for all `*:json profiles` ?

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonProtocolHandler.java
```diff
@@ -0,0 +1,48 @@
...
+import org.greenplum.pxf.api.model.RequestContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static org.apache.commons.lang.StringUtils.isNotEmpty;
+
+/**
+ * Implementation of ProtocolHandler for json protocol.
+ */
+public class JsonProtocolHandler implements ProtocolHandler {
```
denalex, line 13: technically `json` is not a protocol, but a format, so calling it `JsonProtocolHandler` is a bit imprecise. But we only have "ProtocolHandler", not "ProfileHandler". We need to address this naming across profiles, maybe in a separate effort.

denalex (approved):

bradfordb-vmware (approved):

hpandeycodeit (approved):

PR #860, opened 2022-09-02
added Maven, Java 8 and pip install for GPDB7 docker images

denalex:
Since we will be running Tinc for testing FDW and external tables for GP7, we needed to add Maven and some python dependencies to run automation. Additionally, current images for GP7 were assuming PXF is compiled with Java11, this was probably experimental, we need to compile with Java 8 to support existing installations. So added Java 8 as well and set JAVA_HOME to use it.

Some changes, like `pip install` are a bit different between Centos and Ubuntu, but the logic is the same across Centos7 for gp6 and gp7 and Ubuntu for gp6 and gp7.

### DIFF
--- a/concourse/docker/pxf-dev-base/gpdb7/centos7/Dockerfile
+++ b/concourse/docker/pxf-dev-base/gpdb7/centos7/Dockerfile
@@ -32,7 +32,7 @@ RUN curl -fsSL -o /tmp/apache-maven.tar.gz ${BASE_URL}/apache-maven-${MAVEN_VERS
     && ln -s /usr/share/maven/bin/mvn /usr/bin/mvn \
     && yum install -y rpm-build python-devel jq sudo java-1.8.0-openjdk java-11-openjdk-devel && yum clean all \
     && cd /tmp && /usr/bin/pip install --upgrade pip==20.3.3 \
-    && /usr/bin/pip install psi paramiko --no-cache-dir
+    && /usr/bin/pip install paramiko --no-cache-dir
 
 # create user gpadmin since GPDB cannot run under root
 RUN ssh-keygen -t rsa -N "" -f /root/.ssh/id_rsa \
--- a/concourse/docker/pxf-dev-base/gpdb7/centos7/Dockerfile
+++ b/concourse/docker/pxf-dev-base/gpdb7/centos7/Dockerfile
@@ -20,9 +20,19 @@ RUN useradd -s /sbin/nologin -d /opt/minio minio \
     && chmod +x /opt/minio/bin/minio \
     && chown -R minio:minio /opt/minio
 
-# install java 8 and 11 and dependencies that are missing on the base images
-# java 8 is required to run hadoop. Do not use java 8 to build / run PXF
-RUN yum install -y rpm-build sudo java-1.8.0-openjdk java-11-openjdk-devel jq
+ARG MAVEN_VERSION=3.6.3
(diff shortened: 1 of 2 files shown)


concourse/docker/pxf-dev-base/gpdb7/centos7/Dockerfile
```diff
@@ -20,9 +20,19 @@ RUN useradd -s /sbin/nologin -d /opt/minio minio \
...
+
+# install dependencies that are missing on the base images
+RUN curl -fsSL -o /tmp/apache-maven.tar.gz ${BASE_URL}/apache-maven-${MAVEN_VERSION}-bin.tar.gz \
+    && mkdir -p /usr/share/maven \
+    && tar -xzf /tmp/apache-maven.tar.gz -C /usr/share/maven --strip-components=1 \
+    && rm -f /tmp/apache-maven.tar.gz \
+    && ln -s /usr/share/maven/bin/mvn /usr/bin/mvn \
+    && yum install -y rpm-build python-devel jq sudo java-1.8.0-openjdk java-11-openjdk-devel && yum clean all \
+    && cd /tmp && /usr/bin/pip install --upgrade pip==20.3.3 \
+    && /usr/bin/pip install psi paramiko --no-cache-dir
```
ashuka24, line 35: why are we bringing back psi and paramiko?

bradfordb-vmware:
> Additionally, current images for GP7 were assuming PXF is compiled with Java11, this was probably experimental, we need to compile with Java 8 to support existing installations. So added Java 8 as well and set JAVA_HOME to use it.

Do we need to compile with Java 8 or do we need to update PXF's build definition to target Java 8?

concourse/docker/pxf-dev-base/gpdb7/centos7/Dockerfile
bradfordb-vmware: It looks like this Dockerfile is copying what is done in the gpdb6 Dockerfiles. However, we patched `tinc` to not require `psi` as part of adding support for RHEL8 (e3aabd6b). Additionally, both RHEL7 and Ubuntu 18.04 have packages for paramiko:

```shell
# RHEL7
yum install python-paramiko

# Ubuntu 18.04
apt-get install python-paramiko
```

denalex: I will remove `psi`, but why is it better to install `paramiko` via `yum` vs `pip` ?

denalex:
while we can target Java 8 and use JDK11 to build, I think it is easier and more error-proof to just build using JDK8 if our target is Java 8 runtime.

concourse/docker/pxf-dev-base/gpdb7/centos7/Dockerfile
bradfordb-vmware: There are a lot of little reasons:

1. The yum and apt repos will give you a consistent, stable version with security patches; pip without any version constraints will grab the latest available.
2. Installing with pip will drop files in locations typically owned by the system but that the system is unaware of; installing with the yum/apt leaves the system consistently managed
3. You don't have to install/upgrade pip - one less step

denalex: discussed off-line and decided to preserve `pip` here and update all images at once later

bradfordb-vmware (approved):

ashuka24 (approved):

PR #864, opened 2022-09-13
docs - misc updates/additions to the orc write docs

lisakowen:
misc additions and updates to the ORC write docs.  i assumed that the data type mapping info is correct as-is, didn't review that.  also includes a few misc edits for formatting and consistency.

let me know if you want me to pull together a review site.

### DIFF
--- a/docs/content/hdfs_orc.html.md.erb
+++ b/docs/content/hdfs_orc.html.md.erb
@@ -156,6 +156,17 @@ The PXF `hdfs:orc` profile supports the following *read* options. You specify th
 | IGNORE_MISSING_PATH | A Boolean value that specifies the action to take when \<path-to-hdfs-file\> is missing or invalid. The default value is `false`, PXF returns an error in this situation. When the value is `true`, PXF ignores missing path errors and returns an empty fragment. |
 | MAP_BY_POSITION | A Boolean value that, when set to `true`, specifies that PXF should map an ORC column to a Greenplum Database column by position. The default value is `false`, PXF maps an ORC column to a Greenplum column by name. |
 
+The PXF `hdfs:orc` profile supports a single compression-related write option; you specify this option in the `CREATE WRITABLE EXTERNAL TABLE` `LOCATION` clause:
+
+| Write Option  | Value Description |
+|-------|-------------------------------------|
+| COMPRESSION_CODEC    | The compression codec alias. Supported compression codecs for writing ORC data include: `lz4`, `lzo`, `zstd`, `snappy`, `zlib`, and `none` . If this option is not specified, PXF compresses the data using `zlib` compression. |
+
+## <a id="write"></a>About Writing ORC data
+
+When you insert records into a writable external table, the block(s) of data that you insert are written to one or more files in the directory that you specify in the `LOCATION` clause.
+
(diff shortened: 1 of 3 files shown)


denalex (approved):

docs/content/hdfs_orc.html.md.erb
```diff
@@ -156,6 +156,17 @@ The PXF `hdfs:orc` profile supports the following *read* options. You specify th
...
 
+The PXF `hdfs:orc` profile supports a single compression-related write option; you specify this option in the `CREATE WRITABLE EXTERNAL TABLE` `LOCATION` clause:
+
+| Write Option  | Value Description |
+|-------|-------------------------------------|
+| COMPRESSION_CODEC    | The compression codec alias. Supported compression codecs for writing ORC data include: `lz4`, `lzo`, `zstd`, `snappy`, `zlib`, and `none` . If this option is not specified, PXF compresses the data using `zlib` compression. |
+
+## <a id="write"></a>About Writing ORC data
+
+When you insert records into a writable external table, the block(s) of data that you insert are written to one or more files in the directory that you specify in the `LOCATION` clause.
```
bradfordb-vmware, line 167: In the past there has been some confusion about the number of files written when writing from GPDB to PXF. Would it be clearer if this was worded as "...written to one file per segment in the directory..."

docs/content/hdfs_orc.html.md.erb
```diff
@@ -184,10 +195,10 @@ Procedure:
 
     ``` shell
     $ echo '{"location": "Prague", "month": "Jan","num_orders": 101, "total_sales": 4875.33, "items_sold": ["boots", "hats"]}
```
bradfordb-vmware, line 197: Rather than give this as a shell command, would it be better to just have this as a JSON block and tell them to copy-paste to a file called `sampledata.json`?

docs/content/hdfs_orc.html.md.erb
```diff
@@ -268,23 +279,26 @@ Procedure:
 
 ## <a id="write_example"></a>Example: Writing to an ORC File on HDFS
 
-1. Use the `hdfs:orc` profile to create a writable external table that references `sample_orc`:
+In this example, you create a writable external table to write some data to the directory referenced by the `sample_orc` table.
+
+1. Use the `hdfs:orc` profile to create a writable external table that references the `/data/pxf_examples/orc_example` directory on HDFS:
```
bradfordb-vmware, line 284: Nit-pick: The `hdfs:orc` profile isn't creating the table but rather an option of the table the user creates.

```suggestion
1. Create an external table, using `hdfs:orc` as the profileprofile, and the HDFS directory `/data/pxf_examples/orc_example` in the location URL path.
```

bradfordb-vmware (approved):

ashuka24 (approved):

docs/content/hdfs_orc.html.md.erb
lisakowen: i used to do the copy/paste steps long ago, but got input that an echo command is less steps for the user and less steps to document.  we can revisit this, will discuss with team and address in a separate PR.

bradfordb-vmware: Thanks for that context -- in that case I'd leave it as it is.

PR #868, opened 2022-09-15
ci: allow existing artifacts to remain if there is no pivnet product

ashuka24:
In the case that there is no pivnet artifact to pull down, do not error out. Instead, keep whatever artifact was already present in the bucket.

### DIFF
--- a/concourse/scripts/get_product_files.bash
+++ b/concourse/scripts/get_product_files.bash
@@ -68,15 +68,15 @@ for ((i = 0; i < ${#product_files[@]}; i++)); do
 
 	if [[ -z "${id}" ]]; then
 		echo "Did not find '${file}' in product files for GPDB '${gpdb_version}'"
-		rhel_regex='^.*rhel7.*$'
-		ubuntu_regex='^.*ubuntu18.*$'
-		if [[ $file =~ ${ubuntu_regex} ]]; then
-			existing_file=$(find ${product_dirs[$i]}/ -name *ubuntu18*.rpm)
-		elif [[ $file =~ ${rhel_regex} ]]; then
-			existing_file=$(find ${product_dirs[$i]}/ -name *el7*.rpm)
-		else
-			existing_file=$(find ${product_dirs[$i]}/ -name *el8*.rpm)
-		fi
+
+		case "${file}" in
+			*rhel7*) existing_file="$(find ${product_dirs[$i]}/ -name *rhel7*.rpm)" ;;
+			*rhel8*) existing_file="$(find ${product_dirs[$i]}/ -name *rhel8*.rpm)" ;;
+			*ubuntu18*) existing_file="$(find ${product_dirs[$i]}/ -name *ubuntu18*.deb)" ;;
+			*)
+				echo "Unexpected file: ${file}"
+				exit 1;;
+		esac
 
 		echo "Keeping existing file: ${existing_file}"
 		continue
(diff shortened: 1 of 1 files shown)


concourse/scripts/get_product_files.bash
```diff
@@ -68,12 +68,18 @@ for ((i = 0; i < ${#product_files[@]}; i++)); do
...
+			existing_file=$(find ${product_dirs[$i]}/ -name *ubuntu18*.rpm)
+		elif [[ $file =~ ${rhel_regex} ]]; then
+			existing_file=$(find ${product_dirs[$i]}/ -name *el7*.rpm)
+		else
+			existing_file=$(find ${product_dirs[$i]}/ -name *el8*.rpm)
 		fi
-		exit 1
+
+		echo "Keeping existing file: ${existing_file}"
+		continue
```
denalex, line 82: wouldn't there will be many files matching these regex, one of each version previously put there, since we do not delete files from the bucket. I am not sure if you wanted to print the whole list, if one file, maybe sort the list first and take first element ?  I would remove regex at all, since it is only used for printing a message.

concourse/scripts/get_product_files.bash
```diff
@@ -68,12 +68,18 @@ for ((i = 0; i < ${#product_files[@]}; i++)); do
...
 	if [[ -z "${id}" ]]; then
 		echo "Did not find '${file}' in product files for GPDB '${gpdb_version}'"
-		os_regex='^.*rhel8.*$'
-		if [[ $file =~ ${os_regex} ]]; then
-			echo "RHEL 8 artifact unavailable for the given GPDB version. Keeping existing rpm: $(find ${product_dirs[$i]}/ -name *rhel8*.rpm)"
-			continue
+		rhel_regex='^.*rhel7.*$'
+		ubuntu_regex='^.*ubuntu18.*$'
+		if [[ $file =~ ${ubuntu_regex} ]]; then
+			existing_file=$(find ${product_dirs[$i]}/ -name *ubuntu18*.rpm)
```
denalex, line 74:
```suggestion
			existing_file=$(find ${product_dirs[$i]}/ -name *ubuntu18*.deb)
```

concourse/scripts/get_product_files.bash
bradfordb-vmware: The bucket has multiple files, but the local directory in Concourse should only have one.

concourse/scripts/get_product_files.bash
```diff
@@ -68,12 +68,18 @@ for ((i = 0; i < ${#product_files[@]}; i++)); do
...
+		elif [[ $file =~ ${rhel_regex} ]]; then
+			existing_file=$(find ${product_dirs[$i]}/ -name *el7*.rpm)
+		else
```
bradfordb-vmware, line 77: This `if`-statement implicitly assumes only three supported platforms: RHEL7, RHEL8, Ubuntu. Do we want to make that explicit; additionally, I think this could be replaced with a `case`-statement

```sh
case "${file}" in
*rhel7*) existing_file="$(find ${product_dirs[$i]}/ -name *rhel7*.rpm)" ;;
*rhel8*) existing_file="$(find ${product_dirs[$i]}/ -name *rhel8*.rpm)" ;;
*ubuntu18*) existing_file="$(find ${product_dirs[$i]}/ -name *ubuntu18*.deb)" ;;
*)
    echo "Unexpected file: ${file}
    exit 1
    ;;
esac
```

bradfordb-vmware (approved):
Approved pending the suggested fix from Alex.

concourse/scripts/get_product_files.bash
ashuka24: As the existing files are whatever exist already in the bucket, which is automated by the same pipeline. I don't believe it is possible for the default case to exist. However, it doesn't hurt to add it in just in case. Thanks for the suggestion :)

PR #870, opened 2022-09-19
Relax the requirement that C string length matches Java string length

bradfordb-vmware:
Commit 94f8ccad added a check that the length of received strings (as determined by strnlen) matched the length of the string as determined by the Java-based PXF service; if they did not, this was treated as a fatal error. Some users have reported that they have ORC/Parquet files with strings that contain ASCCII NUL-bytes. These strings would not have a strnlen calculate length that matches with the Java string lenght.

This commit removes the requirements that the lengths be equal and instead logs a debug message when they do not match.


bradfordb-vmware:
@divyabhargov @ashwinstar FYI

automation/src/test/resources/data/orc/README.md
```diff
@@ -103,3 +103,13 @@ If desired, you can copy down the CSV file by running the following command:
 mv "./orc_multidim_list_types/000000_0" "./${CSV_FILENAME}"
 ```
 
+
```
ashuka24, line 106: nit: extra lines

external-table/src/gpdbwritableformatter.c
```diff
@@ -894,19 +894,23 @@ gpdbwritableformatter_import(PG_FUNCTION_ARGS)
...
-				/* outlen[i] includes the terminating nul-byte */
-				if (actual_len != myData->outlen[i] - 1)
-					ereport(FATAL,
-							(errcode(ERRCODE_EXTERNAL_ROUTINE_EXCEPTION),
-									errmsg("data for column %d of row %d has invalid length", i + 1, myData->lineno),
-									errdetail("expected column %d to have length %d, actual length is %ld", i + 1, myData->outlen[i] - 1, actual_len)));
+				/*
+				 * Compare the length as returned by strnlen with the length as determined by
+				 * the PXF service. If the source data includes ASCII NUL-byte in the string,
+				 * then these two values will not match. Its possible that this will result in
```
ashuka24, line 907:
```suggestion
				 * then these two values will not match. It's possible that this will result in
```

automation/src/test/java/org/greenplum/pxf/automation/features/orc/OrcReadTest.java
```diff
@@ -146,6 +154,12 @@ public void orcReadMultiDimensionalLists() throws Exception {
         runTincTest("pxf.features.orc.read.multidim_list_types.runTest");
     }
 
+    @Test(groups = {"features", "gpdb", "security", "hcfs"})
+    public void orcReadStringsContainingNulByte() throws Exception {
```
denalex, line 158:
```suggestion
    public void orcReadStringsContainingNullByte() throws Exception {
```
nit, since we call everything here as `null`

external-table/src/gpdbwritableformatter.c
```diff
@@ -894,19 +894,23 @@ gpdbwritableformatter_import(PG_FUNCTION_ARGS)
...
+				 * the Postgres input function truncating/mis-reading the value. Rather than
+				 * treat this as an error here, we log a debug message.
+				 */
+				if (actual_len != myData->outlen[i]-1)
+					ereport(DEBUG1,
```
ashuka24, line 912: With a log as debug, does this affect our ability to get core dumps for cases such as corrupted data streams, etc?

ashuka24 (approved):
nit: commit message has misspellings of both "ASCII" and "length"

external-table/src/gpdbwritableformatter.c
```diff
@@ -894,19 +894,23 @@ gpdbwritableformatter_import(PG_FUNCTION_ARGS)
 				 * Read string value from data_buf and convert it to internal representation
 				 *
 				 * PXF service should send all strings with a terminating nul-byte, so we can
-				 * determine the length of the string and compare it with the length that PXF
-				 * service computed and sent. Since we've checked that outlen[i] is no larger
-				 * than the number of bytes left for this tuple in data_buf, we use it as max
-				 * number of bytes to scan.
+				 * determine the number of bytes that the input function will read and compare
+				 * it withe length that the PXF service computed and sent. Since we've checked
```
denalex, line 898:
```suggestion
				 * it with the length that the PXF service computed and sent. Since we've checked
```

denalex (approved):

external-table/src/gpdbwritableformatter.c
hpandeycodeit: so the `if ` check remains as it is and to avoid the query erroring out we are logging it as as `DEBUG` ?

hpandeycodeit (approved):

external-table/src/gpdbwritableformatter.c
bradfordb-vmware: @ashuka24 Yes, but only here. The SIGSEGV check earlier in the file is still at the FATAL level.

@hpandeycodeit The main reason we are keeping the check and the debug log is that technically this condition could lead to truncated data being read by GPDB/Postgres. The log messages are there to help identify that. For example, reading the string `hello\u0000world` would only return `hello` in GPDB.

ashuka24: Is the truncating of data new behavior? or is it old behavior that we never logged before?

bradfordb-vmware: It is old behavior that was not logged.

yimngli-vmware (approved):

PR #872, opened 2022-09-20
Enable automation tests to run against PXF FDW extension

denalex:
This PR introduces basic support for the existing automation testing framework to run against the PXF FDW extension (instead of PXF external table extension). It achieves that by pre-creating FOREIGN SERVER objects and creating a FOREIGN TABLE instead of an EXTERNAL TABLE when requested. The change mostly impacts data and table preparation logic as most of the tests do SELECT FROM and do not care whether the table is "external" or "foreign".

A new CI environment variable `USE_FDW` is introduced to denote that the test suite should run against PXF FDW extension. Based on this the right extension is created in the database and the automation framework behaves as described above.

This is essentially a work-in-progress as having all tests running against FDW will take some time, but is complete enough to be able to merged. Two new annotations `@WorksWithFDW` and `@FailsWithFDW` are introduced to demarcate each test case appropriately. Currently only test cases annotated with `@WorksWithFDW` are executed (in `USE_FDW=true` mode). The rest of test cases are skipped. In the future, when the majority of the test cases do work with FDW, the annotation will be retired and only test cases annotated with `@FailsWithFDW` will be skipped.

This PR also fixes CSV serialization for timestamp values and introduces `GP7 on RHEL8 with FDW tests` development test swim-lane. If the solution is adopted, a production `GP7 on RHEL8 with FDW tests` swim-lane will be created in a separate PR.


automation/src/main/java/org/greenplum/pxf/automation/components/gpdb/Gpdb.java
```diff
@@ -141,6 +148,33 @@ private void createExtension(String extensionName, boolean ignoreFail) throws Ex
...
+		"default_gs",
+		"default_adl",
+		"default_wasbs",
+		"s3_s3",
+		"hdfs-non-secure_hdfs",
+		"hdfs-secure_hdfs",
+		"hdfs-ipa_hdfs");
+
+		for (String server : servers) {
+			String foreignServerName = server.replace("-", "_");
```
bradfordb-vmware, line 168: The logic inside this `for`-loop looks like a candidate for extracting into either a method or some kind of helper class. It seems to be encoding some testing convention for how things are named.

automation/src/main/java/org/greenplum/pxf/automation/structures/tables/pxf/ForeignTable.java
```diff
@@ -0,0 +1,206 @@
+package org.greenplum.pxf.automation.structures.tables.pxf;
+
+import org.apache.commons.lang.StringUtils;
+
+public class ForeignTable extends WritableExternalTable {
```
bradfordb-vmware, line 5: I know this abstraction isn't added by this PR, but it looks like `WritableExternalTable` extends `ReadableExternalTable` which seems a little weird since the two are not interchangeable in GPDB.

automation/src/main/java/org/greenplum/pxf/automation/structures/tables/pxf/ForeignTable.java
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    @Override
+    protected String createHeader() {
+        return "CREATE FOREIGN TABLE " + getFullName();
+    }
+
+    protected String createServer() {
+        String[] serverParameters = StringUtils.defaultIfBlank(getServer(), "default").split("=");
+        // getServer() might return a string "server=<..>", strip the prefix
```
bradfordb-vmware, line 38: Looking at how `server` is used in `ExternalTable` and how all the other location parameters are handled, would it make sense to "fix" our usage so that it doesn't include `SERVER=<...>`?

server/pxf-service/src/main/java/org/greenplum/pxf/service/BridgeOutputBuilder.java
```diff
@@ -478,20 +478,22 @@ private String fieldListToCSVString(List<OneField> fields) {
...
                         // for Text or Custom Format types, it should \\
                         String hexPrepend = gpdbTableformat.equalsIgnoreCase("csv") ? "\\x" : "\\\\x";
                         return hexPrepend + Hex.encodeHexString((byte[]) field.val);
-                    } else if (field.type == DataType.NUMERIC.getOID() || !DataType.isTextForm(field.type))
+                    } else if (field.type == DataType.NUMERIC.getOID() || !DataType.isTextForm(field.type)) {
                         return field.val.toString();
-                    else if (field.type == DataType.TIMESTAMP.getOID())
-                        return ((Timestamp) field.val).toLocalDateTime().format(GreenplumDateTime.DATETIME_FORMATTER);
-                    else if (field.type == DataType.DATE.getOID())
+                    } else if (field.type == DataType.TIMESTAMP.getOID()) {
```
bradfordb-vmware, line 490: I'm assuming this is the fix for

> CSV serialization for timestamp values and introduces

Is this needed as part of adding support for FDW to automation tests?

bradfordb-vmware (approved):

automation/src/main/java/org/greenplum/pxf/automation/structures/tables/pxf/ForeignTable.java
```diff
@@ -0,0 +1,206 @@
...
+        }
+        builder.append(optionName)
+                .append(" '")
+                .append(optionValue)
+                .append("'");
+    }
+
+    private String getFormatOption() {
+        // FDW format option is a second part of profile
+        if (getProfile() == null) {
```
ashuka24, line 141: seems like we are doing this `getProfile` and splitting it twice. Once here in `getFormatOptions` and below in `getProtocol` can it be refactored out into different function?

automation/src/test/resources/data/avro/array.json
```diff
@@ -1,12 +1,12 @@
...
-{"type_int":-200000002,"type_int_array":[-1.1,-2.2],"type_double":-2.2,"type_double_array":[-0.1,-0.2,-0.3],"type_string":"testing all supported types in AvroResolver 2","type_string_array":["and","arrays","too"],"type_float":-2.2,"type_float_array":[-0.1,0.2,-0.3333],"type_long":-200000000002,"type_long_array":[-300000000003,-500000000005],"type_bytes":"1","type_bytes_array":["3","2"],"type_boolean":false,"type_boolean_array":[false,false,true]}
+{"type_int":10,"type_int_array":[1,2],"type_double":1.1,"type_double_array":[0.1,0.2,0.3],"type_string":"testing all supported types in AvroResolver 1","type_string_array":["and","also","arrays"],"type_float":1.1,"type_float_array":[0.1,0.2,0.3333],"type_long":100000000001,"type_long_array":[200000000002,400000000004],"type_bytes":"12345","type_bytes_array":["2","3"],"type_boolean":true,"type_boolean_array":[true,false,false]}
```
ashuka24, line 1: why did this change?

concourse/scripts/pxf_common.bash
```diff
@@ -50,6 +50,9 @@ function inflate_dependencies() {
 	if [[ -f regression-tools/regression-tools.tar.gz ]]; then
 		tarballs+=(regression-tools/regression-tools.tar.gz)
 	fi
+	if [[ -f gp6-python-libs/gp6-python-libs.tar.gz ]]; then
```
ashuka24, line 53: can we have a comment here why this was necessary?

server/pxf-service/src/test/java/org/greenplum/pxf/service/BridgeOutputBuilderTest.java
```diff
@@ -633,4 +695,5 @@ public void writeUTF(String s) throws IOException {
             throw new IOException("not implemented");
         }
     }
+
```
ashuka24, line 698: nit: extra new line

PR #876, opened 2022-10-03
Parquet write array

yimingli-vmware:


automation/src/test/java/org/greenplum/pxf/automation/features/parquet/ParquetWriteTest.java
```diff
@@ -0,0 +1,397 @@
...
+        String jdbcUrl = HIVE_JDBC_URL_PREFIX + hive.getHost() + ":10000/default";
+        ExternalTable exHiveJdbcTable = TableFactory.getPxfJdbcReadableTable(
+                gpdbTableName + "_readable", PARQUET_PRIMITIVE_ARRAYS_TABLE_COLUMNS_READ_FROM_HIVE,
+                hiveTable.getName() + "_ctas", HIVE_JDBC_DRIVER_CLASS, jdbcUrl, null);
+        exHiveJdbcTable.setHost(pxfHost);
+        exHiveJdbcTable.setPort(pxfPort);
+        gpdb.createTableAndVerify(exHiveJdbcTable);
+
+        // use PXF hive:jdbc profile to read the data
+//        runTincTest("pxf.features.parquet.write.primitive_array_types_with_hive.runTest");
```
bradfordb-vmware, line 271: It looks like this test isn't actually running the tinc test case.

automation/src/test/java/org/greenplum/pxf/automation/features/parquet/ParquetWriteTest.java
```diff
@@ -0,0 +1,397 @@
...
+        }
+        prepareReadableExternalTable(readTableName,
+                PARQUET_PRIMITIVE_TABLE_COLUMNS, hdfsPath + filename);
+        gpdb.runQuery("CREATE OR REPLACE VIEW parquet_view AS SELECT s1, s2, n1, d1, dc1, " +
+                "CAST(tm AS TIMESTAMP WITH TIME ZONE) AT TIME ZONE 'PDT' as tm, " +
+                "f, bg, b, tn, sml, vc1, c1, bin FROM " + readTableName);
+
+        runTincTest("pxf.features.parquet.primitive_types.runTest");
+    }
+    @Test(groups = {"features", "gpdb", "security", "hcfs"})
```
bradfordb-vmware, line 317: We should have all test methods first and all helper methods at the end. Here you have the helper method `runWritePrimitiveScenario` in between two tests.

automation/tincrepo/main/pxf/features/parquet/write/primitive_array_types_with_hive/expected/query01.ans
```diff
@@ -0,0 +1,42 @@
+-- start_ignore
+-- end_ignore
+-- @description query01 for writing parquet primitive array data types using the Hive profile
+-- SET bytea_output=hex;
+-- SET
+SELECT * FROM orc_primitive_arrays_readable ORDER BY id;
+ id | bool_arr |      bigint_arr      | smallint_arr | int_arr | text_arr | real_arr  |  float_arr  | char_arr | varchar_arr |   varchar_arr_nolimit   |   date_arr   |         timestamp_arr          |       numeric_arr
```
bradfordb-vmware, line 7: It looks like all the arrays in this test are single element arrays. We should have a mix of arrays with one element, two elements, and several elements (e.g. 5).

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetFileAccessor.java
```diff
@@ -482,38 +484,104 @@ private MessageType generateParquetSchema(List<ColumnDescriptor> columns) {
...
+                            .optionalElement(PrimitiveTypeName.INT96);
+                    break;
                 default:
                     throw new UnsupportedTypeException(
                             String.format("Type %d is not supported", columnTypeCode));
             }
 
-            fields.add(builder.named(columnName));
-        }
+            if (primitiveBuilder != null) {
```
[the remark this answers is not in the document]
yimingli-vmware: I'll use the second option since other code handle primitive and complex separately, so i think the second option will make it consistent with other code.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
```diff
@@ -199,11 +189,88 @@ private void fillGroup(int index, OneField field, Group group, Type type) throws
...
+        GroupType listType = type.asGroupType();
+        //Get the repeated list group type and schema
+        GroupType repeatedType = listType.getType(0).asGroupType();
+        //Get the element type
+        Type elementType = repeatedType.getType(0).asPrimitiveType();
+        //parse parquet values into a postgres Object list
+        List<Object> vals = parquetUtilities.parsePostgresArray(field.val.toString(), elementType.asPrimitiveType().getPrimitiveTypeName(), elementType.getLogicalTypeAnnotation());
+        Group arrayGroup = new SimpleGroup(listType);
+
+        for (int i = 0; i < vals.size(); i++) {
```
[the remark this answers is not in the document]
yimingli-vmware: you mean like this?
`private void fillPrimitiveGroup(int index, Object value, Group group, Type type, boolean isPrimitive) throws IOException`
and I'll call it like this:
`fillPrimitiveGroup(index, field.val, group, type, true);`

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
```diff
@@ -199,11 +189,88 @@ private void fillGroup(int index, OneField field, Group group, Type type) throws
...
+    private void fillListGroup(int index, OneField field, Group group, Type type) throws IOException {
+        if (field.val == null)
+            return;
+        //Get the LIST group type and schema
+        GroupType listType = type.asGroupType();
+        //Get the repeated list group type and schema
+        GroupType repeatedType = listType.getType(0).asGroupType();
+        //Get the element type
+        Type elementType = repeatedType.getType(0).asPrimitiveType();
+        //parse parquet values into a postgres Object list
```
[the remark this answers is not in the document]
yimingli-vmware: I changed it into `// Decode Postgres String representation of an array into a list of Objects of the required element type`

automation/tincrepo/main/pxf/features/parquet/write/primitive_array_types_with_hive/expected/query01.ans
yimingli-vmware: For now I decide not to work on running the tinc test. I'll correct the test after I finish the read array support

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetUtilitiesTest.java
```diff
@@ -0,0 +1,243 @@
...
+        String value = "{{\"1985-01-01\", \"1990-04-30\"},{\"1995-08-14\", \"2020-12-05\"}}";
+
+        Exception exception = assertThrows(PxfRuntimeException.class, () -> parquetUtilities.parsePostgresArray(value, PrimitiveType.PrimitiveTypeName.INT32, LogicalTypeAnnotation.dateType()));
+        assertEquals("Error parsing array element: {\"1985-01-01\", \"1990-04-30\"} was not of expected type INT32", exception.getMessage());
+        assertEquals("Column value \"{{\"1985-01-01\", \"1990-04-30\"},{\"1995-08-14\", \"2020-12-05\"}}\" is a multi-dimensional array, PXF does not support multi-dimensional arrays for writing Parquet files.", ((PxfRuntimeException) exception).getHint());
+
+    }
+
+    @Test
+    public void testParsePostgresArrayMultiDimensionalArrayTimestamp() {
```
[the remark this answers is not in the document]
yimingli-vmware: > This test is failing locally for me, for the same reason as the previous test.

fixed both tests and added 1-D array test for both date and timestamp

automation/src/test/java/org/greenplum/pxf/automation/features/parquet/ParquetWriteTest.java
```diff
@@ -0,0 +1,397 @@
...
+
+    @Test(groups = {"features", "gpdb", "security", "hcfs"})
+    public void parquetWritePrimitives() throws Exception {
+        runWritePrimitiveScenario("pxf_parquet_write_primitives", "pxf_parquet_read_primitives", PARQUET_WRITE_PRIMITIVES, null);
+    }
+
+    @Test(groups = {"features", "gpdb"})
+    public void parquetWriteArrays() throws Exception {
+        // init only here, not in beforeClass() method as other tests run in environments without Hive
+        hive = (Hive) SystemManagerImpl.getInstance().getSystemObject("hive");
```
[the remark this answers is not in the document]
yimingli-vmware: I want to leave it as what it is now, since after I finish the read support, I will no longer test write functions with Hive

> Instead having `hive` be a field of the test class (that may or may not be initialized), should it be local to this test?

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/ParquetWriteTest.java
```diff
@@ -42,7 +46,7 @@
 import java.util.Collections;
 import java.util.List;
 
-import static java.math.BigDecimal.ROUND_UNNECESSARY;
+
```
ashuka24, line 49: nit: extra empty line

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/ParquetWriteTest.java
```diff
@@ -925,23 +926,108 @@ public void testWriteMultipleTypes() throws Exception {
...
+                .build();
+
+        for (int i = 0; i < 10; i++) {
+            Type outerType = schema.getType(0);
+            assertNotNull(outerType.getLogicalTypeAnnotation());
+            assertEquals(LogicalTypeAnnotation.listType(), outerType.getLogicalTypeAnnotation());
+
+            // get the outer group
+            Group outerGroup = fileReader.read();
+            if (i != 9) {
```
ashuka24, line 965: some of these if loops look duplicated across the tests. can we pull it out into a helper function?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetFileAccessor.java
```diff
@@ -446,77 +447,165 @@ private MessageType generateParquetSchema(List<ColumnDescriptor> columns) {
...
-                case BPCHAR:
-                case TEXT:
-                    builder = Types.optional(PrimitiveTypeName.BINARY)
-                            .as(stringType());
-                    break;
-                default:
-                    throw new UnsupportedTypeException(
-                            String.format("Type %d is not supported", columnTypeCode));
+            if (dataType.isArrayType()) {
+                generateParquetListSchema(column, fields);
```
denalex, line 454: since `fields` is declared in this method, I'd keep modifying it here as well. I think it's better to have these functions return `Type` object, call the functions `generateParquetXXXSchemaType`, not pass fields parameter and have `fields.add(generatedType)` after this `if` block.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
```diff
@@ -120,76 +127,56 @@ public OneRow setFields(List<OneField> record) throws IOException {
...
         switch (type.asPrimitiveType().getPrimitiveTypeName()) {
             case BINARY:
-                if (type.getLogicalTypeAnnotation() instanceof StringLogicalTypeAnnotation)
-                    group.add(index, (String) field.val);
-                else
-                    group.add(index, Binary.fromReusedByteArray((byte[]) field.val));
+                if (type.getLogicalTypeAnnotation() instanceof StringLogicalTypeAnnotation) {
+                    group.add(index, (String) value);
+                } else if (isPrimitive) {
+                    byte[] bytes=(byte[]) value;
```
denalex, line 145: why is this line needed, seems `bytes` is not used ?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
```diff
@@ -120,76 +127,56 @@ public OneRow setFields(List<OneField> record) throws IOException {
...
+                    group.add(index, Binary.fromReusedByteArray((byte[]) value));
+                } else {
+                    group.add(index, Binary.fromReusedByteArray(((ByteBuffer) value).array(),0,((ByteBuffer) value).limit()));
```
denalex, line 148: when would we pass value as a ByteBuffer ?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
```diff
@@ -199,11 +186,91 @@ private void fillGroup(int index, OneField field, Group group, Type type) throws
...
-                throw new IOException("Not supported type " + type.asPrimitiveType().getPrimitiveTypeName());
+                throw new IOException("Not supported primitive type " + type.asPrimitiveType().getPrimitiveTypeName());
+        }
+    }
+
+    private void fillComplexGroup(int index, OneField field, Group group, Type type) throws IOException {
+        if (field.val == null)
+            return;
+
+        if(type.asGroupType().getOriginalType() != LogicalTypeAnnotation.listType().toOriginalType()){
```
denalex, line 200:
```suggestion
        if (type.asGroupType().getOriginalType() != LogicalTypeAnnotation.listType().toOriginalType()) {
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
```diff
@@ -199,11 +186,91 @@ private void fillGroup(int index, OneField field, Group group, Type type) throws
...
+                group.add(index, (Boolean) value);
                 break;
             default:
-                throw new IOException("Not supported type " + type.asPrimitiveType().getPrimitiveTypeName());
+                throw new IOException("Not supported primitive type " + type.asPrimitiveType().getPrimitiveTypeName());
+        }
+    }
+
+    private void fillComplexGroup(int index, OneField field, Group group, Type type) throws IOException {
+        if (field.val == null)
```
denalex, line 197: we already checked for this in `fillGroup`

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
```diff
@@ -199,11 +186,91 @@ private void fillGroup(int index, OneField field, Group group, Type type) throws
...
+
+        if(type.asGroupType().getOriginalType() != LogicalTypeAnnotation.listType().toOriginalType()){
+            throw new IOException("Not supported complex type " + type.asGroupType().getOriginalType());
+        }
+        fillListGroup(index, field, group, type);
+    }
+
+
+    private void fillListGroup(int index, OneField field, Group group, Type type) throws IOException {
+        if (field.val == null)
```
denalex, line 208: we already checked for this in `fillGroup`

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
```diff
@@ -199,11 +186,91 @@ private void fillGroup(int index, OneField field, Group group, Type type) throws
...
+        GroupType repeatedType = listType.getType(0).asGroupType();
+        Type elementType = repeatedType.getType(0).asPrimitiveType();
+        // Decode Postgres String representation of an array into a list of Objects of the required element type
+        List<Object> vals = parquetUtilities.parsePostgresArray(field.val.toString(), elementType.asPrimitiveType().getPrimitiveTypeName(), elementType.getLogicalTypeAnnotation());
+        Group arrayGroup = new SimpleGroup(listType);
+
+        for (Object value : vals) {
+            Group repeatedGroup = new SimpleGroup(repeatedType);
+            fillPrimitiveGroup(0, value, repeatedGroup, elementType, false);
+            arrayGroup.add(0, repeatedGroup);
```
denalex, line 230: why are we always adding at index 0, would it overwrite the previous elements and result in array only having one repeated subgroup ?

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetUtilitiesTest.java
```diff
@@ -0,0 +1,264 @@
...
+    }
+
+    @Test
+    public void testParsePostgresArrayInValidBooleanArrayPXFGeneratedSchema() {
+        // this situation should never happen as the toString method of booleans (boolout in GPDB) should not return a string in this format
+        String value = "{true,false,true}";
+
+        Exception exception = assertThrows(PxfRuntimeException.class, () -> parquetUtilities.parsePostgresArray(value, PrimitiveType.PrimitiveTypeName.BOOLEAN, null));
+        assertEquals("Error parsing array element: true was not of expected type BOOLEAN", exception.getMessage());
+        assertEquals("Unexpected state since PXF generated the ORC schema.", ((PxfRuntimeException) exception).getHint());
```
denalex, line 154: copy & paste ? ORC does not belong here

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetFileAccessor.java
```diff
@@ -446,77 +447,165 @@ private MessageType generateParquetSchema(List<ColumnDescriptor> columns) {
...
+                    scale = columnTypeModifiers[1];
+                }
+                builder = Types.optionalList()
+                        .optionalElement(PrimitiveTypeName.FIXED_LEN_BYTE_ARRAY)
+                        .length(PRECISION_TO_BYTE_COUNT[precision - 1])
+                        .as(DecimalLogicalTypeAnnotation.decimalType(scale, precision));
+                break;
+            case VARCHARARRAY:
+            case BPCHARARRAY:
+            case TEXTARRAY:
```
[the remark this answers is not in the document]
yimingli-vmware: I mentioned about that once in the standup, PM has gathered MS's datatype which doesn't contain UUID. But I prefer to adding UUID and UUIDARRAY. I'll discuss it again in tomorrow's standup

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetUtilities.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    /**
+     * @param val
+     * @param primitiveTypeName
+     * @param logicalTypeAnnotation
+     * @return
+     */
+    private Object decodeString(String val, PrimitiveType.PrimitiveTypeName primitiveTypeName, LogicalTypeAnnotation logicalTypeAnnotation) {
+        if (val == null) {
```
hpandeycodeit, line 68: The `val` will not be null here.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
```diff
@@ -120,76 +127,56 @@ public OneRow setFields(List<OneField> record) throws IOException {
...
             return;
+        if (type.isPrimitive()) {
+            fillPrimitiveGroup(index, field.val, group, type, true);
+        } else {
+            fillComplexGroup(index, field, group, type);
+        }
+    }
+
+    private void fillPrimitiveGroup(int index, Object value, Group group, Type type, boolean isPrimitive) throws IOException {
+        if (value == null)
```
hpandeycodeit, line 138: `fillGroup` is already checking if `field.val` is null. This check here isn't needed.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetFileAccessor.java
```diff
@@ -439,82 +440,183 @@ private MessageType readSchemaFile(String schemaFile)
...
             }
+            fields.add(generatedType);
+        }
+        return new MessageType("hive_schema", fields);
+    }
 
-            fields.add(builder.named(columnName));
+    /**
+     * Generate parquet schema type for primitive types only
+     *
```
hpandeycodeit, line 469: nit: extra line

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetFileAccessor.java
```diff
@@ -439,82 +440,183 @@ private MessageType readSchemaFile(String schemaFile)
     }
 
     /**
-     * Generate parquet schema using column descriptors
+     * Generate parquet schema for all the supported types using column descriptors
+     *
```
hpandeycodeit, line 444: nit: extra line

automation/src/test/java/org/greenplum/pxf/automation/features/parquet/ParquetReadTest.java
```diff
@@ -3,35 +3,24 @@
...
 import org.greenplum.pxf.automation.structures.tables.pxf.ReadableExternalTable;
-import org.greenplum.pxf.automation.structures.tables.pxf.WritableExternalTable;
 import org.greenplum.pxf.automation.utils.system.ProtocolEnum;
 import org.greenplum.pxf.automation.utils.system.ProtocolUtils;
 import org.testng.annotations.Test;
 
 import java.io.File;
-import java.util.List;
 
-import static java.lang.Thread.sleep;
```
hpandeycodeit, line 12: nit: extra lines

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
yimingli-vmware: For a `LIST` group, it can have many repeated groups. But for each repeated group, it can have at most one element. Take two examples from parquet-read-array, if we want to write a char array`["this is exactly","fifteen chars."]`, the data value and schema would be like:
```
// schema
optional group char_arr (LIST) {
  repeated group bag {
    optional binary array_element (STRING);
  }
}

// value
bag
  array_element: this is exactly
bag
  array_element:  fifteen chars.
```

if we want to write an int array `[7,null,8]`, the data value and schema would be like:

```
// schema
optional group int_arr (LIST) {
  repeated group bag {
    optional int32 array_element;
  }
}

// value
bag
  array_element: 7
bag
bag
  array_element: 8
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetUtilities.java
yimingli-vmware: Why `val` will not be null here? Is that because of the input format of Greenplum? Here I check whether `val` is null is to handle parsing a string representation of an array like `"{7,null,8}"`

> The `val` will not be null here.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
```diff
@@ -112,98 +120,169 @@ public OneRow setFields(List<OneField> record) throws IOException {
...
+                group.add(columnIndex, (Boolean) fieldValue);
                 break;
             default:
-                throw new IOException("Not supported type " + type.asPrimitiveType().getPrimitiveTypeName());
+                throw new UnsupportedTypeException("Not supported primitive type " + primitiveType.asPrimitiveType().getPrimitiveTypeName());
+        }
+    }
+
+    private byte[] getFixedLenByteArray(String value, Type type) {
+        // From org.apache.hadoop.hive.ql.io.parquet.write.DataWritableWriter.DecimalDataWriter#decimalToBinary
```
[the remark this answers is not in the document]
yimingli-vmware: I didn't add anything to this method. This method was part of the `NUMERIC` case in `fillGroup`. But I felt the `NUMERIC` was too long, so I extract this part to a single method, and kept all the comments in the newly created method.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetFileAccessor.java
```diff
@@ -482,38 +490,47 @@ private MessageType generateParquetSchema(List<ColumnDescriptor> columns) {
...
                     throw new UnsupportedTypeException(
                             String.format("Type %d is not supported", columnTypeCode));
             }
 
-            fields.add(builder.named(columnName));
-        }
+            Type type;
+            if (!isArray) {
+                type = length == null ?
+                        Types.optional(primitiveTypeName).as(logicalTypeAnnotation).named(columnName) :
```
denalex, line 524: logical annotation can also be null, need to check for it

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetFileAccessor.java
```diff
@@ -482,38 +490,47 @@ private MessageType generateParquetSchema(List<ColumnDescriptor> columns) {
...
+                        Types.optional(primitiveTypeName).length(length).as(logicalTypeAnnotation).named(columnName);
+            }else{
```
denalex, line 526:
```suggestion
            } else {
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetFileAccessor.java
```diff
@@ -482,38 +490,47 @@ private MessageType generateParquetSchema(List<ColumnDescriptor> columns) {
...
+                type = length == null ?
+                        Types.optionalList().optionalElement(primitiveTypeName).as(logicalTypeAnnotation).named(columnName) :
+                        Types.optionalList().optionalElement(primitiveTypeName).length(length).as(logicalTypeAnnotation).named(columnName);
+            }
 
+            fields.add(type);
```
denalex, line 532:
```suggestion
            Types.BasePrimitiveBuilder<? extends Type,?> builder;
            if (!isArray) {
                builder = Types.optional(primitiveTypeName);
            } else {
                builder = Types.optionalList().optionalElement(primitiveTypeName);
            }
            if (length != null) {
                builder.length(length);
            }
            if (logicalTypeAnnotation != null) {
                builder.as(logicalTypeAnnotation);
            }

            fields.add(builder.named(columnName));
```
try if this works

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetFileAccessor.java
```diff
@@ -429,49 +429,101 @@ private void createParquetWriter() throws IOException, InterruptedException {
...
+        return schema;
+    }
+
+    /**
+     * Validate parquet schema parsed from user provided schema file
+     *
+     * @param schema schema parsed from user provided schema file
+     */
+    private void validateParsedSchema(MessageType schema, List<ColumnDescriptor> columns) {
+        for (int i = 0; i < columns.size(); i++) {
```
bradfordb-vmware, line 451: We should generally prefer for-each style of `for`-loops when iterating over lists:

```java
int i = -1;
for (ColumnDescriptor columnDescriptor : columns) {
    i += 1;
    Type type = schema.getType(i);
    ...
}
```

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetUtilitiesTest.java
```diff
@@ -0,0 +1,258 @@
...
+        String value = "{\"2013-07-13 21:00:05-07\",\"2013-07-13 21:00:05-07\"}";
+
+        // nothing is thrown here because we don't decode strings and check for multi-dimensional-ness. This check is done later
+        List<Object> result = parquetUtilities.parsePostgresArray(value, PrimitiveType.PrimitiveTypeName.INT96, null);
+        assertEquals(2, result.size());
+        assertEquals(Arrays.asList("2013-07-13 21:00:05-07","2013-07-13 21:00:05-07"), result);
+    }
+    @Test
+    public void testParsePostgresArrayByteaArrayEscapeOutput() {
+        // parquet bytea is a BINARY with no annotation
```
bradfordb-vmware, line 90:
```suggestion
        // GPDB bytea is Parquet BINARY with no annotation
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetUtilities.java
```diff
@@ -0,0 +1,119 @@
...
+    /**
+     * Provide more detailed error message
+     *
+     * @param isMultiDimensional Whether this error is caused by inserting a multi-dimensional array
+     * @param val                The string representation we are currently trying to decode
+     * @return The generated error message
+     */
+    private String createErrorHintFromValue(boolean isMultiDimensional, String val) {
+        if (isMultiDimensional) {
+            return String.format("Column value \"%s\" is a multi-dimensional array, PXF does not support multi-dimensional arrays for writing Parquet files.", val);
```
bradfordb-vmware, line 114: Style-nit:

```suggestion
            return String.format("Column value \"%s\" is a multi-dimensional array, PXF does not support multi-dimensional arrays when writing Parquet files.", val);
```

This is a weak suggestion because the message as it is now matches with the message in `OrcUtilities` but the phrase `for writing Parquet files` sounds awkward to me. Maybe a better error message would be

```suggestion
            return String.format("Column value \"%s\" is a multi-dimensional array; PXF does not support writing multi-dimensional arrays to Parquet files.", val);
```

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/ParquetWriteTest.java
```diff
@@ -906,6 +915,667 @@ public void testWriteNumeric() throws Exception {
...
+
+        accessor.setRequestContext(context);
+        accessor.afterPropertiesSet();
+        resolver.setRequestContext(context);
+        resolver.afterPropertiesSet();
+
+        assertTrue(accessor.openForWrite());
+
+        String[] values = new String[]{"0", "1", "2", "3", "4", "5", "6", "7", "8", "9"};
+        insertParquetArrayData(DataType.INT4ARRAY.getOID(), values, resolver, accessor);
```
bradfordb-vmware, line 935: It's not obvious to me what the data we are writing looks like; looking at the code for` insertParquetArrayData`, I think it looks like

```
    col1    
------------
 {NULL,1,1}
 {NULL,2,2}
 ...
 {NULL,8,8}
 {NULL,9,9}
(10 rows)
```

I think it would make the test easier to read and understand if we didn't build up the string representation of the array with code but just had the array literals in the code.

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/ParquetWriteTest.java
```diff
@@ -906,6 +915,667 @@ public void testWriteNumeric() throws Exception {
...
+        for (int i = 0; i < values.length; i++) {
+            if (i == values.length - 1) {
+                continue;
+            }
+            for (int j = 0; j < 3; j++) {
+                if (j != 0) {
+                    expectedValues[i][j] = Integer.parseInt(values[i]);
+                }
+            }
+        }
```
bradfordb-vmware, line 959: Echoing my previous comment, it's not obvious to me what the expected values are. Could we just write this out:

```java
Object[][] expectedValues = new Object[][] {
    new Object[] { ... },
    ...
};
```

automation/src/test/java/org/greenplum/pxf/automation/features/parquet/ParquetReadTest.java
```diff
@@ -3,13 +3,14 @@
 import org.greenplum.pxf.automation.features.BaseFeature;
 import org.greenplum.pxf.automation.structures.tables.basic.Table;
 import org.greenplum.pxf.automation.structures.tables.utils.TableFactory;
+import org.greenplum.pxf.automation.structures.tables.pxf.ReadableExternalTable;
```
bradfordb-vmware, line 6: It looks like these imports are unused and not needed.

automation/src/test/java/org/greenplum/pxf/automation/features/parquet/ParquetWriteTest.java
```diff
@@ -122,40 +190,52 @@ public void parquetWritePaddedChar() throws Exception {
...
     }
 
-
     @Test(groups = {"features", "gpdb", "security", "hcfs"})
     public void parquetWriteUndefinedPrecisionNumeric() throws Exception {
+
         String filename = "parquet_write_undefined_precision_numeric";
-        prepareWritableExternalTable("pxf_parquet_write_undefined_precision_numeric",
-                UNDEFINED_PRECISION_NUMERIC, hdfsPath + filename, null);
+        exTable = new WritableExternalTable("pxf_parquet_write_undefined_precision_numeric",
```
bradfordb-vmware, line 215: I don't think we want to create `WritableExternalTable`s directly but instead use the static methods on `TableFactory`. This should help future proof this code when we start testing with FDW. It's also not clear to me why we can't still use `prepareWritableExternalTable` here (and `prepareReadableExternalTable` below).

automation/src/test/java/org/greenplum/pxf/automation/features/parquet/ParquetWriteTest.java
```diff
@@ -166,12 +246,119 @@ public void parquetWriteNumericWithPrecisionAndScale() throws Exception {
...
+    public void parquetWriteLists() throws Exception {
+        String writeTableName = "pxf_parquet_write_list";
+        String readTableName = "pxf_parquet_read_list";
+        String fullTestPath = hdfsPath + PARQUET_WRITE_LIST;
+
+        prepareWritableExternalTable(writeTableName, PARQUET_LIST_TABLE_COLUMNS, fullTestPath, null);
+        gpdb.runQuery("INSERT INTO " + exTable.getName() + " SELECT id, bool_arr, smallint_arr, int_arr, bigint_arr, real_arr, " +
+                "double_arr, text_arr, bytea_arr, char_arr, varchar_arr, date_arr, numeric_arr FROM " + PXF_PARQUET_LIST_TYPES);
+
+        if (protocol != ProtocolEnum.HDFS && protocol != ProtocolEnum.FILE) {
```
bradfordb-vmware, line 259: It looks like this `if`-block is repeated at least three times in this test class, should we extract this bit of logic into a helper method?

automation/src/test/java/org/greenplum/pxf/automation/features/parquet/ParquetWriteTest.java
```diff
@@ -166,12 +246,119 @@ public void parquetWriteNumericWithPrecisionAndScale() throws Exception {
...
+            }
+        }
+
+        prepareReadableExternalTable(readTableName, PARQUET_TIMESTAMP_LIST_TABLE_COLUMNS, fullTestPath);
+
+        runTincTest("pxf.features.parquet.write_list.timestamp_list.runTest");
+    }
+
+
+//    @Test(groups = {"features", "gpdb"})
```
bradfordb-vmware, line 304: This test is currently disabled. Do we need to uncomment this line and run the test? If not, can we remove the test method?

server/pxf-hdfs/src/test/resources/parquet/unsupported_map_type.schema
```diff
@@ -0,0 +1,8 @@
+message spark_schema {
+  optional group unsupported_map (MAP) {
+    repeated group map {
+      required int32 key;
+      optional int32 value;
+    }
+  }
+}
```
hpandeycodeit, line 8: does these schema files have to be separate or can we merge them to one ?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetFileAccessor.java
```diff
@@ -429,49 +429,101 @@ private void createParquetWriter() throws IOException, InterruptedException {
...
-     * Generate parquet schema using column descriptors
+     * Generate parquet schema for all the supported types using column descriptors
+     *
+     * @param columns contains Greenplum data type and column name
+     * @return the generated parquet schema used for write
      */
     private MessageType generateParquetSchema(List<ColumnDescriptor> columns) {
         LOG.debug("{}-{}: Generating parquet schema for write using {}", context.getTransactionId(),
                 context.getSegmentId(), columns);
         List<Type> fields = new ArrayList<>();
```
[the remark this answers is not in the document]
yimingli-vmware: yes, they should be 1:1 mapping

automation/src/test/java/org/greenplum/pxf/automation/features/parquet/ParquetWriteTest.java
yimingli-vmware: changed them into `prepareWritableExternalTable` and `prepareReadableExternalTable`, still can pass the test locally

automation/src/test/java/org/greenplum/pxf/automation/features/parquet/ParquetWriteTest.java
```diff
@@ -0,0 +1,420 @@
...
+                int attempts = 0;
+                while (!hdfs.doesFileExist(file) && attempts++ < 20) {
+                    sleep(1000);
+                }
+            }
+        }
+        prepareReadableExternalTable(readTableName,
+                PARQUET_PRIMITIVE_TABLE_COLUMNS, hdfsPath + filename);
+        gpdb.runQuery("CREATE OR REPLACE VIEW parquet_view AS SELECT s1, s2, n1, d1, dc1, " +
+                "CAST(tm AS TIMESTAMP WITH TIME ZONE) AT TIME ZONE 'PDT' as tm, " +
```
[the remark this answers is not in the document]
yimingli-vmware: It's formatted by auto formatting tool

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetFileAccessor.java
```diff
@@ -446,77 +447,165 @@ private MessageType generateParquetSchema(List<ColumnDescriptor> columns) {
...
+                // of day. It must annotate an int32 that stores the number
+                // of days from the Unix epoch, 1 January 1970. The sort
+                // order used for DATE is signed.
+                builder = Types.optional(PrimitiveTypeName.INT32)
+                        .as(dateType());
+                break;
+            case TIME:
+            case VARCHAR:
+            case BPCHAR:
+            case TEXT:
```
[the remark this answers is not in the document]
yimingli-vmware: I think we should add support for UUID. But how Parquet handle UUID is not clear to me and I may need to spend some time to research. So I decide not to add UUID for now

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/PgUtilities.java
```diff
@@ -69,7 +69,7 @@ public String[] splitArray(String str) {
      * Split the outer-most Postgres array into its constituent elements
      *
      * Re-used from GPHDFS, commit 3b0bfdc169
-     * @param value string with a Postgres array
+     * @param value     string with a Postgres array
```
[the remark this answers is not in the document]
yimingli-vmware: I think they are added by auto formatting tool

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetFileAccessor.java
```diff
@@ -429,92 +427,174 @@ private void createParquetWriter() throws IOException, InterruptedException {
...
+        }
+
+        for (int i = 0; i < schema.getFieldCount(); i++) {
+            Type type = schema.getType(i);
+            if (!type.isPrimitive()) {
+                validateComplexType(type.asGroupType());
+            }
+            // TODO: Need to check ordering and type matching between schema and columns
+        }
+    }
```
denalex, line 461: It's good you switched for iterating over schema and not GPDB columns for now. However, we still want to do iterator based traversal, not index-based access, so:
```suggestion
        for (Type type : schema.getFields()) {
            if (!type.isPrimitive()) {
                validateComplexType(type.asGroupType());
            }
            // TODO: Need to check ordering and type matching between schema and columns
        }
    }
```

denalex (approved):

automation/tincrepo/main/pxf/features/parquet/write_list/write_list_read_with_hive/expected/query01.ans
```diff
@@ -0,0 +1,44 @@
+-- start_ignore
+-- end_ignore
+-- @description query01 for writing Parquet List data then reading with Hive (except Timestamp List and Binary List)
+SET bytea_output=hex;
+SET
+SELECT id,bool_arr,smallint_arr,int_arr,bigint_arr,real_arr,double_arr,text_arr,char_arr,varchar_arr,date_arr,numeric_arr
```
ashuka24, line 6: can you add proper spacing here and an explanation why bytea_arr and timestamp where not added?

automation/src/test/java/org/greenplum/pxf/automation/features/parquet/ParquetWriteTest.java
```diff
@@ -420,38 +452,21 @@ private void insertArrayDataWithoutNulls(String exTable, int numRows) throws Exc
...
-                    // 15 characters
-                    .add(String.format("[\"%s%s\"]", i, StringUtils.repeat(" ", i > 9 ? 13 : 14)))  // DataType.BPCHARARRAY
-                    .add(String.format("[\"var%02d\"]", i))                                             // DataType.VARCHARARRAY
-                    .add(String.format("[\"2010-01-%02d\"]", (i % 30) + 1))                             // DataType.DATEARRAY
-                    .add(String.format("[%s]", numericString))                                           // DataType.NUMERICARRAY
                     ;
 
             // Only 1 bytea element in bytea_array. Need to convert the bytea result in the array into a hex string
-            String byteaArrayString = queryResultData.get(i).get(8);
+            String byteaArrayString = queryResultData.get(i).get(1);
```
ashuka24, line 465: why was this changed from 8 to 1

automation/src/test/java/org/greenplum/pxf/automation/features/parquet/ParquetWriteTest.java
```diff
@@ -420,38 +452,21 @@ private void insertArrayDataWithoutNulls(String exTable, int numRows) throws Exc
...
-                    .add(String.valueOf(i))    // always not-null row index, column index starts with 0 after it
-                    .add(String.format("[%b]", i % 2 != 0))                                             // DataType.BOOLEANARRAY
-                    .add(String.format("[%d]", 10L + i % 32000))                                        // DataType.INT2ARRAY
-                    .add(String.format("[%d]", 100L + i))                                               // DataType.INT4ARRAY
-                    .add(String.format("[%d]", 123456789000000000L + i))                                // DataType.INT8ARRAY
-                    .add(String.format("[%.4f]", Float.valueOf(i + 0.00001f * i)))                   // DataType.FLOAT4ARRAY
-                    .add(String.format("[%f]", i + Math.PI))                                            // DataType.FLOAT8ARRAY
-                    .add(String.format("[\"row-%02d\"]", i))                                            // DataType.TEXTARRAY
+                    .add(String.valueOf(i))    // always not-null row index, column index starts with 0 after it                    // DataType.TEXTARRAY
```
ashuka24, line 460: the value here is for DataType.BYTEAARRAY correct?

automation/src/test/java/org/greenplum/pxf/automation/features/parquet/ParquetWriteTest.java
```diff
@@ -98,6 +100,22 @@ public class ParquetWriteTest extends BaseFeature {
...
+    private static final String[] PARQUET_PRIMITIVE_ARRAYS_TABLE_COLUMNS_READ_FROM_HIVE = {
+            "id                   int",
+            "bool_arr             text",           // DataType.BOOLARRAY
+            "smallint_arr         text",           // DataType.INT2ARRAY
+            "int_arr              text",           // DataType.INT4ARRAY
+            "bigint_arr           text",           // DataType.INT8ARRAY
+            "real_arr             text",           // DataType.FLOAT4ARRAY
+            "double_arr           text",           // DataType.FLOAT8ARRAY
+            "text_arr             text",           // DataType.TEXTARRAY
+            "bytea_arr            text",           // DataType.BYTEAARRAY
```
ashuka24, line 113: you should be able to just remove this column here and get the correct results with a select *

automation/tincrepo/main/pxf/features/parquet/write_list/write_list_read_with_hive/sql/query01.sql
```diff
@@ -0,0 +1,5 @@
+-- @description query01 for writing Parquet List data then reading with Hive (except Timestamp List and Binary List)
+SET bytea_output=hex;
```
ashuka24, line 2: since we do not deal with bytea, it is not necessary to set this here

automation/src/test/java/org/greenplum/pxf/automation/features/parquet/ParquetWriteTest.java
yimingli-vmware: before byteaarray was at index 8, now it's at index 1

automation/src/test/java/org/greenplum/pxf/automation/features/parquet/ParquetWriteTest.java
yimingli-vmware: the first column is id, the second one is BYTEAARRAY

automation/src/test/java/org/greenplum/pxf/automation/features/parquet/ParquetWriteTest.java
yimingli-vmware: I think it's not easy for me to do that. I still need `bytea_arr ` in my hive table to compare its value with expected values. If I remove this line in the DDL, the data source  table of my pxf_readable_table should not contain `bytea_arr` neither. So I may need to create another table without `bytea_arr` column

PR #877, opened 2022-10-06
Update shell scripts to use a more consistent style

bradfordb-vmware:
* All functions should use the `function` keyword, with the opening brace on the same line as the function name
* Indents should be four spaces
  * Indents for "Here Documents" where the redirection operator is `<<-` should use tabs since bash only removes leading tab characters
* Style can be checked with [`shfmt`][0] however it should not be used to actually adjust the formatting since its not possible to preserve the de-facto conventions of spaces around redirect operators and padding for alignment

    ```sh
    shfmt --diff --indent 4 --case-indent <script-file>
    ```

[0]: https://github.com/mvdan/sh

### DIFF
--- a/server/pxf-service/src/scripts/kill-pxf.sh
+++ b/server/pxf-service/src/scripts/kill-pxf.sh
@@ -1,18 +1,18 @@
 #!/bin/bash
 
-log() {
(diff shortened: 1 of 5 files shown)


ashuka24 (approved):

denalex (approved):
is there any artifact for this style that we can import into IntelliJ ? If yes, we should check it into this repo in some dev folder.

PR #878, opened 2022-10-10
Modifying Tinc Tests for GP7x 

hpandeycodeit:
Modifying the tinc tests so that they can pass for GP7x as well as continue to pass for earlier versions of GPDB ( 6x, 5x). 

- In some of the tests, a GUC `set extra_float_digits=0;` is set. This GUC has a default value of `1` in PG12 and thus the columns of type `REAL` displays one digit extra. So to keep the behavior consistent, setting this to `0`. 

- Some of the answer test files I have modified the `Order by` clause to add another column to maintain the same order of the records across versions because the ordering for the results is little different in GP7 coming out a little different than the earlier versions. 

- There is a change in ProjectionInfo in PG12. i.e, how the plan is getting generated and how it's defining the projection columns Ref (https://github.com/greenplum-db/gpdb/commit/b10967eddf964f8c0a11060cf3f366bbdd1235f6) . So changed a couple of test cases to accordingly. 

`GpupgradeTest.testGpdbUpgradeScenario` test is not passing due to python3 in GP7 vs we are using python2 in our tinc tests. So skipping this test.


automation/README.md
```diff
@@ -103,6 +103,17 @@ Note: if you get an error saying that the jar does not exist, ensure that you ha
...
+### Running automation tests for GPDB7
+Tinc still runs on Python2 so it won't work with Python libraries that come with GP7. To overcome that, here is a small workaround in place.
+We will copy the GP6 python libraries and then run the automation for GP7.
+
+To support running the tests for GPDB 7, we have created this new directory, `tinc-python-deps` under `/pxf/automation/tinc/main`
+
+Steps:
+1. Download the GP6 Python libs `gsutil cp gs://data-gpdb-ud-pxf-build-resources/automation-dependencies/gp6-python-libs.tar.gz ~/workspace/pxf/automation/tinc/main/tinc-python-deps`
+2. Unzip the file `tar -xvzf ~/workspace/pxf/automation/tinc/main/tinc-python-deps/gp6-python-libs.tar.gz`
+3. export PXFAUTOMATIONTINCDEPS `export PXFAUTOMATIONTINCDEPS=~/workspace/pxf/automation/tinc/main/tinc-python-deps/`
```
ashuka24, line 115: could we add underscores to make this a little more readable? `PXF_AUTOMATION_TINC_DEPS`

automation/README.md
```diff
@@ -103,6 +103,17 @@ Note: if you get an error saying that the jar does not exist, ensure that you ha
...
 b) only have 1 jar file inside `/usr/local/pxf/application/`
 
+### Running automation tests for GPDB7
+Tinc still runs on Python2 so it won't work with Python libraries that come with GP7. To overcome that, here is a small workaround in place.
+We will copy the GP6 python libraries and then run the automation for GP7.
+
+To support running the tests for GPDB 7, we have created this new directory, `tinc-python-deps` under `/pxf/automation/tinc/main`
+
+Steps:
+1. Download the GP6 Python libs `gsutil cp gs://data-gpdb-ud-pxf-build-resources/automation-dependencies/gp6-python-libs.tar.gz ~/workspace/pxf/automation/tinc/main/tinc-python-deps`
```
ashuka24, line 113: how was this tar generated? are we expected to maintain up-to-date versions for this jar?

automation/src/main/java/org/greenplum/pxf/automation/components/tinc/Tinc.java
```diff
@@ -28,8 +28,23 @@ public void init() throws Exception {
 
 		super.init();
 
-		runCommand("cd " + gphome);
-		runCommand("source greenplum_path.sh");
+		// This property is for setting PYTHONPATH in local env.
+		// This change is for running the Automation tests in GP7.
+		// Since, Tinc still runs on Python2, It won't work with Python3 libs which are present with GP7, so we need to set the PYTHONPATH
+		// pointing towards the Python2 libs.Copy the GP6 Python libs from your local installation GP6
```
ashuka24, line 34: nit 
```suggestion
		// pointing towards the Python2 libs. Copy the GP6 Python libs from your local installation GP6
```

automation/src/test/java/org/greenplum/pxf/automation/features/gpupgrade/GpupgradeTest.java
```diff
@@ -34,6 +34,10 @@ protected void afterMethod() throws Exception {
 
     @Test(groups = {"features", "gpdb"})
     public void testGpdbUpgradeScenario() throws Exception {
+        // TODO: need to resolve tinc's python2 deps for GP7
```
ashuka24, line 37: I'm a little confused why this is still here. It seems like we have resolved these dependencies in by bringing in the `tinc-python-deps`

automation/tincrepo/main/pxf/features/columnprojection/checkColumnProjection/expected/query01.ans
```diff
@@ -220,20 +222,20 @@ SELECT t0, colprojvalue FROM test_column_projection WHERE a1 <= 5 ORDER BY t0;
...
-SELECT t0, colprojvalue FROM test_column_projection GROUP BY t0, colprojvalue HAVING AVG(a1) < 5 ORDER BY t0;
- t0 |    colprojvalue
-----+--------------------
- A  | t0|a1|colprojvalue
- B  | t0|a1|colprojvalue
- C  | t0|a1|colprojvalue
- D  | t0|a1|colprojvalue
- E  | t0|a1|colprojvalue
+SELECT t0, b2, colprojvalue FROM test_column_projection GROUP BY t0, colprojvalue, b2 HAVING AVG(a1) < 5 ORDER BY t0;
+ t0 | b2 |     colprojvalue
```
ashuka24, line 226: this test looks like it's producing different results. is this to be expected?

automation/tincrepo/main/pxf/features/columnprojection/checkColumnProjection/sql/query01.sql
```diff
@@ -4,6 +4,7 @@ CREATE TABLE t0_values(key char(1), value int) DISTRIBUTED BY (key);
 INSERT INTO t0_values VALUES('A', 50);
 -- end_ignore
 -- @description query01 for PXF Column Projection Support
+SET extra_float_digits=0;
```
ashuka24, line 7: it looks like we set this in multiple different files. Could we add this set command on the java side while setting up the initial system instead of in each individual file?

concourse/pipelines/templates/dev_build_pipeline-tpl.yml
```diff
@@ -350,15 +371,6 @@ resources:
     json_key: ((ud/pxf/secrets/pxf-storage-service-account-key))
     versioned_file: build-dependencies/extaccess.h
 
-
-- name: pxf-build-dependencies
```
ashuka24, line 354: why are the build dependences no longer needed here? it does not seem like we removed them from the rest of the pipeline?

concourse/pipelines/templates/dev_build_pipeline-tpl.yml
```diff
@@ -125,6 +125,14 @@ resources:
     json_key: ((ud/pxf/secrets/pxf-storage-service-account-key))
     versioned_file: automation-dependencies/pxf-automation-dependencies.tar.gz
 
+- name: gp6-python-libs
```
ashuka24, line 128: i see the changes here in the `dev_build_pipeline-tpl`. Should these changes also be ported to the `build_pipeline-tpl` as well?

concourse/scripts/test.bash
```diff
@@ -313,7 +306,8 @@ fi
   # we will set PYTHONPATH to point to the set of python libs compiled with Python2 for GP6
   if [[ ${GP_VER} == 7 ]]; then
     local gp6_python_libs=~gpadmin/python
-    echo "export PYTHONPATH=${gp6_python_libs}" >> /usr/local/greenplum-db/greenplum_path.sh
+    #echo "export PYTHONPATH=${gp6_python_libs}" >> /usr/local/greenplum-db/greenplum_path.sh
```
ashuka24, line 309: since these are commented out, can we just delete these 2 lines?

external-table/src/gpdbwritableformatter.c
```diff
@@ -478,9 +481,30 @@ gpdbwritableformatter_export(PG_FUNCTION_ARGS)
...
+        Relation rel = FORMATTER_GET_RELATION(fcinfo);
+        if(rel != NULL)
+        {
+            ExtTableEntry *exttbl = GetExtTableEntry(rel->rd_id);
+            if (exttbl->encoding != PG_UTF8) {
+                ereport(ERROR, (errcode(ERRCODE_EXTERNAL_ROUTINE_EXCEPTION),
+                        errmsg("gpdbwritable formatter can only export UTF8 formatted data. Define the external table with ENCODING UTF8")));
+            }
+        }
+        // TODO: If for some reason the Relation is null here, then shall we throw an error anyways ??
```
ashuka24, line 502: if so, i think the error should be updated to reflect that it was thrown due to the relation being null.

external-table/src/gpdbwritableformatter.c
```diff
@@ -478,9 +481,30 @@ gpdbwritableformatter_export(PG_FUNCTION_ARGS)
...
 	{
-		if (FORMATTER_GET_EXTENCODING(fcinfo) != PG_UTF8)
-			ereport(ERROR, (errcode(ERRCODE_EXTERNAL_ROUTINE_EXCEPTION),
-							errmsg("gpdbwritable formatter can only export UTF8 formatted data. Define the external table with ENCODING UTF8")));
+        // In GP7 FORMATTER_GET_EXTENCODING(fcinfo) gets the database encoding which may not match the table encoding
+        // and thus results in exception here. So getting the table encoding from the ExtTableEntry
+#if PG_VERSION_NUM < 120000
+        if (FORMATTER_GET_EXTENCODING(fcinfo) != PG_UTF8)
+        {
+                 ereport(ERROR, (errcode(ERRCODE_EXTERNAL_ROUTINE_EXCEPTION),
```
ashuka24, line 489: it looks like the same error and err message are used in multiple places. can we pull this out into a variable?

external-table/src/pxfheaders.c
```diff
@@ -557,49 +557,43 @@ add_projection_desc_httpheader_pg12(CHURL_HEADERS headers,
 	int			   number;
 	int			   numTargetList;
 	char			long_number[sizeof(int32) * 8];
-	//int				*varNumbers = projInfo->pi_varNumbers;
+    // In versions < 120000, projInfo->pi_varNumbers contains atttribute numbers of SimpleVars
+    // Since,this pi_varNumbers doesn't exist in PG12 and above, we can add the attribute numbers by
```
ashuka24, line 561: can you add a comment on the relation between PG12 and GP version here for clarity

external-table/src/pxfheaders.c
```diff
@@ -609,6 +603,7 @@ add_projection_desc_httpheader_pg12(CHURL_HEADERS headers,
             add_attnums_from_targetList( (Node *) gstate, l);
 		}
 
+        int i=0;
```
ashuka24, line 606: nit: spacing

external-table/src/pxfheaders.c
```diff
@@ -617,6 +612,8 @@ add_projection_desc_httpheader_pg12(CHURL_HEADERS headers,
 				add_projection_index_header(headers,
 											formatter, attno - 1, long_number);
 				numTargetList++;
+                varNumbers[i] = attno;
```
ashuka24, line 615: nit: spacing

ashuka24:
it also looks like there's an empty `.gitkeep` file. Is this file necessary?

automation/src/test/java/org/greenplum/pxf/automation/features/gpupgrade/GpupgradeTest.java
hpandeycodeit: This test isn't passing in GP7 so skipping it. I have updated the comment here.

automation/tincrepo/main/pxf/features/columnprojection/checkColumnProjection/expected/query01.ans
hpandeycodeit: Yes GP7 produces it little differently. So I have updated it so that it remains consistent across all the PXF versions.

concourse/pipelines/templates/dev_build_pipeline-tpl.yml
hpandeycodeit: That was a duplicate block `pxf-build-dependencies` still in the file.

external-table/src/gpdbwritableformatter.c
hpandeycodeit: Since the next few comments belong to the other PR, I will address these as a part of #857

automation/README.md
hpandeycodeit: This is the set of Python libs from GP6 and are present in our gcloud bucket. Just downloading these to the above directory is enough to get the Test running in local env.

concourse/pipelines/templates/dev_build_pipeline-tpl.yml
```diff
@@ -174,6 +182,7 @@ resources:
 {% set gp_ver = None %}
 
 {% for gp_ver in range(5, 8) %}
+{% if gp_ver != 7 %}
```
yimingli-vmware, line 185: Seems like gp5, gp6 and gp8 would get images through docker? How does gp7 get image?

concourse/pipelines/templates/dev_build_pipeline-tpl.yml
```diff
@@ -448,14 +451,15 @@ jobs:
...
       GROUP: gpdb,proxy,profile
 {% endif %}
       SECRET_ACCESS_KEY: ((tf-machine-secret-access-key))
 {% if slack_notification %}
     <<: *slack_alert
 {% endif %}
+{% endif %}
 {% endfor %}
 
 {% if oel7 %}
```
yimingli-vmware, line 465: What is `oel7`?

automation/tincrepo/main/pxf/features/general/alter/pxfwritable_import/without_column_projection/sql/query01.sql
```diff
@@ -8,8 +8,8 @@
 -- s/DETAIL/CONTEXT/
 --
 --
--- m/ALTER/
--- s/ALTER EXTERNAL TABLE/ALTER FOREIGN TABLE/g
+-- m/ALTER FOREIGN TABLE/
```
yimingli-vmware, line 11: Are line 11 and 12 descriptions? What are `m/` and `s/`?

automation/tinc/main/ext/qautils/gppylib/gpparseopts.py
```diff
@@ -73,13 +73,13 @@ class OptChecker(Option):
     def datetimeCheck(option, opt, value):
         try:
             return str_to_datetime(value)
-        except DatetimeValueError, e:
+        except DatetimeValueError as e:
```
yimingli-vmware, line 76: I'm not familiar with python. What is the difference between `except ... , ...` and `except ... as ...`?

automation/README.md
```diff
@@ -103,6 +103,17 @@ Note: if you get an error saying that the jar does not exist, ensure that you ha
 a) installed the PXF server, and
 b) only have 1 jar file inside `/usr/local/pxf/application/`
 
+### Running automation tests for GPDB7
+Tinc still runs on Python2 so it won't work with Python libraries that come with GP7. To overcome that, here is a small workaround in place.
+We will copy the GP6 python libraries and then run the automation for GP7.
```
yimingli-vmware, line 108: So after copying the python2 lib, GP7 has 2 versions of python, and python2 is only for running tinc tests. Am i understanding right?

concourse/pipelines/templates/dev_build_pipeline-tpl.yml
```diff
@@ -578,67 +584,103 @@ jobs:
...
+    file: pxf_src/concourse/tasks/build.yml
+    params:
+      LICENSE: ((ud/pxf/common/rpm-license))
+      VENDOR: ((ud/pxf/common/rpm-vendor))
+      GOOGLE_CREDENTIALS: ((ud/pxf/secrets/ccp-ci-service-account-key))
+      GOOGLE_PROJECT_ID: ((ud/pxf/common/google-project-id))
+      GOOGLE_ZONE: ((ud/pxf/common/google-zone))
+  - put: pxf_gp[[gp_ver]]_tarball_rhel8
+    params:
+      file: dist/pxf-gp[[gp_ver]]-*.el8.tar.gz
```
denalex, line 654: we can fold this build job into the new loop on line 550

concourse/pipelines/templates/dev_build_pipeline-tpl.yml
```diff
@@ -578,67 +584,103 @@ jobs:
...
+      - get: singlecluster
+        resource: singlecluster-hdp2
+  - task: Test PXF-GP[[gp_ver]]-HDP2 on RHEL8
+    file: pxf_src/concourse/tasks/test.yml
+    image: gpdb[[gp_ver]]-pxf-dev-rhel8-image
+    params:
+      ACCESS_KEY_ID: ((tf-machine-access-key-id))
+      GP_VER: [[gp_ver]]
+      GROUP: gpdb,proxy,profile
+      SECRET_ACCESS_KEY: ((tf-machine-secret-access-key))
```
denalex, line 681: this test job will also fold into the new loop

concourse/pipelines/templates/dev_build_pipeline-tpl.yml
```diff
@@ -659,6 +701,7 @@ SECRET_ACCESS_KEY: ((tf-machine-secret-access-key))
     - get: pxf-automation-dependencies
     - get: singlecluster
       resource: singlecluster-hdp2
+    - get: gp6-python-libs
```
denalex, line 704: no need here and everywhere below ?

automation/README.md
bradfordb-vmware: Since the tar file contains RHEL-specific artifacts that won't work on macOS and might not work on Linux distributions other than RHEL7, should we replace these steps with instructions for locating and copying the required python modules from the local GP6 installation? I think something like this:

```sh
cp -a "$(${GPHOME}/bin/pg_config --libdir)/python" ./tinc/main/tinc-python-deps
```

But if we are just setting `PXF_AUTOMATION_TINC_DEPS` to the location, you might not even need to copy. Just doing

```sh
export PXF_AUTOMATION_TINC_DEPS="${GP6HOME}/lib/python"
```

might be sufficient.

automation/src/main/java/org/greenplum/pxf/automation/components/tinc/Tinc.java
```diff
@@ -28,8 +28,19 @@ public void init() throws Exception {
...
 
-		runCommand("cd " + gphome);
-		runCommand("source greenplum_path.sh");
+		// This property is for setting PYTHONPATH in local env.
+		// This change is for running the Automation tests in GP7.
+		// Since, Tinc still runs on Python2, It won't work with Python3 libs which are present with GP7, so we need to set the PYTHONPATH
+		// pointing towards the Python2 libs. Copy the GP6 Python libs from your local installation GP6
+		String  pxfAutomationTincDeps = System.getenv("PXF_AUTOMATION_TINC_DEPS");
+		String pxfAutomationTincDeps_cmd;
+		if(pxfAutomationTincDeps != null ) {
```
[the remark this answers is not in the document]
bradfordb-vmware: I'm wondering if we should use `StringUtils.isNotBlank()` since this will handle both `null` and empty-string correctly.

automation/src/main/java/org/greenplum/pxf/automation/components/tinc/Tinc.java
```diff
@@ -28,8 +28,19 @@ public void init() throws Exception {
...
+			pxfAutomationTincDeps_cmd = "export PYTHONPATH=" + pxfAutomationTincDeps + "/python";
```
bradfordb-vmware, line 38: I would not append anything to the environment variable, it's easy enough for the user to put the full path.

automation/src/main/java/org/greenplum/pxf/automation/components/tinc/Tinc.java
```diff
@@ -28,8 +28,19 @@ public void init() throws Exception {
...
+		}
+		else{
+			pxfAutomationTincDeps_cmd = "export PYTHONPATH=/home/gpadmin/python";
```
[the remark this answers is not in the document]
bradfordb-vmware: I'd vote for `/usr/local/greenplum-db/lib/python` since that should be location if Greenplum is installed from an official RPM.

automation/tinc/main/ext/qautils/gppylib/gpparseopts.py
bradfordb-vmware: This is a syntax change from Python 2 to Python 3. I don't think we should be making any changes to the tinc code.

automation/tincrepo/main/pxf/features/hive/orc_repeating/expected/query01.ans
```diff
@@ -1,34 +1,9 @@
 -- start_ignore
 -- end_ignore
 -- @description query01 for PXF Hive ORC vectorized with repeating data cases
+
 \pset null 'NIL'
 Null display is "NIL".
-\d pxf_hivevectorizedorc_repeating_no_nulls
```
ashuka24, line 6: why was describing the table removed? was there a reason it was there in the first place?

automation/src/main/java/org/greenplum/pxf/automation/components/tinc/Tinc.java
```diff
@@ -28,8 +29,19 @@ public void init() throws Exception {
 
 		super.init();
 
-		runCommand("cd " + gphome);
-		runCommand("source greenplum_path.sh");
```
bradfordb-vmware, line 32: It looks like we still need to source `greenplum_path.sh` because we need `LD_LIBRARY_PATH` to be set and pointing to `$GP7HOME/lib`.

automation/src/main/java/org/greenplum/pxf/automation/components/tinc/Tinc.java
```diff
@@ -28,8 +29,19 @@ public void init() throws Exception {
...
+		// Since, Tinc still runs on Python2, It won't work with Python3 libs which are present with GP7, so we need to set the PYTHONPATH
+		// pointing towards the Python2 libs.
+		String pxfAutomationTincDeps = System.getenv("PXF_AUTOMATION_TINC_DEPS");
+		String pxfAutomationTincDepsCommand;
+		if (StringUtils.isNotBlank(pxfAutomationTincDeps)) {
+			pxfAutomationTincDepsCommand = "export PYTHONPATH=" + pxfAutomationTincDeps;
+		}
+		else {
+			pxfAutomationTincDepsCommand = "export PYTHONPATH=/home/gpadmin/python";
+		}
```
bradfordb-vmware, line 43: Since it looks like we still need to source `$GPHOME/greenplum_path.sh` and this will set `PYHTONPATH` to the correct location when testing against GP6, we don't need to set this variable then. Let's only override `PYTHONPATH` if the user has set it.

```suggestion
		String pxfAutomationTincDeps = System.getenv("PXF_AUTOMATION_TINC_DEPS");
		if (StringUtils.isNotBlank(pxfAutomationTincDeps)) {
			runCommand("export PYTHONPATH=" + pxfAutomationTincDeps);
		}
```

automation/tincrepo/main/pxf/features/hive/orc_repeating/expected/query01.ans
hpandeycodeit: Not sure why the `\d` was there but I don't think it was serving any purpose other than the table description. Keeping it there wouldn't work since we have to deal with different table description formats for GP6 and GP7 and will be hard to match. 

For eg: GP6: 

```
gpadmin=# \d+ test
                    External table "public.test"
 Column |  Type   | Modifiers | Storage | Stats target | Description 
--------+---------+-----------+---------+--------------+-------------
 t1     | integer |           | plain   |              | 
Type: readable
Encoding: UTF8
Format type: text
Format options: delimiter '	' null '\N' escape '\'
External options: {}
External location: pxf://data/id.txt?PROFILE=HDFS:TEXT
Execute on: all segments
```

GP7:

```
postgres=# \d+ test
                                      Foreign table "public.test"
 Column |  Type   | Collation | Nullable | Default | FDW options | Storage | Stats target | Description 
--------+---------+-----------+----------+---------+-------------+---------+--------------+-------------
 t1     | integer |           |          |         |             | plain   |              | 
FDW options: (format 'text', delimiter '	', "null" E'\\N', escape E'\\', format_type 't', location_uris 'pxf://data/id.txt?PROFILE=HDFS:TEXT', execute_on 'ALL_SEGMENTS', log_errors 'f', encoding '6', is_writable 'false')

```

automation/tincrepo/main/pxf/features/general/alter/pxfwritable_import/without_column_projection/sql/query01.sql
ashuka24: m stands for match
s stands for sub
so we match on line 11 and we substitute it for what is on line 12

automation/src/test/java/org/greenplum/pxf/automation/features/hcfs/HcfsGlobbingTest.java
```diff
@@ -92,7 +92,11 @@ public void testMatchAStringFromStringSet() throws Exception {
 
     @Test(groups = {"gpdb", "hcfs", "security"})
     public void testJavaRegexSpecialChars() throws Exception {
-        prepareTestScenario("java_regex_special_chars", "(.|+)bc", "abc", null, null, "(.|+)*");
+        // The earlier glob (.|+)* is failing for GP7 with this error:
+        // select * from hcfs_glob_java_regex_special_chars;
+        // ERROR:  invalid URI '+)*?PROFILE=hdfs:text' : undefined structure
+        // Since it doesn't like the pipe | in the URI, changing this test a bit to remove the pipe 
+        prepareTestScenario("java_regex_special_chars", "(.+)bc", "abc", null, null, "(.+)*");
```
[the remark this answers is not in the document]
denalex: this is due to the logic in `gp_exttable_fdw` that treats | symbol as a separator for the urls

automation/src/test/java/org/greenplum/pxf/automation/features/gpupgrade/GpupgradeTest.java
```diff
@@ -36,6 +36,11 @@ protected void afterMethod() throws Exception {
 
     @Test(groups = {"features", "gpdb"})
     public void testGpdbUpgradeScenario() throws Exception {
+
+        // Skipping this test for GP7 since this isn't passing for GP7
```
ashuka24, line 40: do we know the reason it is not passing in GP7 or what it would take to get this passing for GP7?

concourse/pipelines/templates/build_pipeline-tpl.yml
```diff
@@ -434,7 +457,7 @@ resources:
   icon: docker
   source:
     repository: google/cloud-sdk
-    tag: slim
+    tag: 402.0.0-slim
```
ashuka24, line 460: why is this change needed? why can we not use latest?

automation/src/test/java/org/greenplum/pxf/automation/features/hcfs/HcfsGlobbingTest.java
```diff
@@ -92,7 +92,11 @@ public void testMatchAStringFromStringSet() throws Exception {
 
     @Test(groups = {"gpdb", "hcfs", "security"})
     public void testJavaRegexSpecialChars() throws Exception {
-        prepareTestScenario("java_regex_special_chars", "(.|+)bc", "abc", null, null, "(.|+)*");
+        // The earlier glob (.|+)* is failing for GP7 with this error:
+        // select * from hcfs_glob_java_regex_special_chars;
+        // ERROR:  invalid URI '+)*?PROFILE=hdfs:text' : undefined structure
+        // Since it doesn't like the pipe | in the URI, changing this test a bit to remove the pipe
```
bradfordb-vmware, line 98: Based on my understanding from the last team conversation, the issue is that GP7 is treating the `LOCATION` URL as a pipe-delimited list of URLs. I think this additional context would be useful to include here.

concourse/pipelines/templates/build_pipeline-tpl.yml
```diff
@@ -10,16 +10,18 @@ groups:
...
   - Test PXF-GP[[gp_ver]]-HDP2 on RHEL7
 {% endfor %}
-  - Test PXF-GP6-HDP2 on OEL7
 {% set gp_ver = 6 %}
-  - Test PXF-GP[[gp_ver]]-HDP2 on Ubuntu18
+  - Test PXF-GP[[gp_ver]]-HDP2 on OEL7
   - Build PXF-GP[[gp_ver]] on Ubuntu18
-  - Test PXF-GP[[gp_ver]]-HDP2 on RHEL8
+  - Test PXF-GP[[gp_ver]]-HDP2 on Ubuntu18
+{% for gp_ver in range(6, 8) %}
```
bradfordb-vmware, line 21: This is inside a `set gp_ver = 6` block. Should this be moved out of that block and have its own block?

concourse/pipelines/templates/build_pipeline-tpl.yml
bradfordb-vmware: Newer versions of the `google/cloud-sdk:slim` image are based on Debian Bullseye and this was causing the CI job to fail with the following error:

> ls: cannot access 'pxf_artifacts': Operation not permitted

This is a somewhat known issue with the version of Concourse that we are using and newer container images. We've been pinned to this image in the web UI since the last release. I imagine this is just making that pinning explicit in the pipeline definition.

bradfordb-vmware (approved):

automation/src/test/java/org/greenplum/pxf/automation/features/columnprojection/ColumnProjectionTest.java
```diff
@@ -50,6 +50,13 @@ public void checkColumnProjection() throws Exception {
         pxfExternalTable.setResolver("org.greenplum.pxf.plugins.hdfs.StringPassResolver");
         gpdb.createTableAndVerify(pxfExternalTable);
 
-        runTincTest("pxf.features.columnprojection.checkColumnProjection.runTest");
+        // TODO: revert when 2 queries with GP7 planner start propagating projection info to foreign scans
```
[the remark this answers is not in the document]
denalex: there are no Github issues created for these

automation/src/test/java/org/greenplum/pxf/automation/features/gpupgrade/GpupgradeTest.java
denalex: that test is for upgrade GP5 --> GP6, we will need to take a closer look / modify it when we start addressing GP6 ->GP7 upgrade scenario

concourse/pipelines/templates/build_pipeline-tpl.yml
denalex: `set gp_ver = 6` is not a block, it is just an instruction, so the `for` loop goes after that instruction and overrides `gp_ver` with a different value for each iteration, which removes the effects of that instruction.

I guess we can add `{% set gp_ver = None %}` right before that to "mentally" close the "block", even though it is not needed technically and I actually wanted to clean up things later by removing those "closing" instructions where not necessary.

concourse/pipelines/templates/build_pipeline-tpl.yml
denalex: yes, that makes the "pinning" explicit, since "pinning" is CI UI concept and requires a pipeline to have been run with the pinned image being seen, which is not the case for newly deployed pipelines.

PR #879, opened 2022-10-11
Refactor JsonRecordReader and fix incorrect parsing of JSON objects for multi-line JSON

ashuka24:
This PR is a continuation of ttps://github.com/greenplum-db/pxf/pull/858.

This commit refactors the JsonRecordReader to internally use the LineRecordReader when handling multi-line JSON files. This is done to avoid incorrectly parsing JSON objects, especially those that contain special characters. 

It adds logic for a new table parameter "USE_PARALLEL_READ" which allows the users to toggle between the default HdfsDataFragmenter (true) and the HdfsFileFragmenter (false).


automation/src/test/java/org/greenplum/pxf/automation/features/json/JsonTest.java
```diff
@@ -268,6 +299,14 @@ public void malformedRecordWithRejectLimit() throws Exception {
         exTable.setSegmentRejectLimit(2);
         exTable.setErrorTable("true");
         gpdb.createTableAndVerify(exTable);
+
+        prepareExternalTable("jsontest_malformed_record_with_reject_limit", TWEETS_FIELDS, hdfsPath + FILENAME_BROKEN + SUFFIX_JSON, "custom");
```
denalex, line 303: does this table name need `_filefrag` suffix ?

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
```diff
@@ -55,9 +55,25 @@ public class JsonRecordReader implements RecordReader<LongWritable, Text> {
     private long start;
     private long pos;
     private long end;
+    private String[] hosts;
     private int maxObjectLength;
-    private InputStream is;
     private PartitionedJsonParser parser;
+    private LineRecordReader lineRecordReader;
+    private long linePos;
+    private Text currentLine;
```
denalex, line 63: let's add comments of what new variables track

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParser.java
```diff
@@ -39,185 +37,138 @@
  */
```
denalex, line 37: we need to update the Javadoc for accuracy is the parser might be doing things differently now

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
```diff
@@ -73,25 +89,27 @@ public JsonRecordReader(JobConf conf, FileSplit split) throws IOException {
...
         final CompressionCodec codec = compressionCodecs.getCodec(file);
 
         // openForWrite the file and seek to the start of the split
         FileSystem fs = file.getFileSystem(conf);
         FSDataInputStream fileIn = fs.open(split.getPath());
         if (codec != null) {
-            is = codec.createInputStream(fileIn);
             start = 0;
             end = Long.MAX_VALUE;
+            FileSplit codecSplit = new FileSplit(file, start, Long.MAX_VALUE, hosts);
```
denalex, line 103:
```suggestion
            FileSplit codecSplit = new FileSplit(file, start, Long.MAX_VALUE, split.getLocations());
```
`hosts` member is not used anywhere else, so might as well just inline it here

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
```diff
@@ -100,27 +118,53 @@ public JsonRecordReader(JobConf conf, FileSplit split) throws IOException {
     @Override
     public boolean next(LongWritable key, Text value) throws IOException {
 
-        while (pos < end) {
+        while (!inNextSplit) { // split level. if out of split, then return false
+            int i;
```
denalex, line 122: move this to line 139 closer to where it is used

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
```diff
@@ -100,27 +118,53 @@ public JsonRecordReader(JobConf conf, FileSplit split) throws IOException {
...
+            boolean completedObject = false;
+            // scan to first start brace object.
+            boolean foundBeginObject = scanToFirstBeginObject();
 
-            String json = parser.nextObjectContainingMember(jsonMemberName);
-            pos = start + parser.getBytesRead();
-            if (json == null) {
+            if (!foundBeginObject) {
+                // if we've read all the way to the end and didn't find something, update the pos and move on
+                pos = linePos;
```
denalex, line 131: please explain the logic on how the position is updated

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
```diff
@@ -100,27 +118,53 @@ public JsonRecordReader(JobConf conf, FileSplit split) throws IOException {
...
                 return false;
             }
 
-            long jsonStart = pos - json.getBytes(StandardCharsets.UTF_8).length;
+            // found an object, so we will either return a completed object or be mid-object
```
denalex, line 135: what does it mean to be `mid-object` ?

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
```diff
@@ -100,27 +118,53 @@ public JsonRecordReader(JobConf conf, FileSplit split) throws IOException {
...
+            // found a start brace so begin a new json object
+            parser.startNewJsonObject();
 
-            // if the "begin-object" position is after the end of our split, we should ignore it
-            if (jsonStart >= end) {
-                return false;
+            // read through the file until the object is completed
+            while (!completedObject && (i = readNextChar()) != EOF) { // in the split, create the object
+                if (i == END_OF_SPLIT) {
+                    if (currentLineBuffer == null || currentLineIndex >= currentLineBuffer.length()) {
```
denalex, line 142: can you explain each condition ?

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
```diff
@@ -100,27 +118,53 @@ public JsonRecordReader(JobConf conf, FileSplit split) throws IOException {
...
+                            LOG.debug("JSON object incomplete, moving onto next split to finish");
+                            getNextSplit();
+                            // continue the while loop to complete the object
+                            continue;
+                    }
```
denalex, line 147: there are no else statements ? if so, can 2 `if` statements be combined with && ?

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
```diff
@@ -100,27 +118,53 @@ public JsonRecordReader(JobConf conf, FileSplit split) throws IOException {
...
+                }
+
+                char c = (char) i;
+                completedObject = parser.buildNextObjectContainingMember(c, jsonObject);
             }
```
denalex, line 153: we could exit the `while loop` when `i == EOF`, do we need to check for it here ?

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
```diff
@@ -100,27 +118,53 @@ public JsonRecordReader(JobConf conf, FileSplit split) throws IOException {
...
+            // position of the JsonRecordReader
+            long unreadCharsInBuffer = currentLineBuffer.length() - currentLineIndex;
+            pos = linePos - unreadCharsInBuffer;
+
+            if (completedObject && parser.foundObjectWithIdentifier()) {
+                String json = jsonObject.toString();
+                if (json.length() > maxObjectLength) {
+                    LOG.warn("Skipped JSON object of size " + json.length() + " at pos " + pos);
+                } else {
+                    key.set(pos);
```
denalex, line 164: I don't think key is meaningfully used, but it seems before it would be a file offset where a JSON object started. Is it the same now ? I looks to me (perhaps incorrectly) that this is now the position of where the object ends.

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
```diff
@@ -169,4 +213,96 @@ public float getProgress() throws IOException {
...
+    /**
+     * Reads the next line of the file in to begin parsing the characters
+     * @return
+     * @throws IOException
+     */
+    private boolean getNextLine() throws IOException {
+        currentLine.clear();
+        long currentPos = lineRecordReader.getPos();
+        boolean getNext = lineRecordReader.next(lineRecordReader.createKey(), currentLine);
+        linePos = lineRecordReader.getPos();
```
denalex, line 288: if this is an overall fileOffset , better call it `filePos` ?

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
```diff
@@ -169,4 +213,96 @@ public float getProgress() throws IOException {
...
+
+    /**
+     * Reads the next line of the file in to begin parsing the characters
+     * @return
+     * @throws IOException
+     */
+    private boolean getNextLine() throws IOException {
+        currentLine.clear();
+        long currentPos = lineRecordReader.getPos();
+        boolean getNext = lineRecordReader.next(lineRecordReader.createKey(), currentLine);
```
denalex, line 287: do we need to create key all the time ? can we reuse it or use null ?

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
```diff
@@ -169,4 +213,96 @@ public float getProgress() throws IOException {
...
+            if (delta == 1) {
+                currentLine.append(NEW_LINE, 0, NEW_LINE.length);
+            }
+            if (delta >= 3) {
+                LOG.debug("LineRecordReader removed more characters than expected");
+            }
+            currentLineBuffer = new StringBuffer(currentLine.toString());
+            currentLineIndex = 0;
+        }
+        return getNext;
```
denalex, line 306: it will return false if there was end of split or end of file (end of the very last split), right ? Is there a way to differentiate between these 2 ? is there a need to differentiate (I think you're checking for EOF somewhere).

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
```diff
@@ -169,4 +213,96 @@ public float getProgress() throws IOException {
...
+     *   Closes the current LineRecordReader and opens a new one that starts at the end of the current split
+     *  The end of the new split is set to Long.MAX
+     * @throws IOException
+     */
+    private void getNextSplit() throws IOException {
+        // close the old lineRecordReader
+        lineRecordReader.close();
+        // we need to move into the next split, so create one that starts at the end of the current split
+        // and goes until the end of the file
+        FileSplit nextSplit = new FileSplit(file, end, Long.MAX_VALUE - end, hosts);
```
denalex, line 274: reusing hosts is tricky as a new splits might be on different hosts than the original

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
```diff
@@ -169,4 +213,96 @@ public float getProgress() throws IOException {
...
+     * Read through the characters until we hit starting bracket that indicates the start of a JSON object
+     * @return true when an open bracket '{' is found, false otherwise
+     * @throws IOException
+     */
+    private boolean scanToFirstBeginObject() throws IOException {
+        // assumes each line is a valid json line
+        // seek until we hit the first begin-object
+        boolean inString = false;
+        int i;
+        while ((i = readNextChar()) > EOF) {
```
denalex, line 249: would you ever hit EOF if you're reading chars from the line buffer ?

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/parser/JsonRecordReaderSeekNoSeekTest.java
```diff
@@ -0,0 +1,154 @@
...
+
+
+        for (File jsonDir : dirs) {
+            File childFile = new File(jsonDir, "input.json");
+
+            File[] jsonObjectFiles = jsonDir.listFiles(new FilenameFilter() {
+                public boolean accept(File file, String s) {
+                    return s.contains("expected");
+                }
+            });
```
bradfordb-vmware, line 79: Putting this comment here on the first occurrence, but you can replace all occurrences of `FilenameFilter` with a lambda like this

```suggestion
            File[] jsonObjectFiles = jsonDir.listFiles((dir, name) -> name.contains("expected"));
```

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/parser/JsonRecordReaderSeekNoSeekTest.java
```diff
@@ -0,0 +1,154 @@
...
+        int start = 0;
+        if (seek) {
+            try (InputStream jsonInputStream = new FileInputStream(jsonFile)) {
+                start = seekToStart(jsonInputStream);
+            }
+
+        }
+
+        Path path = new Path(jsonFile.getPath());
+        FileSplit fileSplit = new FileSplit(path, start, 2000, hosts);
```
bradfordb-vmware, line 111: Is there anything special about the value 2000? Could we use a descriptively named static field instead?

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/parser/JsonRecordReaderSeekNoSeekTest.java
```diff
@@ -0,0 +1,154 @@
...
+        JsonRecordReader jsonRecordReader = new JsonRecordReader(jobConf, fileSplit);
+        LongWritable key = new LongWritable();
+        Text data = new Text();
+
+        Arrays.sort(jsonObjectFiles);
+        if (jsonObjectFiles.length == 0) {
+            jsonRecordReader.next(key, data);
+            String result = data.getLength() == 0 ? null : data.toString();
+            assertNull(result, "File " + jsonFile.getAbsolutePath() + " got result '" + result + "'");
```
bradfordb-vmware, line 120: Is there any reason to not just assert that `data.getLength()` is equal to zero instead of nulling it and asserting on that?

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/parser/JsonRecordReaderSeekNoSeekTest.java
```diff
@@ -0,0 +1,154 @@
...
+            LOG.info("File " + jsonFile.getAbsolutePath() + " passed");
+        } else {
+            for (File jsonObjectFile : jsonObjectFiles) {
+                String expected = trimWhitespaces(FileUtils.readFileToString(jsonObjectFile, Charset.defaultCharset()));
```
bradfordb-vmware, line 124: I usually think of `trim`-methods as removing leading and/or trailing whitespace. This method is replacing all sequences of whitespace characters with a single space character. I;d rename this to either `replaceAllWhitespace` or `normalizeWhitespace`.

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/parser/JsonRecordReaderSeekNoSeekTest.java
```diff
@@ -0,0 +1,154 @@
...
+        JsonRecordReader jsonRecordReader = new JsonRecordReader(jobConf, fileSplit);
+        LongWritable key = new LongWritable();
+        Text data = new Text();
+
+        Arrays.sort(jsonObjectFiles);
+        if (jsonObjectFiles.length == 0) {
+            jsonRecordReader.next(key, data);
+            String result = data.getLength() == 0 ? null : data.toString();
+            assertNull(result, "File " + jsonFile.getAbsolutePath() + " got result '" + result + "'");
+            LOG.info("File " + jsonFile.getAbsolutePath() + " passed");
```
bradfordb-vmware, line 121: Do tests need logging statements?

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
ashuka24: Yes. it is possible to hit `EOF` or `END_OF_SPLIT` while reading characters from the buffer

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
ashuka24: LineRecordReader returns false if it read 0 bytes into the line. In the case of the `EOF` character, it will return true and we will read the `EOF` in the buffer. 

In the case of EOF, it is also the end of the split, so we handle all false cases as if it was the end of the split in `readNextChar`

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonProtocolHandler.java
```diff
@@ -35,6 +35,10 @@ public String getResolverClassName(RequestContext context) {
         return context.getResolver();
     }
 
+    public boolean useParallelRead(RequestContext context) {
+        return context.getOption("use_parallel_read", true);
```
hpandeycodeit, line 39: `use_parallel_read` will always be set to true? Or is it only be used when using the compressed  filed ( like with codec) ?

ashuka24: This will be always default to true. My understanding is that this will be a "hidden" parameter in that we will not be adding to the official PXF docs.

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
```diff
@@ -169,4 +219,94 @@ public float getProgress() throws IOException {
...
+        currentLine.clear();
+        long currentPos = lineRecordReader.getPos();
+        boolean didReturnLine = lineRecordReader.next(key, currentLine);
+        linePos = lineRecordReader.getPos();
+        if (didReturnLine) {
+            // lineRecordReader removes the new lines and carriage returns when it does the read
+            // we want to track that delta so we know the proper size of the line that was returned
+            long delta = linePos - currentPos - currentLine.getLength();
+            // append the removed chars back for proper accounting
+            if (delta == 2) {
```
hpandeycodeit, line 300: Why this delta will always be 2 or 1? How we calculated it? And if  it does removes more characters than expected, then what will happen?

ashuka24: We expect the delta to always be 1 or 2 because LineRecordReader should only be removing newline(LF)/carriage return(CR) characters. These characters are each 1 byte each. Typically it is CRLF (windows) or LF (unix) so we need to bring those back here. 

For now, if it is more, than we log a debug message and don't append any chars. (Those chars would not be read).

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/parser/JsonRecordReaderSeekNoSeekTest.java
ashuka24: Since these tests run multiple files, it would be helpful to know which files passed and which ones didn't.

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/parser/JsonRecordReaderSeekNoSeekTest.java
ashuka24: I used 2000 because I knew that no file was that large. I have since changed it to use the length of the file itself.

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
ashuka24: I have added the explanation in a comment. Please let me know if I need to clarify more.

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
```diff
@@ -169,4 +213,96 @@ public float getProgress() throws IOException {
...
+                currentLine.append(CARRIAGERETURN_NEWLINE, 0, CARRIAGERETURN_NEWLINE.length);
+            }
+            if (delta == 1) {
+                currentLine.append(NEW_LINE, 0, NEW_LINE.length);
+            }
+            if (delta >= 3) {
+                LOG.debug("LineRecordReader removed more characters than expected");
+            }
```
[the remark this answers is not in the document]
ashuka24: Nope. I have combined it into if-else statements

automation/src/test/java/org/greenplum/pxf/automation/features/json/JsonTest.java
```diff
@@ -215,6 +227,13 @@ public void missingIdentifier() throws Exception {
     @Test(groups = {"features", "gpdb", "security", "hcfs"})
     public void exceedsMaxSize() throws Exception {
         prepareExternalTable("jsontest_max_size", TWEETS_FIELDS, hdfsPath + FILENAME_PRETTY_PRINT + SUFFIX_JSON, "custom");
+        exTable.setUserParameters(new String[]{
+                "IDENTIFIER=created_at",
+                "MAXLENGTH=566",
+                "USE_PARALLEL_READ=false"});
+        gpdb.createTableAndVerify(exTable);
+
+        prepareExternalTable("jsontest_max_size_filefrag", TWEETS_FIELDS, hdfsPath + FILENAME_PRETTY_PRINT + SUFFIX_JSON, "custom");
```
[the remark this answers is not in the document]
ashuka24: Thanks for catching that. It's been updated.

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParser.java
```diff
@@ -19,205 +19,167 @@
...
+					break;
 				}
-				break;
 			}
+			break;
 		}
-		endOfStream = true;
-		return null;
+
+		return isCompletedObject;
```
denalex, line 167: I posted a question above about the semantic of this returned value.

server/pxf-json/src/test/resources/parser-tests/offset/README.md
```diff
@@ -0,0 +1,32 @@
...
+psql
+\pset tuples_only on
+\o big_data.json
+SELECT '{"cüstömerstätüs":"' || CASE WHEN (((i+47)%29)=0) THEN 'invälid' ELSE 'välid' END || '","name": "äää", "year": "20' || (i%89+10)::text || '", "address": "söme city", "zip": "' || (i%1000 + 90000)::text || '" },'
+FROM generate_series (1,30000000)i;
+```
+
+Then going in and adding the starting and ending array brackets in the data itself so that the final file looks something like this
+```
+[
```
bradfordb-vmware, line 20:
```suggestion
Then go in and add the starting and ending array brackets in the data itself so that the final file looks something like this

```json
[
```

server/pxf-json/src/test/resources/parser-tests/offset/README.md
```diff
@@ -0,0 +1,32 @@
...
+# Generate the big_data.json.bz2 file
+
+This data set contains 30,000,000 rows of data zipped up using the BZip2 codec.
+`big_data.json` and `big_data.json.bz2` were created using the following
+
+```sql
+psql
+\pset tuples_only on
+\o big_data.json
+SELECT '{"cüstömerstätüs":"' || CASE WHEN (((i+47)%29)=0) THEN 'invälid' ELSE 'välid' END || '","name": "äää", "year": "20' || (i%89+10)::text || '", "address": "söme city", "zip": "' || (i%1000 + 90000)::text || '" },'
```
bradfordb-vmware, line 14: For some reason this errors out when I copy-paste this as is. I had to munge this a little

```sql
SELECT '{"cüstömerstätüs":"' || CASE WHEN (((i+47)%29)=0) THEN 'invälid' ELSE 'välid' END || '","name": "äää", "year": "20' || (i%89+10)::text || '", "address": "söme city", "zip": "' || (i%1000 + 90000)::text || '" },' FROM generate_series(1,30000000)i;
```

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParser.java
```diff
@@ -19,205 +19,167 @@
...
-			char c = (char) i;
-			if (c == START_BRACE && prev != BACKSLASH) {
-				lexer.setState(JsonLexer.JsonLexerState.BEGIN_OBJECT);
-				return true;
-			}
-			prev = c;
-		}
-		endOfStream = true;
-		return false;
+		this.objectStack = new ArrayList<Integer>();
```
denalex, line 62:
```suggestion
		this.objectStack = new ArrayList<>();
```

server/pxf-json/src/test/resources/parser-tests/offset/README.md
```diff
@@ -0,0 +1,32 @@
+# Generate JSON files for testing
+
+These instructions will help you generate some of the JSON files required for testing. 
+
+# Generate the big_data.json.bz2 file
+
+This data set contains 30,000,000 rows of data zipped up using the BZip2 codec.
+`big_data.json` and `big_data.json.bz2` were created using the following
```
bradfordb-vmware, line 8: _Commenting here since I can't comment in the file itself._

I know that testing with this file helped to uncover a bug in the JSON profile and that including this going forward is important for catching regressions. However, this file is 92 MiB in size which is over 9x the size of the next largest file in the repo. Do we really want to include a file that large in the git repository?

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParser.java
```diff
@@ -19,205 +19,167 @@
...
-		// You need to wrap the InputStream with an InputStreamReader, so that it can encode the incoming byte stream as
-		// UTF-8 characters
-		this.inputStreamReader = new InputStreamReader(is, StandardCharsets.UTF_8);
+		this.memberName = memberName;
+		this.memberState = MemberSearchState.SEARCHING;
 
-		this.uncountedCharsReadFromStream = new StringBuilder(CHARS_READ_LIMIT);
-	}
+		this.objectCount = 0;
+		this.currentObject = new StringBuilder();
```
denalex, line 58: without initial capacity large enough the default one of 16 will be used and appending character by character will resize the internal buffer on almost each append and will be very costly. We need larger initial capacity and perhaps some intelligent method of resizing the buffer.

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParser.java
```diff
@@ -19,205 +19,167 @@
...
+		this.currentStringLiteral = new StringBuilder();
```
denalex, line 59: same comment on capacity, but the initial might be smaller, but perhaps not 16.

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParser.java
```diff
@@ -19,205 +19,167 @@
...
-		int i;
-		int objectCount = 0;
-		StringBuilder currentObject = new StringBuilder();
-		StringBuilder currentString = new StringBuilder();
-		MemberSearchState memberState = MemberSearchState.SEARCHING;
+		this.objectCount = 0;
+		this.currentObject = new StringBuilder();
+		this.currentStringLiteral = new StringBuilder();
+		this.objectStack = new ArrayList<Integer>();
+		this.isCompletedObject = false;
```
denalex, line 85: same comments on capacities. Also these 4 lines and the ones in constructor seem the same, extract into a separate method ?

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParser.java
```diff
@@ -19,205 +19,167 @@
...
-						objectStack.remove(objectStack.size() - 1);
-					}
-					if (objectStack.size() == 0) {
-						currentObject.setLength(0);
-					}
+	}
+	/**
+	 * @param c character to parse
+	 *            Indicates the member name used to determine the encapsulating object to return.
+	 * @return true  if the JSON object has been completed, false otherwise
```
denalex, line 93: which JSON object ? Having a good Javadoc here will probably answer the questions I was asking before

denalex (changes requested):
A few things to review before merging:
- reuse of hosts in file split constructor, maybe use null ? ideally test on a hadoop cluster with replication factor less than number of nodes
- use charset when converting string -> byte[]
- can we use `Text` object instead of building additional `StringBuffer` for each line ?
- initial (and maybe additional) capacity for StringBuilders, is this the right type to use, performance numbers for changes there

server/pxf-s3/src/main/java/org/greenplum/pxf/plugins/s3/S3ProtocolHandler.java
```diff
@@ -128,8 +129,10 @@ private boolean useS3Select(RequestContext context) {
         }
     }
 
-    public boolean useMultilineJson(RequestContext context) {
-        return isNotEmpty(context.getOption("identifier"));
+    public boolean useFileReadForMultilineJson(RequestContext context) {
+        boolean JsonFormat = StringUtils.equalsIgnoreCase("JSON", context.getFormat());
+        boolean splitByFile = context.getOption("split_by_file", false);
+        return JsonFormat && isNotEmpty(context.getOption("identifier")) && splitByFile;
```
denalex, line 135: same comment here is `split_by_file` option should take effect just by itself without further checks.

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
```diff
@@ -100,27 +106,60 @@ public JsonRecordReader(JobConf conf, FileSplit split) throws IOException {
...
+                // however, if we are at the end of the split, then we need to get the next split before we can read the line
+                if (i == END_OF_SPLIT && (currentLineBuffer == null || currentLineIndex >= currentLineBuffer.length())) {
+                    LOG.debug("JSON object incomplete, moving onto next split to finish");
+                    getNextSplit();
+                    // continue the while loop to complete the object
+                    continue;
+                }
+
+                char c = (char) i;
+                isObjectComplete = parser.parse(c);
```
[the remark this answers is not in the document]
ashuka24: Correct. Completed means that the parser found a matching `}` to the first `{`. It will return true in all such scenarios, regardless of whether or not we found the matching identifier.

server/pxf-s3/src/main/java/org/greenplum/pxf/plugins/s3/S3ProtocolHandler.java
ashuka24: As `SPLIT_BY_FILE` is currently only a parameter that takes affect on JSON files, I will be leaving the JsonFormat check as this is only related to JSON for now. I will be removing the check for the identifier though.

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/parser/JsonRecordReaderSeekNoSeekTest.java
ashuka24: I took a look into this. We have logging set up, but when you debug it seems like we do not actually log to any file. 

If a test were to fail, we would see something like: 
`~/workspace/pxf/server/pxf-json/src/test/resources/parser-tests/seek/multi/input.json/expected.1.json` and from this we could determine the file that is causing problems. 

As such, I will be removing the logging component of this test suite.

server/pxf-json/src/test/resources/parser-tests/offset/README.md
ashuka24: I'm not sure I see the difference. Is the only difference that it is not on 2 lines?

server/pxf-json/src/test/resources/parser-tests/offset/README.md
ashuka24: Since we have the instructions for creating the file, we could not include it in the github, but then add a script when something that would first generate this file before running the unit test?

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
```diff
@@ -100,27 +106,60 @@ public JsonRecordReader(JobConf conf, FileSplit split) throws IOException {
...
+                long unreadCharsInBuffer = totalChars - readChars;
+                pos = filePos - unreadCharsInBuffer;
             } else {
-                key.set(jsonStart);
-                value.set(json);
-                return true;
+                pos = filePos;
+            }
+
+            if (isObjectComplete && parser.foundObjectWithIdentifier()) {
```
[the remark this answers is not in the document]
ashuka24: The parser would report completed at line 5 and only return the inner nested object (lines 4-5)

However, if the matching key was in the outer object, the parser would return completed at line 7 and give the full object (lines 1 - 7)

This matches behavior that is present on pxf main. 

The 2 cases for a "completed object" are 
1) when there is no matching value, then we mark completed at line 7. 
2) when there is a matching value, the we mark completed when the object containing the matching value is completed, even if that object is a nested object.

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
```diff
@@ -169,4 +208,95 @@ public float getProgress() throws IOException {
...
+            long delta = filePos - currentPos - currentLine.getLength();
+            // append the removed chars back for proper accounting
+            if (delta == 2) {
+                currentLine.append(CARRIAGERETURN_NEWLINE, 0, CARRIAGERETURN_NEWLINE.length);
+            } else if (delta == 1) {
+                currentLine.append(NEW_LINE, 0, NEW_LINE.length);
+            } else {
+                LOG.warn("LineRecordReader removed more characters than expected while parsing a JSON file.");
+            }
+            currentLineBuffer = new StringBuffer(currentLine.toString());
```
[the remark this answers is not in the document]
ashuka24: The `Text` object API does some internal logic handling when calling `charAt` that is not ideal for our situation. It also underlying does `ByteBuffer` wrapping and splicing for each character which creates new `HeapByteBuffers`. In this case, it is not less expensive than just creating one buffer.

server/pxf-json/src/test/resources/parser-tests/offset/README.md
ashuka24: As discussed, for the unit test case, we will be reducing the size of the file.

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParser.java
ashuka24: I've removed the instances from the constructor, as we do not need them there, instead we instantiate when we start a new json object.

bradfordb-vmware (approved):

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
```diff
@@ -169,4 +200,108 @@ public float getProgress() throws IOException {
...
+    /**
+     * Reads the next character in the buffer. It will pull the next line as necessary
+     * @return the int value of a character
+     * @throws IOException
+     */
+    private int readNextChar() throws IOException {
+        // if the currentLineBuffer is null, nothing has been read yet, so we need to read the next line
+        // if the currentLineIndex is greater than the length,  we are at the end of the buffer, read the next line
+        if (currentLineBuffer == null || currentLineIndex >= currentLineBuffer.length()) {
+            if (!getNextLine()) {
```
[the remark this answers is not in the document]
ashuka24: I moved things into a boolean and condensed the if statement to improve readability.

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParser.java
```diff
@@ -19,205 +19,204 @@
...
-	private final StringBuilder uncountedCharsReadFromStream;
-
-	public PartitionedJsonParser(InputStream is) {
+	private String memberName;
+
+	private MemberSearchState memberState;
+	private StringBuilder currentObject;
+	private StringBuilder currentStringLiteral;
+	private int objectCount;
+	private ArrayDeque<Integer> objectStack;
```
denalex, line 44: would be great to explain what that integer value in stack element represents

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParser.java
```diff
@@ -19,205 +19,204 @@
...
+	private boolean isCompletedObject;
+	// track the average size of JSON objects produced to avoid extraneous buffer copies
+	private int averageObjectSize;
+	private int numObjectsRead;
+	public PartitionedJsonParser(String memberName) {
```
denalex, line 49: please insert a new line before the constructor and javadoc it

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParser.java
denalex: no need to refer to variables with `this.` prefix

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParser.java
```diff
@@ -19,205 +19,204 @@
...
-	private int readNextChar() throws IOException {
-		int i = inputStreamReader.read();
-
-		if (i != EOF) {
-			uncountedCharsReadFromStream.append((char) i);
-			if (uncountedCharsReadFromStream.length() == CHARS_READ_LIMIT) {
-				bytesRead += countBytesInReadChars();
-			}
+	private void updateAverageJsonObjectSize() {
+		int currentObjectSize = this.currentObject.toString().getBytes(StandardCharsets.UTF_8).length;
```
denalex, line 213:
```suggestion
		int currentObjectSize = currentObject.toString().getBytes(StandardCharsets.UTF_8).length;
```

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/parser/PartitionedJsonParser.java
```diff
@@ -19,205 +19,204 @@
...
+		if (numObjectsRead == 1) {
+			averageObjectSize = currentObjectSize;
+		} else {
+			// recalculate the average: (averageObjectSize*(numObjectsRead - 1) + currentObjectSize) / numObjectsRead
+			//                        = averageObjectSize + (currentObjectSize - averageObjectSize) / numObjectsRead
+			averageObjectSize += (currentObjectSize - averageObjectSize)/numObjectsRead;
```
denalex, line 219: might be good to log the results of the computations

server/pxf-s3/src/main/java/org/greenplum/pxf/plugins/s3/S3ProtocolHandler.java
```diff
@@ -128,8 +129,10 @@ private boolean useS3Select(RequestContext context) {
         }
     }
 
-    public boolean useMultilineJson(RequestContext context) {
-        return isNotEmpty(context.getOption("identifier"));
+    public boolean useFileReadForJson(RequestContext context) {
```
denalex, line 132: does it need to be public ? please Javadoc it as well

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
```diff
@@ -169,4 +200,105 @@ public float getProgress() throws IOException {
...
+            // lineRecordReader removes the new lines and carriage returns when it does the read
+            // we want to track that delta, so we know the proper size of the line that was returned
+            long delta = filePos - currentPos - currentLine.getLength();
+            // append the removed chars back for proper accounting
+            if (delta == 2) {
+                currentLine.append(CARRIAGERETURN_NEWLINE, 0, CARRIAGERETURN_NEWLINE.length);
+            } else if (delta == 1) {
+                currentLine.append(NEW_LINE, 0, NEW_LINE.length);
+            } else if (delta == 0) {
+                // likely end of file, do not do anything.
```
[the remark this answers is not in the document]
ashuka24: If the file had 2 bytes, and we read it and then got EOF back, the final filePos after EOF would be 2.

denalex (approved):

yimingli-vmware (approved):

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
```diff
@@ -100,27 +104,54 @@ public JsonRecordReader(JobConf conf, FileSplit split) throws IOException {
...
+
+            if (isObjectComplete && parser.foundObjectWithIdentifier()) {
+                String json = parser.getCompletedObject();
+                // check the char length of the json against the MAXLENGTH
+                long jsonLength = json.length();
+                if (jsonLength > maxObjectLength) {
+                    LOG.warn("Skipped JSON object of size " + json.length() + " at pos " + pos);
+                } else {
+                    // the key is set to beginning of the json object
+                    key.set(pos - jsonLength);
```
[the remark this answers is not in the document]
ashuka24: I will update the pos in line 148 and the position in 151 to be jsonStart `pos - json.getBytes(UTF8).length()`

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
```diff
@@ -169,4 +200,105 @@ public float getProgress() throws IOException {
...
+            long delta = filePos - currentPos - currentLine.getLength();
+            // append the removed chars back for proper accounting
+            if (delta == 2) {
+                currentLine.append(CARRIAGERETURN_NEWLINE, 0, CARRIAGERETURN_NEWLINE.length);
+            } else if (delta == 1) {
+                currentLine.append(NEW_LINE, 0, NEW_LINE.length);
+            } else if (delta == 0) {
+                // likely end of file, do not do anything.
+            } else {
+                throw new IllegalStateException("LineRecordReader removed more characters than expected while parsing a JSON file.");
```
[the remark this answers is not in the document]
ashuka24: Based off our understanding of LineRecordReader and tests with various iterations of carriage returns/new lines, we do NOT expect the delta to be greater than 2. Delta accounting is used purely for the length calculating when it comes to maxObjectLength inside JsonRecordReader. As it is not and should not stop processing, this will be updated to a `LOG.warn`

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
```diff
@@ -73,25 +89,15 @@ public JsonRecordReader(JobConf conf, FileSplit split) throws IOException {
...
-            if (start != 0) {
-                fileIn.seek(start);
-            }
-            is = fileIn;
-        }
-        parser = new PartitionedJsonParser(is);
+        file = split.getPath();
+        hosts = split.getLocations();
+        lineRecordReader =  new LineRecordReader(conf, split);
+        this.conf = conf;
```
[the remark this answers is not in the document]
ashuka24: to prevent shadowing, since the private member is the same as the passed in parameter, we need to explicitly say `this.conf`

PR #881, opened 2022-10-13
Improve PXF CLI reliability

bradfordb-vmware:
1. Check the status of the PXF service by checking the status of the PXF JVM process and by checking the HTTP health endpoint; if the PXF webapp fails to load for some reason, the JVM exits.

2. Disable actuator shutdown endpoint
    * The recommended way for shutting down a Spring Boot application is to send the JVM SIGTERM. This commit disables the shutdown endpoint since we can always assume that PXF is being shutdown locally and not remotely.

3. Add logging function to PXF CLI
    * PXF CLI writes log messages to `$PXF_LOGDIR/admin/pxf-cli-YYYY-MM.log`
    * Log lines are prefixed with a timestamp and the PID of the shell script to help identify all log messages from a single invocation of the PXF CLI
  
4. Remove PXF CLI's dependency on `curl` command
    * the `curl` command was previously required because of its use in shutting down and checking the status of PXF

### DIFF
--- a/server/pxf-service/src/main/resources/application.properties
+++ b/server/pxf-service/src/main/resources/application.properties
@@ -2,10 +2,10 @@
 # 1. health: returns the status of the application {"status":"UP"}
 # 2. info: returns information about the build {"build":{"version":"X.X.X","artifact":"pxf-service","name":"pxf-service","group":"org.greenplum.pxf","time":"timestamp"}}
 # 3. shutdown: allows shutting down the application
-# 4. metrics: shows ‘metrics’ information for the application
+# 4. metrics: shows 'metrics' information for the application
 # 5. prometheus: exposes metrics in a format that can be scraped by a Prometheus server
-management.endpoints.web.exposure.include=health,info,shutdown,metrics,prometheus
-management.endpoint.shutdown.enabled=true
+management.endpoints.web.exposure.include=health,info,metrics,prometheus
+management.endpoint.shutdown.enabled=false
 management.endpoint.health.probes.enabled=true
 
 # common tags applied to all metrics
--- a/server/pxf-service/src/scripts/pxf
+++ b/server/pxf-service/src/scripts/pxf
@@ -53,6 +53,10 @@ export JAVA_HOME=${JAVA_HOME:=/usr/java/default}
 # Path to Log directory
 export PXF_LOGDIR=${PXF_LOGDIR:=${PXF_BASE}/logs}
 
+export PXF_CLI_LOGDIR="${PXF_LOGDIR}/admin"
+PXF_CLI_LOGFILE="${PXF_CLI_LOGDIR}/pxf-cli-$(date +%Y-%m).log"
+export PXF_CLI_LOGFILE
+
 # Path to Run directory
 export PXF_RUNDIR=${PXF_RUNDIR:=${PXF_BASE}/run}
 
@@ -118,13 +122,19 @@ function failWithHint() {
     exit 1
 }
 
+function log() {
+    printf "%s\t[%d]\t%s\n" "$(date '+%Y-%m-%d %H:%M:%S.%3N %Z')" "${BASHPID}" "$*" >>"${PXF_CLI_LOGFILE}"
+}
+
 # determines whether the application is running
 function isRunning() {
+    log "Checking if PXF process with PID=$1 exists"
     ps -o pid,command -p "$1" | grep 'pxf-app' &> /dev/null
 }
 
 # creates the runtime directory structure inside $PXF_BASE
 function createPxfBaseDirectories() {
+    log "Creating PXF runtime directory structure in ${PXF_BASE}"
     install -m 744 -d "${PXF_BASE}/conf"
     install -m 744 -d "${PXF_BASE}/lib"
     install -m 744 -d "${PXF_BASE}/lib/native"
@@ -137,6 +147,7 @@ function createPxfBaseDirectories() {
 
 # copies configuration files inside $PXF_BASE
 function copyConfigurationFilesToPxfBase() {
+    log "Copying PXF configuration files from ${PXF_HOME} to ${PXF_BASE}"
     cp -Lv "${PXF_HOME}/conf/pxf-application.properties" "${PXF_BASE}/conf"
     cp -Lv "${PXF_HOME}/conf/pxf-env.sh"                 "${PXF_BASE}/conf"
     cp -Lv "${PXF_HOME}/conf/pxf-log4j2.xml"             "${PXF_BASE}/conf"
@@ -322,6 +333,7 @@ function doStart() {
 }
 
 function do_start() {
+    log "Starting PXF service..."
     if [[ -f "$pid_file" ]]; then
         pid=$(cat "$pid_file")
         isRunning "$pid" && { echoYellow "PXF is already running [pid=$pid]"; return 0; }
@@ -354,6 +366,8 @@ function do_start() {
     disown $pid
     echo "$pid" > "$pid_file"
 
+    log "Started PXF service on port ${PXF_PORT} - PID=${pid}"
+
     [[ -z $pid ]] && fail "Failed to start PXF"
     echoGreen "PXF started [pid=$pid]. Listening on port $PXF_PORT";
 }
@@ -369,11 +383,21 @@ function doStop() {
     checkJavaHome
     warnUserEnvScript
 
+    log "Stopping PXF service..."
     working_dir=$(dirname "$jarfile")
     pushd "$working_dir" > /dev/null
-    [[ -f $pid_file ]] || { echoYellow "Not running (pidfile not found). Is PXF running? Stop aborted."; return 0; }
+    if [[ ! -f $pid_file ]]; then
+        log "No pidfile=${pid_file} for PXF found"
+        echoYellow "Not running (pidfile not found). Is PXF running? Stop aborted."
+        return 0
+    fi
     pid=$(cat "$pid_file")
-    isRunning "$pid" || { echoYellow "Not running (process ${pid}). Removing stale pid file '$pid_file'."; rm -f "$pid_file"; return 0; }
+    if ! isRunning "$pid"; then
+        log "No running PXF process with PID=${pid} found; removing stale pidfile"
+        echoYellow "Not running (process ${pid}). Removing stale pid file '$pid_file'."
+        rm -f "$pid_file"
+        return 0
+    fi
 
     echoYellow "Stopping PXF [pid=$pid]..."
     do_stop "$pid" "$pid_file" || do_stop "$pid" "$pid_file" "-f"
@@ -384,12 +408,18 @@ function do_stop() {
     local killcmd=(kill)
     [[ $3 == -f ]] && killcmd+=(-9)
     killcmd+=("$1")
+    log "Attempting to stop PXF with '${killcmd[*]}'"
     "${killcmd[@]}" &> /dev/null || { echoRed "Unable to terminate process $1"; return 1; }
     for i in $(seq 1 $STOP_WAIT_TIME); do
         isRunning "$1" || { echoGreen "PXF stopped [$1]"; rm -f "$2"; return 0; }
-        [[ $i -eq STOP_WAIT_TIME/2 ]] && "${killcmd[@]}" &> /dev/null
+        log "PXF process with PID=$1 is still running..."
+        if [[ $i -eq STOP_WAIT_TIME/2 ]]; then
+            log "Attempting to stop PXF again with '${killcmd[*]}'"
+            "${killcmd[@]}" &> /dev/null
+        fi
         sleep 1
     done
+    log "Unable to stop process PID=$1 with '${killcmd[*]}'"
     echoRed "Unable to terminate process $1"
     return 1
 }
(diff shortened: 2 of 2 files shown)


server/pxf-service/src/main/resources/application.properties
```diff
@@ -2,10 +2,10 @@
 # 1. health: returns the status of the application {"status":"UP"}
 # 2. info: returns information about the build {"build":{"version":"X.X.X","artifact":"pxf-service","name":"pxf-service","group":"org.greenplum.pxf","time":"timestamp"}}
 # 3. shutdown: allows shutting down the application
-# 4. metrics: shows ‘metrics’ information for the application
+# 4. metrics: shows 'metrics' information for the application
 # 5. prometheus: exposes metrics in a format that can be scraped by a Prometheus server
-management.endpoints.web.exposure.include=health,info,shutdown,metrics,prometheus
-management.endpoint.shutdown.enabled=true
+management.endpoints.web.exposure.include=health,info,metrics,prometheus
+management.endpoint.shutdown.enabled=false
```
ashuka24, line 8: can we leave a comment why this is now false?

server/pxf-service/src/scripts/pxf
```diff
@@ -53,6 +53,10 @@ export JAVA_HOME=${JAVA_HOME:=/usr/java/default}
 # Path to Log directory
 export PXF_LOGDIR=${PXF_LOGDIR:=${PXF_BASE}/logs}
 
+export PXF_CLI_LOGDIR="${PXF_LOGDIR}/admin"
+PXF_CLI_LOGFILE="${PXF_CLI_LOGDIR}/pxf-cli-$(date +%Y-%m).log"
```
ashuka24, line 57: for cli logs, we do not do any log rotation, instead a new log file is created every month? 
ex: if 4 commands were run on different days of oct 2022 they would all be `pxf-cli-2022-10.log` and if a 5th command were run on Nov 1, then that 5th command and its log statements would go into `pxf-cli-2022-11.log`?

bradfordb-vmware: That is correct. The GPDB management utilities uses a similar approach but with _daily_ instead of _monthly_; however, the GPDB management utilities are more verbose than the PXF CLI. I thought that using monthly for PXF would make sense so that we don't have a larger number of log files with only a few lines each.

server/pxf-service/src/scripts/pxf
```diff
@@ -417,28 +383,23 @@ function doStop() {
...
-                sleep 1
-            else
-                echoGreen "PXF stopped [pid=$pid]"
-                rm -f $pid_file
-                return 0
-            fi
-        done
+    if ! isRunning "$pid"; then
+        log "No running PXF process with PID=${pid} found; removing stale pidfile"
+        echoYellow "Not running (process ${pid}). Removing stale pid file '$pid_file'."
```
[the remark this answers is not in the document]
bradfordb-vmware: I agree they are similar but I think they are serving different purposes -- the first line is a log message which contains information about the outcome of an operation, while the second line is output to the user and is an answer to the question "What is the status of PXF?".

server/pxf-service/src/main/resources/application.properties
bradfordb-vmware: I put some context in the commit message.

server/pxf-service/src/scripts/pxf
```diff
@@ -53,6 +53,10 @@ export JAVA_HOME=${JAVA_HOME:=/usr/java/default}
 # Path to Log directory
 export PXF_LOGDIR=${PXF_LOGDIR:=${PXF_BASE}/logs}
 
+export PXF_CLI_LOGDIR="${PXF_LOGDIR}/admin"
```
denalex, line 56: do we need pre-create this directory during "prepare" in `createPxfBaseDirectories` function ?

server/pxf-service/src/scripts/pxf
denalex, line 57: `pxf-cli-` feels incomplete and when we would ask customers to provide us "pxf cli logs" it sounds a bit unusual. Would it be better to call it `pxf-command-` instead to reflect that this is a log of command outputs ? Then we can refer to these logs as "command logs"

server/pxf-service/src/scripts/pxf
```diff
@@ -53,6 +53,10 @@ export JAVA_HOME=${JAVA_HOME:=/usr/java/default}
...
+PXF_CLI_LOGFILE="${PXF_CLI_LOGDIR}/pxf-cli-$(date +%Y-%m).log"
+export PXF_CLI_LOGFILE
```
denalex, line 58: this breaks a pattern of exporting and assigning a variable on a single line. I know shell check advises us to do this separately, not sure what's the best there. I don't think date command will ever fail, so maybe doing it on a single line here is ok ?

server/pxf-service/src/scripts/pxf
```diff
@@ -118,52 +122,19 @@ function failWithHint() {
...
-#
-function checkWebapp() {
-    waitForSpringBoot "$1" "$2" || return 1
-
-    curlResponse=$($curl -s "http://localhost:${PXF_PORT}/actuator/health")
-
-    [[ $curlResponse == {\"status\":\"UP\"* ]] || fail "PXF is inaccessible. Check logs ($PXF_LOGDIR) for more information"
-    return 0
+function log() {
+    printf "%s\t[%d]\t%s\n" "$(date '+%Y-%m-%d %H:%M:%S.%3N %Z')" "${BASHPID}" "$*" >>"${PXF_CLI_LOGFILE}"
```
denalex, line 126: can we document in a comment here what the target line will look like to help future developers to parse these tokens ?

server/pxf-service/src/scripts/pxf
```diff
@@ -118,52 +122,19 @@ function failWithHint() {
...
 }
 
 # determines whether the application is running
 function isRunning() {
+    log "Checking if PXF process with PID=$1 exists"
     ps -o pid,command -p "$1" | grep 'pxf-app' &> /dev/null
```
denalex, line 132: should we log the result of this check along with the exact command used (including pid) ?

server/pxf-service/src/scripts/pxf
```diff
@@ -176,19 +147,13 @@ function createPxfBaseDirectories() {
```
denalex, line 147: should we log the directories created (full paths) ?

server/pxf-service/src/scripts/pxf
```diff
@@ -176,19 +147,13 @@ function createPxfBaseDirectories() {
 
 # copies configuration files inside $PXF_BASE
 function copyConfigurationFilesToPxfBase() {
+    log "Copying PXF configuration files from ${PXF_HOME} to ${PXF_BASE}"
     cp -Lv "${PXF_HOME}/conf/pxf-application.properties" "${PXF_BASE}/conf"
     cp -Lv "${PXF_HOME}/conf/pxf-env.sh"                 "${PXF_BASE}/conf"
     cp -Lv "${PXF_HOME}/conf/pxf-log4j2.xml"             "${PXF_BASE}/conf"
     cp -Lv "${PXF_HOME}/conf/pxf-profiles.xml"           "${PXF_BASE}/conf"
```
denalex, line 154: should we log the files created (full paths) ?

server/pxf-service/src/scripts/pxf
```diff
@@ -401,8 +366,9 @@ function do_start() {
     disown $pid
     echo "$pid" > "$pid_file"
 
+    log "Started PXF service on port ${PXF_PORT} - PID=${pid}"
+
     [[ -z $pid ]] && fail "Failed to start PXF"
-    checkWebapp 300 10 || return 1
     echoGreen "PXF started [pid=$pid]. Listening on port $PXF_PORT";
```
denalex, line 372: should we still wait until Boot app is fully initialized before returning success ? It might not be ready to serve request for a minute or so.

server/pxf-service/src/scripts/pxf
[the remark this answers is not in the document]
denalex: we need to discuss this. ideally we would log everything we show to user exactly as it was presented and maybe more info for extra troubleshooting.

server/pxf-service/src/scripts/pxf
```diff
@@ -447,21 +408,39 @@ function do_stop() {
     local killcmd=(kill)
     [[ $3 == -f ]] && killcmd+=(-9)
     killcmd+=("$1")
+    log "Attempting to stop PXF with '${killcmd[*]}'"
     "${killcmd[@]}" &> /dev/null || { echoRed "Unable to terminate process $1"; return 1; }
     for i in $(seq 1 $STOP_WAIT_TIME); do
         isRunning "$1" || { echoGreen "PXF stopped [$1]"; rm -f "$2"; return 0; }
```
denalex, line 414: need to log when pxf was determined to have stopped

server/pxf-service/src/scripts/pxf
```diff
@@ -447,21 +408,39 @@ function do_stop() {
...
     echoRed "Unable to terminate process $1"
     return 1
 }
 
 function doStatus() {
     checkJavaHome
     warnUserEnvScript
-    checkWebapp 1 0 || return 1
+
+    echoYellow 'Checking if PXF is up and running...'
```
denalex, line 431: was this supposed to be a log statement rater than a message to the user ? "status" command is not smth a user should be prepared/warned to wait for

server/pxf-service/src/scripts/pxf
```diff
@@ -447,21 +408,39 @@ function do_stop() {
...
+
+    working_dir="$(dirname "${jarfile}")"
+    pushd "${working_dir}" >/dev/null
+    if [[ ! -s "${pid_file}" ]]; then
+        log "PXF pidfile '${pid_file}' does not exist"
+        echoRed 'ERROR: PXF is down - the application is not running'
+        return 3
```
denalex, line 438: do we need to `popd` before returning ? here and at the end of the function

server/pxf-service/src/scripts/pxf
```diff
@@ -447,21 +408,39 @@ function do_stop() {
...
 
 function doStatus() {
     checkJavaHome
     warnUserEnvScript
-    checkWebapp 1 0 || return 1
+
+    echoYellow 'Checking if PXF is up and running...'
+
+    working_dir="$(dirname "${jarfile}")"
+    pushd "${working_dir}" >/dev/null
```
denalex, line 434: why do we need to change directories ?

server/pxf-service/src/scripts/pxf
```diff
@@ -447,21 +408,39 @@ function do_stop() {
...
+    if [[ ! -s "${pid_file}" ]]; then
+        log "PXF pidfile '${pid_file}' does not exist"
+        echoRed 'ERROR: PXF is down - the application is not running'
```
denalex, line 437: absence of the pid file is not necessarily an indication that a process is not running. the pid file could have been deleted. I believed we talked about having some identifier and doing `ps | grep pxf | grep <identifier>` to see if the process is actually running to allow for multiple pxf processes on the same host?

server/pxf-service/src/scripts/pxf
```diff
@@ -447,21 +408,39 @@ function do_stop() {
...
+        return 3
+    fi
+    pid="$(cat "${pid_file}")"
+    isRunning "${pid}" || { echoRed "ERROR: PXF is down - process ${pid} not found"; return 1; }
```
denalex, line 441: if we determined that the process is not running but the pid file exists, do we want to remove the pid file here ?

server/pxf-service/src/scripts/pxf
```diff
@@ -492,8 +473,9 @@ pxf_script_command=$1
 
 silent=false
 
+[[ -d "${PXF_CLI_LOGDIR}" ]] || mkdir "${PXF_CLI_LOGDIR}"
```
denalex, line 476: this feels too low-level for this place in the code, should we make a function for it ? also we can use `mkdir -p` to be safe. I think we should include this in prep command but still leave it at this place in the flow as well to account for cases when users get new binaries (script) after they have already prep-ed their `PXF_BASE`.

denalex (commented):
generally, I think for every operation we should be logging:
- "Starting ..."
- full commands being run with variables substituted
- output of those commands
- anything printed out to the user
- anything calculated or derived as part of "if" statements
- outcome of the operation
- "Finished ..."

server/pxf-service/src/scripts/pxf
bradfordb-vmware: No. This message was previously printed to the user by `waitForSpringBoot` which was called by `checkWebapp`. Since those methods are being removed, that message was moved to here to maintain the current output.

server/pxf-service/src/scripts/pxf
bradfordb-vmware: Good point, I can add that in.

server/pxf-service/src/scripts/pxf
bradfordb-vmware: I'm not sure that we need to log either the result or the exact command. We log the value of the PID in the line above and presumably the caller of this method will make a decision based on what this returns and includes logging messages for either case.

server/pxf-service/src/scripts/pxf
bradfordb-vmware: My initial thought is no. This PR updates the CLI to only consider the state of the JVM process when checking if the status of PXF, so it seems this script should only be concerned with the process state. Additionally, (while not included in this PR) systemd will only consider the process state when starting and reporting the status of the PXF service.

ashuka24 (approved):

ashuka24:
if this PR goes in, would https://github.com/greenplum-db/pxf/pull/831 be closed or would its contents need to be added to this PR?

bradfordb-vmware:
Based on discussion with the rest of the team, I think the scope and direction of this work is changing. I'm going to close this PR for now while we work on this topic some more.

PR #886, opened 2022-10-24
Update jackson-databind to 2.13.4.1

ashuka24:
This commit updates the `jackson-databind` version to 2.13.4.1 in the automation framework as a precautionary measure to mitigate CVE-2022-42003. 

PXF pulls in `jackson-databind` through Springboot v2.5.12. The CVE in question is related to `UNWRAP_SINGLE_VALUE_ARRAYS` which is disabled by default in `jackson-databind` and stays disabled in Springboot v2.5.12. We remove explicit declarations of these dependencies in the subproject `build.gradle` files.

### DIFF
--- a/automation/pom.xml
+++ b/automation/pom.xml
@@ -247,7 +247,7 @@
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
-            <version>2.12.6.1</version>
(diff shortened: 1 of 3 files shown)


denalex (approved):
Can we double check whether removal of the explicit declaration (with transitive set to false) will result in transitive dependencies (if any) of these modules now being imported and packaged ?

PR #890, opened 2022-11-01
Adding Install and installcheck for the fdw test target

hpandeycodeit:
Currently, for GP7, if we run the `make test` it runs only `fdw -C make installcheck` and not `make install` due to which `fdw` doesn't gets installed properly and fails the fdw tests. This PR adds `install and installcheck` for the `test` target.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,13 +22,15 @@ default: all
 
 .PHONY: all external-table fdw cli server install install-server stage tar rpm rpm-tar deb deb-tar clean test it help
 
-all: external-table cli server
+all: external-table cli server fdw
 
 external-table:
 	make -C external-table
 
 fdw:
+ifneq ($(FDW_SUPPORT),)
 	make -C fdw
+endif
 
 cli:
 	make -C cli/go/src/pxf-cli
@@ -44,7 +46,7 @@ clean:
 
 test:
 ifneq ($(FDW_SUPPORT),)
-	make -C fdw installcheck
+	make -C fdw test
 endif
 	make -C cli/go/src/pxf-cli test
 	make -C server test
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ default: all
 
 .PHONY: all external-table fdw cli server install install-server stage tar rpm rpm-tar deb deb-tar clean test it help
 
-all: external-table cli server fdw
+all: fdw external-table cli server
 
 external-table:
 	make -C external-table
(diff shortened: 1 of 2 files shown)


fdw/Makefile
```diff
@@ -35,3 +35,5 @@ stage: pxf_fdw.so
 .PHONY: clean-all
 clean-all: clean
 	rm -rf build
+
+test: install installcheck
```
ashuka24, line 39: do we want to add back the new line at the end of the file that was there previously?

Makefile
```diff
@@ -22,13 +22,15 @@ default: all
...
 .PHONY: all external-table fdw cli server install install-server stage tar rpm rpm-tar deb deb-tar clean test it help
 
-all: external-table cli server
+all: external-table cli server fdw
 
 external-table:
 	make -C external-table
 
 fdw:
+ifneq ($(FDW_SUPPORT),)
```
ashuka24, line 31: is there supposed to be a comma here?

ashuka24 (approved):

Makefile
```diff
@@ -22,13 +22,15 @@ default: all
...
 	make -C fdw
+endif
```
bradfordb-vmware, line 33: This is conditional on `FDW_SUPPORT` being set to some non-empty value; if it is empty, we silently skip this step. I think it would be helpful if we included a message that we are skipping `fdw` and why

```suggestion
ifneq ($(FDW_SUPPORT),)
	make -C fdw
else
	@echo "The variable FDW_SUPPORT was empty; skipping building the fdw/ directory"
endif
```

Makefile
```diff
@@ -44,7 +46,7 @@ clean:
 
 test:
 ifneq ($(FDW_SUPPORT),)
-	make -C fdw installcheck
+	make -C fdw test
```
bradfordb-vmware, line 49: Same comment as before, especially since this would skip running tests.

Makefile
```diff
@@ -22,13 +22,15 @@ default: all
 
 .PHONY: all external-table fdw cli server install install-server stage tar rpm rpm-tar deb deb-tar clean test it help
 
-all: external-table cli server
+all: external-table cli server fdw
```
denalex, line 25:
```suggestion
all: fdw external-table cli server
```
let's have C-side projects first, then the rest

fdw/Makefile
denalex, line 39: do you need to add `test` to `.PHONY` list ?

Makefile
hpandeycodeit: yeah as per the syntax. It's like checking `if FDW_SUPPORT != ""`

bradfordb-vmware (approved):

denalex (approved):

yimingli-vmware (approved):

PR #891, opened 2022-11-01
Replace email with Slack in pxf-build

bradfordb-vmware:


ashuka24 (approved):

concourse/Makefile
```diff
@@ -125,7 +122,6 @@ set-dev-build-pipeline:
 		--config "$${PIPELINE_FILE}" \
 		--load-vars-from=$(HOME)/workspace/pxf/concourse/settings/pxf-multinode-params.yml \
 		--var=pxf-git-branch=${BRANCH} \
-		--var=slack-integration-webhook-url=$${SLACK_INTEGRATION_WEBHOOK_URL} \
```
denalex, line 128: why was this removed ?

bradfordb-vmware: I removed this because the variable (`slack-integration-webhook-url`) isn't actually used. Looking at the git history, this variable wasn't actually ever used.

yimingli-vmware (approved):

denalex (approved):

PR #895, opened 2022-11-08
Parquet read array

yimingli-vmware:
This PR is for adding read support for Parquet LIST type. 

- Parquet LIST schema can be find [here](https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#lists)

- For now our Parquet doesn't support the following Greenplum type: `UUID`, `UUIDARRAY`, `TIMESTAMP_WITH_TIMEZONE`, and `TIMESTAMP_WITH_TIME_ZONE_ARRAY`. Since Parquet TIMESTAMP doesn't store timezone information, we treat each timestamp as timestamp with local timezone, then store it as UTC timestamp. The local timezone information is from pxf jvm timezone.

- Provide a README in automation test for how to generate test dataset. Since for now HIVE doesn't support Parquet `TIMESTAMP LIST`, the test dataset for `TIMESTAMP LIST` is generated using Spark.


server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/ParquetResolverTest.java
```diff
@@ -590,7 +839,8 @@ private List<ColumnDescriptor> getColumnDescriptorsFromSchema(MessageType schema
         return schema.getFields()
                 .stream()
                 .map(f -> {
-                    ParquetTypeConverter converter = ParquetTypeConverter.from(f.asPrimitiveType());
+                    Type type = f.isPrimitive() ? f.asPrimitiveType() : f.asGroupType();
```
denalex, line 842: can this logic reside in ParquetTypeConverter.from(type) method ?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetTypeConverter.java
```diff
@@ -219,20 +222,161 @@ public Object getValue(Group group, int columnIndex, int repeatIndex, Type type)
...
+                LOG.error(e.getMessage());
+            }
+            return null;
+        }
 
-    // ********** PUBLIC INTERFACE **********
-    public abstract DataType getDataType(Type type);
+        @Override
+        public Object getValue(Group group, int columnIndex, int repeatIndex, Type type) {
+            PgUtilities pgUtilities = new PgUtilities();
```
denalex, line 279: I think we would want to create PgUtilities once per class, not per function call.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetTypeConverter.java
```diff
@@ -219,20 +222,161 @@ public Object getValue(Group group, int columnIndex, int repeatIndex, Type type)
...
+            pgArrayBuilder.startArray();
+
+            Group listGroup = group.getGroup(columnIndex, repeatIndex);
+            int repetitionCount = listGroup.getFieldRepetitionCount(0);
+            for (int i = 0; i < repetitionCount; i++) {
+                Group repeatedGroup = listGroup.getGroup(0, i);
+                if (repeatedGroup.getFieldRepetitionCount(0) == 0) {
+                    pgArrayBuilder.addElement((String) null);
+                } else {
+                    PrimitiveType elementType = type.asGroupType().getType(0).asGroupType().getType(0).asPrimitiveType();
```
denalex, line 290: can this be done outside the loop since all elements should be of the same type ?

server/pxf-hdfs/src/test/resources/parquet/generate_parquet_list_types_without_null.hql
```diff
@@ -0,0 +1,34 @@
...
+
+INSERT INTO parquet_list_types_without_null_tmp SELECT 2, array(false, true, boolean(1), boolean(0)), array(tinyint(128),tinyint(96)), array(2,3), array(bigint(-9223372036854775808), bigint(223372036854775808)), array(float(-123456.987654321), float(123456.987654321)), array(double(1.0), double(-99.9)), array('this is a string with "special" characters', 'this is a string without'), array(unhex('DEADBEEF'), unhex('ADBEEF')), array(cast('this is exactly' as char(15)), cast(' fifteen chars.' as char(15))), array(cast('this is exactly' as varchar(15)), cast(' fifteen chars.' as varchar(15))), array(cast('2022-10-07' as date),cast('2022-10-08' as date)), array(cast(1.23456 as decimal(38,18)), cast(1.23456 as decimal(38,18))) FROM foo;
+
+DROP TABLE IF EXISTS parquet_list_types_without_null;
+
+CREATE TABLE parquet_list_types_without_null STORED AS PARQUET as SELECT * FROM parquet_list_types_without_null_tmp ORDER BY id;
```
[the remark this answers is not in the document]
yimingli-vmware: These 2 `INSERT` will generate 2 parquet files. So I create another table and l select all the data from the first table, then I have only 1 parquet file. Do we have a better way to do that?

> not sure why you need CTAS since you're already inserting into PARQUET table?

server/pxf-api/src/main/java/org/greenplum/pxf/api/io/DataType.java
```diff
@@ -90,6 +90,23 @@ public enum DataType {
         TIMESTAMPARRAY.typeElem = TIMESTAMP;
         TIMESTAMP_WITH_TIMEZONE_ARRAY.typeElem = TIMESTAMP_WITH_TIME_ZONE;
 
+        SMALLINT.typeArray = INT2ARRAY;
```
ashuka24, line 93: why do we need a reverse mapping when we have the typeElem mapping for arrays above?

server/pxf-api/src/main/java/org/greenplum/pxf/api/io/DataType.java
```diff
@@ -158,4 +176,6 @@ public int getOID() {
     public DataType getTypeElem() {
         return typeElem;
     }
+
+    public DataType getTypeArray() { return typeArray; }
```
ashuka24, line 180: can we match styling as above?
```suggestion
    public DataType getTypeArray() { 
        return typeArray; 
    }

File: server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetTypeConverter.java [hunk 7]
```diff
@@ -219,28 +232,69 @@ public Object getValue(Group group, int columnIndex, int repeatIndex, Type type)
...
+    },
 
-    // ********** PUBLIC INTERFACE **********
-    public abstract DataType getDataType(Type type);
+    LIST {
+        @Override
+        public DataType getDataType(Type type) {
+            Type elementType = type.asGroupType().getType(0).asGroupType().getType(0);
+            if (!elementType.isPrimitive()) {
+                throw new PxfRuntimeException(String.format("List of %s is not supported.", elementType.getOriginalType()));
```
ashuka24, line 242:
```suggestion
                throw new PxfRuntimeException(String.format("List of type %s is not supported.", elementType.getOriginalType()));

File: server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetTypeConverter.java [hunk 8]
```diff
@@ -219,28 +232,69 @@ public Object getValue(Group group, int columnIndex, int repeatIndex, Type type)
...
+     * @return a {@link ParquetTypeConverter} containing the ENUM value
+     */
+    public static ParquetTypeConverter from(Type type) {
+        if (type.isPrimitive()) {
+            return valueOf(type.asPrimitiveType().getPrimitiveTypeName().name());
+        }
+        try {
+            return valueOf(type.getOriginalType().name());
+        } catch (IllegalArgumentException e) {
+            throw new PxfRuntimeException(String.format("Complex type %s is not supported", type.getOriginalType().name()), e);
```
ashuka24, line 294:
```suggestion
            throw new PxfRuntimeException(String.format("Parquet complex type %s is not supported", type.getOriginalType().name()), e);

File: server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetFilterPushDownTest.java [hunk 9]
```diff
@@ -769,7 +769,6 @@ private void assertValues(List<OneField> fieldList, final int row) {
         } else {
             assertNull(fieldList.get(1).val, "Row " + row);
         }
-
```
ashuka24, line 772: nit: let's remove this unnecessary change

server/pxf-hdfs/src/test/resources/parquet/README.md
```diff
@@ -0,0 +1,106 @@
...
+```shell script
+./generate_parquet_types.bash
+```
+
+The `parquet_types.parquet` file will be copied to the directory where you ran the
+script.
+
+## Generate the parquet_list_types.parquet file using Hive
+
+Identify your HDFS and HIVE commands. Identify the Hive warehouse path where
```
ashuka24, line 37: I have the same comments here as the first README: can we remove some of the duplication in this file?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetTypeConverter.java
```diff
@@ -66,6 +68,17 @@ public Object getValue(Group group, int columnIndex, int repeatIndex, Type type)
                 return group.getString(columnIndex, repeatIndex);
             }
         }
+
+        @Override
+        public void addValueToArray(Group group, int columnIndex, int repeatIndex, PrimitiveType primitiveType, PgArrayBuilder pgArrayBuilder) {
+            Object value = getValue(group, 0, 0, primitiveType);
```
ashuka24, line 74: is there ever a case where the col index or the repeat index value will not be 0?

nit:
```suggestion
                                        Object value = getValue(group, 0 /*col index*/, 0 /*repeat index*/, primitiveType);
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
```diff
@@ -228,10 +230,12 @@ private OneField resolveField(Group group, int columnIndex, Type type) {
         ParquetTypeConverter converter = ParquetTypeConverter.from(type);
         // determine how many values for the field are present in the column
         int repetitionCount = group.getFieldRepetitionCount(columnIndex);
+        // repetitionCount would only be 0 or 1
```
ashuka24, line 233:
```suggestion
        // repetitionCount can only be 0 or 1

File: server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetTypeConverter.java [hunk 13]
```diff
@@ -219,28 +233,69 @@ public Object getValue(Group group, int columnIndex, int repeatIndex, Type type)
...
+    },
 
-    // ********** PUBLIC INTERFACE **********
-    public abstract DataType getDataType(Type type);
+    LIST {
+        @Override
+        public DataType getDataType(Type type) {
+            Type elementType = type.asGroupType().getType(0).asGroupType().getType(0);
+            if (!elementType.isPrimitive()) {
+                throw new UnsupportedTypeException(String.format("List of type %s is not supported.", elementType.getOriginalType().name()));
```
ashuka24, line 243:
```suggestion
                throw new UnsupportedTypeException(String.format("Parquet list of type %s is not supported.", elementType.getOriginalType().name()));

File: server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java [hunk 14]
```diff
@@ -52,11 +53,10 @@ public class ParquetResolver extends BasePlugin implements Resolver {
     // used to distinguish string pattern between type "timestamp" ("2019-03-14 14:10:28")
     // and type "timestamp with time zone" ("2019-03-14 14:10:28+07:30")
     public static final Pattern TIMESTAMP_PATTERN = Pattern.compile("[+-]\\d{2}(:\\d{2})?$");
-
+    private final ObjectMapper mapper = new ObjectMapper();
```
hpandeycodeit, line 56: This object is only used inside the else block Lines 241 - 251. So may be move it there?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
```diff
@@ -218,42 +216,41 @@ private void validateSchema() {
...
-            field.val = repetitionCount == 0 ? null : converter.getValue(group, columnIndex, 0, type);
-        } else if (type.getRepetition() == REPEATED) {
-            // repeated primitive at any level will convert into JSON
+            field.val = null;
+        } else if (type.getRepetition() != REPEATED) {
+            // here the repetition count can only be 1
+            field.type = converter.getDataType(type).getOID();
+            field.val = converter.getValue(group, columnIndex, 0, type);
+        } else {
+            // repeated primitive will be converted into JSON
```
hpandeycodeit, line 242:
```suggestion
            // repeated primitives will be converted into JSON
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetTypeConverter.java
```diff
@@ -219,28 +233,69 @@ public Object getValue(Group group, int columnIndex, int repeatIndex, Type type)
...
-
+    },
 
-    // ********** PUBLIC INTERFACE **********
-    public abstract DataType getDataType(Type type);
+    LIST {
+        @Override
+        public DataType getDataType(Type type) {
+            Type elementType = type.asGroupType().getType(0).asGroupType().getType(0);
+            if (!elementType.isPrimitive()) {
```
hpandeycodeit, line 242: Will this elementType always be non-null ?

automation/src/test/resources/data/parquet/generate_parquet_types.bash
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+# Requires Hive 2.3+
+
+set -euxo pipefail
+
+# Run this command to generate the parquet_types.parquet file
+
+SRC_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
+HDFS_CMD=${HDFS_CMD:-~/workspace/singlecluster/bin/hdfs}
```
denalex, line 10: in a file above the default was `~/workspace/singlecluster-HDP3/bin/hdfs`, so it is not consistent

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
denalex: it can only be 0 or 1 for non REPEATED repetition type, it can be higher for REPEATED type

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
denalex: it is used for resolving each field, we don't want to create this object for every field, so I think it is ok being here

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
```diff
@@ -218,42 +216,41 @@ private void validateSchema() {
...
-                    throw new RuntimeException("Failed to serialize repeated parquet type " + type.asPrimitiveType().getName(), e);
-                }
-            } else {
-                // just return the array node within OneField container
-                field.val = jsonArray;
+            field.type = DataType.TEXT.getOID();
+            try {
+                field.val = mapper.writeValueAsString(jsonArray);
+            } catch (Exception e) {
+                throw new RuntimeException("Failed to serialize repeated parquet type " + type.asPrimitiveType().getName(), e);
```
denalex, line 251: I think we lost recursion and the concept of `level`, need to look closer as to why and whether that is ok

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/ParquetResolverTest.java
denalex: I think we already implemented this, so that ParquetTypeConverter.from(type) can take any type, primitive or complex, so can you just do below:
```
ParquetTypeConverter converter = ParquetTypeConverter.from(f);
```

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/ParquetResolverTest.java
```diff
@@ -521,20 +527,218 @@ public void testGetFields_Primitive_RepeatedInt() throws IOException {
...
+        for (Object expectedValue : expectedValues) {
+            if (expectedValue == null) {
+                pgArrayBuilder.addElement((String) null);
+            } else if (type == DataType.BYTEAARRAY) {
+                pgArrayBuilder.addElementNoEscaping(pgUtilities.encodeAndEscapeByteaHex(ByteBuffer.wrap(((Binary) expectedValue).getBytes())));
+            } else {
+                pgArrayBuilder.addElement(String.valueOf(expectedValue));
+            }
         }
+        pgArrayBuilder.endArray();
```
denalex, line 739: this assertion relies on pgArrayBuilder logic to provide expected values, it would be preferable to hardcode them to avoid not catching potential regressions with that logic

denalex (commented):
generally looks good, a few minor comments and these items we should revisit:
- support for timestamp with timezone
- using timezone casts for tests
- should we use hardcoded expected values
- do we have enough unit test coverage ? I have not seen tests for error cases for ParquetTypeConverter (when we would try to resolve unsupported types)
- we lost recursion and level, need to understand why

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
yimingli-vmware: From the old code, this old `resolvePrimitive` method was not called recursively. And the only place we call `resolvePrimitive` was in line 81 in the old code: 
```
oneField = resolvePrimitive(group, columnIndex, schema.getType(columnIndex), 0);
```
The `level` variable was always `0` and it wasn't changed. We would never go into this `else` block in line 248:
```
else {
                // just return the array node within OneField container
                field.val = jsonArray;
 }
```
So I think we can remove this `level` variable and the `else` block.

automation/src/test/java/org/greenplum/pxf/automation/features/parquet/ParquetReadTest.java
```diff
@@ -0,0 +1,190 @@
...
+        gpdb.copyFromFile(gpdbNumericWithPrecisionScaleTable, new File(localDataResourcesFolder
+                + "/numeric/" + NUMERIC_FILENAME), "E','", true);
+
+        prepareReadableExternalTable(PXF_PARQUET_TABLE, PARQUET_TABLE_COLUMNS, hdfsPath + PARQUET_PRIMITIVE_TYPES);
+    }
+
+    @Test(groups = {"features", "gpdb", "security", "hcfs"})
+    public void parquetReadPrimitives() throws Exception {
+        gpdb.runQuery("CREATE OR REPLACE VIEW parquet_view AS SELECT s1, s2, n1, d1, dc1, " +
+                "CAST(tm AS TIMESTAMP WITH TIME ZONE) AT TIME ZONE 'PDT' as tm, " +
```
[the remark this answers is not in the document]
yimingli-vmware: We need to cast it since our concourse timezone is different from our local timezone.

For the write case, if we insert a timestamp, this timestamp is our local timestamp. It will be converted into UTC, then stored into Parquet format since Parquet doesn't support `timestamp with timezone`.

For the read case, if we want to query a timestamp, we want the the timestamp to be displayed in our local timezone. Parquet only stores UTC timestamp, and our server stores timezone information. 

Back to this code, if we don't cast the timestamp into a specific timezone, our timestamp output would be different in our local env and concourse env. Then we would failed in the automation test since our expected output is hardcoded.

yimingli-vmware:
>

yimingli-vmware:
> generally looks good, a few minor comments and these items we should revisit:
> 
> * support for timestamp with timezone
> * using timezone casts for tests
> * should we use hardcoded expected values
> * do we have enough unit test coverage ? I have not seen tests for error cases for ParquetTypeConverter (when we would try to resolve unsupported types)
> * we lost recursion and level, need to understand why

- support for timestamp with timezone
   I have unit tests for writing timestamp with timezone. From the write side, user can write timestamp with any timezone since we'll convert it into UTC nano time, then store it as byte array form into parquet. But from the read side, a method `bytesToTimestamp` in `ParquetTypeConverter` convert parquet byte array to java timestamp IN LOCAL SERVER'S TIME ZONE. This timezone information is hard-coded. So we only support timestamp with local timezone

- using timezone casts for tests
  Since we read timestamp with local timezone, and we have different timezones for dev environment and concourse environment, we need to convert the output timestamp into a specific timezone to match the expected results.

- should we use hardcoded expected values?
  will discuss

- test coverage
  I've added tests for testing unsupported element types for LIST and unsupported complex type, and our exceptions are working.
  
  -  we lost recursion and level
    will discuss

automation/src/test/resources/data/parquet/README.md
```diff
@@ -0,0 +1,92 @@
+# Generate Parquet files for testing
```
bradfordb-vmware, line 1: This file is the same as `server/pxf-hdfs/src/test/resources/parquet/README.md`. Rather than repeat the content here, would it make sense to just have a Markdown link to the other one?

automation/src/test/resources/data/parquet/generate_parquet_list_types.bash
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
bradfordb-vmware, line 1: This file is the exact same as the one in `server/pxf-hdfs/src/test/resources/parquet/generate_parquet_list_types.bash`. Is there anyway we can reuse those files without copying them?

server/pxf-api/src/main/java/org/greenplum/pxf/api/io/DataType.java
```diff
@@ -104,6 +121,7 @@ public enum DataType {
 
     private final int OID;
     private DataType typeElem;
+    private DataType typeArray;
```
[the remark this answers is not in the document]
denalex: Here're my personal opinions: 

I see your point, but the "state space" does not seem very concerning to me. This is a type system with primitive and complex types, where each primitive type knows its corresponding array type and each array type knows its corresponding primitive type. It is the presence of this knowledge that makes a type "primitive" or "array" in the eyes of a consumer. If we are concerned about 2 extra states and overall integrity, we can have a unit test that iterates over available types and ensures the invariants are held.

We could have expressed this as a class hierarchy (like Parquet does) with basic `Type` and subclasses or `PrimitiveType` and `ComplexType`, but this will be much more verbose and you'll have to do castings, just like Parquet does. I think what we have here is better and leads to more compact and readable code (IMO).

We could also have the mapping between primitive and complex types in a separate place, but since the mapping is known a priori, I think it belongs with the type system itself, not with a client of the type system, like a converter.

Finally, the mapping is also needed in ORC case, except there we use functions and maps of types to functions. Again, I personally like this self-documenting enum pattern much more that maintaining multiple separate maps and conversion functions.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetTypeConverter.java
```diff
@@ -309,4 +364,20 @@ private static Binary getBinaryFromZonedDateTime(String timestampString, ZonedDa
...
     }
+
+    // ********** PUBLIC INTERFACE **********
+    public abstract DataType getDataType(Type type);
+
+    public abstract Object getValue(Group group, int columnIndex, int repeatIndex, Type type);
+
+    public abstract void addValueToJsonArray(Group group, int columnIndex, int repeatIndex, Type type, ArrayNode jsonNode);
+
+    public DataType getArrayDataType(Type type) {
```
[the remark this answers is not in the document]
denalex: yes, this can be done, but I personally think maintaining the mapping in `DataType` is better (see another comment). The drawback of this here:
- the mapping `Type -> DataType (array)` can not be precomputed here since `Type` might have different logical annotations available only per-instance, 
- so we will have to pre-compute `DataType (prim) -> DataType (array)`, but this is a property of a type system, not the converter and is not Parquet-specific.
- iteration over a list will likely be less performant as opposed to reading value of enum field

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/AvroResolverTest.java
```diff
@@ -459,6 +459,7 @@ public void testTimeStamps() throws Exception {
         List<OneField> oneFields = resolver.getFields(new OneRow(null, genericRecord));
         List<Schema.Field> schemaFields = schema.getFields();
         // Default System TimeZone
+        TimeZone defaultTimeZone = TimeZone.getDefault();
```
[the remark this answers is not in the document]
yimingli-vmware: I have to add this line to make my PR pipeline green. But i'll remove it when merging to `main`

server/pxf-hdfs/src/test/resources/parquet/README.md
```diff
@@ -0,0 +1,92 @@
...
+./generate_parquet_list_types_without_null.bash
+```
+
+The `parquet_list_types_without_null.parquet` file will be copied to the directory where you ran the script.
+
+## Generate Parquet TIMESTAMP LIST Testing Data using Spark
+
+According to the latest version of[Hive](https://github.com/apache/hive/blob/4e4e39c471094567dcdfd9840edbd99d7eafc230/ql/src/java/org/apache/hadoop/hive/ql/io/parquet/vector/VectorizedParquetRecordReader.java#L578),
+Hive doesn't support TIMESTAMP LIST. Therefore, we use Spark to generate TIMESTAMP LIST dataset. Note that the input timestamps are in
+local time zone, and Parquet will store them in UTC time zone.
```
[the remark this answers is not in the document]
yimingli-vmware: The parquet fill will be the same. If we read the parquet file using parquet tools, the timestamps are all UTC timestamps. But if we try to read it out through greenplum, we would convert the timestamp based on server's timezone. So NY users and SF users can see the same parquet file using parquet tool, but they would see different values in greenplum external table. 

I just checked this morning that although our parquet write doesn't limit timezone, our parquet read hard-code the timezone to be local timezone, which means we now only support timestamp with local timezone.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
```diff
@@ -218,42 +216,41 @@ private void validateSchema() {
...
-        // determine how many values for the primitive are present in the column
+        // get type converter based on the field type
+        ParquetTypeConverter converter = ParquetTypeConverter.from(type);
+        // determine how many values for the field are present in the column
         int repetitionCount = group.getFieldRepetitionCount(columnIndex);
-
-        // at the top level (top field), non-repeated primitives will convert to typed OneField
-        if (level == 0 && type.getRepetition() != REPEATED) {
+        // repetitionCount would only be 0 or 1
+        if (repetitionCount == 0) {
```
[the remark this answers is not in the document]
yimingli-vmware: I updated the comments:
- For REPEATED type, repetitionCount can be any non-negative number, the element will be converted into JSON
- For non-REPEATED type, repetitionCount can only be 0 or 1. `repetitionCount == 0` means this is a null LIST/Primitive

I re-ordered these conditions. first condition is `type.getRepetition() == REPEATED`, the second is `repetitionCount == 0`, the third is supported array & primitives with repetitionCount==1

automation/src/test/resources/data/parquet/generate_parquet_list_types.bash
yimingli-vmware: I kept all these bash files under `automation` directory, and they now can generate data files for both automation tests and unit tests.

automation/src/test/resources/data/parquet/generate_undefined_precision_numeric_parquet.hql
```diff
@@ -1,3 +1,8 @@
+!connect jdbc:hive2://localhost:10000/
+
+DROP TABLE IF EXISTS undefined_precision_numeric_csv;
```
ashuka24, line 3: same here, why is this duplicated?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
```diff
@@ -250,6 +243,15 @@ private OneField resolveField(Group group, int columnIndex, Type type) {
             } catch (Exception e) {
                 throw new RuntimeException("Failed to serialize repeated parquet type " + type.asPrimitiveType().getName(), e);
             }
+        }else if (repetitionCount == 0){
```
ashuka24, line 246:
```suggestion
        } else if (repetitionCount == 0) {

File: server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java [hunk 30]
```diff
@@ -250,6 +243,15 @@ private OneField resolveField(Group group, int columnIndex, Type type) {
...
+            // For non-REPEATED type, repetitionCount can only be 0 or 1
+            // repetitionCount == 0 means this is a null LIST/Primitive
+            field.type = converter.getDataType(type).getOID();
+            field.val = null;
+        }else{
```
ashuka24, line 251:
```suggestion
        } else {
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
```diff
@@ -217,29 +215,24 @@ private void validateSchema() {
...
-        if (repetitionCount == 0) {
-            field.type = converter.getDataType(type).getOID();
-            field.val = null;
-        } else if (type.getRepetition() != REPEATED) {
-            // here the repetition count can only be 1
-            field.type = converter.getDataType(type).getOID();
-            field.val = converter.getValue(group, columnIndex, 0, type);
-        } else {
-            // repeated primitive will be converted into JSON
+        if (type.getRepetition() == REPEATED){
```
ashuka24, line 233:
```suggestion
        if (type.getRepetition() == REPEATED) {

File: server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/ParquetResolverTest.java [hunk 32]
```diff
@@ -491,7 +495,7 @@ public void testGetFields_Primitive_Repeated_Synthetic() {
         List<OneField> fields = assertRow(groups, 0, 16);
 
         assertField(fields, 0, "[\"row1-1\",\"row1-2\"]", DataType.TEXT);
-        assertField(fields, 1, null, DataType.TEXT);
+        assertField(fields, 1, "[]", DataType.TEXT);
```
ashuka24, line 498: why did this change from null to empty array?

automation/src/test/resources/data/parquet/README.md
```diff
@@ -8,7 +8,7 @@ follow the instructions below.
 
 - Hadoop CLI commands
```
ashuka24, line 9: can we specify here that we are using the `singlecluster-HDP3` and that all the bash scripts we have in this directory use the HDP3 cluster to generate files.

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/ParquetWriteTest.java
```diff
@@ -42,7 +42,6 @@
 import java.util.Collections;
 import java.util.List;
 
-import static java.math.BigDecimal.ROUND_UNNECESSARY;
```
ashuka24, line 45: is this change necessary? it seems like an unnecessary change in regards to this PR (similar to the one in `ParquetFilterPushDownTest.java`)

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetTypeConverter.java
yimingli-vmware: added a null pointer error handling

server/pxf-api/src/main/java/org/greenplum/pxf/api/io/DataType.java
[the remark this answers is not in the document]
yimingli-vmware: added a unit test for testing valid combinations `(typeElem, typeArray)`

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/ParquetResolverTest.java
yimingli-vmware: Before I introduced a slight logical change and the original assertion with `"[]"` failed. So I change it to `null`. But now I fix the logic and change this assertion back to `"[]"`

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
```diff
@@ -218,42 +214,45 @@ private void validateSchema() {
...
-                    throw new RuntimeException("Failed to serialize repeated parquet type " + type.asPrimitiveType().getName(), e);
-                }
-            } else {
-                // just return the array node within OneField container
-                field.val = jsonArray;
+            field.type = DataType.TEXT.getOID();
+            try {
+                field.val = mapper.writeValueAsString(jsonArray);
+            } catch (Exception e) {
+                throw new RuntimeException("Failed to serialize repeated parquet type " + type.asPrimitiveType().getName(), e);
```
denalex, line 244: we don't know here whether the type was actually a primitive, do we ?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
```diff
@@ -218,42 +214,45 @@ private void validateSchema() {
         }
     }
 
-    private OneField resolvePrimitive(Group group, int columnIndex, Type type, int level) {
-
+    /**
+     * Resolve the Parquet data at the columnIndex into a Postgres type in String representation
+     *
```
Aayushs4, line 219: nit: extra line

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
```diff
@@ -218,42 +214,45 @@ private void validateSchema() {
...
+     * @param group       contains parquet schema and data of a {@link OneRow}
```
Aayushs4, line 220: nit: extra space

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetTypeConverter.java
```diff
@@ -219,28 +235,70 @@ public Object getValue(Group group, int columnIndex, int repeatIndex, Type type)
...
     private static final Logger LOG = LoggerFactory.getLogger(ParquetTypeConverter.class);
 
+    /**
+     * Retrieve corresponding ENUM value according to the input parquet type
+     *
+     * @param type is the input parquet type
+     * @return a {@link ParquetTypeConverter} containing the ENUM value
+     */
+    public static ParquetTypeConverter from(Type type) {
+        if (type.isPrimitive()) {
```
Aayushs4, line 292: Can type be NULL?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetTypeConverter.java
```diff
@@ -309,4 +367,45 @@ private static Binary getBinaryFromZonedDateTime(String timestampString, ZonedDa
...
+     *   }
+     * }
+     *
+     * - The outer-most level must be a group annotated with `LIST` that contains a single field named `list`. The repetition of this level must be either `optional` or `required` and determines whether the list is nullable.
+     * - The middle level, named `list`, must be a repeated group with a single field named `element`.
+     * - The `element` field encodes the list's element type and repetition. Element repetition must be `required` or `optional`.
+     */
+    private static Type getElementType(Type type) {
+        if (type.asGroupType().getFields().size() != 1 || type.asGroupType().getType(0).asGroupType().getFields().size() != 1) {
+            String invalidListSchema = type.asGroupType().getFields().size() != 1 ? type.asGroupType().toString() : type.asGroupType().getType(0).asGroupType().toString();
```
denalex, line 385: can we format this as 3 lines for easy reading ?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetTypeConverter.java
```diff
@@ -309,4 +367,45 @@ private static Binary getBinaryFromZonedDateTime(String timestampString, ZonedDa
...
+            String validListSchema = "<list-repetition> group <name> (LIST) {\n"
+                    + "  repeated group list {\n"
+                    + "    <element-repetition> <element-type> element;\n"
+                    + "  }\n"
+                    + "}";
+            throw new PxfRuntimeException(String.format("Invalid Parquet List schema:\n %s. \nThe valid Parquet List schema should be:\n %s.", invalidListSchema, validListSchema));
```
denalex, line 391: please do not make error messages multi-line as we prefer the log file where they end up to have 1 line per entry. The likelihood of this ever being shown to users is very low as this is a very edge case.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetTypeConverter.java
```diff
@@ -309,4 +367,45 @@ private static Binary getBinaryFromZonedDateTime(String timestampString, ZonedDa
...
+        }
+        return type.asGroupType().getType(0).asGroupType().getType(0);
```
denalex, line 393: this is a long chain of access and this call is somewhat on a critical pass. Can we store and reuse intermediate results ?

server/pxf-hdfs/src/test/resources/parquet/README.md
```diff
@@ -0,0 +1,5 @@
+# Generate Parquet files for testing
+
+These instructions will help you generate the Parquet files required for testing.
+The parquet files are pre-generated, but if you want to generate these files again,
+please refer to this [README](https://github.com/greenplum-db/pxf/blob/3c9a0810de49b1b59b9960d90c21d5202480c233/automation/src/test/resources/data/parquet/README.md)
```
denalex, line 5: can we use relative path here ?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
yimingli-vmware: I checked other files, seems like we are all following the format when the java doc has a description and parameter explanation:
```
/**
 * Some high level description
 *
 * @param myParam1     explanation for myParam1 
 * @param myParam2     explanation for myParam2
 * / 
```
we need that line between the high level description and params.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
yimingli-vmware: These spaces are generated by the auto-formatting tool. The explanation of each @param should be vertically aligned.

automation/src/test/resources/data/parquet/generate_parquet_timestamp_list.py
```diff
@@ -0,0 +1,45 @@
+from pyspark.sql import SparkSession
+from pyspark.sql.functions import expr
+from pyspark.sql.types import IntegerType
+import shutil,os
```
bradfordb-vmware, line 4: Style-nit: 

```suggestion
import shutil, os
```

I don't think the team has an existing convention for Python code style, but I generally run [`black`](https://github.com/psf/black) on all my Python code so that it is consistent and automatically formatted for me.

automation/src/test/resources/data/parquet/generate_parquet_timestamp_list.py
```diff
@@ -0,0 +1,45 @@
...
+
+spark = SparkSession.builder.master("local[2]").appName("generate_parquet_timestamp_list").getOrCreate()
```
bradfordb-vmware, line 6: Style-nit:

```suggestion
spark = (
    SparkSession.builder.master("local[2]")
    .appName("generate_parquet_timestamp_list")
    .getOrCreate()
)
```

automation/src/test/resources/data/parquet/generate_parquet_timestamp_list.py
```diff
@@ -0,0 +1,45 @@
...
+
+# prepare the timestamp array dataset in string type
+data = [
+    (1, ["2022-10-05 11:30:00", "2022-10-06 12:30:00", "2022-10-07 13:30:00"]),
+    (2, ["2022-10-05 11:30:00","2022-10-05 11:30:00","2022-10-07 13:30:00"]),
```
bradfordb-vmware, line 11: Style-nit:

```suggestion
    (2, ["2022-10-05 11:30:00", "2022-10-05 11:30:00", "2022-10-07 13:30:00"]),
```

automation/src/test/resources/data/parquet/generate_parquet_timestamp_list.py
```diff
@@ -0,0 +1,45 @@
...
+    (3, [None, "2022-10-05 11:30:00", "2022-10-05 11:30:00"]),
+    (4, [None]),
+    (5, []),
+    (6, None)
```
bradfordb-vmware, line 15: Style-nit:

```suggestion
    (6, None),
```

bradfordb-vmware (approved):

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetTypeConverter.java
```diff
@@ -309,4 +367,45 @@ private static Binary getBinaryFromZonedDateTime(String timestampString, ZonedDa
...
+     *     <element-repetition> <element-type> element;
+     *   }
+     * }
+     *
+     * - The outer-most level must be a group annotated with `LIST` that contains a single field named `list`. The repetition of this level must be either `optional` or `required` and determines whether the list is nullable.
+     * - The middle level, named `list`, must be a repeated group with a single field named `element`.
+     * - The `element` field encodes the list's element type and repetition. Element repetition must be `required` or `optional`.
+     */
+    private static Type getElementType(Type type) {
+        if (type.asGroupType().getFields().size() != 1 || type.asGroupType().getType(0).asGroupType().getFields().size() != 1) {
```
ashuka24, line 384: Looks like we do `type.asGroupType().getType(0).asGroupType()` or similar iterations a lot. It may be beneficial to pull this out into a variable and reference this in all locations. would help with readability too. 
ex: 
```
Group underlyingGroup = type.asGroupType().getType(0).asGroupType()

underlyingGroup.getFields().size();
underlyingGroup.toString()
etc
```

ashuka24 (approved):

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
```diff
@@ -202,7 +198,7 @@ private void fillGroup(int index, OneField field, Group group, Type type) throws
                 group.add(index, (Boolean) field.val);
                 break;
             default:
-                throw new IOException("Not supported type " + type.asPrimitiveType().getPrimitiveTypeName());
+                throw new UnsupportedTypeException("Not supported type " + type.asPrimitiveType().getPrimitiveTypeName());
```
hpandeycodeit, line 201: Can we change it a bit so it's in sync with the other error messages like `Parquet type %s is not supported`

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetTypeConverter.java
```diff
@@ -309,4 +367,45 @@ private static Binary getBinaryFromZonedDateTime(String timestampString, ZonedDa
...
+            String invalidListSchema = type.asGroupType().getFields().size() != 1 ? type.asGroupType().toString() : type.asGroupType().getType(0).asGroupType().toString();
+            String validListSchema = "<list-repetition> group <name> (LIST) {\n"
+                    + "  repeated group list {\n"
+                    + "    <element-repetition> <element-type> element;\n"
+                    + "  }\n"
+                    + "}";
```
hpandeycodeit, line 390: This validListSchema contains the newlines. Will the invalidListSchema also contains the newlines or no?

automation/src/test/resources/data/parquet/generate_parquet_types.hql
```diff
@@ -1,3 +1,6 @@
+!connect jdbc:hive2://localhost:10000/
+
+DROP TABLE IF EXISTS parquet_types_csv;
```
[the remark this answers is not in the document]
yimingli-vmware: It was because the first `DROP` is treated as an input for username and password, so only the second `DROP` did the actual drop table. Now I move the open connection commands into bash scripts.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetTypeConverter.java
yimingli-vmware: I think it will not really happen, but still add an error handling for it.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
yimingli-vmware: changed into `type.getOriginalType().name()`

automation/src/test/resources/data/parquet/generate_parquet_timestamp_list.py
yimingli-vmware: formatted using `black`

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetTypeConverter.java
yimingli-vmware: Alex said it's better not to have newlines in error messages, so I removed the validSchema and only report the invalid schema. The valid `LIST` schema can be easily looked up in parquet README.

The invalid schema contains newlines but we cannot control its format

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
yimingli-vmware: It's the code for parquet write one. I'll change that in my write branch

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/ParquetWriteTest.java
```diff
@@ -42,6 +42,7 @@
 import java.util.Collections;
 import java.util.List;
 
+import static java.math.BigDecimal.ROUND_UNNECESSARY;
```
denalex, line 45: the only change here, is it needed ?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetTypeConverter.java
```diff
@@ -73,13 +73,14 @@ public Object getValue(Group group, int columnIndex, int repeatIndex, Type type)
...
 
         @Override
-        public void addValueToArray(Group group, int columnIndex, int repeatIndex, PrimitiveType primitiveType, PgArrayBuilder pgArrayBuilder) {
+        public String getValueFromList(Group group, int columnIndex, int repeatIndex, PrimitiveType primitiveType) {
             Object value = getValue(group, columnIndex, repeatIndex, primitiveType);
             if (primitiveType.getLogicalTypeAnnotation() == null) {
                 ByteBuffer byteBuffer = ByteBuffer.wrap((byte[]) value);
-                pgArrayBuilder.addElementNoEscaping(pgUtilities.encodeAndEscapeByteaHex(byteBuffer));
+                String escapedByteaHex = pgUtilities.encodeAndEscapeByteaHex(byteBuffer);
+                return pgUtilities.unescapeArrayElement(escapedByteaHex);
```
denalex, line 81: I think this is not the best approach — why do we need to escape and then unescape the element ? Seems like doing extra work on the super-critical path (handling of each array element). I understand it makes the refactoring easier, but we should find a better way if we want to change the original logic (and I am not sure why the original logic needed to be changed)

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetFilterPushDownTest.java
```diff
@@ -877,3 +877,4 @@ private void assertBigDecimal(String message, Double expectedDouble, Object actu
     }
 
 }
+
```
hpandeycodeit, line 880: nit:  new line

hpandeycodeit (approved):

server/pxf-api/src/test/java/org/greenplum/pxf/api/io/DataTypeTest.java
```diff
@@ -31,4 +31,26 @@ public void enumElementArrayMapping() {
             }
         }
     }
+
+    @Test
+    public void testNeedsEscapingInArray() {
+        for (DataType dataType : DataType.values()) {
+            if (dataType.isArrayType()) {
```
denalex, line 38: this is confusing -- all array types are currently using the default of `true`, so why do we need to check them individually ?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetTypeConverter.java
```diff
@@ -269,7 +269,11 @@ public Object getValue(Group group, int columnIndex, int repeatIndex, Type type)
                     // add the non-null element into array
                     PrimitiveType elementType = getElementType(type.asGroupType()).asPrimitiveType();
                     String elementValue = from(elementType).getValueFromList(repeatedGroup, 0, 0, elementType);
-                    pgArrayBuilder.addElement(elementValue);
+                    if (elemGPType.getNeedsEscapingInArray()) {
+                        pgArrayBuilder.addElement(elementValue);
+                    } else {
+                        pgArrayBuilder.addElementNoEscaping(elementValue);
+                    }
```
denalex, line 276: we can use
```
pgArrayBuilder.addElement(elementValue, elemGPType.getNeedsEscapingInArray());
```
if we create this new method which takes a boolean second parameter on whether escaping is needed.

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/ParquetResolverTest.java
```diff
@@ -766,40 +765,23 @@ private List<OneField> assertRow(List<Group> groups, int desiredRow, int numFiel
...
-        pgArrayBuilder.startArray();
-        for (Object expectedValue : expectedValues) {
-            if (expectedValue == null) {
-                pgArrayBuilder.addElement((String) null);
-            } else if (type == DataType.BYTEAARRAY) {
-                pgArrayBuilder.addElementNoEscaping(pgUtilities.encodeAndEscapeByteaHex(ByteBuffer.wrap(((Binary) expectedValue).getBytes())));
-            } else {
-                pgArrayBuilder.addElement(String.valueOf(expectedValue));
-            }
+    private void assertList(List<OneField> fields, int index, String expectedValue) {
```
denalex, line 779: do we want to assert type of `OneField` object ? Why did we lose all type assertions as we had them in `assertField` ?

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/ParquetResolverTest.java
```diff
@@ -766,40 +765,23 @@ private List<OneField> assertRow(List<Group> groups, int desiredRow, int numFiel
...
+        if (expectedValue.equals("NULL")) {
```
denalex, line 780: why not use Java `null` as before instead of artificial `"NULL"` string ?

server/pxf-api/src/test/java/org/greenplum/pxf/api/io/DataTypeTest.java
yimingli-vmware: I want to test all array types are set to be true by default, and their corresponding primitive types are set correctly

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetTypeConverter.java
```diff
@@ -240,9 +240,11 @@ public void addValueToJsonArray(Group group, int columnIndex, int repeatIndex, T
     LIST {
         @Override
         public DataType getDataType(Type type) {
-            Type elementType = getElementType(type);
+            Type elementType = getElementType(type.asGroupType());
```
ashuka24, line 243: why is the group type needed here now instead of just `type`

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/PgUtilities.java
```diff
@@ -210,15 +210,21 @@ public String escapeArrayElement(String str) {
...
+     *
      * @param data a byte buffer to convert to bytea hex format
      * @return a string containing Postgres bytea hex format
      */
     public String encodeAndEscapeByteaHex(ByteBuffer data) {
-        return escapeArrayElement(String.format("\\x%s", Hex.encodeHexString(data)));
+        return escapeArrayElement(encodeByteaHex(data));
+    }
+
+    public String encodeByteaHex(ByteBuffer data) {
```
ashuka24, line 221: can we add javadocs here

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/PgArrayBuilder.java
```diff
@@ -30,8 +30,21 @@ public void addElement(String elem) {
...
      *
+     * @param elem the element to be added to the array
+     * @param needsEscapingInArray whether the element need to be escaped before being added to the array
+     */
+    public void addElement(String elem, boolean needsEscapingInArray) {
+        if(needsEscapingInArray){
+            addElement(elem);
+        }else{
+            addElementNoEscaping(elem);
+        }
```
denalex, line 43: formatting please

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/ParquetResolverTest.java
```diff
@@ -539,34 +538,34 @@ public void testGetFields_List() throws IOException {
...
-        assertList(fields, 5, "{1.11}");                                                                  // DataType.FLOAT4ARRAY
-        assertList(fields, 6, "{1.7E308}");                                                               // DataType.FLOAT8ARRAY
-        assertList(fields, 7, "{\"this is a test string\"}");                                             // DataType.TEXTARRAY
-        assertList(fields, 8, "{\"\\\\xdeadbeef\"}");                                                     // DataType.BYTEAARRAY
-        assertList(fields, 9, "{hello}");                                                                 // DataType.BPCHARARRAY, right trimmed spaces
-        assertList(fields, 10, "{hello}");                                                                // DataType.VARCHARARRAY
-        assertList(fields, 11, "{2022-10-07}");                                                           // DataType.DATEARRAY
-        assertList(fields, 12, "{" + BigDecimal.valueOf(1234560000000000000L, 18) + "}");// DataType.NUMERICARRAY
+        assertList(fields, 0, "1", DataType.INTEGER);
```
denalex, line 541: how can this be asserted as a list ? Should this be `assertField` ? How is it working ?

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/ParquetResolverTest.java
```diff
@@ -539,34 +538,34 @@ public void testGetFields_List() throws IOException {
...
-        assertList(fields, 6, "{1.0,-99.9}");                                                                         // DataType.FLOAT8ARRAY
-        assertList(fields, 7, "{\"this is a string with \\\"special\\\" characters\",\"this is a string without\"}"); // DataType.TEXTARRAY
-        assertList(fields, 8, "{\"\\\\xdeadbeef\",\"\\\\xadbeef\"}");                                                 // DataType.BYTEAARRAY
-        assertList(fields, 9, "{\"this is exactly\",\" fifteen chars.\"}");                                           // DataType.BPCHARARRAY
-        assertList(fields, 10, "{\"this is exactly\",\" fifteen chars.\"}");                                          // DataType.VARCHARARRAY
-        assertList(fields, 11, "{2022-10-07,2022-10-08}");                                                            // DataType.DATEARRAY
-        assertList(fields, 12, "{" + BigDecimal.valueOf(1234560000000000000L, 18) + "," + BigDecimal.valueOf(1234560000000000000L, 18) + "}");  // DataType.NUMERICARRAY
+        assertList(fields, 0, "2", DataType.INTEGER);
```
denalex, line 556: same here

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetTypeConverter.java
yimingli-vmware: `getElementType` is only called by `LIST` case, so the type here should always be `GroupType`. By passing in a `GroupType`, we don't need to do type casting `type.asGroupType()` inside `getElementType`. That makes the code in `getElementType` a little bit shorter and cleaner

automation/src/test/resources/data/parquet/generate_parquet_types.hql
```diff
@@ -1,9 +1,6 @@
-!connect jdbc:hive2://localhost:10000/
-
-DROP TABLE IF EXISTS parquet_types_csv;
 DROP TABLE IF EXISTS parquet_types_csv;
 
-CREATE EXTERNAL TABLE IF NOT EXISTS parquet_types_csv (
```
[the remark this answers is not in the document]
yimingli-vmware: When we want manually execute the `CREATE EXTERNAL TABLE` query locally
- if we use `CREATE EXTERNAL TABLE IF NOT EXISTS` then `INSERT` data into table, we will have duplicated rows.
- if we use `CREATE EXTERNAL TABLE`, it will forces us to do the `DROP` first then create an empty one.

server/pxf-api/src/test/java/org/greenplum/pxf/api/io/DataTypeTest.java
```diff
@@ -0,0 +1,56 @@
...
+                if (needsEscaping(dataType.getOID())) {
+                    assertTrue(dataType.getTypeElem().getNeedsEscapingInArray());
+                } else {
+                    assertFalse(dataType.getTypeElem().getNeedsEscapingInArray());
+                }
+                assertTrue(dataType.getNeedsEscapingInArray());
+            }
```
bradfordb-vmware, line 45: I think this `if`-block skips over non-array types; if `dataType.isArrayType()` returns false, then no assertion is checked. Is that intended?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetTypeConverter.java
```diff
@@ -219,28 +235,82 @@ public Object getValue(Group group, int columnIndex, int repeatIndex, Type type)
...
+     *
+     * @param type is the input parquet type
+     * @return a {@link ParquetTypeConverter} containing the ENUM value
+     */
+    public static ParquetTypeConverter from(Type type) {
+        if (type.isPrimitive()) {
+            if (type.asPrimitiveType().getPrimitiveTypeName() == null) {
+                throw new PxfRuntimeException("Invalid Parquet Primitive schema.");
+            }
+            return valueOf(type.asPrimitiveType().getPrimitiveTypeName().name());
```
bradfordb-vmware, line 304:
```suggestion
        if (type.isPrimitive()) {
            PrimitiveType.PrimitiveTypeName primitiveTypeName = type.asPrimitiveType().getPrimitiveTypeName();
            if (primitiveTypeName == null) {
                throw new PxfRuntimeException("Invalid Parquet Primitive schema.");
            }
            return valueOf(primitiveTypeName.name());
        }
```

server/pxf-api/src/test/java/org/greenplum/pxf/api/io/DataTypeTest.java
yimingli-vmware: I think we only use `needsEscapingInArray` when the coming data is an array enum. So I only do assertions on array enums and their element enum, not doing assertions directly on element enum. I've added another `else if` case for `UNSUPPORTED_TYPE`.

server/pxf-api/src/test/java/org/greenplum/pxf/api/io/DataTypeTest.java
denalex: then maybe rename `needsEscaping` to smth like `needsEscapingForElementType` since the array types themselves do not need escaping (even though we are setting their flags to true)

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetTypeConverter.java
```diff
@@ -219,28 +235,86 @@ public Object getValue(Group group, int columnIndex, int repeatIndex, Type type)
...
+        public Object getValue(Group group, int columnIndex, int repeatIndex, Type type) {
+            PgArrayBuilder pgArrayBuilder = new PgArrayBuilder(pgUtilities);
+            pgArrayBuilder.startArray();
+
+            Group listGroup = group.getGroup(columnIndex, repeatIndex);
+            // a listGroup can have any number of repeatedGroups
+            int repetitionCount = listGroup.getFieldRepetitionCount(0);
+            boolean elementNeedsEscapingInArray = getDataType(type).getTypeElem().getNeedsEscapingInArray();
+            PrimitiveType elementType = getElementType(type.asGroupType()).asPrimitiveType();
+            ParquetTypeConverter converter = from(elementType);
```
denalex, line 263: it is a bit easier to reason about this logic if I rewrite it as such (avoiding multiple lookups and deriving it from elementType):
```suggestion
            PrimitiveType elementType = getElementType(type.asGroupType()).asPrimitiveType();
            ParquetTypeConverter elementConverter = from(elementType);
            boolean elementNeedsEscapingInArray = elementConverter.getDataType(elementType).getNeedsEscapingInArray();
```
Will this still be equivalent and correct ?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetTypeConverter.java
```diff
@@ -219,28 +235,86 @@ public Object getValue(Group group, int columnIndex, int repeatIndex, Type type)
...
+        @Override
+        public Object getValue(Group group, int columnIndex, int repeatIndex, Type type) {
+            PgArrayBuilder pgArrayBuilder = new PgArrayBuilder(pgUtilities);
+            pgArrayBuilder.startArray();
+
+            Group listGroup = group.getGroup(columnIndex, repeatIndex);
+            // a listGroup can have any number of repeatedGroups
+            int repetitionCount = listGroup.getFieldRepetitionCount(0);
+            boolean elementNeedsEscapingInArray = getDataType(type).getTypeElem().getNeedsEscapingInArray();
+            PrimitiveType elementType = getElementType(type.asGroupType()).asPrimitiveType();
```
denalex, line 262: what if we have List of Lists in Parquet which we do not support here, but would `getElementType(type.asGroupType())` be non-primitive ? Do we have a test case that will prove we will not get NPE and rather a proper Exception ?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetTypeConverter.java
```diff
@@ -309,4 +383,58 @@ private static Binary getBinaryFromZonedDateTime(String timestampString, ZonedDa
...
+
+    public abstract void addValueToJsonArray(Group group, int columnIndex, int repeatIndex, Type type, ArrayNode jsonNode);
+
+    /**
+     * Get DataType based on the element type of Parquet LIST
+     *
+     * @param type is the element type of the Parquet LIST
+     * @return the array DataType
+     */
+    public DataType getArrayDataType(Type type) {
```
denalex, line 423: this function is used only from one place and is not applicable for complex types, it will be better to remove this function from the public API and just inline the logic to where it is called from

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetTypeConverter.java
yimingli-vmware: Our code will throw an exception far before we reach this step. I have a unit test called `testGetFields_Unsupported_Complex` in `ParquetResolverTest`. The exception will be thrown when we try to get columnDescriptors based on the input schema.  These unsupported non-primitive list schema will be taken care of in my `ParquetTypeConverter.from`. In my unit test, I now also add a `LIST of LIST` situation.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetTypeConverter.java
```diff
@@ -309,4 +385,61 @@ private static Binary getBinaryFromZonedDateTime(String timestampString, ZonedDa
...
+    public String getValueFromList(Group group, int columnIndex, int repeatIndex, PrimitiveType primitiveType) {
+        return String.valueOf(getValue(group, columnIndex, repeatIndex, primitiveType));
+    }
+
+    /**
+     * Validate whether the element type in Parquet List type is supported by pxf
+     *
+     * @param elementType the element type of Parquet List type
+     */
+    public void validateElementTypeInListType(Type elementType) {
```
denalex, line 437: this should be private function outside of "public interface" block.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetTypeConverter.java
```diff
@@ -309,4 +385,61 @@ private static Binary getBinaryFromZonedDateTime(String timestampString, ZonedDa
...
+        if (!elementType.isPrimitive()) {
+            String complexTypeName = elementType.asGroupType().getOriginalType() == null ?
+                    "customized struct" :
+                    elementType.asGroupType().getOriginalType().name();
```
denalex, line 441: these 2 lines repeat elsewhere, can we extract the logic into `getComplexTypeName` helper private function ?

denalex (approved):

PR #896, opened 2022-11-09
Minor fixes for IPA dev cluster provision script and templates

denalex:
- updated terraform template and added designation of some values as sensitive to make it work with the latest versions of terraform
- added a flag to not error out when cleanup tries to remove files that might not exist during the first run

### DIFF
--- a/concourse/terraform/ipa-multinode-hadoop/outputs.tf
+++ b/concourse/terraform/ipa-multinode-hadoop/outputs.tf
@@ -5,14 +5,16 @@ output "ssh_config" {
     namenode = google_compute_instance.namenode
     datanode = google_compute_instance.datanode
   })
-
+  sensitive = true
 }
+
 output "ansible_inventory" {
   value = templatefile("${path.module}/templates/inventory.ini.tpl", {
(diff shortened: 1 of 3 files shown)


hpandeycodeit (approved):

concourse/terraform/ipa-multinode-hadoop/outputs.tf
```diff
@@ -5,14 +5,16 @@ output "ssh_config" {
     namenode = google_compute_instance.namenode
     datanode = google_compute_instance.datanode
   })
-
+  sensitive = true
```
ashuka24, line 8: why is this needed?

ashuka24 (approved):

concourse/terraform/ipa-multinode-hadoop/outputs.tf
denalex: without this the newer versions of terraform fail, they want us to explicitly declare the intention of saving a sensitive information to local disk to prevent accidental developer errors.

edespino:
Closing this PR as it was incorporated into PR #899.

PR #897, opened 2022-11-11
Add close connection if statementRead is null

RomaZe:
Sometimes the connection to the external data source is still open and doesn't close.
One of the examples is when the user runs named query with readable external table. If the file with query is not present on the segment, the query finishes with error but the session will not be closed.
Here is a simple logic of the issue:
1. The user runs query to the external readable table with named query;
2. The method `openForRead()` opens the connection to the external data source;
3. The method `getQueryText()` tries to read query from the file and receives the error that file doesn't exist. The field `statementRead` is still `null`
4. The method `closeForRead()` executes the `closeStatementAndConnection(statementRead)` method that checks whether the `statementRead` is `null` or not. If it is `null` we don't close neither statement nor connection.
5. The connection is still open in that case until the idle timeout.

This commit adds try-catch block to catch the error if some exceptions happen before the statement execution. The connection will be closed in case of the error.

### DIFF
--- a/server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcAccessor.java
+++ b/server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcAccessor.java
@@ -22,7 +22,6 @@ package org.greenplum.pxf.plugins.jdbc;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang.StringUtils;
 import org.greenplum.pxf.api.OneRow;
-import org.greenplum.pxf.api.error.PxfRuntimeException;
 import org.greenplum.pxf.api.model.Accessor;
 import org.greenplum.pxf.api.model.ConfigurationFactory;
 import org.greenplum.pxf.api.security.SecureLogin;
@@ -36,12 +35,7 @@ import org.slf4j.LoggerFactory;
 import java.io.File;
 import java.io.IOException;
 import java.nio.charset.Charset;
-import java.sql.Connection;
-import java.sql.PreparedStatement;
-import java.sql.ResultSet;
-import java.sql.SQLException;
-import java.sql.SQLTimeoutException;
-import java.sql.Statement;
+import java.sql.*;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.concurrent.ExecutorService;
@@ -101,8 +95,18 @@ public class JdbcAccessor extends JdbcBasePlugin implements Accessor {
         if (statementRead != null && !statementRead.isClosed()) {
             return true;
         }
-
         Connection connection = super.getConnection();
+        try {
+            return openForReadInner(connection);
+        } catch (Throwable e) {
+            if (statementRead == null) {
+                JdbcBasePlugin.closeConnection(connection);
+            }
+            throw new RuntimeException(e.getMessage(), e);
+        }
+    }
+
+    private boolean openForReadInner(Connection connection) throws SQLException {
         SQLQueryBuilder sqlQueryBuilder = new SQLQueryBuilder(context, connection.getMetaData(), getQueryText());
 
         // Build SELECT query
@@ -225,7 +229,7 @@ public class JdbcAccessor extends JdbcBasePlugin implements Accessor {
         return true;
     }
 
-     /**
+    /**
      * writeNextObject() implementation
      * <p>
      * If batchSize is not 0 or 1, add a tuple to the batch of statementWrite
--- a/server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcAccessor.java
+++ b/server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcAccessor.java
@@ -22,6 +22,7 @@ package org.greenplum.pxf.plugins.jdbc;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang.StringUtils;
 import org.greenplum.pxf.api.OneRow;
+import org.greenplum.pxf.api.error.PxfRuntimeException;
 import org.greenplum.pxf.api.model.Accessor;
 import org.greenplum.pxf.api.model.ConfigurationFactory;
 import org.greenplum.pxf.api.security.SecureLogin;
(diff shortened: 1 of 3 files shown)


server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcAccessor.java
```diff
@@ -36,12 +35,7 @@
 import java.io.File;
 import java.io.IOException;
 import java.nio.charset.Charset;
-import java.sql.Connection;
-import java.sql.PreparedStatement;
-import java.sql.ResultSet;
-import java.sql.SQLException;
-import java.sql.SQLTimeoutException;
-import java.sql.Statement;
+import java.sql.*;
```
denalex, line 38: please restore previous imports, our style is to always use explicit imports.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcAccessor.java
```diff
@@ -101,8 +95,18 @@ public boolean openForRead() throws SQLException, SQLTimeoutException {
         if (statementRead != null && !statementRead.isClosed()) {
             return true;
         }
-
         Connection connection = super.getConnection();
+        try {
+            return openForReadInner(connection);
+        } catch (Throwable e) {
+            if (statementRead == null) {
+                JdbcBasePlugin.closeConnection(connection);
```
denalex, line 103:
```suggestion
                closeConnection(connection);

```
this method is inherited, so it is not necessary to refer to JdbcBasePlugin to invoke it

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcAccessor.java
```diff
@@ -101,8 +95,18 @@ public boolean openForRead() throws SQLException, SQLTimeoutException {
...
         }
-
         Connection connection = super.getConnection();
+        try {
+            return openForReadInner(connection);
+        } catch (Throwable e) {
+            if (statementRead == null) {
+                JdbcBasePlugin.closeConnection(connection);
+            }
+            throw new RuntimeException(e.getMessage(), e);
```
denalex, line 105:
```suggestion
            throw new PxfRuntimeException(e.getMessage(), e);

```

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
```diff
@@ -479,7 +479,7 @@ private Connection getConnectionInternal() throws Exception {
      * @param connection connection to close
      * @throws SQLException throws when a SQLException occurs
      */
-    private static void closeConnection(Connection connection) throws SQLException {
+    static void closeConnection(Connection connection) throws SQLException {
```
denalex, line 482:
```suggestion
    protected static void closeConnection(Connection connection) throws SQLException {

```
to allow it to be called by subclasses only

server/pxf-jdbc/src/test/java/org/greenplum/pxf/plugins/jdbc/JdbcAccessorTest.java
```diff
@@ -82,7 +82,7 @@ public void testReadFromQueryFailsWhenServerDirectoryIsNotSpecified() throws SQL
         context.setDataSource("query:foo");
         accessor.setRequestContext(context);
         accessor.afterPropertiesSet();
-        Exception e = assertThrows(IllegalStateException.class,
+        Exception e = assertThrows(RuntimeException.class,
```
denalex, line 85:
```suggestion
        Exception e = assertThrows(PxfRuntimeException.class,
```

denalex, line 85: It also would be great to verify that method `closeConnection` has been called on the accessor

denalex (commented):
Thank you for finding the issue and submitting the PR, I added a few minor recommendations.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcAccessor.java
RomaZe: Restored explicit imports

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcAccessor.java
RomaZe: Removed redundant reference to the class

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcAccessor.java
RomaZe: Replaced

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
RomaZe: Changed class modifier

server/pxf-jdbc/src/test/java/org/greenplum/pxf/plugins/jdbc/JdbcAccessorTest.java
RomaZe: Replaced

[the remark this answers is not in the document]
RomaZe: The original was thrown if the configuration server directory was not found: [here](https://github.com/greenplum-db/pxf/blob/main/server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcAccessor.java#L341)
I think we should check it to be on a safe site.

denalex (approved):

ashuka24 (approved):

PR #899, opened 2022-11-15
Minor fixes for IPA dev cluster provision script and templates

edespino:
I have the following updates to allow `ipa-install-bash` to run in an Ubuntu 18.04 dev environment.

Here is an overview of the change:
-     SSH Config file - non-append approach
-     Terraform v1.3.4 - Sensitive = true
-     Removal of files errors (if they do not exist)
-     Portable Sed command alternative
-     Curl noise (-s)
-     For consistency use, expand TF_VAR_env_name use within the script.

### DIFF
--- a/dev/ipa-cluster.bash
+++ b/dev/ipa-cluster.bash
@@ -135,9 +135,9 @@ function terraform_destroy() {
 
 # cleanup ansible artifacts that might exist from previous runs
 function cleanup_existing_artifacts() {
-    rm "${ansible_play_path}"/*.p12
-    rm "${ansible_play_path}"/inventory.ini
-    rm "${ansible_play_path}"/config.yml
+    rm -f "${ansible_play_path}"/*.p12
+    rm -f "${ansible_play_path}"/inventory.ini
+    rm -f "${ansible_play_path}"/config.yml
 }
 
 function generate_keystores() {
--- a/dev/ipa-cluster.bash
+++ b/dev/ipa-cluster.bash
@@ -73,7 +73,7 @@ function check_pre_requisites() {
 }
 
 function create_firewall_resource() {
-    local my_ip=$(curl ifconfig.co)
+    local my_ip=$(curl -s ifconfig.co)
     cat << EOF > "${terraform_dir}"/local-firewall.tf
 resource "google_compute_firewall" "$USER-access" {
   name    = "$USER-access"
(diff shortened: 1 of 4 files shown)


dev/ipa-cluster.bash
```diff
@@ -164,11 +164,18 @@ function generate_keystores() {
...
-# setup ssh_config with host alias, users, and identity files
+# setup ssh ipa_config with host alias, users, and identity files
 function setup_ssh_config() {
     mkdir -p ~/.ssh
-    jq <"${metadata_path}" -r '.ssh_config.value' >>~/.ssh/config
+
+    [ ! -f ~/.ssh/config ] && touch ~/.ssh/config
+
+    # If necessary, update ~/.ssh/config file with Include directive
+    grep -qxF 'Include "ipa_config"' ~/.ssh/config || echo 'Include "ipa_config"' >> ~/.ssh/config
```
bradfordb-vmware, line 174: On my Debian systems, I've observed that `Include` doesn't work if it isn't at the top of the file; for example:

```
Include first_include

Host  foo
    HostName 127.0.0.1
    User foo

Include second_include
```

won't work. To get this to work, I have to do

```
Include first_include
Include second_include

Host  foo
    HostName 127.0.0.1
    User foo
```

Rather than try and pre-pend to the config file, the pattern I've been using is to take advantage of wildcards

```
Include config.d/*

Host foo
    HostName 127.0.0.1
    User foo
```

This way the script just creates `~/.ssh/config.d/ipa_config`; you can add a `grep` for `Include config.d/*` in `~/.ssh/config` as part of `check_pre_requisites`.

bradfordb-vmware (approved):
When I was running this `ipa-cluster.bash` script last week on my macOS machine, I was having issues with the following lines:

```shell
scp "${hadoop_namenode}:\$HADOOP_PREFIX/etc/hadoop/*-site.xml" ipa_env_files/conf/
scp "${hive_node}:\$HIVE_HOME/conf/hive-site.xml" ipa_env_files/conf/
```

For some reason `HADOOP_PREFIX` and `HIVE_HOME` were not being expanded on the remote side. I worked around this by replacing the variable references with the actual path. I don't know if this is an issue with my local environment setup or the the version of SSH being used on macOS Ventura. Just wanted to get your input on replacing the variable references.

denalex (approved):
Looks good, thank you for the improvements. A few points:
- you will need to update the `IPA.md` file to reflect changes to ssh_config
- would be good to update the title of the PR (that will become squash commit message) with the action word and remove GP UD (as it is not an OSS concept)
- I assume it was tested on a MacOS ?
- as this extends a previously opened PR (https://github.com/greenplum-db/pxf/pull/896) it would be good to add the appropriate comments and cherry-pick original commits for proper attribution.

dev/ipa-cluster.bash
edespino: @bradfordb-vmware I have updated the flow to incorporate your suggestion. The change now adds the `Include config.d/*` directive as the first line in `.ssh/config`. The IPA-specific ssh config file is added in `.ssh/config.d/ipa_config`.

edespino:
@denalex I have incorporated your changes:

- IPA.md updated to reflect ssh config changes
- I will update the PR title as suggested
- Tested on macOS Monterey
- cherry-picked PR #896

PR #901, opened 2022-11-22
Replace shell parameter expansions with Make functions

bradfordb-vmware:
The various Makefiles used to build PXF contain a number of shell parameter expansions that either work or cause Make to fail with a syntax error, depending on the version of bash and Make that are installed. This commit replaces these parameter expansions witl Make text functions that are more portable.

### DIFF
--- a/cli/go/src/pxf-cli/Makefile
+++ b/cli/go/src/pxf-cli/Makefile
@@ -3,7 +3,7 @@ include ../../../../common.mk
 .PHONY: all build stage install test clean help
 .DEFAULT_GOAL := build
 
-TARGET := $(shell echo $${PWD\#\#*/})
+TARGET := $(notdir $(CURDIR))
 PXF_ROOT_DIR := $(shell cd ../../../.. && pwd)
 PXF_SERVER_DIR := ${PXF_ROOT_DIR}/server
 PXF_CLI_DIR := ${PXF_ROOT_DIR}/cli
--- a/server/Makefile
+++ b/server/Makefile
@@ -48,11 +48,11 @@ compile:
 GRADLEW_TEST_PARAMS = test
 ifneq "$(TEST)" ""
 	# find test file, remove leading "./" and trailing ".java"
-	TEST_FILE = $(shell : "$$(find . -name $(TEST).java)"; : "$${_%.java}"; echo "$${_\#./}")
+	TEST_FILE = $(subst ./,,$(subst .java,,$(shell find . -name $(TEST).java)))
 	# parse out the gradle project by grabbing top-level dir name
-	PROJECT   = $(shell : '${TEST_FILE}'; echo "$${_%%/*}")
+	PROJECT = $(word 1,$(subst /, ,$(TEST_FILE)))
 	# get java-style path with dots, starting at "org"
-	TEST_PATH = $(shell : '${TEST_FILE}'; : "$${_\#${PROJECT}/src/test/java/}"; echo "$${_//\//.}")
+	TEST_PATH = $(subst /,.,$(subst $(PROJECT)/src/test/java/,,$(TEST_FILE)))
 	GRADLEW_TEST_PARAMS = :$(PROJECT):test --rerun-tasks --tests $(TEST_PATH)
 endif
 


cli/go/src/pxf-cli/Makefile
```diff
@@ -3,7 +3,7 @@ include ../../../../common.mk
 .PHONY: all build stage install test clean help
 .DEFAULT_GOAL := build
 
-TARGET := $(shell echo $${PWD\#\#*/})
+TARGET := $(notdir ${PWD})
```
edespino, line 6: I believe the lack of using the `shell` command causes an interesting side-effect. With the PR version, if the Makefile is referenced using the make `-C` option, the TARGET value will vary dependent on the directory in which the make command is invoked.

Below, I have provided three examples to consider. The first is the original version. the second is the PR version, and the third is an alternate version. The output below demonstrates 

```
======================================================================
Original version:

Notice how the value for TARGET is IDENTICAL in the following cases

- The make command is invoked from the same directory as the Makefile
- The make command is invoked from another directory while referencing
  the directory of the Makefile via a make -C option.
----------------------------------------------------------------------
TARGET := $(shell echo $${PWD\#\#*/})
======================================================================

(base) eespino:workspace/ $ pwd
/Users/eespino/workspace
(base) eespino:workspace/ $ cat Makefile
TARGET := $(shell echo $${PWD\#\#*/})

target_is:
        echo ${TARGET}
(base) eespino:workspace/ $ make target_is
echo workspace
workspace
(base) eespino:workspace/ $ cd ..
(base) eespino:~/ $ pwd
/Users/eespino
(base) eespino:~/ $ make -C workspace target_is
echo workspace
workspace
(base) eespino:~/ $





======================================================================
PR version:

Notice how the value for TARGET is DIFFERENT in the following cases

- The make command is invoked from the same directory as the Makefile
- The make command is invoked from another directory while referencing
  the directory of the Makefile via a make -C option.
----------------------------------------------------------------------
TARGET := $(notdir ${PWD})
======================================================================

(base) eespino:workspace/ $ pwd
/Users/eespino/workspace
(base) eespino:workspace/ $ cat Makefile
TARGET := $(notdir ${PWD})

target_is:
        echo ${TARGET}
(base) eespino:workspace/ $ make target_is
echo workspace
workspace
(base) eespino:workspace/ $ cd ..
(base) eespino:~/ $ pwd
/Users/eespino
(base) eespino:~/ $ make -C workspace target_is
echo eespino
eespino





======================================================================
Alternate version (to consider):

Notice how the value for TARGET is IDENTICAL in the following cases

- The make command is invoked from the same directory as the Makefile
- The make command is invoked from another directory while referencing
  the directory of the Makefile via a make -C option.
----------------------------------------------------------------------
TARGET := $(notdir $(shell pwd))
======================================================================

(base) eespino:workspace/ $ pwd
/Users/eespino/workspace
(base) eespino:workspace/ $ cat Makefile
TARGET := $(notdir $(shell pwd))

target_is:
        echo ${TARGET}
(base) eespino:workspace/ $ make target_is
echo workspace
workspace
(base) eespino:workspace/ $ cd ..
(base) eespino:~/ $ pwd
/Users/eespino
(base) eespino:~/ $ make -C workspace target_is
echo workspace
workspace
```

edespino (changes requested):

cli/go/src/pxf-cli/Makefile
bradfordb-vmware: Good catch with `${PWD}` not updating when invoking as `make -C <path>`. It looks like Make as a variable `$(CURDIR)` that it will update so I've updated the PR to use this.

cli/go/src/pxf-cli/Makefile
```diff
@@ -3,7 +3,7 @@ include ../../../../common.mk
 .PHONY: all build stage install test clean help
 .DEFAULT_GOAL := build
 
-TARGET := $(shell echo $${PWD\#\#*/})
+TARGET := $(notdir $(CURDIR))
```
edespino, line 6: Nice! I have verified CURDIR is working as expected. Thank you.
```
GNU Make 4.3
GNU Make 4.1
GNU Make 3.81
```

edespino (approved):
+1 - with CURDIR use.

denalex (approved):

PR #904, opened 2022-12-01
Add script for setting up a dataproc cluster for local testing

bradfordb-vmware:

### DIFF
--- a/dev/dataproc-cluster.bash
+++ b/dev/dataproc-cluster.bash
@@ -1,0 +1,260 @@
+#!/usr/bin/env bash
+
+set -o errexit
+# uncomment the following line for debugging
+# set -o xtrace
+
+# This script uses the gcloud CLI to create and delete GCP resources in your
+# default GCP project. If you want/need to create the resources in a specific
+# project, export CLOUDSDK_CORE_PROJECT before running this script.
+CLUSTER_USER="${CLUSTER_USER:-${USER}}"
+CLUSTER_NAME="${CLUSTER_NAME:-${CLUSTER_USER}-local-cluster}"
+REGION="${REGION:-us-west1}"
+NUM_WORKERS="${NUM_WORKERS:-2}"
+NETWORK="${NETWORK:-default}"
+# valid choices for this can be found
+# https://cloud.google.com/dataproc/docs/concepts/versioning/dataproc-version-clusters
+IMAGE_VERSION="${IMAGE_VERSION:-2.0-debian10}"
+
+function check_pre_requisites() {
+    if ! type gcloud &>/dev/null; then
+        >&2 echo 'gcloud is not found, did you install and configure it?'
+        >&2 echo 'see https://cloud.google.com/sdk/docs/install and https://cloud.google.com/sdk/docs/initializing'
+        exit 1
+    fi
+
+    if ! type xmlstarlet &>/dev/null; then
+        >&2 echo 'xmlstarlet is not found, did you install it (e.g. "brew install xmlstarlet") ?'
+        exit 1
+    fi
+
+    : "${PXF_HOME?"PXF_HOME is required. export PXF_HOME=[YOUR_PXF_INSTALL_LOCATION]"}"
+
+    if [[ -d dataproc_env_files ]]; then
+        >&2 echo "dataproc_env_files already exists; remove before re-running this script..."
+        exit 1
+    fi
+}
+
+function create_dataproc_cluster() {
+    local extra_hadoop_config="${1}"
+
+    if gcloud dataproc clusters describe "${CLUSTER_NAME}" --region="${REGION}" &>/dev/null; then
+        echo "Cluster ${CLUSTER_NAME} already exists; skipping create..."
+        return 0
+    fi
+
+    local -a create_cmd
+    create_cmd=(gcloud dataproc clusters create "${CLUSTER_NAME}"
+        --region="${REGION}"
+        --master-machine-type=n1-standard-4
+        --worker-machine-type=n1-standard-4
+        --tags="${CLUSTER_USER}-only"
+        --num-workers="${NUM_WORKERS}"
+        --image-version="${IMAGE_VERSION}"
+        --network="${NETWORK}")
+
+    if [[ -n $1 ]]; then
+        create_cmd+=("--properties=${extra_hadoop_config}")
+    fi
+
+    "${create_cmd[@]}"
+
+    echo "Cluster ${CLUSTER_NAME} has been created"
+}
+
+function delete_dataproc_cluster() {
+    if ! gcloud dataproc clusters describe "${CLUSTER_NAME}" --region "${REGION}" &>/dev/null; then
+        echo "Cluster ${CLUSTER_NAME} does not exist, skipping delete..."
+        return 0
+    fi
+
+    gcloud dataproc clusters delete "${CLUSTER_NAME}" --region="${REGION}" --quiet
+
+    echo "Cluster ${CLUSTER_NAME} has been deleted"
+}
+
+function create_firewall_rule() {
+    local firewall_rule_name="${1}"
+
+    if gcloud compute firewall-rules describe "${firewall_rule_name}" &>/dev/null; then
+        echo "Firewall rule ${firewall_rule_name} already exists; skipping create..."
+        return 0
+    fi
+
+    local local_external_ip
+    local_external_ip="$(curl -s https://ipinfo.io/ip)"
+
+    # Web UI                Port URL
+    # YARN ResourceManager  8088 http://${CLUSTER_NAME}-m:8088
+    # HDFS NameNode         9870 http://${CLUSTER_NAME}-m:9870
+    #
+    # https://cloud.google.com/dataproc/docs/concepts/accessing/cluster-web-interfaces
+    gcloud compute firewall-rules create "${firewall_rule_name}" \
+        --description="Allow incoming HDFS traffic from ${USER}'s home office" \
+        --network="${NETWORK}" \
+        --allow=tcp:8020,tcp:8088,tcp:9083,tcp:9866,tcp:9870,tcp:10000 \
+        --direction=INGRESS \
+        --target-tags="${CLUSTER_USER}-only" \
+        --source-ranges="${local_external_ip}/32"
+}
+
+function delete_firewall_rule() {
+    local firewall_rule_name="${1}"
+
+    if ! gcloud compute firewall-rules describe "${firewall_rule_name}" &>/dev/null; then
+        echo "Firewall rule ${firewall_rule_name} does not exist; skipping delete..."
+        return 0
+    fi
+
+    gcloud compute firewall-rules delete "${firewall_rule_name}" --quiet
+}
+
+function create_dataproc_env_files() {
+    local zoneUri
+    zoneUri="$(gcloud dataproc clusters describe "${CLUSTER_NAME}" --region="${REGION}" --format='get(config.gceClusterConfig.zoneUri)')"
+    # the zoneUri field may contain a
+    #
+    #   * full URL    (https://www.googleapis.com/compute/v1/projects/[projectId]/zones/[zone])
+    #   * partial URI (projects/[projectId]/zones/[zone])
+    #   * short name  ([zone])
+    #
+    # <https://cloud.google.com/dataproc/docs/reference/rest/v1/ClusterConfig#gceclusterconfig>
+    #
+    # we need the just the short name when running
+    #
+    #     gcloud compute instances describe
+    #
+    # use bash parameter expansion to remove the longest matching prefix
+    # (everything up to the last forward-slash)
+    local zone="${zoneUri##*/}"
+
+    mkdir -p dataproc_env_files/conf
+    printf "# BEGIN LOCAL DATAPROC SECTION\n" >dataproc_env_files/etc_hostfile
+    external_ip="$(gcloud compute instances describe "${CLUSTER_NAME}-m" --zone="${zone}" --format='get(networkInterfaces[0].accessConfigs[0].natIP)')"
+    printf "%s\t%s %s\n" "${external_ip}" "${CLUSTER_NAME}-m" "${CLUSTER_NAME}-m.c.data-gpdb-ud.internal" >>dataproc_env_files/etc_hostfile
+
+    for ((i = 0; i < NUM_WORKERS; i++)); do
+        instance_name="${CLUSTER_NAME}-w-${i}"
+        external_ip="$(gcloud compute instances describe "${instance_name}" --zone="${zone}" --format='get(networkInterfaces[0].accessConfigs[0].natIP)')"
+        printf "%s\t%s %s\n" "${external_ip}" "${instance_name}" "${instance_name}.c.data-gpdb-ud.internal" >>dataproc_env_files/etc_hostfile
+    done
+    printf "# END LOCAL DATAPROC SECTION\n" >>dataproc_env_files/etc_hostfile
+
+    gcloud compute scp \
+        --zone="${zone}" \
+        "${CLUSTER_NAME}-m:/etc/hadoop/conf/*-site.xml" \
+        "${CLUSTER_NAME}-m:/etc/hive/conf/*-site.xml" \
+        dataproc_env_files/conf/
+
+    # set Hadoop client to use hostnames for datanodes instead of IP addresses (which are internal in GCP network)
+    xmlstarlet ed --inplace --pf --append '/configuration/property[last()]' --type elem -n property -v "" \
+        --subnode '/configuration/property[last()]' --type elem -n name -v "dfs.client.use.datanode.hostname" \
+        --subnode '/configuration/property[last()]' --type elem -n value -v "true" dataproc_env_files/conf/hdfs-site.xml
+
+    cp "${PXF_HOME}/templates/pxf-site.xml" dataproc_env_files/conf
+    # set impersonation property to false for the PXF server
+    xmlstarlet ed --inplace --pf --update "/configuration/property[name = 'pxf.service.user.impersonation']/value" -v false dataproc_env_files/conf/pxf-site.xml
+
+    echo "Cluster config for ${CLUSTER_NAME} has been written in dataproc_env_files"
+}
+
+function delete_dataproc_env_files() {
+    rm -rf dataproc_env_files
+}
+
+function prompt_for_confirmation() {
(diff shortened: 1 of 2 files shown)


dev/dataproc-cluster.sh
```diff
@@ -0,0 +1,205 @@
...
+EOF
+}
+
+function print_usage() {
+    cat <<EOF
+NAME
+    dataproc-cluster.sh - manage a Google Cloud Dataproc cluster
+
+SYNOPSIS
+    dataproc-cluster.sh create [<optional-options-for-dataproc>]
```
denalex, line 167: the existing `ipa-cluster.bash` is similar in spirit and purpose. It would be good to have them both follow the same style for both APIs and coding. Without saying what is right or preferred, so far I see the following differences:
- extension `.sh` vs. `.bash`
- create/delete vs. create/destroy
- this script does not require `--` prefix for commands, the other one does
- lack of exit $? at the end

dev/dataproc-cluster.sh
```diff
@@ -0,0 +1,205 @@
...
+    fi
+
+    if ! type xmlstarlet &>/dev/null; then
+        echo 'xmlstarlet is not found, did you install it (e.g. "brew install xmlstarlet") ?'
+        exit 1
+    fi
+
+    : "${PXF_HOME?"PXF_HOME is required. export PXF_HOME=[YOUR_PXF_INSTALL_LOCATION]"}"
+
+    if [[ -d dataproc_env_files ]]; then
```
denalex, line 35: this will look for the files in the current working directory (in case the script is run from somewhere else), do we want to make sure they are only supposed to be located here in `/dev` ?

dev/dataproc-cluster.sh
```diff
@@ -0,0 +1,205 @@
...
+
+set -o errexit
+# uncomment the following line for debugging
+# set -o xtrace
+
+# BEGIN USER CONFIGURATION
+# This script uses the gcloud CLI to create and delete GCP resources in your
+# default GCP project. If you want/need to create the resources in a specific
+# project, export CLOUDSDK_CORE_PROJECT before running this script.
+cluster_name="${USER}-local-cluster"
```
denalex, line 11: should we allow to have the user to specify this before starting the script and take the current value as default ?

dev/dataproc-cluster.sh
```diff
@@ -0,0 +1,205 @@
...
+        echo "Cluster ${cluster_name} already exists; skipping create..."
+        return 0
+    fi
+
+    local -a create_cmd
+    create_cmd=(gcloud dataproc clusters create "${cluster_name}"
+        --region="${region}"
+        --master-machine-type=n1-standard-4
+        --worker-machine-type=n1-standard-4
+        --tags="${USER}-only"
```
denalex, line 52: does `-only` have any special meaning ?

dev/dataproc-cluster.sh
```diff
@@ -0,0 +1,205 @@
...
+    custom value for dfs.namenode.fs-limits.min-block-size.
+
+EOF
+
+}
+
+# --- main script logic ---
+
+script_command="$1"
+custom_dataproc_properties="$2"
```
denalex, line 186: if there are multiple properties, are they supposed to be comma separated ? Can we either give an example or document this in the description

dev/dataproc-cluster.sh
```diff
@@ -0,0 +1,205 @@
...
+    fi
+
+    gcloud compute firewall-rules delete "${firewall_rule_name}" --quiet
+}
+
+function create_dataproc_env_files() {
+    local zoneUri
+    zoneUri="$(gcloud dataproc clusters describe "${cluster_name}" --region us-west1 --format='get(config.gceClusterConfig.zoneUri)')"
+    local zone="${zoneUri##*/}"
+
```
denalex, line 108: can we describe what the logic below does and why ?

dev/dataproc-cluster.sh
ashuka24: When we create the new firewall rule, we want it to only apply to the cluster that was created. So we tag the cluster with `{USER}-only` and the firewall rule with `{USER}-only` so that only that cluster will have access to your local development and vice versa.

dev/dataproc-cluster.sh
ashuka24: You mean something like 
`CLUSTER_NAME=${CLUSTER_NAME:="${USER}-local-cluster"}?

dev/dataproc-cluster.sh
bradfordb-vmware: There is nothing special about the suffix `-only` in the tag. The tag could be anything as long as it matches the tag we use when creating the firewall rule. The firewall rule is setup to _only_ allow access from the (external) IP address of machine running the script.

dev/dataproc-cluster.sh
bradfordb-vmware: The variable name here (`custom_dataproc_properties`) was chosen because at the time we wanted to provide a custom argument of the form `--properties=[PREFIX:PROPERTY=VALUE,...]` to the `gcloud dataproc clusters create` command. The way it's implemented is that we take this value as-is and append it to the command. This means that its not just limited to `--properties=...` but could be any valid option to `gcloud dataproc clusters create` (e.g., adding an expiration time via `--expiration-time=...`). However, this only allows you to add **one** extra option -- you couldn't do both `--expiration-time` and `--properties`).

With that context, I wonder if we should rename this to something like `extra_hadoop_config` and form the `--properties` option like

```sh
if [[ -n ${extra_hadoop_config} ]]; then
    create+=("--properties=${extra_hadoop_config}")
```

This would save the caller from having to type `--properties`. For documenting the syntax, we could then refer to the man page for `gcloud dataproc clusters create`.

dev/dataproc-cluster.sh
bradfordb-vmware: I don't think we need to restrict `dataproc_env_files` to be in `pxf_src/dev`; creating files inside the working directory is pretty standard.

dev/dataproc-cluster.sh
ashuka24: When we create the dataproc cluster, we provide the region, but allow GCP to assign a zone for resource management purposes (called "autozone" https://cloud.google.com/dataproc/docs/concepts/configuring-clusters/auto-zone). However, zone is more specific and better suited when trying to pull information from the cluster (ex: external IP addresses).

dev/dataproc-cluster.sh
bradfordb-vmware: We could do that, but it raises (in my mind at least) the question why not do that for all the other options in the user configuration section? If the user is using this script to create two running clusters, I'd assume they were changing at least one other option (num workers, image_version, etc) and editing this file anyways.

dev/dataproc-cluster.sh
bradfordb-vmware: I've update the option and script usage to better describe how to use this optional argument.

edespino (changes requested):
I have a few suggested changes:

1: Similar to other recommendations, allow overriding the hardcoded values by passing in a corresponding environment variable. If I run two clusters, I generally run from two separate directories where different environments will be activated.
2. At times, I stop the dataproc cluster. External IPs can change when the dataproc cluster is restared . Introduce an additional option (suggested: update_env_files) that will allow the regeneration of the config files. Change append to overwrite for dataproc_env_files/etc_hostfile.
3. Copy the hive-site.xml file (required when using hive services). FYI: With these changes, I could use hive table data.
4. Add port 8088 to TCP firewall rule so the user can access YARN ResourceManager Web UI
5. Add port 9083 to TCP firewall rule to access Hive service

Here are my recommended changes:

```
diff --git a/dev/dataproc-cluster.sh b/dev/dataproc-cluster.sh
index d0e9218a..f8f22284 100755
--- a/dev/dataproc-cluster.sh
+++ b/dev/dataproc-cluster.sh
@@ -8,15 +8,16 @@ set -o errexit
 # This script uses the gcloud CLI to create and delete GCP resources in your
 # default GCP project. If you want/need to create the resources in a specific
 # project, export CLOUDSDK_CORE_PROJECT before running this script.
-cluster_name="${USER}-local-cluster"
-region="us-west1"
-num_workers=2
+cluster_user="${cluster_user:-${USER}}"
+cluster_name="${cluster_name:-${cluster_user}-local-cluster}"
+region="${region:-us-west1}"
+num_workers=${num_workers:-2}
 network="default"

 # valid choices for this can be found
 # https://cloud.google.com/dataproc/docs/concepts/versioning/dataproc-version-clusters
 # TODO: re-run with 1.5-debian10 to test Hadoop 2.10
-image_version="2.0-debian10"
+image_version="${image_version:-2.0-debian10}"
 # END USER CONFIGURATION

 function check_pre_requisites() {
@@ -51,7 +52,7 @@ function create_dataproc_cluster() {
         --region="${region}"
         --master-machine-type=n1-standard-4
         --worker-machine-type=n1-standard-4
-        --tags="${USER}-only"
+        --tags="${cluster_user}-only"
         --num-workers="${num_workers}"
         --image-version="${image_version}"
         --network="${network}")
@@ -84,11 +85,11 @@ function create_firewall_rule() {
     local_external_ip="$(curl -s https://ipinfo.io/ip)"

     gcloud compute firewall-rules create "${firewall_rule_name}" \
-        --description="Allow incoming HDFS traffic from ${USER}'s home office" \
+        --description="Allow incoming HDFS traffic from ${cluster_user}'s home office" \
         --network="${network}" \
-        --allow=tcp:8020,tcp:9866,tcp:9870 \
+        --allow=tcp:8020,tcp:8088,tcp:9083,tcp:9866,tcp:9870 \
         --direction=INGRESS \
-        --target-tags="${USER}-only" \
+        --target-tags="${cluster_user}-only" \
         --source-ranges="${local_external_ip}/32"
 }

@@ -123,7 +124,7 @@ function create_dataproc_env_files() {
     local zone="${zoneUri##*/}"

     mkdir -p dataproc_env_files/conf
-    printf "# BEGIN LOCAL DATAPROC SECTION\n" >>dataproc_env_files/etc_hostfile
+    printf "# BEGIN LOCAL DATAPROC SECTION\n" >dataproc_env_files/etc_hostfile
     external_ip="$(gcloud compute instances describe "${cluster_name}-m" --zone="${zone}" --format='get(networkInterfaces[0].accessConfigs[0].natIP)')"
     printf "%s\t%s %s\n" "${external_ip}" "${cluster_name}-m" "${cluster_name}-m.c.data-gpdb-ud.internal" >>dataproc_env_files/etc_hostfile

@@ -135,7 +136,11 @@ function create_dataproc_env_files() {
     printf "# END LOCAL DATAPROC SECTION\n" >>dataproc_env_files/etc_hostfile

     # TODO: copy Hive config and setup
-    gcloud compute scp --zone="${zone}" "${cluster_name}-m:/etc/hadoop/conf/*-site.xml" dataproc_env_files/conf/
+    gcloud compute scp \
+           --zone="${zone}" \
+           "${cluster_name}-m:/etc/hadoop/conf/*-site.xml" \
+           "${cluster_name}-m:/etc/hive/conf/*-site.xml" \
+           dataproc_env_files/conf/
     # set Hadoop client to use hostnames for datanodes instead of IP addresses (which are internal in GCP network)
     xmlstarlet ed --inplace --pf --append '/configuration/property[last()]' --type elem -n property -v "" \
         --subnode '/configuration/property[last()]' --type elem -n name -v "dfs.client.use.datanode.hostname" \
@@ -217,6 +222,10 @@ case "${script_command}" in
     delete_dataproc_cluster
     print_user_instructions_for_delete
     ;;
+'update_env_files')
+    create_dataproc_env_files
+    print_user_instructions_for_create
+    ;;
 *)
     print_usage
     exit 2
```

dev/dataproc-cluster.sh
```diff
@@ -0,0 +1,224 @@
...
+extra_hadoop_config="$2"
+case "${script_command}" in
+'create')
+    check_pre_requisites
+    create_dataproc_cluster "${extra_hadoop_config}"
+    create_firewall_rule "${cluster_name}-external-access"
+    create_dataproc_env_files
+    print_user_instructions_for_create
+    ;;
+'delete')
```
edespino, line 214: Please consider adding a delete operation user acknowledgment. You can use a separate script variable (which can be overridden) to put the script into non-interactive mode.

dev/dataproc-cluster.sh
bradfordb-vmware: I have renamed the script to use the `.bash` extension and updated the commands to use `--create` and `--destroy`. We don't need the `exit $?` at the end of the script for two reasons: 1) we have `set -o errexit` at the top of the script which causes the script to exit immediately if any command exits with a non-zero status, and 2) In bash, the exit code of the script is implicitly the exit code of the last run command.

dev/dataproc-cluster.sh
bradfordb-vmware: Per feedback from @edespino, `CLUSTER_USER`, `CLUSTER_NAME`, `REGION`, `NUM_WORKERS`, `NETWORK`, and `IMAGE_VERSION` are now all configurable via environment variables.

dev/dataproc-cluster.sh
bradfordb-vmware: Per feedback from @edespino

> If I run two clusters, I generally run from two separate directories where different environments will be activated.

Creating the `dataproc_env_files` in the current working directory (instead of `$PXF_SRC/dev`) allows for creating two different clusters at the same time.

dev/dataproc-cluster.bash
```diff
@@ -0,0 +1,247 @@
...
+'--destroy')
+    [[ -n ${NON_INTERACTIVE} ]] || prompt_for_confirmation
+    delete_dataproc_env_files
+    delete_firewall_rule "${CLUSTER_NAME}-external-access"
+    delete_dataproc_cluster
+    print_user_instructions_for_delete
+    ;;
+'---update_env_files')
+    create_dataproc_env_files
+    print_user_instructions_for_create
```
ashuka24, line 241: this is misleading for update as the print instructions for create include the sentence "Cluster ${CLUSTER_NAME} has been created, now do the following:"

the first sentence in the case of update should perhaps reflect what has been updated? Maybe something like:
"Cluster XX has been updated to include YY"

bradfordb-vmware: I've reworded the output so that `Cluster XX has been created` is printed by the function that creates the cluster. Now when you run `--update_env_files` it should print:

```text
Cluster config for ${CLUSTER_NAME} has been written in dataproc_env_files
Now do the following:
```

edespino (approved):
Aside from the one change "tcp:9083", nice work. And THANKS!

dev/dataproc-cluster.bash
```diff
@@ -0,0 +1,247 @@
...
+
+    # Web UI                Port URL
+    # YARN ResourceManager  8088 http://${CLUSTER_NAME}-m:8088
+    # HDFS NameNode         9870 http://${CLUSTER_NAME}-m:9870
+    #
+    # https://cloud.google.com/dataproc/docs/concepts/accessing/cluster-web-interfaces
+    gcloud compute firewall-rules create "${firewall_rule_name}" \
+        --description="Allow incoming HDFS traffic from ${USER}'s home office" \
+        --NETWORK="${NETWORK}" \
+        --allow=tcp:8020,tcp:8088,tcp:9866,tcp:9870 \
```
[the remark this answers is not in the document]
bradfordb-vmware: I've pushed a fixup commit with this port added to the firewall rule. I've also include port 10000 that is needed for JDBC connections to Hive (e.g., for use with the Hive CLI included in singlecluster).

dev/dataproc-cluster.bash
```diff
@@ -0,0 +1,259 @@
...
+    print_user_instructions_for_create
+    ;;
+'--destroy')
+    [[ -n ${NON_INTERACTIVE} ]] || prompt_for_confirmation
+    delete_dataproc_env_files
+    delete_firewall_rule "${CLUSTER_NAME}-external-access"
+    delete_dataproc_cluster
+    print_user_instructions_for_delete
+    ;;
+'---update_env_files')
```
edespino, line 251: @bradfordb-vmware - you have introduced an extra dash in the new option.  `---update_env_files` --> `--update_env_files`

dev/dataproc-cluster.bash
```diff
@@ -158,31 +164,37 @@ function delete_dataproc_env_files() {
...
     while true; do
-        read -p "Destroy Dataproc cluster and firewall rule (y/n)? " yn
+        read -r -p "Destroy Dataproc cluster and firewall rule (y/n)? " yn
         case $yn in
-        [Yy]* ) ;;
-        [Nn]* ) exit ;;
-        * ) echo "Please answer yes or no.";;
+        [Yy]*) break ;;
+        [Nn]*) exit ;;
+        *) echo "Please answer yes or no." ;;
```
ashuka24, line 171: This prompt makes it seem like we need to type "yes" or "no", when we should be using "y" or "n", can we make that more clear?

dev/dataproc-cluster.bash
```diff
@@ -0,0 +1,259 @@
...
+function create_dataproc_cluster() {
+    local extra_hadoop_config="${1}"
+
+    if gcloud dataproc clusters describe "${CLUSTER_NAME}" --region="${REGION}" &>/dev/null; then
+        echo "Cluster ${CLUSTER_NAME} already exists; skipping create..."
+        return 0
+    fi
+
+    local -a create_cmd
+    create_cmd=(gcloud dataproc clusters create "${CLUSTER_NAME}"
```
ashuka24, line 47: Something I noticed is that I cannot just run this without any setup.  This command assumes 2 things: 
1. The user is already logged in to gcloud via `gcloud auth login`
2. The project id has been set via `gcloud config set project PROJECT_ID`
Should we add these 2 checks in to the pre-reqs as well besides just having gcloud installed?

denalex (approved):

hpandeycodeit (approved):

dev/dataproc-cluster.bash
bradfordb-vmware: In bash, `case` matches against patterns so this is seeing if the user input something that begins with `Y` or `y` so that `Y`, `y`, `YES`, `yes`, `Yes`, ... etc. will work.

dev/dataproc-cluster.bash
bradfordb-vmware: I've updated the pre-reqs check to instruct users to install and configure gcloud.

PR #913, opened 2022-12-21
ignore parquetWriteListReadWithHive test

yimingli-vmware:
For this test, we use HIVE to read out Parquet Data we write using pxf. This test can pass on HDP2 and HDP3, but it fails on CDH. CDH is using HIVE1.1, which doesn't support Parquet Date. Parquet Date is supported by Hive1.2+. So for now, we decide to temporarily ignore this test.


hpandeycodeit (approved):
nit: unused import on line 14

denalex (approved):

PR #914, opened 2022-12-21
Bump version to 6.5.0

denalex:


Aayushs4 (approved):

lisakowen (approved):
looks good to me.

hpandeycodeit (approved):

.github/workflows/create-release-on-tag.yml
```diff
@@ -22,14 +22,25 @@ jobs:
...
           release_name: PXF Version ${{ github.ref }}
           body: |
-            ## 6.4.2 (09/19/2022)
+            ## 6.5.0 (12/21/2022)
 
             ### Enhancements:
 
+            - [#872](https://github.com/greenplum-db/pxf/pull/872) Enable automation tests to run against PXF FDW extension
+            - [#876](https://github.com/greenplum-db/pxf/pull/876) Parquet write array
+            - [#877](https://github.com/greenplum-db/pxf/pull/877) Update shell scripts to use a more consistent style
```
ashuka24, line 31: i don't think this one needs to be here? as it's not a product enhancement but more a developer enhancement.

.github/workflows/create-release-on-tag.yml
```diff
@@ -22,14 +22,25 @@ jobs:
...
+            - [#879](https://github.com/greenplum-db/pxf/pull/879) Add new SPLIT_BY_FILE option for json profiles
```
ashuka24, line 32: Are we matching commit names here? if so this should be
```suggestion
            - [#879](https://github.com/greenplum-db/pxf/pull/879) Refactor JsonRecordReader and fix incorrect parsing of JSON objects for multi-line JSON

File: .github/workflows/create-release-on-tag.yml [hunk 3]
```diff
@@ -22,14 +22,25 @@ jobs:
...
+            - [#885](https://github.com/greenplum-db/pxf/pull/885) Prevent out-of-bounds buffer access
+            - [#886](https://github.com/greenplum-db/pxf/pull/886) Update jackson-databind to 2.13.4.1
+            - [#895](https://github.com/greenplum-db/pxf/pull/895) Parquet read array
+            - [#903](https://github.com/greenplum-db/pxf/pull/903) Bump postgresql from 42.4.1 to 42.4.3
 
             ### Bug Fixes:
 
-            - [#870](https://github.com/greenplum-db/pxf/pull/870) Relax the requirement that C string length matches Java string length
+            - [#875](https://github.com/greenplum-db/pxf/pull/875) Fix projection of boolean qualifiers whose attrs are not in SELECT list
+            - [#879](https://github.com/greenplum-db/pxf/pull/879) Refactor JsonRecordReader and fix incorrect parsing of JSON objects for multi-line JSON
```
ashuka24, line 42: I see that this is listed in both enhancement and bug fixes... is that intentional?

.github/workflows/create-release-on-tag.yml
denalex: I explicitly changed from PR name here to commit name as the same PR fixed a bug and introduced a new enhancement, so for this section the enhancement only was relevant

.github/workflows/create-release-on-tag.yml
denalex: yes, intentional as the same PR fixed a bug and introduced a new URL option

PR #916, opened 2022-12-22
docs - pxf supports writing parquet lists

lisakowen:
docs for https://github.com/greenplum-db/pxf/pull/876.

update the data type mapping for parquet write operations.  enhance the description of the SCHEMA write option (more work to do here).  also add a topic about schemas.  i pulled the schema content from the avro topic, let me know if parquet doesn't behave the same.  i have yet to update the nfs/file topic with a special description of specifying SCHEMA there.

i want to confirm that SCHEMA is available only for writing, that is what we currently say in the docs.

note that i updated the example, which includes creating a writable external table, in the read PR, so no changes to the example in this PR.

doc review site link (behind vpn) - NOTE: the read PR has not been merged, so that content will not be reflected in the site:  https://docs-staging.vmware.com/en/VMware-Tanzu-Greenplum-Platform-Extension-Framework/pqlistw/tanzu-greenplum-platform-extension-framework/GUID-hdfs_parquet.html#write-mapping-3


Aayushs4 (approved):

lisakowen:
question about file:parquet and SCHEMA - what is the default if there is no pxf.fs.basePath set?  or is this required?

docs/content/hdfs_parquet.html.md.erb
```diff
@@ -70,27 +70,61 @@ PXF uses the following data type mapping when reading Parquet data:
...
+Parquet is a columnar storage format. A Parquet data file contains a compact binary representation of the data. The *schema* defines the structure of the data, and is composed of the same primitive and complex types identified in the data type mapping section above.
+
+A Parquet data file may include an embedded schema. Or, you can specify the schema that PXF uses to write the data to HDFS via the `SCHEMA` custom option in the `CREATE EXTERNAL TABLE` `LOCATION` clause \(described below\):
+
+| External Table Type  | `SCHEMA` Specified? | Behaviour |
+|-------|--------------------------|-----------|
+| writable   | yes | PXF uses the specified schema. |
+| writable   | no | PXF creates the Parquet schema based on the external table definition. |
+
+When you provide the Parquet schema file to PXF, the file must reside in the same location on each Greenplum Database host **or** the file may reside on the Hadoop file system. PXF first searches for an absolute file path on the Greenplum hosts. If PXF does not find the schema file there, it searches for the file relative to the PXF classpath. If PXF cannot find the schema file locally, it searches for the file on HDFS.
```
yimingli-vmware, line 125: I'm not sure if there is a searching logic here

docs/content/hdfs_parquet.html.md.erb
```diff
@@ -135,7 +169,7 @@ The PXF `hdfs:parquet` profile supports encoding- and compression-related write
 | ENABLE\_DICTIONARY | A boolean value that specifies whether or not to enable dictionary encoding. The default value is `true`; dictionary encoding is enabled when PXF writes Parquet files. |
 | DICTIONARY\_PAGE\_SIZE | When dictionary encoding is enabled, there is a single dictionary page per column, per row group. `DICTIONARY_PAGE_SIZE` is similar to `PAGE_SIZE`, but for the dictionary. The default dictionary page size is `1 * 1024 * 1024` bytes. |
 | PARQUET_VERSION | The Parquet version; PXF supports the values `v1` and `v2` for this option. The default Parquet version is `v1`. |
-| SCHEMA | The location of the Parquet schema file on the file system of the specified `SERVER`. |
+| SCHEMA | The absolute path to the Parquet schema file on the Greenplum host or on HDFS, or the relative path to the schema file on the host. |
```
yimingli-vmware, line 172: Users need to provide the relative path only when user use `FILE` protocol. When they use HDFS, ADL, they should provide absolute path. GS and S3 still need to confirm. But I don't know what should be a proper way to describe this part.

docs/content/hdfs_parquet.html.md.erb
lisakowen: removed that text.

docs/content/hdfs_parquet.html.md.erb
lisakowen: ok, i added a section to the nfs/file topic about this.  i can also add to the object store topic if/when we identify what is required.

docs/content/hdfs_parquet.html.md.erb
```diff
@@ -70,27 +70,59 @@ PXF uses the following data type mapping when reading Parquet data:
 
 PXF uses the following data type mapping when writing Parquet data:
 
-| PXF/Greenplum Data Type | Original Type | Parquet Data Type |
+| PXF/Greenplum Data Type | Parquet Logical Type | Parquet Physical Type |
```
ashuka24, line 73: I realize that this is already present in current documentation, it's a little strange to me that the order here is `PXF | Parquet Logical | Parquet Physical` because if there is no logical type, the reader must skip the middle column to find the mapping. Would it make more sense to have 
``` PXF/GPDB Datatype | Parquet Physical Type | Parquet Logical Type```?

File: docs/content/hdfs_parquet.html.md.erb [hunk 4]
```diff
@@ -70,27 +70,59 @@ PXF uses the following data type mapping when reading Parquet data:
...
 </br><sup>1</sup>&nbsp;PXF localizes a <code>Timestamp</code> to the current system time zone and converts it to universal time (UTC) before finally converting to <code>int96</code>.
 </br><sup>2</sup>&nbsp;PXF converts a <code>Timestamptz</code> to a UTC <code>timestamp</code> and then converts to <code>int96</code>. PXF loses the time zone information during this conversion.
 
+PXF can write a one-dimensional `LIST` of certain Parquet data types. The supported mappings follow:
+
+| PXF/Greenplum Data Type |  Parquet Data Type |
+|-------------------|--------------------------|
+| Bigint[] | list of \<int64> |
+| Boolean[] | list of \<boolean> |
+| Bpchar[] |  list of \<binary> (UTF8) |
```
ashuka24, line 100: do we need a comment here about how parquet does not save the length, and so in parquet, it will be a list of text with undefined length?

docs/content/nfs_pxf.html.md.erb
```diff
@@ -214,3 +214,9 @@ Perform the following procedure to insert some data into the `ex1` directory and
 
     When you select from the `pxf_read_nfs` table here, PXF reads the `somedata.csv` file and the new files that it added to the `ex1/` directory in the previous step.
 
+## <a id="about_schema"></a>About Specifying a Parquet Schema File Location
+
+If you use the `file:parquet` profile to write to an external table that references a Parquet file and you want to provide the Parquet schema, specify the `SCHEMA` custom option in the `LOCATION` clause when you create the writable external table. Refer to the [Creating the External Table](hdfs_parquet.html#profile_cet) discussion in the PXF HDFS Parquet documentation for more information on the options available when you create an external table.
+
+You must set `SCHEMA` to the location of the Parquet schema file on the file system of the specified `SERVER=<server_name>`. When the `<server_name>` configuration includes a [`pxf.fs.basePath`](cfg_server.html#pxf-fs-basepath) property setting, PXF considers the file that you specify to be relative to the base path specified.
```
ashuka24, line 221:
```suggestion
You must set `SCHEMA` to the location of the Parquet schema file on the file system of the specified `SERVER=<server_name>`. When the `<server_name>` configuration includes a [`pxf.fs.basePath`](cfg_server.html#pxf-fs-basepath) property setting, PXF considers the schema file that you specify to be relative to the base path specified.

File: docs/content/hdfs_parquet.html.md.erb (hunk 4 above)
Comment (yimingli-vmware) [reply docs/content/hdfs_parquet.html.md.erb:100]: Yeah, I think it's better to have a comment

Comment (lisakowen) [reply docs/content/hdfs_parquet.html.md.erb]: added a note to bpchar in both the primitive and list tables.

File: docs/content/hdfs_parquet.html.md.erb (hunk 3 above)
Comment (lisakowen) [reply docs/content/hdfs_parquet.html.md.erb]: ok, the physical column is now before the logical column.

**Comment by lisakowen** [comment]
thank you for reviewing @yimingli-vmware and @ashuka24 .  i've addressed most of the outstanding comments and republished the review site.

**Review by yimingli-vmware — APPROVED** [review]

File: docs/content/hdfs_parquet.html.md.erb [hunk 6]
```diff
@@ -70,27 +70,60 @@ PXF uses the following data type mapping when reading Parquet data:
...
+| Timestamp[]<sup>2</sup> | list of \<int96> |
+| Timestamptz[]<sup>3</sup> | list of \<int96> |
+| Varchar[] | list of \<binary> (UTF8) |
+| OTHERS | UNSUPPORTED |
+
+### <a id="schema "></a>About Parquet Schemas and Data
+
+Parquet is a columnar storage format. A Parquet data file contains a compact binary representation of the data. The *schema* defines the structure of the data, and is composed of the same primitive and complex types identified in the data type mapping section above.
+
+A Parquet data file may include an embedded schema. Or, you can specify the schema that PXF uses to write the data to HDFS via the `SCHEMA` custom option in the `CREATE WRITABLE EXTERNAL TABLE` `LOCATION` clause \(described below\):
```
denalex, line 119: I think the parquet file will always contain the schema, so the OR statement here is not accurate. PXF will generate parquet schema form the Greenplum external table definition using the type mappings above OR a user can provide the schema in the schema file and pass its location in the SCHEMA URL option when creating the external table.

docs/content/hdfs_parquet.html.md.erb
```diff
@@ -70,27 +70,60 @@ PXF uses the following data type mapping when reading Parquet data:
...
+
+| External Table Type  | `SCHEMA` Specified? | Behaviour |
+|-------|--------------------------|-----------|
+| writable   | yes | PXF uses the specified schema. |
+| writable   | no | PXF creates the Parquet schema based on the external table definition. |
+
+When you provide the Parquet schema file to PXF, you must specify the absolute path, and the file must reside in the same location on each Greenplum Database host **or** the file may reside on the Hadoop file system.
```
denalex, line 126: we do not support (unlike Avro) having schema file on local file system when connecting to Hadoop. The schema file should be on the same file system as the server used (In Hadoop when connecting to Hadoop, in cloud when connecting to cloud, on local file system when reading local files). For `file:parquet` profile the path should be relative to the mount point (as I mentioned previously)

docs/content/hdfs_parquet.html.md.erb
```diff
@@ -135,7 +168,7 @@ The PXF `hdfs:parquet` profile supports encoding- and compression-related write
 | ENABLE\_DICTIONARY | A boolean value that specifies whether or not to enable dictionary encoding. The default value is `true`; dictionary encoding is enabled when PXF writes Parquet files. |
 | DICTIONARY\_PAGE\_SIZE | When dictionary encoding is enabled, there is a single dictionary page per column, per row group. `DICTIONARY_PAGE_SIZE` is similar to `PAGE_SIZE`, but for the dictionary. The default dictionary page size is `1 * 1024 * 1024` bytes. |
 | PARQUET_VERSION | The Parquet version; PXF supports the values `v1` and `v2` for this option. The default Parquet version is `v1`. |
-| SCHEMA | The location of the Parquet schema file on the file system of the specified `SERVER`. |
+| SCHEMA | The absolute path to the Parquet schema file on the Greenplum host or on HDFS. |
```
denalex, line 171: unless it is `file:parquet` profile, where the path is relative to mount point

lisakowen: this topic is specifically for accessing HDFS (using the `hdfs:parquet` profile).  i updated the nfs/file topic with the mount point info.

lisakowen:
thank you all for reviewing.  i am going to merge this so that i can publish the docs.  i will address any further comments.  i also plan to revisit the SCHEMA-related info as we enhance the feature.

PR #917, opened 2022-12-24
Fix parquet write read with hive for singlecluster CDH

ashuka24:
CDH 5.12 uses Apache Hive version 1.1. Parquet Date is not a supported type with Hive 1.1, so we have modified this test to skip over the columns with the date type when we are using CDH for testing.


yimingli-vmware (approved):

automation/src/test/java/org/greenplum/pxf/automation/features/parquet/ParquetWriteTest.java
```diff
@@ -81,11 +83,12 @@ public class ParquetWriteTest extends BaseFeature {
...
             "varchar_arr          VARCHAR(15)[]",         // DataType.VARCHARARRAY
-            "date_arr             DATE[]"       ,         // DataType.DATEARRAY
-            "numeric_arr          NUMERIC[]"              // DataType.NUMERICARRAY
+            "numeric_arr          NUMERIC[]"    ,         // DataType.NUMERICARRAY
+            "date_arr             DATE[]"                 // DataType.DATEARRAY
     };
 
-    private static final String[] PARQUET_PRIMITIVE_ARRAYS_TABLE_COLUMNS_HIVE = {
+    // CDH (Hive 1.1) does not support date, so we will add the date_arr column as needed in the test case
+    private static ArrayList<String> PARQUET_PRIMITIVE_ARRAYS_TABLE_COLUMNS_HIVE = new ArrayList<>(Arrays.asList(
```
denalex, line 91:
```suggestion
    private static List<String> PARQUET_PRIMITIVE_ARRAYS_TABLE_COLUMNS_HIVE = new ArrayList<>(Arrays.asList(
```
try avoiding specific types for variable declarations and use interfaces instead

automation/src/test/java/org/greenplum/pxf/automation/features/parquet/ParquetWriteTest.java
```diff
@@ -472,7 +487,34 @@ private void assertHiveByteaArrayData(List<List<String>> queryResultData) {
...
+            StringJoiner rowBuilder = new StringJoiner(", ", "[", "]")
+                    .add(String.valueOf(i))    // always not-null row index, column index starts with 0 after it
+                    .add(String.format("[\"2010-01-%02d\"]", (i % 30) + 1)) // DataType.DATEARRAY
+                    ;
+
+            assertEquals(rowBuilder.toString(), queryResultData.get(i).toString());
+        }
+    }
+
+    private boolean checkForCdhCluster() throws Exception {
```
denalex, line 506: can we add the comment with the example from versions.txt file and which specific line we expect here ?

denalex (approved):

automation/src/test/java/org/greenplum/pxf/automation/features/parquet/ParquetWriteTest.java
```diff
@@ -287,11 +295,11 @@ public void parquetWriteListsReadWithHive() throws Exception {
 
         // load the data into hive to check that PXF-written Parquet files can be read by other data
         String hiveExternalTableName = writeTableName + "_external";
-        hiveTable = new HiveExternalTable(hiveExternalTableName, PARQUET_PRIMITIVE_ARRAYS_TABLE_COLUMNS_HIVE, "hdfs:/" + fullTestPath);
+        hiveTable = new HiveExternalTable(hiveExternalTableName, PARQUET_PRIMITIVE_ARRAYS_TABLE_COLUMNS_HIVE.toArray(new String[0]), "hdfs:/" + fullTestPath);
```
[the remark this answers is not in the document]
ashuka24: `toArray` requires an array type to know what type of array to return. We input `String[0]` so that the function knows to return a String[]. The size that we give it does not matter as the underlying function will resize it appropriately.

automation/src/test/java/org/greenplum/pxf/automation/features/parquet/ParquetWriteTest.java
```diff
@@ -81,11 +83,12 @@ public class ParquetWriteTest extends BaseFeature {
...
             "varchar_arr          VARCHAR(15)[]",         // DataType.VARCHARARRAY
-            "date_arr             DATE[]"       ,         // DataType.DATEARRAY
-            "numeric_arr          NUMERIC[]"              // DataType.NUMERICARRAY
+            "numeric_arr          NUMERIC[]"    ,         // DataType.NUMERICARRAY
+            "date_arr             DATE[]"                 // DataType.DATEARRAY
     };
 
-    private static final String[] PARQUET_PRIMITIVE_ARRAYS_TABLE_COLUMNS_HIVE = {
+    // CDH (Hive 1.1) does not support date, so we will add the date_arr column as needed in the test case
+    private static List<String> PARQUET_PRIMITIVE_ARRAYS_TABLE_COLUMNS_HIVE = new ArrayList<>(Arrays.asList(
```
bradfordb-vmware, line 91: `Arrays.asList` returns a `List` so you don't need to wrap with `new ArrayList()`

automation/src/test/java/org/greenplum/pxf/automation/features/parquet/ParquetWriteTest.java
bradfordb-vmware: We have a comment at the call site for this method that indicates that we are checking for a CDH cluster because the version of Hive (1.1) used in CDH doesn't support Parquet DATE logical type. I think it would make sense to rename this method to something like `isDateArraySupported()` and move the comment about CHD w/ Hive 1.1 to be with the method.

Additionally, this method now couples the test to the structure of the single cluster tar but that isn't documented in the build for the single cluster tar. What happens if I am running this test against a Hadoop cluster that isn't started from a single cluster tar and/or there is no `versions.txt`? On my local single-cluster-HDP3, it looks like the `versions.txt` is messed up (note the missing hive version):

```text
build number: root
single_cluster-2.1.0
HDP-3.1.4.0-centos7-tars-tarball
hive
hadoop-3.1.1.3.1.4.0-315
zookeeper-3.4.6.3.1.4.0-315
hbase-2.0.2.3.1.4.0-315
tez-0.9.1.3.1.4.0-315-minimal
```

Could we instead get the Hive version from either `SELECT version();`?

automation/src/test/java/org/greenplum/pxf/automation/features/parquet/ParquetWriteTest.java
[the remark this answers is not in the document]
bradfordb-vmware: I don't think it resizes it (although it doesn't matter much) -- according to the JavaDoc for `toArray`

> the array into which the elements of this list are to be stored, if it is big enough; otherwise, a new array of the same runtime type is allocated for this purpose.

I wonder if it would be clearer to do

```java
String[] fields = new String[PARQUET_PRIMITIVE_ARRAYS_TABLE_COLUMNS_HIVE.size()];
fields = PARQUET_PRIMITIVE_ARRAYS_TABLE_COLUMNS_HIVE.toArray(fields);
hiveTable = new HiveExternalTable(hiveExternalTableName, fields, "hdfs:/" + fullTestPath);
```

automation/src/test/java/org/greenplum/pxf/automation/features/parquet/ParquetWriteTest.java
```diff
@@ -472,7 +487,34 @@ private void assertHiveByteaArrayData(List<List<String>> queryResultData) {
...
+
+    private boolean checkForCdhCluster() throws Exception {
+        File versionFile = new File(cluster.getPhdRoot() + "/versions.txt");
+        Scanner versionScanner =  new Scanner(versionFile);
+        while (versionScanner.hasNext()) {
+            String line = versionScanner.nextLine();
+            if (line.contains("CDH")) {
+                return true;
+            } else if (line.contains("HDP")) {
+                // if HDP, bail early so we don't have to scan the entire file
```
hpandeycodeit, line 523: The file expected to contain many version strings?

hpandeycodeit (approved):

automation/src/test/java/org/greenplum/pxf/automation/features/parquet/ParquetWriteTest.java
ashuka24: Yes. The versions.txt file typically contains the CDH/HDP version, the Hive, Hadoop, Zookeeper, Hbase and Tez versions. Since we only care about the CDH/HDP version, bailing early gives a slight optimization so that we don't need to read the rest of the file.

automation/src/test/java/org/greenplum/pxf/automation/features/parquet/ParquetWriteTest.java
ashuka24: I do not check for the Hive version, I am only checking the version of the HDP/CDH as it is implied that CDH contains Hive 1.1 and HDP2/3 contains Hive 1.2+.

An additional check can be added for the Hive version if needed.

automation/src/test/java/org/greenplum/pxf/automation/features/parquet/ParquetWriteTest.java
ashuka24: Since `List` is an interface and I do not define this anywhere else, I need to specify the type of list it is. If I do not include `new ArrayList`, this fails later on when I try to add to the list with the following err: 

```
java.lang.UnsupportedOperationException: null
	at java.base/java.util.AbstractList.add(AbstractList.java:153)
	at java.base/java.util.AbstractList.add(AbstractList.java:111)
	at org.greenplum.pxf.automation.features.parquet.ParquetWriteTest.parquetWriteListsReadWithHive(ParquetWriteTest.java:292)

bradfordb-vmware: Yep; according to [StackOverflow](https://stackoverflow.com/a/1624163) we can use `Lists.newArrayList()`.

yimingli-vmware (approved):

bradfordb-vmware (approved):

PR #920, opened 2023-01-06
Simplify build for pxf-cli

bradfordb-vmware:
Now that the pxf-cli is using Go modules, we don't need to maintain a separate in-tree GOPATH. This allows us to move the pxf-cli source to a top-level directory.

Other changes:
* upgrade to go 1.19.4
* run go mod tidy
* format go code with gofmt


pxf-cli/go.mod
```diff
@@ -1,6 +1,6 @@
 module pxf-cli
 
-go 1.17
+go 1.19
```
denalex, line 3: only 2 parts needed ? We are using 1.19.4 elsewhere

Makefile
```diff
@@ -35,15 +35,15 @@ else
 endif
 
 cli:
-	make -C cli/go/src/pxf-cli
+	make -C pxf-cli
```
denalex, line 38: now that `pxf-cli` is a top-level child , the `pxf-` prefix looks redundant. Maybe this is the time to search for a better name ? It is not a `client`, since a `client` would be a tool that calls PXF APIs, not manages cluster lifecycle. This is more an `admin`, but we also have the main `pxf` script elsewhere. This ended up to be a cluster tool only, not sure if `cluster` or `cluster-tool` is a better name though. Any other suggestions ?

pxf-cli/go.mod
bradfordb-vmware: Looking at the [Go Modules Reference - `go` directive](https://go.dev/ref/mod#go-mod-file-go), it looks like this can only be

> a positive integer followed by a dot and a non-negative integer

We can use `golang:1.19` (instead of `golang:1.19.4`) for container images so that we track the latest patch release for a given Go release version.

concourse/docker/pxf-build-base/cloudbuild.yaml
```diff
@@ -31,18 +31,20 @@ steps:
...
     args:
       - '-c'
       - |
         mkdir -p ./.go-mod-cached-sources
-        cp -r ./.go-mod-cached-sources ./cli/go/pkg/mod
-        go install github.com/onsi/ginkgo/ginkgo@v1.16.5
-        make -C ./cli/go/src/pxf-cli build
+        cp -r ./.go-mod-cached-sources $(go env GOMODCACHE)
+        cd ./pxf-cli
+        go install github.com/onsi/ginkgo/ginkgo
```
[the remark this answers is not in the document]
bradfordb-vmware: The version is captured in the `go.mod` file; by running the `go install` from inside the `pxf-cli` directory, it will use the version from there. I was trying to reduce the number of places where we would need to update this version if/when it changes.

Makefile
ashuka24: +1 for possibly renaming this. would `cluster-admin` work as an option as well? `cluster-tool` seems to make sense to me too.

ashuka24 (approved):

pxf-cli/end_to_end/end_to_end_suite_test.go
```diff
@@ -1,12 +1,11 @@
 package end_to_end_test
 
 import (
-	"fmt"
-	. "github.com/onsi/ginkgo"
-	. "github.com/onsi/gomega"
-	"os"
```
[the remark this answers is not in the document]
bradfordb-vmware: We had been using `os` to get the environment variable `GOPATH` but we aren't doing that anymore so the import is unused. We had been using `fmt.Sprintf` to turn `[]byte` into a `string` but that can/should be done with a type conversion (i.e., `string([]byte)`) so the import is unused now.

yimingli-vmware (approved):

Makefile
bradfordb-vmware: We discussed this as a team and decided to leave the top-level directory just `cli` for now.

README.md
```diff
@@ -95,7 +95,8 @@ To build PXF, you must have:
     Once you have installed Go, you will need the `ginkgo` tool which runs Go tests,
     respectively. Assuming `go` is on your `PATH`, you can run:
     ```shell
-    go install github.com/onsi/ginkgo/ginkgo@v1.16.5
+    cd ~/workspace/pxf/cli
+    go install github.com/onsi/ginkgo/ginkgo
```
denalex, line 99: does this need to be done while in `pxf/cli` directory ? If yes, this is a bit inconvenient as people would not be able to run `make test` just by checking out the project. Can we automate this in a Makefile ?

cli/go/.gitignore
```diff
@@ -1,11 +0,0 @@
-# Ignore everything under go/
-*
-# but allow recursion into subtrees
-!*/
-# then allow everything in pxf-cli
-!src/pxf-cli/**
-
-# reignore some stuff
-.idea
-vendor/
```
denalex, line 10: do we still not want to ignore vendor if someone runs a command that creates it ?

denalex (approved):
added a few minor comments. 

A bigger question (not directly related to changes here, but could be addressed as a part of this PR) is whether moving tests into a separate module will make excluding test dependencies easier and more reliable (when their versions change) for doing license scans.

README.md
bradfordb-vmware: It doesn't need to be run in the `pxf/cli` directory, but then we would need to specify the version (e.g., `go install github.com/onsi/ginkgo/ginkgo@vx.y.z`) and there is a risk of this becoming out of date with what is specified in `go.mod`.

cli/Makefile
```diff
@@ -26,23 +25,22 @@ help:
...
+		ln -s ${HOME}/.go-mod-cached-sources $(GOMODCACHE); \
 	fi
-	go build -v $(LDFLAGS)
-	go install $(LDFLAGS)
+	go build -v $(LDFLAGS) -o build/ $(MODULE_NAME)
 
 stage: test
 	mkdir -p $(PXF_CLI_DIR)/build/stage/pxf/bin
-	cp $(GOPATH)/bin/pxf-cli $(PXF_CLI_DIR)/build/stage/pxf/bin
+	cp build/pxf-cli build/stage/pxf/bin
```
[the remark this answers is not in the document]
bradfordb-vmware: Currently, our module `pxf-cli` only contains a single main package (`pxf-cli.go`) and so we have the module name is the same as the build executable. It is possible for a module to have multiple main packages and therefore multiple built binaries. So, yes we can use the `$(MODULE_NAME)` to avoid repeating `pxf-cli`. If we every added another executable though (e.g., a PXF HA helper utility), we'd have to revisit this. It also means that if we decide that we want to rename the current executable, we'd potentially have to rename the module and update all import statements in the source.

What if we re-introduce the `TARGET` variable as `pxf-cli` and use that to avoid duplication of `pxf-cli` but without coupling the name of the built executable to the import path prefix (i.e., module name)?

README.md
denalex: yes, so can we maybe automate this in the Makefile ? The `test` goal will use `command -v ginkgo` to check if it is already installed and will run the install command if it is not ?

cli/Makefile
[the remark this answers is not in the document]
bradfordb-vmware: I've added back `TARGET := pxf-cli` for setting the name of the built executable.

cli/README.md
```diff
@@ -4,10 +4,10 @@
 
 1. Ensure you are set up for PXF development by following the README.md at the root of this repository. This tool requires Go version 1.9 or higher. Follow the directions [here](https://golang.org/doc/) to get the language set up.
 
-1. Go to the pxf-cli folder and install dependencies
+1. Go to the cli folder and install dependencies
    ```
-   cd pxf/cli/go/src/pxf-cli
-   go get github.com/onsi/ginkgo/ginkgo
+   cd pxf/cli
+   go install github.com/onsi/ginkgo/ginkgo
```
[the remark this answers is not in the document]
bradfordb-vmware: I've added a new Make target `install-tools` to cli/Makefile and added it as a prereq for the `test` target. This will install `ginkgo` into `pxf/cli/bin` if it doesn't already exist.

README.md
bradfordb-vmware: The Makefile in the cli directory has been updated to install `ginkgo` to `pxf/cli/bin` if its not already there.

cli/Makefile
```diff
@@ -0,0 +1,58 @@
+include ../common.mk
```
ashuka24, line 1: why is this needed?

concourse/docker/pxf-build-base/cloudbuild.yaml
```diff
@@ -31,18 +31,20 @@ steps:
...
     entrypoint: 'bash'
     args:
       - '-c'
       - |
         mkdir -p ./.go-mod-cached-sources
-        cp -r ./.go-mod-cached-sources ./cli/go/pkg/mod
-        go install github.com/onsi/ginkgo/ginkgo@v1.16.5
-        make -C ./cli/go/src/pxf-cli build
+        cp -r ./.go-mod-cached-sources $(go env GOMODCACHE)
+        cd ./cli
```
ashuka24, line 42: is there a reason you chose to use `cd` instead of `pushd` and `popd`?

cli/README.md
```diff
@@ -37,7 +31,7 @@ Because it's hard to mock out a Greenplum cluster, it's useful to debug on a rea
 1. Install `dlv` command, see [here](https://github.com/go-delve/delve/blob/master/Documentation/installation/linux/install.md) for more details:
 
 ```bash
-go get -u github.com/go-delve/delve/cmd/dlv
+go install github.com/go-delve/delve/cmd/dlv@latest
```
ashuka24, line 34: instead of having a manual install with a go install here, can we do something similar to what we did with ginkgo and have this import in?

ashuka24 (approved):

cli/Makefile
bradfordb-vmware: The top-level `common.mk` changes Make's shell from POSIX `sh` to `bash` and it contains logic for setting a default value of `PXF_HOME` which is required for the install target.

cli/README.md
bradfordb-vmware: We can but the reason I didn't add it is that our build and test isn't sensitive/dependent on the version of `dlv` -- any version should work and if I'm firing up `dlv` to debug `pxf-cli`, I probably want the latest sets of features. On the other hand, we are tied to a particular version of ginkgo -- using ginkgo 2+ to run our tests will fail with weird messages. So I wanted to couple the ginkgo version to the one in our `go.mod`.

concourse/docker/pxf-build-base/cloudbuild.yaml
bradfordb-vmware: It's super minor, but in `bash` the command `pushd` will print the current directory stack when it changes directories. That's why you'll often see `pushd &>/dev/null`. I tend to use `pushd` and `popd` in longer scripts where the extra redirect is minimal. For this inline YAML, I used the simpler `cd`.

PR #931, opened 2023-02-14
Fix pxf-dev-base build configration

bradfordb-vmware:
According to the Cloud Build documentation [1], the substitutions field is a map of string keys to string values. Unfortunately in YAML, an unquoted value of 1.19 is parsed as a number, causing the build to fail with

> json: cannot unmarshal number into Go value of type string

This commit quotes the substitution value of 1.19 to force it to be parsed as a string.

[1]: https://cloud.google.com/build/docs/build-config-file-schema

### DIFF
--- a/concourse/docker/pxf-dev-base/cloudbuild.yaml
+++ b/concourse/docker/pxf-dev-base/cloudbuild.yaml
@@ -228,7 +228,7 @@ steps:
     - gpdb7-ubuntu18.04-test-pxf-image-cache
 
 substitutions:
-  _GO_VERSION: 1.19 # default values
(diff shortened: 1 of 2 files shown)


ashuka24 (approved):

ashuka24:
Why was this not an issue before?

bradfordb-vmware:
> Why was this not an issue before?

Previously, we specified the Go version with major, minor, and patch (e.g., 1.17.5) which we could cause it to be parsed as a string. Now we are just pinning to the minor version so it is ambiguous if 1.19 is a number or a string.

denalex (approved):

PR #940, opened 2023-03-03
Gpud 163 Write Parquet Numeric Decimal with Overflow 

yimingli-vmware:
This PR is a first-step fix for writing parquet numeric with overflowing values.

- Overflow contains 3 cases, this PR has tested all of these cases
  - pecision overflow (precision defined in table DDL like NUMERIC(90), and data precision overflow with undefined precision NUMERIC)
  - integer digit count overflow (precision - scale > maxPrecision - maxScale)
  - scale overflow

- Current implementation of writing parquet numeric utilizes HiveDecimal class, which limits the precision to be up to 38 and scale up to 38.
  - When table DDL uses NUMERIC, and data precision overflow or integer digit count overflow happens, current pxf behavior is setting the overflowing value to empty, and query goes through.
  - When table DDL uses NUMERIC(precision), and precision overflow happens, current pxf behavior is 
  -  throwing an `ArrayIndexOutOfBoundaryException` 

- As it's a first-step fix, this PR introduces a pxf server configuration `pxf.parquet.write.decimal.overflow` with value `[error|ignore]` 
  - By setting the property to error, when precision data overflow or integer digit count overflow happens in a NUMERIC column, or precision overflow happens in a NUMERIC(precison) column, an error will be thrown, and the query will fail.
  - By setting the property to ignore, when precision data overflow or integer digit count overflow happens in a NUMERIC column, overflowing values will be set to empty and query will success. When precision overflow happens in a NUMERIC(precison) column, an error will still be thrown, and the query will fail


server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetFileAccessor.java
```diff
@@ -552,6 +552,13 @@ private Type getTypeForColumnDescriptor(ColumnDescriptor columnDescriptor) {
                     precision = columnTypeModifiers[0];
                     scale = columnTypeModifiers[1];
                 }
+
+                // precision is defined but precision >  HiveDecimal.MAX_PRECISION
```
hpandeycodeit, line 556:
```suggestion
                // precision is defined but precision >  HiveDecimal.MAX_PRECISION i.e., 38
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetWriteDecimalOverflowOption.java
```diff
@@ -0,0 +1,20 @@
+package org.greenplum.pxf.plugins.hdfs.parquet;
+
+/**
+ * Parquet server configuration options for property pxf.parquet.write.decimal.overflow
+ */
+public enum ParquetWriteDecimalOverflowOption {
+    ERROR("error"),
+    IGNORE("ignore");
+
+    private String value = "";
```
hpandeycodeit, line 10:
```suggestion
    private String value ;
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetWriteDecimalOverflowOption.java
```diff
@@ -0,0 +1,20 @@
...
+
+    private ParquetWriteDecimalOverflowOption(String decimalOverflowOption) {
```
hpandeycodeit, line 12:
```suggestion
    ParquetWriteDecimalOverflowOption(String decimalOverflowOption) {
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
```diff
@@ -257,10 +262,37 @@ private void fillGroupWithPrimitive(Group group, int columnIndex, Object fieldVa
     }
 
     private byte[] getFixedLenByteArray(String value, Type type) {
+        if (!NumberUtils.isNumber(value)) {
```
denalex, line 265: why is this extra check needed ? GP has a value which is NUMERIC that it sent us, so we can assume it is a proper number

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
```diff
@@ -257,10 +262,37 @@ private void fillGroupWithPrimitive(Group group, int columnIndex, Object fieldVa
...
+        }
+
+        int integerDigitCount = dataPrecision - dataScale;
+        int maxIntegerDigitCount = precision - scale;
+        //  data integer digit count overflow and decimal overflow option is error
+        if (integerDigitCount > maxIntegerDigitCount && decimalOverflowOption.equals(ParquetWriteDecimalOverflowOption.ERROR.getValue())) {
+            throw new UnsupportedTypeException(String.format("Integer digit count of data %s overflows. " +
+                    "Your integer digit count is %d. The max integer digit count is %d", value, integerDigitCount, maxIntegerDigitCount));
+        }
+
```
denalex, line 295: why all this logic above ? I thought we would call `HiveDecimal.create(value)` and if the incoming data was not null but we got back NULL, then we have a problem and need to throw the exception (conditionally).

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/parquet/ParquetWriteDecimalOverflowOption.java
```diff
@@ -0,0 +1,20 @@
+package org.greenplum.pxf.plugins.hdfs.parquet;
+
+/**
+ * Parquet server configuration options for property pxf.parquet.write.decimal.overflow
+ */
+public enum ParquetWriteDecimalOverflowOption {
```
denalex, line 6: I don't think we need this class, just having 2 constants in the class that uses the property will be sufficient.

server/pxf-service/src/templates/templates/pxf-site.xml
```diff
@@ -84,4 +84,14 @@
         </description>
     </property>
 
+    <property>
+        <name>pxf.parquet.write.decimal.overflow</name>
+        <value>error</value>
+        <description>
+            Specifies whether the PXF Parquet writer should error out or ignore values when writing decimal values with overflow.
+            If set to error, an error will be thrown when seeing an overflowing value, and the writing is incomplete.
```
denalex, line 92:
```suggestion
            If set to error, an error will be reported when a numeric value exceeds maximum column precision and the transaction will fail, potentially leaving incomplete dataset in the external system.
```

server/pxf-service/src/templates/templates/pxf-site.xml
```diff
@@ -84,4 +84,14 @@
...
+            If set to ignore, the overflowing value will be skipped, and writing will success but some data are missing.
```
denalex, line 93:
```suggestion
            If set to ignore, NULL values will be stored for numeric values that exceeds maximum column precision, the transaction will succeed, but the resulting dataset in the external system will have NULL values instead of the original ones where modifications took place.
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
```diff
@@ -257,10 +262,37 @@ private void fillGroupWithPrimitive(Group group, int columnIndex, Object fieldVa
...
         int precision = Math.min(HiveDecimal.MAX_PRECISION, typeAnnotation.getPrecision());
         int scale = Math.min(HiveDecimal.MAX_SCALE, typeAnnotation.getScale());
+
+        BigDecimal bigDecimal = NumberUtils.createBigDecimal(value);
+        int dataPrecision = bigDecimal.precision();
+        int dataScale = bigDecimal.scale();
+
+        String decimalOverflowOption = parseDecimalOverflowOption(configuration);
+        // data precision overflow and decimal overflow option is error
+        if (dataPrecision > HiveDecimal.MAX_PRECISION) {
```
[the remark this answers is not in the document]
yimingli-vmware: If we want to get a precision from `typeAnnotation.getPrecision()` different from `HiveDecimal.MAX_PRECISION`, we must define the column type as `NUMERIC(precision, scale)`. And GPDB forces user to insert data with data precision and scale not greater than the provided values. Therefore, GPDB will throw an `numeric overflow` error when try to insert such data

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
yimingli-vmware: There is another case that `HiveDecimal.create()` can return null when column is defined as `NUMERIC`. `HiveDecimal.create` mainly to see if the incoming value can fit into the provided precision. 
- If the integer digit count of the incoming value > provided precision, cannot place all the integer digits -> return null
- if the integer digit count of the incoming value <= provided precision
  - if overall data precision > provided precision -> round off the decimal part in the data
  - if overall data precision <= provided precision -> create a HiveDecimal with the same value

And `HiveDecimal.enforcePrecisionScale` mainly to enforce Decimal(precision) into Decimal(precision, scale). Although `HiveDecimal.create` has ensured the data precision, it is still possible that integer digit count is still greater than the maximum supported integer digit count. Then null value is returned in this case.

For example, a column `num` is defined as `NUMERIC`, and pxf will treat it as `DECIMAL(38,18)` in parquet schema. 
- When we want to insert a value like `1234567890123456789012345678901234567890.123`, it has 40 integer digits which is greater than precision 38, HiveDecimal.create will return `null` and this data will be stored as null
- When we want to insert a value like `123456789012345678901234567890.1234567890`, its precision is 40 which is greater than 38, but its integer digit count is less than 38. The value will be rounded off into `123456789012345678901234567890.12345679`. But DECIMAL(38,18) only allow 20 integer digits, that means this value still cannot fit in this requirement. So `HiveDecimal.enforcePrecisionScale` will return null and this value will be stored as null.

Columns defined as `NUMERIC(precision, scale)` don't need to care about these cases since GPDB has ensured the value can fit in `DECIMAL(precision, scale)` when inserting.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
```diff
@@ -261,21 +275,95 @@ private byte[] getFixedLenByteArray(String value, Type type) {
...
 
+        BigDecimal accurateDecimal = new BigDecimal(value);
+        // At this point data may be rounded off
+        if ((isDecimalOverflowOptionError || isDecimalOverflowOptionRound) && accurateDecimal.compareTo(hiveDecimal.bigDecimalValue()) != 0) {
+            if (isDecimalOverflowOptionError) {
+                throw new UnsupportedTypeException(String.format("The scale of the numeric data %s is %d, which exceeds the maximum supported scale %d. Data accuracy is lost. Query failed.",
+                        value, accurateDecimal.scale(), scale));
+            }
+
+            LOG.trace(String.format("The scale of the numeric data %s is %d, which exceeds the maximum supported scale %d. Data will be rounded off.",
```
ashuka24, line 358: this log here is at `LOG.trace` but in the other cases, they are at `LOG.warn` do we want to be consistent here?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
```diff
@@ -59,17 +62,33 @@ public class ParquetResolver extends BasePlugin implements Resolver {
...
+    private boolean isPrecisionOverflowWarningLogged;
+    private boolean isIntegerDigitCountOverflowWarningLogged;
+    private boolean isScaleOverflowWarningLogged;
+
     @Override
     public void afterPropertiesSet() {
         super.afterPropertiesSet();
         columnDescriptors = context.getTupleDescription();
+        isDecimalOverflowOptionError = parseDecimalOverflowOption(configuration).equals(PXF_PARQUET_WRITE_DECIMAL_OVERFLOW_OPTION_ERROR);
+        isDecimalOverflowOptionRound = parseDecimalOverflowOption(configuration).equals(PXF_PARQUET_WRITE_DECIMAL_OVERFLOW_OPTION_ROUND);
```
denalex, line 87: you should call `parse` function once, store the result and then compare, no reason to parse it twice

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
```diff
@@ -59,17 +62,33 @@ public class ParquetResolver extends BasePlugin implements Resolver {
...
+        isPrecisionOverflowWarningLogged = false;
+        isIntegerDigitCountOverflowWarningLogged = false;
+        isScaleOverflowWarningLogged = false;
+
```
denalex, line 91: nit: extra line

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
```diff
@@ -261,21 +280,107 @@ private byte[] getFixedLenByteArray(String value, Type type) {
...
+        HiveDecimal.create will return the same decimal value as provided.
+        For example, 123456.123456 can fit in DECIMAL(38,18) without and data loss, so the data will be created as the same
+         */
+        // HiveDecimal.create will return a decimal value which can fit in DECIMAL(38)
+        HiveDecimal parsedValue = HiveDecimal.create(value);
+
+        if (parsedValue == null) {
+            if (isDecimalOverflowOptionError || isDecimalOverflowOptionRound) {
+                throw new UnsupportedTypeException(String.format("The integer digit count of data %s in " +
+                                "a NUMERIC column exceeds the maximum supported precision %d. Query failed.",
```
denalex, line 313: we need to report columnName, declare as function parameter, if needed.
```suggestion
                                "a NUMERIC column %s exceeds the maximum supported precision %d. Query failed.",
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
```diff
@@ -261,21 +280,107 @@ private byte[] getFixedLenByteArray(String value, Type type) {
...
+            }
+
+            LOG.trace(String.format("The integer digit count of  data %s in " +
+                            "a NUMERIC column exceeds the maximum supported precision %d. Data will be stored as NULL.",
+                    value, HiveDecimal.MAX_PRECISION));
+
+            if (!isPrecisionOverflowWarningLogged) {
+                LOG.warn("There are some uncleaned data in a column defined as NUMERIC with undefined precision. " +
+                        "The integer digit counts of these data exceed the maximum supported precision. " +
+                        "Data will be stored as NULL.");
```
denalex, line 324:
```suggestion
                LOG.warn("There are rows where for the NUMERIC column %s the values exceed maximum precision %d and have been stored as NULL. Enable TRACE log level for row-level details.");
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
```diff
@@ -261,21 +280,107 @@ private byte[] getFixedLenByteArray(String value, Type type) {
...
+
+        if (parsedValue == null) {
+            if (isDecimalOverflowOptionError || isDecimalOverflowOptionRound) {
+                throw new UnsupportedTypeException(String.format("The integer digit count of data %s in " +
+                                "a NUMERIC column exceeds the maximum supported precision %d. Query failed.",
+                        value, HiveDecimal.MAX_PRECISION));
+            }
+
+            LOG.trace(String.format("The integer digit count of  data %s in " +
+                            "a NUMERIC column exceeds the maximum supported precision %d. Data will be stored as NULL.",
```
denalex, line 318:
```suggestion
            LOG.trace(String.format("The value %d for the NUMERIC column %s exceeds maximum precision %d and has been stored as NULL.",
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
```diff
@@ -261,21 +280,107 @@ private byte[] getFixedLenByteArray(String value, Type type) {
...
         if (hiveDecimal == null) {
-            // When precision is higher than HiveDecimal.MAX_PRECISION
-            // and enforcePrecisionScale returns null, it means we
-            // cannot store the value in Parquet because we have
-            // exceeded the precision. To make the behavior consistent
-            // with Hive's behavior when storing on a Parquet-backed
-            // table, we store the value as null.
+            if (isDecimalOverflowOptionError || isDecimalOverflowOptionRound) {
+                throw new UnsupportedTypeException(String.format("The integer digit count of data %s in a NUMERIC column exceeds " +
+                                "the maximum supported integer digit count %d. Query failed.",
```
denalex, line 349:
```suggestion
                throw new UnsupportedTypeException(String.format("The value %d for the NUMERIC column %s exceeds maximum precision (%d,%d)",
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
```diff
@@ -261,21 +280,107 @@ private byte[] getFixedLenByteArray(String value, Type type) {
...
+                        value, precision - scale));
+            }
+
+            LOG.trace(String.format("The integer digit count of data %s in a NUMERIC column exceeds " +
+                            "the maximum supported integer digit count %d. Data will be stored as NULL.",
```
denalex, line 354:
```suggestion
            LOG.trace(String.format("The value %d for the NUMERIC column %s exceeds maximum precision (%d,%d) and has been stored as NULL.",
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
```diff
@@ -261,21 +280,107 @@ private byte[] getFixedLenByteArray(String value, Type type) {
...
+                    value, precision - scale));
+
+            if (!isIntegerDigitCountOverflowWarningLogged) {
+                LOG.warn("There are some uncleaned data in a column defined as NUMERIC with undefined precision. " +
+                        "The integer digit counts of these data exceed the maximum supported integer digit count. " +
+                        "Data will be stored as NULL.");
```
denalex, line 360: please update to a similar error message as the other case

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
```diff
@@ -261,21 +280,107 @@ private byte[] getFixedLenByteArray(String value, Type type) {
...
+            }
             return null;
         }
 
+        BigDecimal accurateDecimal = new BigDecimal(value);
+        // At this point data can fit in DECIMAL(38,18), but may have been rounded off
+        if ((isDecimalOverflowOptionError || isDecimalOverflowOptionRound) && accurateDecimal.compareTo(hiveDecimal.bigDecimalValue()) != 0) {
+            if (isDecimalOverflowOptionError) {
+                throw new UnsupportedTypeException(String.format("The scale of the numeric data %s in a NUMERIC column " +
+                                "exceeds the maximum supported scale %d. Data accuracy is lost. Query failed.", value, scale));
```
denalex, line 371:
```suggestion
                throw new UnsupportedTypeException(String.format("The scale of the data %s in the NUMERIC column %s " +
                                "exceeds the maximum supported scale %d.", value, columnName, scale));
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
```diff
@@ -261,21 +280,107 @@ private byte[] getFixedLenByteArray(String value, Type type) {
...
+            }
+
+            LOG.trace(String.format("The scale of the numeric data %s in a NUMERIC column exceeds the maximum supported scale %d. " +
+                            "Data will be rounded off.", value, scale));
```
denalex, line 375:
```suggestion
            LOG.trace(String.format("The scale of the data %s in a NUMERIC column %s exceeds the maximum supported scale %d and will be rounded off.", value, columnName, scale));
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
```diff
@@ -261,21 +280,107 @@ private byte[] getFixedLenByteArray(String value, Type type) {
...
+
+            if (!isScaleOverflowWarningLogged) {
+                LOG.warn("There are some uncleaned data in a column defined as NUMERIC with undefined precision. " +
+                        "The scales of these data exceed the maximum supported scale. Data will be rounded off");
```
denalex, line 379: please update to a similar error message as other cases

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
```diff
@@ -356,4 +461,20 @@ private OneField resolveField(Group group, int columnIndex) {
...
     }
+
+    /**
+     * Returns server configuration value for pxf.parquet.write.decimal.overflow.
+     *
+     * @param configuration contains server configuration properties
+     * @return String value for pxf.parquet.write.decimal.overflow. Must be "error", "round" or "ignore"
+     */
+    public String parseDecimalOverflowOption(Configuration configuration) {
+        String decimalOverflowOption = configuration.get(PXF_PARQUET_WRITE_DECIMAL_OVERFLOW_PROPERTY_NAME, "round");
```
denalex, line 472:
```suggestion
        String decimalOverflowOption = configuration.get(PXF_PARQUET_WRITE_DECIMAL_OVERFLOW_PROPERTY_NAME, PXF_PARQUET_WRITE_DECIMAL_OVERFLOW_OPTION_ROUND);
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
```diff
@@ -356,4 +461,20 @@ private OneField resolveField(Group group, int columnIndex) {
...
+        if (StringUtils.equalsIgnoreCase(PXF_PARQUET_WRITE_DECIMAL_OVERFLOW_OPTION_IGNORE, decimalOverflowOption)) {
+            return PXF_PARQUET_WRITE_DECIMAL_OVERFLOW_OPTION_IGNORE;
+        } else if (StringUtils.equalsIgnoreCase(PXF_PARQUET_WRITE_DECIMAL_OVERFLOW_OPTION_ERROR, decimalOverflowOption)) {
+            return PXF_PARQUET_WRITE_DECIMAL_OVERFLOW_OPTION_ERROR;
+        }
```
denalex, line 477: we need to analyze ROUND and error out if the property has been set to anything else.

server/pxf-service/src/templates/templates/pxf-site.xml
```diff
@@ -84,4 +84,18 @@
...
+        <name>pxf.parquet.write.decimal.overflow</name>
+        <value>round</value>
+        <description>
+            Specifies whether the PXF Parquet profiles should error out, or log a warning about values have been rounded, or ignore numeric values (convert to null) that exceed maximum column precision 38.
+            If set to error, an error will be reported when a numeric value exceeds maximum column precision and the transaction will fail,
+            potentially leaving an incomplete dataset in the external system.
+            If set to round, a warning will be reported when a value with precision overflow can be rounded and fit in the provided precision;
+            or an error will be reported if data cannot been rounded, then transaction will fail and potentially leave an incomplete dataset in the external system.
+            If set to ignore, there will be no warnings about rounded values, and NULL values will be stored for numeric values that exceeds maximum column precision.
+            The transaction will succeed, but the resulting dataset in the external system will have NULL values instead of the original ones where modifications took place.
```
denalex, line 97: we need to make this shorter. details can be found in documentation.
```suggestion
            Specifies behavior of the PXF Parquet profiles when writing NUMERIC data that exceeds maximum precision. Valid values are error, round, and ignore. The default value is round.
            If set to error, an error will be reported and the transaction will fail, potentially leaving an incomplete dataset in the external system.
            If set to round, PXF will try to round and write the value or report an error if the value cannot be rounded.
            If set to ignore, PXF will write NULL instead of the value (as in previous PXF versions).
```

automation/src/test/java/org/greenplum/pxf/automation/features/parquet/ParquetWriteTest.java
```diff
@@ -234,6 +263,79 @@ public void parquetWriteNumericWithPrecisionAndScale() throws Exception {
...
+        gpdb.copyFromFile(gpdbUndefinedPrecisionNumericTable, new File(localDataResourcesFolder
+                + "/numeric/undefined_precision_numeric_with_large_scale.csv"), "E','", true);
+
+        String filename = "parquet_write_undefined_precision_numeric_large_scale";
+        prepareWritableExternalTable("parquet_write_undefined_precision_numeric_large_scale",
+                UNDEFINED_PRECISION_NUMERIC, hdfsPath + filename, null);
+        exTable.setHost(pxfHost);
+        exTable.setPort(pxfPort);
+        exTable.setFormatter("pxfwritable_export");
+        exTable.setProfile(ProtocolUtils.getProtocol().value() + ":parquet");
```
denalex, line 323: boilerplate repeated everywhere - there are helper functions in the newer tests that do it for external tables

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
yimingli-vmware: Do we need to error out if ROUND is not set to "round"? Or we treat anything else as "round" instead of erroring out?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
```diff
@@ -256,26 +270,108 @@ private void fillGroupWithPrimitive(Group group, int columnIndex, Object fieldVa
...
+            if (isDecimalOverflowOptionError || isDecimalOverflowOptionRound) {
+                throw new UnsupportedTypeException(String.format("The value %s for the NUMERIC column %s exceeds maximum precision and scale (%d,%d).",
+                        value, columnName, precision, scale));
+            }
+
+            LOG.trace(String.format("The value %s for the NUMERIC column %s exceeds maximum precision and scale (%d,%d) and has been stored as NULL.",
+                    value, columnName, precision, scale));
+
+            if (!isIntegerDigitCountOverflowWarningLogged) {
+                LOG.warn(String.format("There are rows where for the NUMERIC column %s the values exceed maximum precision and scale (%d,%d) " +
```
ashuka24, line 348: is it possible to know how many rows?

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/ParquetWriteTest.java
```diff
@@ -842,8 +845,8 @@ public void testWriteChar() throws Exception {
     @Test
     public void testWriteNumeric() throws Exception {
         String path = temp + "/out/numeric/";
-        // precision is 38 and scale is 18
-        columnDescriptors.add(new ColumnDescriptor("dec1", DataType.NUMERIC.getOID(), 0, "numeric", new Integer[]{38, 18}));
+        int precision = 38, scale = 18;
```
ashuka24, line 848: nit: we typically do not use this styling convention, is this something we want to keep or swap it to something like this: 
```suggestion
        int precision = 38;
        int scale = 18;

File: server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java [hunk 28]
```diff
@@ -356,4 +452,22 @@ private OneField resolveField(Group group, int columnIndex) {
...
+        String decimalOverflowOption = configuration.get(PXF_PARQUET_WRITE_DECIMAL_OVERFLOW_PROPERTY_NAME, PXF_PARQUET_WRITE_DECIMAL_OVERFLOW_OPTION_ROUND).toLowerCase();
+
+        if (decimalOverflowOption.equals(PXF_PARQUET_WRITE_DECIMAL_OVERFLOW_OPTION_ERROR)) {
+            isDecimalOverflowOptionError = true;
+        } else if (decimalOverflowOption.equals(PXF_PARQUET_WRITE_DECIMAL_OVERFLOW_OPTION_ROUND)) {
+            isDecimalOverflowOptionRound = true;
+        } else if (!decimalOverflowOption.equals(PXF_PARQUET_WRITE_DECIMAL_OVERFLOW_OPTION_IGNORE)) {
+            throw new UnsupportedTypeException(String.format("Invalid configuration value %s for " +
+                    "pxf.parquet.write.decimal.overflow. Valid values are error, round, and ignore.", decimalOverflowOption));
+        }
```
denalex, line 471: I'd use a switch here: https://docs.oracle.com/javase/8/docs/technotes/guides/language/strings-switch.html

```suggestion
        switch(configuration.get(PXF_PARQUET_WRITE_DECIMAL_OVERFLOW_PROPERTY_NAME, PXF_PARQUET_WRITE_DECIMAL_OVERFLOW_OPTION_ROUND).toLowerCase()) {
        case PXF_PARQUET_WRITE_DECIMAL_OVERFLOW_OPTION_ERROR:
             isDecimalOverflowOptionError = true;
             break;
         case PXF_PARQUET_WRITE_DECIMAL_OVERFLOW_OPTION_ROUND:
              isDecimalOverflowOptionRound = true;
              break;
          case PXF_PARQUET_WRITE_DECIMAL_OVERFLOW_OPTION_IGNORE:
              break;
          default:
               throw new UnsupportedTypeException(String.format("Invalid configuration value %s for " +
                    "pxf.parquet.write.decimal.overflow. Valid values are error, round, and ignore.", decimalOverflowOption));
        }
```

server/pxf-service/src/templates/templates/pxf-site.xml
```diff
@@ -84,4 +84,15 @@
         </description>
     </property>
 
+    <property>
+        <name>pxf.parquet.write.decimal.overflow</name>
+        <value>round</value>
+        <description>
+            Specifies behavior of the PXF Parquet profiles when writing NUMERIC data that exceeds maximum precision. Valid values are error, round, and ignore. The default value is round.
+            If set to error, an error will be reported and the transaction will fail, potentially leaving an incomplete dataset in the external system.
```
hpandeycodeit, line 92: From the description here, it seems like setting to `error` will not be a good idea as it will be`potentially leaving an incomplete dataset in the external system.` ?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
yimingli-vmware: We have this warning only when we first see the overflow, so it's not possible to know how many rows contain overflow in the warning. But log.trace will record all the overflowing values in the numeric column

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
```diff
@@ -356,4 +452,22 @@ private OneField resolveField(Group group, int columnIndex) {
...
+
+    /**
+     * Sets configuration variables based on server configuration properties of pxf.parquet.write.decimal.overflow.
+     *
+     * @param configuration contains server configuration properties
+     */
+    public void parseDecimalOverflowOption(Configuration configuration) {
+        String decimalOverflowOption = configuration.get(PXF_PARQUET_WRITE_DECIMAL_OVERFLOW_PROPERTY_NAME, PXF_PARQUET_WRITE_DECIMAL_OVERFLOW_OPTION_ROUND).toLowerCase();
+
+        if (decimalOverflowOption.equals(PXF_PARQUET_WRITE_DECIMAL_OVERFLOW_OPTION_ERROR)) {
```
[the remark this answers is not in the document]
yimingli-vmware: Refactored this part to use `switch case`, so no need to have string comparison here

server/pxf-service/src/templates/templates/pxf-site.xml
yimingli-vmware: It depends on which behavior user want to choose. If users want the transaction goes through and don't care too much about data accuracy, then `error` is a bad option. If users want the data stored in the database is restrictedly accurate, then `error` is a good option. If users can accept some degree of accuracy loss, then `round` is a good option since in this case, pxf will try to round off values if overflow occurs. But it is still possible that the dataset can be incomplete when pxf fails to round off values. I will add this in the round option.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
```diff
@@ -256,26 +269,108 @@ private void fillGroupWithPrimitive(Group group, int columnIndex, Object fieldVa
...
+         */
+        // HiveDecimal.create will return a decimal value which can fit in DECIMAL(38)
+        HiveDecimal parsedValue = HiveDecimal.create(value);
+        if (parsedValue == null) {
+            if (isDecimalOverflowOptionError || isDecimalOverflowOptionRound) {
+                throw new UnsupportedTypeException(String.format("The value %s for the NUMERIC column %s exceeds maximum precision %d.",
+                        value, columnName, precision));
+            }
+
+            LOG.trace(String.format("The value %s for the NUMERIC column %s exceeds maximum precision %d and has been stored as NULL.",
```
bradfordb-vmware, line 309: Since this has the potential to be executed for every column of every row, I think we should do either

```java
if (LOG.isTraceEnabled()) {
    LOG.trace(String.format("The value %s for the NUMERIC column %s exceeds maximum precision %d and has been stored as NULL.",
        value, columnName, precision));
}
```

or

```java
LOG.trace("The value {} for the NUMERIC column {} exceeds maximum precision {} and has been stored as NULL.",
    value, columnName, precision);
```

to avoid the overhead of creating a new string that won't be logged unless the log level is set to `TRACE`.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
```diff
@@ -256,26 +269,108 @@ private void fillGroupWithPrimitive(Group group, int columnIndex, Object fieldVa
...
 
+        BigDecimal accurateDecimal = new BigDecimal(value);
+        // At this point data can fit in DECIMAL(38,18), but may have been rounded off
+        if ((isDecimalOverflowOptionError || isDecimalOverflowOptionRound) && accurateDecimal.compareTo(hiveDecimal.bigDecimalValue()) != 0) {
+            if (isDecimalOverflowOptionError) {
+                throw new UnsupportedTypeException(String.format("The value %s for the NUMERIC column %s exceeds maximum scale %d.",
+                        value, columnName, scale));
+            }
+
+            LOG.trace(String.format("The value %s for the NUMERIC column %s exceeds maximum scale %d and has been rounded off.",
```
bradfordb-vmware, line 363: Same comment about the `trace` message formatting as above.

bradfordb-vmware (commented):
I think we need to re-write the logging statements to avoid using `String.format` for performance reasons.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetResolver.java
```diff
@@ -256,26 +269,108 @@ private void fillGroupWithPrimitive(Group group, int columnIndex, Object fieldVa
...
+            if (isDecimalOverflowOptionError || isDecimalOverflowOptionRound) {
+                throw new UnsupportedTypeException(String.format("The value %s for the NUMERIC column %s exceeds maximum precision %d.",
+                        value, columnName, precision));
+            }
+
+            LOG.trace("The value {} for the NUMERIC column {} exceeds maximum precision {} and has been stored as NULL.",
+                    value, columnName, precision);
+
+            if (!isPrecisionOverflowWarningLogged) {
+                LOG.warn(String.format("There are rows where for the NUMERIC column %s the values exceed maximum precision %d " +
```
ashuka24, line 313: you should not need to use String.format here. This is also true for all the other `LOG.warn` instances in this file

ashuka24 (approved):

hpandeycodeit (approved):

PR #943, opened 2023-03-07
Upgrade gp-common-go-libs to 1.0.11

hpandeycodeit:


edespino (approved):
Approved.

A couple of notes for your consideration:
- The **"Build and Test PXF-GP5 on RHEL7"** PR job will continue to fail until the images are rebuilt.
- Please squash these commits manually prior to committing. This will allow you to remove the accidental commits which introduce testing changes.
- The PXF build pipelines should be paused before this is merged. After merging this PR, the Cloud Build processes will build and publish new **_latest_** container images.
- After containers have been built, the PXF build pipelines can be unpaused. Please ensure all jobs are green.

PR #948, opened 2023-03-17
Additional gp7 release support

edespino:
Upon reviewing the `pxf-build (pipeline) --> Publish_PXF-GP5_and_PXF-GP6_Artifacts_to_GP-RelEng (job) --> Get PXF-GP5 and PXF-GP6 Artifacts from Releases Directory (task)`, I noticed the need for the changes in this PR.

Outstanding question: In `Publish_PXF-GP5_and_PXF-GP6_Artifacts_to_GP-RelEng`, I notice the creation of pxf_gp5 and pxf_gp6 tarballs. Is there a need for a gp7 variant?

There are also a few occurrences of jobs and tasks referencing PXF-GP5 and PXF-GP6 and not PXF-GP7 even though it is also part of the workflow. At some point, we should change that to be inclusive in terminology or generic in nature (my preference).  This should happen after the release. If the names are changed now, the previous build job logs will not be available for review. I'm expecting the previous run output to be helpful to diagnose any issues with the new PXF GP7 support.

### DIFF
--- a/concourse/scripts/get_pxf_release_artifacts.bash
+++ b/concourse/scripts/get_pxf_release_artifacts.bash
@@ -44,6 +44,7 @@ artifacts=(
   "${GCS_RELEASES_PATH}/gp6/pxf-gp6-${version}-1.el7.x86_64.rpm"
   "${GCS_RELEASES_PATH}/gp6/pxf-gp6-${version}-1.el8.x86_64.rpm"
   "${GCS_RELEASES_PATH}/gp6/pxf-gp6-${version}-1-ubuntu18.04-amd64.deb"
+  "${GCS_RELEASES_PATH}/gp7/pxf-gp7-${version}-1.el8.x86_64.rpm"
   "${GCS_OSL_PATH}/open_source_license_VMware_Tanzu_Greenplum_Platform_Extension_Framework_${version}_GA.txt"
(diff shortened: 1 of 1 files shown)


denalex:
```
Outstanding question: In Publish_PXF-GP5_and_PXF-GP6_Artifacts_to_GP-RelEng, I notice the creation of pxf_gp5 and pxf_gp6 tarballs. Is there a need for a gp7 variant?
```
@edespino - it has been decided there is no need for PXF-GP7 tarballs as they are not used in the release process.

denalex:
I thikn we also need to rename the task `Publish_PXF-GP5_and_PXF-GP6_Artifacts_to_GP-RelEng` to `Publish_PXF-GP[5|6|7]_Artifacts_to_GP-RelEng`

edespino:
@denalex Thank you for your GP7 tarball insight. Regarding your recommendation to change the Publish job name. I am aware of this. However, if the name is changed, the job with the previous name will not be visible. I would like to have that information available in case I run into any release job issues. The Promote job's name should also be updated. I suggest we rename these to something more generic (Promote_PXF_GP_Artifacts & Publish_PXF_GP_Artifacts_to_GP-RelEng) without GP versions. Unfortunately, the Unmanaged Data team's Concourse instance is using an older version. Concourse v6.7.3 added support to rename a job and reference the old_name and thus allowing one to view the old job details.

denalex (approved):

PR #949, opened 2023-03-20
Support for fixedwidth formatter with new *:fixedwidth PXF profiles

denalex:
This commit enables PXF to support existing GPDB `fixedwidth_in` (for reading) and `fixedwidth_out` (for writing) formatters that will allow PXF to be able to read and write text data where fields are declared to be fixed length.

A new family of HCFS-compatible `*:fixedwidth` PXF profiles is introduced. The new profiles are:
- hdfs:fixedwidth
- s3:fixedwidth
- adl:fixedwidth
- wasbs:fixedwidth
- gs:fixedwidth
- file:fixedwidth

All options of the `fixedwidth` formatters are supported and need to be provided in the formatter specification when defining an external table. However, a special care needs to be taken if the formatter specification defines a `line_delim` option.

By default, the PXF profiles use the value of `\n` (LF) for the new line delimiter. If the the external file also has the same line delimiter, then usually there will be no `line_delim` option specified for the formatter and no special action is required. But if the `line_delim` option is provided to contain either `\r`, \r\n` or a set of custom characters, then the PXF external table needs to include the NEWLINE user option with the values of `CR`, `CRLF` or the set of custom characters, respectively.

Here's an example of using the new profile for reading and writing fixed-length external data file:
```
CREATE EXTERNAL TABLE fixed_read (s text, n integer)
LOCATION ('pxf://path_to_file/?PROFILE=hdfs:fixedwidth&NEWLINE=CLRF')
FORMAT 'custom' (formatter='fixedwidth_in', s='5', n='10', line_delim=E'\r\n');

CREATE WRITABLE EXTERNAL TABLE fixed_write (s text, n integer)
LOCATION ('pxf://path_to_directory/?PROFILE=hdfs:fixedwidth&NEWLINE=CLRF')
FORMAT 'custom' (formatter='fixedwidth_out', s='5', n='10', line_delim=E'\r\n');
```


external-table/src/pxfheaders.c
```diff
@@ -617,32 +618,76 @@ add_location_options_httpheader(CHURL_HEADERS headers, GPHDUri *gphduri)
...
+isFormatterPxfWritable(ExtTableEntry *exttbl)
+{
+	char *formatterNameSearchString = NULL;
+
+#if PG_VERSION_NUM >= 120000
+	// for GP7 the custom formatter name is among all other options of the corresponding foreign table
+	ListCell   *option;
+	foreach (option, exttbl->options) {
+		DefElem *defel = (DefElem *) lfirst(option);
+		if (strcmp(defel->defname, "formatter") == 0) {
```
bradfordb-vmware, line 670: Do we need to worry about the case of the `formatter` option? Most of the PXF documentation gives examples with `FORMATTER` but I think the format options get converted to lower case.

automation/src/test/resources/data/fixed/fixed_small_correct.txt
bradfordb-vmware, line 1: Minor nit: Can we make use `fixedwidth` instead of `fixed`? I initially interpreted this as "corrected".

external-table/src/pxfheaders.c
yimingli-vmware, line 1: This file is on the code path for both read case and write case, right?

server/pxf-service/src/main/resources/pxf-profiles-default.xml
```diff
@@ -261,6 +261,17 @@ under the License.
             <resolver>org.greenplum.pxf.plugins.hdfs.StringPassResolver</resolver>
         </plugins>
     </profile>
+    <profile>
+        <name>hdfs:fixed</name>
```
yimingli-vmware, line 265: As Bradford commented in another place, would it be much more clear to use `fixedwidth` instead of `fixed`?

external-table/src/pxfheaders.c
```diff
@@ -617,32 +618,76 @@ add_location_options_httpheader(CHURL_HEADERS headers, GPHDUri *gphduri)
...
 	{
-		formatName = GpdbWritableFormatName;
+		// need to determine if the formatter is pxfwritable_import or pxfwritable_export which requires
+		// us to send to PXF the value of on-the-wire format as "GPDBWritable"
+		if (isFormatterPxfWritable(exttbl)) {
+			formatName = GpdbWritableFormatName;
+		}
+		else
+		{
+			// treat other custom formatters (such as 'fixed_in') as TEXT for PXF transport
```
yimingli-vmware, line 642:
```suggestion
			// treat other custom formatters (such as 'fixedwidth_in') as TEXT for PXF transport
```

external-table/src/pxfheaders.c
denalex: good point, tested and figured out the case in DDL does not matter. Added an automation test to prove that, the test creates a table with `CREATE EXTERNAL TABLE fixedwidth_in_small_correct_alternating_case_ddl (...) LOCATION (E'pxf://tmp/ ... ?PROFILE=hdfs:fixedwidth') FORMAT 'custom' (FoRmAtTeR='fixedwidth_in', s1='10' ...)`

yimingli-vmware (approved):

hpandeycodeit (approved):

automation/src/main/java/org/greenplum/pxf/automation/structures/tables/pxf/ExternalTable.java
```diff
@@ -183,7 +190,13 @@ public String constructCreateStmt() {
         }
 
         if (getFormatter() != null) {
-            createStatment += " (formatter='" + getFormatter() + "')";
+            String formatterOption = isFormatterMixedCase() ? "FoRmAtTeR" : "formatter";
+            createStatment += String.format(" (%s='%s'", formatterOption, getFormatter());
```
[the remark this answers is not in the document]
denalex: not sure what you mean @hpandeycodeit . this is new function that allows us to mangle the word "formatter" into a mixed case to simulate a similar action from a user and to make sure code still works. we did not have it before and were not checking it and were using all lower case when creating the table. Now we can set this option, check it here and can generated a mixed-case word.

hpandeycodeit (approved):

PR #952, opened 2023-03-23
remove RESOURCE QUEUE from test SQL

yimingli-vmware:
GP7 made a change related to `RESOURCE QUEUE` in this [commit](https://github.com/greenplum-db/gpdb/commit/82851a0b855c0c980422f3541d2f770470cbd8b9), which brings extra warning message when running `CREATE ROLE testuser LOGIN RESOURCE QUEUE pg_default` in GP7 automation tests. 
- Since PXF doesn't care about `RESOURCE QUEUE`. We decide to remove `RESOURCE QUEUE` from our automation testing sql. 
- At the meantime, `CREATE ROLE testuser LOGIN` will generate another `NOTICE` message when running GP6 automation tests, so a `matchsubs` is added in `global_init_file` to ignore all these `NOTICE` messages.

### DIFF
--- a/automation/tincrepo/main/pxf/features/hdfsha/step_1_pre_failover/expected/query02.ans
+++ b/automation/tincrepo/main/pxf/features/hdfsha/step_1_pre_failover/expected/query02.ans
@@ -10,7 +10,7 @@ GRANT
 \set OLD_GP_USER :USER
 DROP ROLE IF EXISTS testuser;
 DROP ROLE
-CREATE ROLE testuser LOGIN RESOURCE QUEUE pg_default;
+CREATE ROLE testuser LOGIN;
 CREATE ROLE
 \connect - testuser
 SELECT * FROM pxf_hdfsha_hdfs_ipa ORDER BY name;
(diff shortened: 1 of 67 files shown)


automation/tincrepo/main/pxf/features/hdfsha/step_1_pre_failover/expected/query02.ans
```diff
@@ -10,7 +10,7 @@ GRANT
 \set OLD_GP_USER :USER
 DROP ROLE IF EXISTS testuser;
 DROP ROLE
-CREATE ROLE testuser LOGIN RESOURCE QUEUE pg_default;
+CREATE ROLE testuser LOGIN;
```
hpandeycodeit, line 13: I think we can add this in the  [Gpdb.java ](https://github.com/greenplum-db/pxf/blob/ee05b2655d5db6ad74287604492909a06407b979/automation/src/main/java/org/greenplum/pxf/automation/components/gpdb/Gpdb.java) instead of adding it to every file.

yimingli-vmware: this query is only used in 5 tests

ashuka24 (approved):

denalex (approved):

hpandeycodeit (approved):

bradfordb-vmware (approved):

PR #954, opened 2023-03-25
Update table options names to not include dash character

denalex:
Existing PXF table options that have "-" character in their names cannot be used with FDW where foreign table options do not allow dashes in the names. Currently, 4 options have been identified:

DATA-SCHEMA
THREAD-SAFE
STATS-SAMPLE-RATIO
STATS-MAX-FRAGMENTS

The THREAD-SAFE option is not used anywhere anymore. The other 3 options have been renamed to:
DATA_SCHEMA
STATS_SAMPLE_RATIO
STATS_MAX_FRAGMENTS

The new names should be used now, but the old ones will still work. If both the new and the old options are defined, the new option's value will take effect.

### DIFF
--- a/automation/src/test/java/org/greenplum/pxf/automation/features/writable/HdfsWritableTextTest.java
+++ b/automation/src/test/java/org/greenplum/pxf/automation/features/writable/HdfsWritableTextTest.java
@@ -1,6 +1,5 @@
 package org.greenplum.pxf.automation.features.writable;
 
-import annotations.WorksWithFDW;
 import org.apache.commons.lang.StringUtils;
 import org.greenplum.pxf.automation.datapreparer.writable.WritableDataPreparer;
 import org.greenplum.pxf.automation.enums.EnumCompressionTypes;
--- a/automation/src/test/java/org/greenplum/pxf/automation/features/writable/HdfsWritableTextTest.java
+++ b/automation/src/test/java/org/greenplum/pxf/automation/features/writable/HdfsWritableTextTest.java
@@ -27,7 +27,6 @@ import static java.lang.Thread.sleep;
 /**
  * Testing cases for PXF Writable feature for Text formats (Text, CSV) and compressions.
  */
-@WorksWithFDW
 public class HdfsWritableTextTest extends BaseWritableFeature {
 
     private static final String COMPRESSION_CODEC = "org.apache.hadoop.io.compress.DefaultCodec";
--- a/automation/src/test/java/org/greenplum/pxf/automation/features/writable/HdfsWritableTextTest.java
+++ b/automation/src/test/java/org/greenplum/pxf/automation/features/writable/HdfsWritableTextTest.java
@@ -84,44 +84,6 @@ public class HdfsWritableTextTest extends BaseWritableFeature {
     }
 
     /**
-     * Insert data using "THREAD-SAFE=FALSE" option literally
-     *
-     * @throws Exception if test fails to run
-     */
-    @Test(groups = {"features", "gpdb", "hcfs", "security"})
-    public void textFormatInsertThreadSafeFalse() throws Exception {
-
-        String hdfsPath = hdfsWritePath + "/text_format_thread_safe_false";
-        writableExTable = prepareWritableTable("pxf_text_format_thread_safe_false", hdfsPath, null, null, new String[]{"THREAD-SAFE=FALSE"});
-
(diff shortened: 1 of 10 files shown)


automation/src/test/java/org/greenplum/pxf/automation/features/writable/HdfsWritableTextTest.java
```diff
@@ -1,5 +1,6 @@
 package org.greenplum.pxf.automation.features.writable;
 
+import annotations.WorksWithFDW;
```
bradfordb-vmware, line 3: Is this import used in this file? I don't see any reference to it elsewhere in the file.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/DataSchemaException.java
```diff
@@ -30,7 +30,7 @@ public class DataSchemaException extends RuntimeException {
     public static enum MessageFmt {
 		SCHEMA_NOT_INDICATED("%s requires a data schema to be specified in the "+
 							 "pxf uri, but none was found. Please supply it" +
-							 "using the DATA-SCHEMA option "),
+							 "using the DATA_SCHEMA option "),
```
bradfordb-vmware, line 33: Will this potentially cause confusion if the user is still using `DATA-SCHEMA` but the error message refers to `DATA_SCHEMA`?

Would it make sense to log a warning when user's use the deprecated option names? As an example, Spark [logs a warning][1] when a deprecated configuration option is used.

[1]: https://github.com/apache/spark/blob/d3338c6879fd3e0b986654889f2e1e6407988dcc/core/src/main/scala/org/apache/spark/SparkConf.scala#L763-L777

bradfordb-vmware (approved):

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/WritableResolver.java
```diff
@@ -60,7 +61,8 @@ public class WritableResolver extends BasePlugin implements Resolver {
      */
     @Override
     public void afterPropertiesSet() {
-        String schemaName = this.context.getOption("DATA-SCHEMA");
+        // DATA-SCHEMA option is deprecated in favor of DATA_SCHEMA to support FDW foreign table definitions
+        String schemaName = StringUtils.defaultString(context.getOption("DATA_SCHEMA"), context.getOption("DATA-SCHEMA"));
```
ashuka24, line 65: should we be showing a deprecation warning?

server/pxf-service/src/main/java/org/greenplum/pxf/service/HttpRequestParser.java
```diff
@@ -152,12 +152,18 @@ public RequestContext parseRequest(MultiValueMap<String, String> requestHeaders,
         String config = params.removeUserProperty("CONFIG");
         context.setConfig(StringUtils.isNotBlank(config) ? config : context.getServerName());
 
-        String maxFrags = params.removeUserProperty("STATS-MAX-FRAGMENTS");
+        // STATS-MAX-FRAGMENTS is deprecated in favor of STATS_MAX_FRAGMENTS for FDW options support
```
ashuka24, line 155: same here, should this comment be a deprecation warning the user can see?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/WritableResolver.java
denalex: there is no reasonable way to show it when the query succeeds and logging the warning into the log file is useless ( IMO) as nobody will look there anyways, especially when there are no errors

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/DataSchemaException.java
denalex: Not in this case, since this message is only shown when neither option has been provided, so we suggest the user here to use the new option name. I responded for the previous comment that logging a warn in the log file is probably useless since it will likely be not seen by users unless troubleshooting some other errors. I can add it if there's a strong feeling we really need to do so.

hpandeycodeit (approved):

PR #956, opened 2023-03-31
Add pxfdelimited_import formatter to support multibyte delimites for TEXT and CSV profiles

ashuka24:
This PR adds a new formatter, `pxfdelimiter_import` which can be used with `:text` and `:csv` profiles. 

This formatter allows users who have text or CSV files with multi-byte delimiters (such as `¤` or `停`) to load and query tables pointing to these types of files via PXF. This formatter also supports multi-character delimiters (such as `DELIM`).

This formatter can be used by creating an external table similar to the following: 
```
CREATE READABLE EXTERNAL TABLE multibyte_example (id int, city text, state text)
LOCATION ('pxf://multibyte_exmaple?PROFILE=hdfs:text')
FORMAT 'CUSTOM'
(FORMATTER='pxfdelimited_import', DELIMITER='¤'); 
```

The following format options are also available for use: 
* `quote` (default value `"`)
* `escape` (default value is the `quote` value)
* `newline` (default value is `LF`, options are `LF`, `CR`, `CRLF`)

The introduction of this PR also bumps the PXF extension version from 2.0 to 2.1, and use of this formatter requires users who already have an existing database with the PXF extension to run the following command; 
```
ALTER EXTENSION pxf UPDATE;
```


automation/src/test/java/org/greenplum/pxf/automation/features/multibytedelimiter/MultibyteDelimiterTest.java
```diff
@@ -0,0 +1,669 @@
...
+        // verify results
+        runTincTest("pxf.features.multibyte_delimiter.multi_char.runTest");
+    }
+
+    @Test(groups = {"features", "gpdb", "hcfs", "security"})
+    public void readTwoByteDelimiterWithCRLF() throws Exception {
+        // set profile and format
+        exTable.setName("pxf_multibyte_twobyte_withcrlf_data");
+        exTable.setProfile(protocol.value() + ":csv");
+        exTable.setFormatterOptions(new String[] {"delimiter='¤'", "NEWLINE='CRLF'"});
```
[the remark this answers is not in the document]
ashuka24: We are only allowing `NEWLINE` values for CR, LF and CRLF. LineRecordReader already knows how to handle these three situations. On the formatter side, there is added logic to map `CR` to `\r`, `LF` to `\n` and `CRLF` to `\r\n`. Thus, we do not need to add anything to the location URI for this.

automation/src/test/java/org/greenplum/pxf/automation/features/multibytedelimiter/MultibyteDelimiterTest.java
```diff
@@ -0,0 +1,669 @@
...
+                '¤', CSVWriter.DEFAULT_QUOTE_CHARACTER, CSVWriter.DEFAULT_ESCAPE_CHARACTER, CSVWriter.DEFAULT_LINE_END);
+        // copy local CSV to HDFS
+        hdfs.copyFromLocal(tempLocalDataPath, hdfsFilePath);
+
+        // verify results
+        // in newer versions of GP6 and in GP7, GPDB calls into the formatter one more time to handle EOF properly
+        // however, this is not the case for GP5 and for versions of GP6 older than 6.24.???
+        // therefore, we must run 2 different sets of tests to check for the expected error
+        // TODO: when a version of GP6 gets released with the custom formatter fix, change this if condition
+        if (gpdb.getVersion() > 6) {
```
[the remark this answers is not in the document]
ashuka24: We do indeed run backwards compatibility testing, but the test suites run in that job do not include this one. We only run very minimal set of tests in those jobs. 

Certification testing is forward looking and thus is a non issue.

automation/src/test/java/org/greenplum/pxf/automation/features/multibytedelimiter/MultibyteDelimiterTest.java
```diff
@@ -0,0 +1,669 @@
...
+        // copy local CSV to HDFS
+        hdfs.copyFromLocal(tempLocalDataPath, hdfsFilePath);
+
+        // verify results
+        runTincTest("pxf.features.multibyte_delimiter.two_byte_with_wrong_escape.runTest");
+    }
+
+    // users should still be able to use a normal delimiter with this formatter
+    @Test(groups = {"features", "gpdb", "hcfs", "security"})
+    public void readOneByteDelimiter() throws Exception {
```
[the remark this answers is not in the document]
ashuka24: As it stands right now, if you provide a one-byte delimiter and use the pxfdelimited_import, we do not detect this and call COPY code, instead we will use the custom delimiter logic as is. In theory, doing the detection and calling the other code could work. However, I am not sure how feasible it would be to change the underlying FORMAT type from `custom` to `text` or something similar as that is two completely different code paths within the COPY framework. We would need to look into it more if we want to do this sort underlying switch.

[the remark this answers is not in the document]
bradfordb-vmware: I don't think we need to do that, I was asking more for my own understanding. If a user has a file that uses a single-byte delimiter, they should just use `CSV` or `TEXT` for the format. But it's good that we have coverage to make sure this scenario produces correct results.

external-table/pxf--2.0--2.1.sql
```diff
@@ -0,0 +1,7 @@
+------------------------------------------------------------------
+-- PXF Protocol/Formatters
+------------------------------------------------------------------
+
+CREATE OR REPLACE FUNCTION pg_catalog.pxfdelimited_import() RETURNS record
+AS 'MODULE_PATHNAME', 'multibyte_delim_import'
```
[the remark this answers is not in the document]
ashuka24: What potential conflict do you think we might hit? We do not match catalog names for `gpdbwritableformatter_import` to `pxfwritable_import`, in order to follow that convention, we could use `pxfmulitbyte_delim_import` if needed.

external-table/src/multibyte_delim.c
```diff
@@ -0,0 +1,645 @@
...
+/**
+ * Find the first occurrence of the target given a pointer to the beginning of the string and a pointer to the end of the string
+ * @param target the character to find
+ * @param left_border where to start searching
+ * @param right_border where to stop searching
+ * @param myData struct containing formatter option information
+ * @return pointer to the character immediately after the first instance of the target
+ */
+static char*
+find_first_ins_for_multiline(char* target, char* left_border, char* right_border, format_delimiter_state *myData)
```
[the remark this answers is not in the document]
ashuka24: `target` is a string. It is used for both finding the first instance of `eol` character but is also used for finding the first instance of `delimiter` values. As such, in order to be able to handle the multi-byte characters necessary such as `\r\n` or whatever multi-byte delimiter value, it needs to be a string. 

`ins` as I have internalized it stands for `instance`

external-table/src/multibyte_delim.c
```diff
@@ -0,0 +1,645 @@
...
+
+    if(fmt_state->escape != NULL && strlen(fmt_state->escape) != 1)
+    {
+        ereport(ERROR, (errcode(ERRCODE_INTERNAL_ERROR), errmsg("escape option must be a single character")));
+    }
+
+    fmt_state->situation = (fmt_state->quote != NULL ? WITH_QUOTE : WITHOUT_QUOTE);
+
+    if(fmt_state->situation == WITH_QUOTE) {
+        fmt_state->quote_delimiter = palloc(strlen(fmt_state->delimiter) + 2);
```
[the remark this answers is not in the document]
ashuka24: We set those values in `get_config` which is called within this same function. However, if there is no `quote` formatter option, we do not set the value to null or any other known value.

external-table/src/multibyte_delim.c
```diff
@@ -0,0 +1,645 @@
...
+ * Helper function to handle any escaping that needs to be done
+ * @param start pointer to the beginning of the buffer
+ * @param len total length of the buffer
+ * @param myData struct containing formatter options
+ * @return a new buffer containing a copy of the string that has been properly escaped
+ */
+static char*
+remove_escape(char* start, int len, format_delimiter_state *myData)
+{
+    char* buf = palloc(len + 1);
```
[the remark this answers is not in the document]
ashuka24: This function is actually doing the opposite. If the incoming string is `"This string has an \\ escaped escape"`

The return would be `"This string has an \ escaped escape"`

external-table/src/multibyte_delim.c
```diff
@@ -0,0 +1,645 @@
...
+
+/**
+ * Helper function to handle any escaping that needs to be done
+ * @param start pointer to the beginning of the buffer
+ * @param len total length of the buffer
+ * @param myData struct containing formatter options
+ * @return a new buffer containing a copy of the string that has been properly escaped
+ */
+static char*
+remove_escape(char* start, int len, format_delimiter_state *myData)
```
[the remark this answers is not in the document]
ashuka24: Removing escaping

automation/src/test/java/org/greenplum/pxf/automation/features/multibytedelimiter/MultibyteDelimiterTest.java
```diff
@@ -0,0 +1,669 @@
...
+        // set profile and format
+        exTable.setName("pxf_multibyte_twobyte_withquote_data");
+        exTable.setProfile(protocol.value() + ":csv");
+        exTable.setFormatterOptions(new String[] {"delimiter='¤'", "quote='\"'"});
+        // create external table
+        gpdb.createTableAndVerify(exTable);
+        // create local CSV file
+        String tempLocalDataPath = dataTempFolder + "/data.csv";
+        CsvUtils.writeTableToCsvFileOptions(dataTable, tempLocalDataPath, StandardCharsets.UTF_8,
+                '¤', CSVWriter.DEFAULT_QUOTE_CHARACTER, CSVWriter.DEFAULT_ESCAPE_CHARACTER, CSVWriter.DEFAULT_LINE_END);
```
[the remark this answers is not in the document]
ashuka24: Yes. If you want there to be no quotes, there is a `CSVWriter.NO_QUOTE_CHARACTER` option.

external-table/src/multibyte_delim.h
```diff
@@ -0,0 +1,54 @@
...
+#include "utils/memutils.h"
+#include "utils/typcache.h"
+#include "utils/syscache.h"
+#include "utils/datetime.h"
+
+#include "lib/stringinfo.h"
+
+/* Do the module magic dance */
+//PG_MODULE_MAGIC;
+PG_FUNCTION_INFO_V1(multibyte_delim_import);
```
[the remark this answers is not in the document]
ashuka24: Yes. I have moved it to the c file for consistency

external-table/src/multibyte_delim.c
```diff
@@ -0,0 +1,645 @@
...
+            t = myData->quote_delimiter;
+        }
+        else
+        {
+            t = myData->quote_eol;
+        }
+    }
+
+    char* start_pos = left_border;
+    while(1)
```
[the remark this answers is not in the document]
ashuka24: After tech discussion, we have agreed to avoid modifying too much core logic. As such, this will stay as is for now.

external-table/src/multibyte_delim.h
```diff
@@ -0,0 +1,54 @@
+// Portions Copyright (c) 2023 VMware, Inc. or its affiliates.
+
+#ifndef MULTIBYTE_DELIM_H
```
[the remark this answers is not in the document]
ashuka24: Typically, I believe we match the `ifdef` with the file name. Currently, there is no other file that has this naming.

external-table/src/multibyte_delim.c
```diff
@@ -0,0 +1,645 @@
...
+    }
+
+    while ( (end - (char*)data) < len)
+    {
+        resetStringInfo(buf);
+        end = (char*)data + len;
+        if (index >= myData->nColumns)
+        {
+            ereport(ERROR, (errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+                    errmsg("unpack_delimited: Column mismatch, index:%d >= nColumns:%d, more columns than expected", index, myData->nColumns)));
```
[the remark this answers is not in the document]
ashuka24: No, we don't I have removed the printing of the function name for all of the internal functions in this file

concourse/pipelines/templates/build_pipeline-tpl.yml
```diff
@@ -529,10 +529,18 @@ resources:
 - name: pxf5_gp6_rhel7_released
   type: gcs
   source:
-    bucket: ((ud/pxf/[[ environment ]]/releng-drop-bucket-name))
+    bucket: ((ud/pxf/prod/releng-drop-bucket-name))
     json_key: ((concourse-gcs-resources-service-account-key))
     regexp: ((ud/pxf/common/releng-drop-path))/gpdb6/pxf-gp6-5.(.*)-2.el7.x86_64.rpm
 
+## ---------- PXF 6.5.1 (for GPDB 6) Artifact ---------------
+- name: pxf6_5_1_gp6_rhel7_released
```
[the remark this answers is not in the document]
ashuka24: I have added a comment here for clarity, I will be updating this value to PXF 6.6.X and the names accordingly

concourse/scripts/upgrade_pxf.bash
```diff
@@ -50,14 +50,25 @@ function upgrade_pxf() {
...
+		echoGreen "ALTER EXTENSION pxf UPDATE - for testupgrade database"
+		ssh "${COORDINATOR_HOSTNAME}" "source ${GPHOME}/greenplum_path.sh && psql -d testupgrade -c 'ALTER EXTENSION pxf UPDATE'"
+	fi
+
+	if [[ ${EXISTING_PXF_VERSION} == "6.5.1" ]]; then
+		echoGreen "ALTER EXTENSION pxf UPDATE - for multibyte delimiter tests"
+		ssh "${COORDINATOR_HOSTNAME}" "source ${GPHOME}/greenplum_path.sh && psql -d template1 -c 'ALTER EXTENSION pxf UPDATE'"
+		ssh "${COORDINATOR_HOSTNAME}" "source ${GPHOME}/greenplum_path.sh && psql -d pxfautomation -c 'ALTER EXTENSION pxf UPDATE'"
+		ssh "${COORDINATOR_HOSTNAME}" "source ${GPHOME}/greenplum_path.sh && psql -d pxfautomation_encoding -c 'SELECT * FROM pg_extension'"
+	fi
```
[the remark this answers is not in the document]
ashuka24: I have moved out the changes for extension upgrading into a different job

concourse/pipelines/templates/build_pipeline-tpl.yml
```diff
@@ -1467,6 +1479,31 @@ jobs:
       GP_VER: [[gp_ver]]
       KERBEROS: ((kerberos-enabled))
       PXF_JVM_OPTS: ((pxf-jvm-opts))
+      PXF_VERSION: 6.5.1
+  - task: Extension Test for PXF 6.5.1
```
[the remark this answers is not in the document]
ashuka24: I have moved out the changes for extension upgrading into a different job

automation/tincrepo/main/pxf/features/multibyte_delimiter/two_byte_with_wrong_quote/expected/query01.ans
```diff
@@ -0,0 +1,6 @@
+-- @description query01 for PXF Multibyte delimiter, 2-byte delim with wrong quote cases
+
+SELECT * from pxf_multibyte_twobyte_wrong_quote_data ORDER BY n1;
+ERROR:  quoted delimiter (|¤) not found
```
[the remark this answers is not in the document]
ashuka24: That is correct, if a quote character is given, then we expect all fields to be quoted regardless of if they _need_ to be. 

Here, quoted delimiter refers to the quote of the previous column + the delimiter value.

external-table/src/multibyte_delim.c
```diff
@@ -0,0 +1,645 @@
...
+remove_escape(char* start, int len, format_delimiter_state *myData)
+{
+    char* buf = palloc(len + 1);
+    int j = 0;
+    int eol_len = strlen(myData->eol);
+    int delimiter_len = strlen(myData->delimiter);
+
+    for(int i = 0; i < len;)
+    {
+        if(start[i] == *(myData->escape))
```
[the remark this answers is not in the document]
ashuka24: As the rest of the while loop uses things such as `start[i+1]` etc, I will leave it as `start[i]`

automation/src/main/java/org/greenplum/pxf/automation/components/gpdb/Gpdb.java
```diff
@@ -123,6 +123,12 @@ public void copyData(Table source, Table target) throws Exception {
                 + source.getName());
     }
 
+	public void connectToDataBase(String schemaName) throws Exception {
```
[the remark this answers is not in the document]
ashuka24: I need to be able to connect to the proper database on the java side in order to alter the table. I intentionally had it there as opposed to the SQL files inside to for more clarity on what we were doing prior to the test. 

What we do underlying when we first create the connection is we also use jdbc to connect (see Gpdb.java:76), however, I did not want to run through the entire init process again, so instead we are just changing the connection to the new database.

server/pxf-service/src/scripts/pxf-pre-gpupgrade
```diff
@@ -112,6 +112,10 @@ psql --no-align --tuples-only --command "SELECT datname FROM pg_catalog.pg_datab
 		CREATE OR REPLACE FUNCTION pg_catalog.pxfwritable_export(record) RETURNS bytea
 		AS 'pxf', 'gpdbwritableformatter_export'
 		LANGUAGE C STABLE;
+
+		CREATE OR REPLACE FUNCTION pg_catalog.pxfdelimited_import() RETURNS record
```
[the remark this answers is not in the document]
ashuka24: The scripts have been changed to first query the database for the version of the PXF extension before it updates that function. I have also introduced a new test to the `GpupgradeTest.java` file to make sure that it works with both the old and new extension versions.

automation/src/test/java/org/greenplum/pxf/automation/features/multibytedelimiter/MultibyteDelimiterTest.java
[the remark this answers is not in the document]
ashuka24: In the case of CR, we would indeed have to add `NEWLINE` to the location URI. I have added a test case for this.

external-table/src/multibyte_delim.c
```diff
@@ -0,0 +1,645 @@
...
+            // if we didn't find the right quote in the buf
+            if(p >= data_border)
+            {
+                return NULL;
+            }
+
+            if(myData->escape != NULL)
+            {
+                int cnt = count_of_escape(p-1, data, *(myData->escape));
+                if(cnt % 2 == 1)
```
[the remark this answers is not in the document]
ashuka24: After tech discussion, we have agreed to avoid modifying too much core logic. As such, despite the inefficiency, we will keep the logic of backtracking to count the number of escapes as is.

external-table/src/multibyte_delim.h
```diff
@@ -0,0 +1,54 @@
...
+#include "utils/syscache.h"
+#include "utils/datetime.h"
+
+#include "lib/stringinfo.h"
+
+/* Do the module magic dance */
+//PG_MODULE_MAGIC;
+PG_FUNCTION_INFO_V1(multibyte_delim_import);
+
+typedef enum DelimitedSituation
```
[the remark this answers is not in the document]
ashuka24: Instead of creating a boolean, I removed enum completely and just do a check on if the quote value != null

external-table/src/multibyte_delim.c
```diff
@@ -0,0 +1,645 @@
...
+            }
+        }
+        else
+        {
+            FORMATTER_SET_BAD_ROW_DATA(fcinfo, data_buf + data_cur, remaining);
+            ereport(ERROR, (errcode(ERRCODE_DATA_EXCEPTION),
+                            errmsg("unexpected end of file (multibyte case)")));
+        }
+
+		FORMATTER_RETURN_NOTIFICATION(fcinfo, FMT_NEED_MORE_DATA);
```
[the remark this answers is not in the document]
ashuka24: I have refactored this and added more comments to hopefully make it more clear

automation/src/main/java/org/greenplum/pxf/automation/utils/csv/CsvUtils.java
```diff
@@ -46,6 +51,24 @@ public static Table getTable(String pathToCsvFile) throws IOException {
...
+	/**
+	 * Update the delimiter in a CSV file. This helper function is required as CSVWriter only allows
+	 * for single-character delimiters. As such, for test cases that have multi-character delimiters
+	 * we have to update the delimiter in the file after the CSV has been created
+	 *
+	 * @param originalDelim Original single char delimiter
+	 * @param newDelimiter Desired multi-char delimiter
+	 * @throws IOException
+	 */
+	public static void updateDelim(String targetCsvFile, char originalDelim, String newDelimiter)
```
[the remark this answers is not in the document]
ashuka24: I have not been able to find another CSV package that can handle multi-character delimiters. As such, I have left this as is for now.

external-table/Makefile
```diff
@@ -1,7 +1,7 @@
 EXTENSION   = pxf
-DATA        = pxf--2.0.sql pxf--1.0--2.0.sql pxf--1.0.sql
+DATA        = pxf--2.1.sql pxf--2.0--2.1.sql pxf--2.1--2.0.sql pxf--2.0.sql pxf--1.0--2.0.sql pxf--1.0.sql
```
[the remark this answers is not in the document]
ashuka24: My understanding is that if there is no direct path available, then the next fastest path would be taken. For example, if there was a `1.0--1.1` and  a `1.0--1.2` and a `1.1--1.2` and a  `1.2--1.3`. The upgrade path from 1.0 to 1.3 would be `1.0--1.2` then `1.2--1.3` not `1.0--1.1`, `1.1--1.2` and then `1.2--1.3`. However, if there is a script file that directly builds the desired version directly, it would still take it.

concourse/scripts/update_pxf_minor_version.bash
```diff
@@ -0,0 +1,50 @@
...
+GPHOME=/usr/local/greenplum-db-devel
+
+# we need word boundary in case of standby coordinator (scdw)
+COORDINATOR_HOSTNAME="cdw"
+PXF_HOME=/usr/local/pxf-gp${GP_VER}
+PXF_BASE_DIR=${PXF_BASE_DIR:-$PXF_HOME}
+
+# ANSI Colors
+echoRed() { echo $'\e[0;31m'"$1"$'\e[0m'; }
+echoGreen() { echo $'\e[0;32m'"$1"$'\e[0m'; }
```
denalex, line 15: why do we need color for concourse scripts ?

concourse/scripts/update_pxf_minor_version.bash
```diff
@@ -0,0 +1,50 @@
...
+	echoGreen "Check the PXF 6 version"
+	su gpadmin -c "${PXF_HOME}/bin/pxf version"
+
+	echoGreen "Register the PXF extension into Greenplum"
+	su gpadmin -c "GPHOME=${GPHOME} ${PXF_HOME}/bin/pxf cluster register"
+
+	if [[ "${PXF_BASE_DIR}" != "${PXF_HOME}" ]]; then
+		echoGreen "Prepare PXF in ${PXF_BASE_DIR}"
+		PXF_BASE=${PXF_BASE_DIR} ${PXF_HOME}/bin/pxf cluster prepare
+		echo \"export PXF_BASE=${PXF_BASE_DIR}\" >> ~gpadmin/.bashrcz
```
denalex, line 34: is it supposed to be `.bashrc` or `.bashrcz` ?

concourse/scripts/update_pxf_minor_version.bash
```diff
@@ -0,0 +1,50 @@
...
+	updated_pxf_version=$(cat $PXF_HOME/version)
+
+	echoGreen "Starting PXF ${updated_pxf_version}"
+
+	if [[ ${existing_pxf_version} > ${updated_pxf_version} ]]; then
+		echoRed "Existing version of PXF (${existing_pxf_version}) is greater than or equal to the new version (${updated_pxf_version})"
+	fi
+
+	su gpadmin -c "PXF_BASE=${PXF_BASE_DIR} ${PXF_HOME}/bin/pxf cluster start"
+
```
denalex, line 45: we need comments on what we are trying to do here. "we just upgraded the whole PXF, that brought in an extension with a higher version, so we need to explicitly upgrade extension installed in GPDB to the new version, ... "

concourse/scripts/update_pxf_minor_version.bash
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
+
+SCRIPT_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
+source "${SCRIPT_DIR}/pxf_common.bash"
+
+GPHOME=/usr/local/greenplum-db-devel
+
+# we need word boundary in case of standby coordinator (scdw)
+COORDINATOR_HOSTNAME="cdw"
```
denalex, line 9: do we need this, is it only gonna run on GP7 ?

concourse/scripts/update_pxf_minor_version.bash
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
```
denalex, line 1: should the script be called `upgrade...` instead of `update...` ? Also, which minor version ? We need more docs here what this script is doing and what it expects as inputs.

external-table/src/pxfdelimited_formatter.h
```diff
@@ -0,0 +1,48 @@
...
+#include "utils/builtins.h"
+#include "utils/memutils.h"
+#include "utils/typcache.h"
+#include "utils/syscache.h"
+#include "utils/datetime.h"
+
+#include "lib/stringinfo.h"
+
+/* Do the module magic dance */
+//PG_MODULE_MAGIC;
```
denalex, line 22: if we do not need this, should we remove the line ?

external-table/src/pxfdelimited_formatter.h
```diff
@@ -0,0 +1,48 @@
...
+	char *quote_delimiter; /* only for searching for border, not in the config file */
+	char *quote_eol; /* only for searching for border, not in the config file */
+	int nColumns;
+
+	int external_encoding;				/* remote side's character encoding */
+	FmgrInfo *enc_conversion_proc;		/* conv proc from exttbl encoding to
+										server or the other way around */
+	bool saw_delim;
+	bool saw_eol;
+} format_delimiter_state;
```
denalex, line 43:
```suggestion
} pxfdelimited_state;
```
I'd prefer this type to have pxf namespace

external-table/src/pxfheaders.c
```diff
@@ -90,6 +92,20 @@ build_http_headers(PxfInputData *input)
 		ListCell   *option;
 		List	   *copyFmtOpts = NIL;
 
+		// in the case of pxfdelimited_import formatter, the only viable profiles are TEXT and CSV.
+		// error out early here if the profile is not accepted
+		if (getFormatterString(exttbl) != NULL &&
```
denalex, line 97:
```suggestion
		if (getFormatterString(exttbl) &&
```
do we need explicit NULL comparisons in C since a pointer will be >0 and evaluate to `true` in conditionals ?

external-table/src/pxfheaders.c
```diff
@@ -90,6 +92,20 @@ build_http_headers(PxfInputData *input)
 		ListCell   *option;
 		List	   *copyFmtOpts = NIL;
 
+		// in the case of pxfdelimited_import formatter, the only viable profiles are TEXT and CSV.
```
denalex, line 95:
```suggestion
		// in the case of pxfdelimited_import formatter, the only viable profiles are '*:test' and '*:csv'.
```

external-table/src/pxfheaders.c
```diff
@@ -90,6 +92,20 @@ build_http_headers(PxfInputData *input)
...
+		// error out early here if the profile is not accepted
+		if (getFormatterString(exttbl) != NULL &&
+		    isFormatterPxfDelimited(exttbl) &&
+			(!strstr(input->gphduri->profile, ":text") &&
+				!strstr(input->gphduri->profile, ":csv")))
```
denalex, line 100: 2 points here:
- profile might not be defined (old style with frag / acc / resolver) specified directly. not sure if you want to error out in this case (probably a good idea to do so)
- if profile is not defined, `strstr(NULL, "...")` might segfault, so better to double check

external-table/src/pxfheaders.c
```diff
@@ -673,11 +691,39 @@ isFormatterPxfWritable(ExtTableEntry *exttbl)
...
+	char *formatterNameSearchString = getFormatterString(exttbl);
+
+	if (!formatterNameSearchString || !strlen(formatterNameSearchString))
+	{
+		ereport(ERROR,
+				(errcode(ERRCODE_INTERNAL_ERROR),
+						errmsg("cannot determine the name of a custom formatter")));
+	}
+
+	return strstr(formatterNameSearchString, "pxfdelimited_import") != NULL;
```
denalex, line 717: we use string literal "pxfdelimited_import" is a few places, should we `#define` it ?

server/pxf-service/src/scripts/pxf-post-gpupgrade
```diff
@@ -47,6 +47,7 @@ EOF
 
 pxf_gpdb_major_version="$(awk 'BEGIN { FS = "=" } /gpdb.major-version/{ print $2 }' "${PXF_HOME}/gpextable/metadata")"
 gp_version="$(psql --no-align --tuples-only --command 'SELECT substring(version(), $$.*Greenplum Database (.*) build.*$$)')"
+pxf_version="$(cat $PXF_HOME/version)"
```
denalex, line 50: why is this needed if it worked before and the new code does not refer to this variable ?

external-table/src/pxfdelimited_formatter.c
```diff
@@ -0,0 +1,715 @@
...
+	}
+
+	while ( (end - (char*)data) < len)
+	{
+		resetStringInfo(buf);
+		end = (char*)data + len;
+		if (index >= myData->nColumns)
+		{
+			ereport(ERROR, (errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+					errmsg("Expected %d columns in row but found %d", myData->nColumns, index),
```
hpandeycodeit, line 450:
```suggestion
					errmsg("Expected %d columns but found %d" in the row, myData->nColumns, index),
```

external-table/src/pxfdelimited_formatter.c
```diff
@@ -0,0 +1,715 @@
...
+		--len;
+	}
+
+	while ( (end - (char*)data) < len)
+	{
+		resetStringInfo(buf);
+		end = (char*)data + len;
+		if (index >= myData->nColumns)
+		{
+			ereport(ERROR, (errcode(ERRCODE_INVALID_PARAMETER_VALUE),
```
hpandeycodeit, line 449: In case of a column mismatch, the database will throw an error. Do we need to provide our own error?

external-table/src/pxfdelimited_formatter.c
```diff
@@ -0,0 +1,715 @@
...
+					errmsg("Expected %d columns in row but found %d", myData->nColumns, index),
+					errhint("Please verify the number of columns in the table definition.")));
+		}
+		if (start == NULL) {
+			ereport(ERROR, (errcode(ERRCODE_NULL_VALUE_NOT_ALLOWED),
+					errmsg("Unexpected null value found while trying to read data"),
+					errhint("Please verify the number of columns in the table definition.")));
```
hpandeycodeit, line 456: This is the same message at two places, I think this can be taken out in a variable.

external-table/src/pxfdelimited_formatter.c
```diff
@@ -0,0 +1,715 @@
...
+		}
+
+		if (myData->quote != NULL && *(start-1) != *myData->quote)
+		{
+			ereport(ERROR, (errcode(ERRCODE_DATA_EXCEPTION),
+					errmsg("Missing quote before some column"),
+					errhint("Please verify that are columns in the data are properly quoted.")));
```
hpandeycodeit, line 463: same with this one.

external-table/src/pxfdelimited_formatter.c
```diff
@@ -0,0 +1,715 @@
...
+ * Main formatter function.
+ * @return
+ */
+Datum
+pxfdelimited_import(PG_FUNCTION_ARGS)
+{
+	HeapTuple		   tuple;
+	TupleDesc		   tupdesc;
+	MemoryContext	   m, oldcontext;
+	pxfdelimited_state		   *myData;
```
hpandeycodeit, line 518: These variables need alignment.

external-table/src/pxfdelimited_formatter.c
```diff
@@ -0,0 +1,715 @@
...
+	data_buf = FORMATTER_GET_DATABUF(fcinfo);
+	data_len = FORMATTER_GET_DATALEN(fcinfo);
+	data_cur = FORMATTER_GET_DATACURSOR(fcinfo);
+
+	int	remaining = data_len - data_cur;
+
+	/*
+	 * NOTE: Unexpected EOF Error Handling
+	 *
+	 * The first time we noticed an unexpected EOF, we'll set the datacursor
```
hpandeycodeit, line 557:
```suggestion
	 * The first time we noticed an unexpected EOF, we'll set the data cursor
```

external-table/src/pxfdelimited_formatter.c
```diff
@@ -0,0 +1,715 @@
...
+	{
+		ereport(ERROR, (errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+				errmsg("Expected %d columns in row but found %d", myData->nColumns, index),
+				myData->saw_delim ? errhint("Please verify the number of columns in the table definition.")
+									: errhint("Is the `delimiter` value in the format options set correctly?")));
+	}
+}
+
+/**
+ * Main formatter function.
```
hpandeycodeit, line 509: Can we have some description here of what this function is doing?

concourse/scripts/test_upgrade_extension.bash
```diff
@@ -0,0 +1,138 @@
...
+	# Must be after installing GPDB to transfer hbase jar
+	setup_hadoop "${GPHD_ROOT}"
+
+	# initialize GPDB as gpadmin user
+	su gpadmin -c "${CWDIR}/initialize_gpdb.bash"
+
+	add_remote_user_access_for_gpdb testuser
+	configure_pxf_server
+
+	local HCFS_BUCKET # team-specific bucket names
```
bradfordb-vmware, line 86: This variable is unused -- do we need it?

concourse/scripts/test_upgrade_extension.bash
```diff
@@ -0,0 +1,138 @@
...
+		time make GROUP=${GROUP} test
+	EOF
+
+	chown gpadmin:gpadmin ~gpadmin/run_pxf_automation_test.sh
+	chmod a+x ~gpadmin/run_pxf_automation_test.sh
+
+	su gpadmin -c ~gpadmin/run_pxf_automation_test.sh
+}
+
+function generate_extras_fat_jar() {
```
bradfordb-vmware, line 51: This function is identical to the one in `test.bash`, so its a good candidate for moving to `common.sh`.

concourse/tasks/upgrade_extension.yml
```diff
@@ -0,0 +1,29 @@
...
+  - name: pxf_src
+  - name: gpdb_package
+  - name: singlecluster
+    optional: true
+  - name: pxf-automation-dependencies
+    optional: true
+
+params:
+  GP_VER:
+  FIRST_GROUP:
```
bradfordb-vmware, line 18: I think we need some kind of doc comment for what `FIRST_GROUP` and `SECOND_GROUP` are and how they are being used. Quickly reading through the code for `test_upgrade_extension.bash`, I get the impression that the script is called twice -- once with `FIRST_GROUP` set and `SECOND_GROUP` unset and then again with `FIRST_GROUP` unset and `SECOND_GROUP` set. But I don't that's the case.

Additionally, I think the `test_upgrade_extension.bash` is specialized to running a specific set of test groups. I don't know that we need to "generalize" for different sets of tests to run before upgrading and after upgrading.

concourse/scripts/test_upgrade_extension.bash
```diff
@@ -0,0 +1,138 @@
...
+
+	# Create fat jar for automation
+	generate_extras_fat_jar
+
+	inflate_dependencies
+
+	ln -s "${PWD}/pxf_src" ~gpadmin/pxf_src
+
+	# Run tests
+	if [[ -n ${FIRST_GROUP} ]]; then
```
bradfordb-vmware, line 99: When would `FIRST_GROUP` ever be unset or set to an empty string?

concourse/scripts/test_upgrade_extension.bash
```diff
@@ -0,0 +1,138 @@
...
+	echo "*                                       Upgrading PXF                                              *"
+	echo '****************************************************************************************************'
+	echo
+	echo
+
+	# Upgrade from older version of PXF to newer version of PXF present in the tarball
+	upgrade_pxf
+
+	# Run tests
+	if [[ -n ${SECOND_GROUP} ]]; then
```
bradfordb-vmware, line 129: When would `SECOND_GROUP` be unset?

concourse/scripts/update_pxf_minor_version.bash
```diff
@@ -0,0 +1,49 @@
...
+
+CWDIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
+source "${CWDIR}/pxf_common.bash"
+
+# we need word boundary in case of standby coordinator (scdw)
```
bradfordb-vmware, line 6: What 'word boundary' is this referring to?

concourse/scripts/update_pxf_minor_version.bash
```diff
@@ -0,0 +1,49 @@
...
+PXF_HOME=/usr/local/pxf-gp${GP_VER}
+PXF_BASE_DIR=${PXF_BASE_DIR:-$PXF_HOME}
+
+# ANSI Colors
+echoRed() { echo $'\e[0;31m'"$1"$'\e[0m'; }
+echoGreen() { echo $'\e[0;32m'"$1"$'\e[0m'; }
+
+function upgrade_pxf() {
+	existing_pxf_version=$(cat $PXF_HOME/version)
```
bradfordb-vmware, line 15: Double quote to prevent globbing and word splitting.

concourse/scripts/update_pxf_minor_version.bash
```diff
@@ -0,0 +1,49 @@
...
+
+	echoGreen "Check the PXF 6 version"
+	su gpadmin -c "${PXF_HOME}/bin/pxf version"
+
+	echoGreen "Register the PXF extension into Greenplum"
+	su gpadmin -c "GPHOME=${GPHOME} ${PXF_HOME}/bin/pxf cluster register"
+
+	if [[ "${PXF_BASE_DIR}" != "${PXF_HOME}" ]]; then
+		echoGreen "Prepare PXF in ${PXF_BASE_DIR}"
+		PXF_BASE=${PXF_BASE_DIR} ${PXF_HOME}/bin/pxf cluster prepare
```
bradfordb-vmware, line 30: Double quote to prevent globbing and word splitting.
```suggestion
		PXF_BASE=${PXF_BASE_DIR} "${PXF_HOME}"/bin/pxf cluster prepare
```

concourse/scripts/update_pxf_minor_version.bash
```diff
@@ -0,0 +1,49 @@
...
+		echo \"export PXF_BASE=${PXF_BASE_DIR}\" >> ~gpadmin/.bashrc
```
bradfordb-vmware, line 31: Why are the double-quotes escaped?This will result in `echo` printing them and `~gpadmin/.bashrc` will contain

```
"export PXF_BASE=..."
```

Did you mean

```sh
echo "export PXF_BASE=\"${PXF_BASE_DIR}\"" >> ~gpadmin/.bashrc
```

so that you'd end up with

```
export PXF_BASE="..."
```

in `~gpadmin/.bashrc`?

concourse/scripts/update_pxf_minor_version.bash
```diff
@@ -0,0 +1,49 @@
...
+	fi
+	updated_pxf_version=$(cat $PXF_HOME/version)
```
bradfordb-vmware, line 33: Double quote to prevent globbing and word splitting.

concourse/scripts/update_pxf_minor_version.bash
bradfordb-vmware: It looks like this file isn't a script but rather a library shell file -- running this file doesn't do anything since it doesn't call the function `upgrade_pxf`. If that's intended, I would rename this script to match the name of the function and make the file not executable (`chmod a-x update_pxf_minor_version.bash`). Is there any reason to not just add this function to `common.sh`?

external-table/pxf--2.0--2.1.sql
```diff
@@ -0,0 +1,7 @@
+------------------------------------------------------------------
+-- PXF Protocol/Formatters
+------------------------------------------------------------------
+
+CREATE OR REPLACE FUNCTION pg_catalog.pxfdelimited_import() RETURNS record
```
bradfordb-vmware, line 5: In the downgrade script, we include an explicit `ALTER EXTENSION pxf DROP FUNCTION`. Do we need to include an explicit `ALTER EXTENSION pxf ADD FUNCTION`? Or is that implied because this is in SQL file for the extension?

server/pxf-service/src/scripts/pxf-post-gpupgrade
```diff
@@ -47,8 +47,9 @@ EOF
...
```
bradfordb-vmware, line 50: Double quote to prevent globbing and word splitting.

server/pxf-service/src/scripts/pxf-post-gpupgrade
```diff
@@ -113,6 +114,15 @@ psql --no-align --tuples-only --command "SELECT datname FROM pg_catalog.pg_datab
 		AS '${PXF_HOME}/gpextable/pxf', 'gpdbwritableformatter_export'
 		LANGUAGE C STABLE;
 	END_OF_SQL
+
+  # PXF extension 2.1 introduces a new function, only update the function if it exists
+  if [[ $(psql --dbname="${dbname}" --no-align --tuples-only --command "SELECT extversion FROM pg_catalog.pg_extension WHERE extname = 'pxf'") > 2.0 ]]; then
```
bradfordb-vmware, line 119: This is potentially problematic code since bash and POSIX sh don't support decimals in numbers [ShellCheck Wiki](https://www.shellcheck.net/wiki/SC2072); however, it probably OK since we are doing string comparison here (`>` instead of `-gt`).

We should either put a [ShellCheck Directive](https://www.shellcheck.net/wiki/Directive) here so that running `shellcheck` on the script doesn't raise any issues. We should also include a comment about why this is OK here.

server/pxf-service/src/scripts/pxf-pre-gpupgrade
```diff
@@ -47,8 +47,9 @@ EOF
 
 pxf_gpdb_major_version="$(awk 'BEGIN { FS = "=" } /gpdb.major-version/{ print $2 }' "${PXF_HOME}/gpextable/metadata")"
 gp_version="$(psql --no-align --tuples-only --command 'SELECT substring(version(), $$.*Greenplum Database (.*) build.*$$)')"
+pxf_version="$(cat $PXF_HOME/version)"
```
bradfordb-vmware, line 50: Double quote to prevent globbing and word splitting.

server/pxf-service/src/scripts/pxf-pre-gpupgrade
```diff
@@ -113,6 +114,15 @@ psql --no-align --tuples-only --command "SELECT datname FROM pg_catalog.pg_datab
 		AS 'pxf', 'gpdbwritableformatter_export'
 		LANGUAGE C STABLE;
 	END_OF_SQL
+
+    # PXF extension 2.1 introduces a new function, only update the function if it exists
+    if [[ $(psql --dbname="${dbname}" --no-align --tuples-only --command "SELECT extversion FROM pg_catalog.pg_extension WHERE extname = 'pxf'") > 2.0 ]]; then
```
bradfordb-vmware, line 119: Same comment as from the post upgrade script.

external-table/pxf--2.0--2.1.sql
ashuka24: My understanding is that it is implied because we create the function by running an `ALTER EXTENSION pxf UPDATE to` so whatever gets created as part of this gets added to the extension. 

If we ran this sql file on it's own, then we would need to run `ALTER EXTENSION pxf ADD FUNCTION` to include it in the extension. 

In the case of downgrade, the function needs to first be removed from the extension before we can delete the function itself from the catalog. Unfortunately `ALTER EXTENSION pxf DROP FUNCTION` does not remove the function itself from the catalog.

automation/src/main/java/org/greenplum/pxf/automation/utils/csv/CsvUtils.java
```diff
@@ -54,12 +77,13 @@ public static Table getTable(String pathToCsvFile) throws IOException {
...
-	public static void writeTableToCsvFile(Table table, String targetCsvFile)
+	public static void writeTableToCsvFile(Table table, String targetCsvFile, Charset charset,
+										   char delimiter, char quotechar,
+										   char escapechar, String eol)
 			throws IOException {
 
 		// create CsvWriter using FileWriter
-		CSVWriter csvWriter = new CSVWriter(new FileWriter(targetCsvFile));
-
+		CSVWriter csvWriter = new CSVWriter(new OutputStreamWriter(new FileOutputStream(targetCsvFile), charset), delimiter, quotechar, escapechar, eol);
```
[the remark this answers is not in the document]
ashuka24: yes. The default value for CSV is indeed the comma, but we can allow for other delimiters here

concourse/scripts/update_pxf_minor_version.bash
ashuka24: I refactored this a bit and have removed the file completely. All the code specific to upgrading is now in test_upgrade_extension.bash and we do not call this file anymore.

concourse/scripts/test_upgrade_extension.bash
ashuka24: Originally, there was no default value set in the job task yaml. However, this has been updated to give a default value so now there should not be a time when `SECOND_GROUP` is unset.

concourse/scripts/test_upgrade_extension.bash
ashuka24: Originally, the task yml did not have a default value set for `FIRST_GROUP` so it was possible that it could be unset. However, I refactored a little bit so now `FIRST_GROUP` is no longer used.

concourse/scripts/test_upgrade_extension.bash
```diff
@@ -0,0 +1,138 @@
+#!/bin/bash
+
+set -exo pipefail
+
+CWDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+# make sure GP_VER is set so that we know what PXF_HOME will be
+: "${GP_VER:?GP_VER must be set}"
```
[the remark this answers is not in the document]
ashuka24: This is a variable that comes from CI and used widely across all yml/script files. For consistency I will leave it as `GP_VER`, maybe in another future story we can do some concourse cleanup and include this change then

concourse/tasks/upgrade_extension.yml
ashuka24: The script was originally setting `GROUP=${FIRST_GROUP}` value and then after upgrading, running tests with `GROUP=${SECOND_GROUP}`. However, this has now been modified to run test.bash (the existing test script) and then we set `GROUP=${SECOND_GROUP}` to run the second set of tests in `test_upgrade_extension.bash`

concourse/tasks/upgrade_extension.yml
```diff
@@ -0,0 +1,34 @@
...
+  - name: pxf_src
+  - name: gpdb_package
+  - name: singlecluster
+    optional: true
+  - name: pxf-automation-dependencies
+    optional: true
+
+params:
+  GP_VER:
+  GROUP: pxfCExtensionVersion2
```
denalex, line 18:
```suggestion
  GROUP: pxfExtensionVersion2
```
"Extension" already means C-side extension, a name without C in it is more readable IMO

server/pxf-service/src/scripts/pxf-post-gpupgrade
```diff
@@ -113,6 +114,18 @@ psql --no-align --tuples-only --command "SELECT datname FROM pg_catalog.pg_datab
 		AS '${PXF_HOME}/gpextable/pxf', 'gpdbwritableformatter_export'
 		LANGUAGE C STABLE;
 	END_OF_SQL
+
+  # PXF extension 2.1 introduces a new function, only update the function if it exists
+  # bash and POSIX sh don't support decimals in numbers, however wwe are doing a string comparison
```
denalex, line 119:
```suggestion
  # bash and POSIX sh don't support decimals in numbers, however we are doing a string comparison
```

external-table/src/pxfheaders.c
ashuka24: I added a check to make sure that profile is non-empty by doing `input->gphduri->profile` (implicit check). Is the ask here to include a different error message if profile is null instead of the one already present?

external-table/src/pxfdelimited_formatter.c
```diff
@@ -0,0 +1,715 @@
+// Portions Copyright (c) 2023 VMware, Inc. or its affiliates.
```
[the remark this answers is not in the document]
ashuka24: From past discussions, this was determined to be acceptable.

automation/tincrepo/main/pxf/features/extension_tests/create_extension_rpm/sql/query01.sql
```diff
@@ -0,0 +1,28 @@
...
+
+SHOW dynamic_library_path;
+
+SELECT p.proname, p.prosrc, p.probin
+FROM pg_catalog.pg_extension AS e
+    INNER JOIN pg_catalog.pg_depend AS d ON (d.refobjid = e.oid)
+    INNER JOIN pg_catalog.pg_proc AS p ON (p.oid = d.objid)
+WHERE d.deptype = 'e' AND e.extname = 'pxf'
+ORDER BY 1;
+
```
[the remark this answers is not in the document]
ashuka24: There is no good reason to do it this way. We already have code in other tests  This sql is very similar to how GPDB creates the query if you were to run `\df pg_catalog.pxf*`

automation/tincrepo/main/pxf/features/extension_tests/upgrade/step_1_create_extension_with_older_pxf_version/sql/query01.sql
```diff
@@ -0,0 +1,26 @@
...
+SHOW dynamic_library_path;
+
+SELECT p.proname, p.prosrc, p.probin
+FROM pg_catalog.pg_extension AS e
+    INNER JOIN pg_catalog.pg_depend AS d ON (d.refobjid = e.oid)
+    INNER JOIN pg_catalog.pg_proc AS p ON (p.oid = d.objid)
+WHERE d.deptype = 'e' AND e.extname = 'pxf'
+ORDER BY 1;
+
+SELECT * FROM pxf_upgrade_test ORDER BY num;
```
yimingli-vmware, line 26: add an empty line

automation/src/test/java/org/greenplum/pxf/automation/features/extension/PxfExtensionTest.java
```diff
@@ -0,0 +1,135 @@
...
+
+        // create an external table with the multibyte formatter
+        createReadablePxfTable("default", location_multi, true);
+        gpdb.runQuery("ALTER EXTENSION pxf UPDATE TO '2.1'");
+        runTincTest("pxf.features.extension_tests.explicit_upgrade.step_2_after_alter_extension.runTest");
+    }
+
+    @Test(groups = {"gpdb", "pxfExtensionVersion2_1"})
+    public void testPxfDowngrade() throws Exception {
+        gpdb.runQuery("CREATE EXTENSION pxf");
```
yimingli-vmware, line 93: By default, the extension will be always installed in the latest version, which is `2.1`, right?

automation/src/test/java/org/greenplum/pxf/automation/features/gpupgrade/GpupgradeTest.java
```diff
@@ -34,20 +34,38 @@ protected void afterMethod() throws Exception {
         super.afterMethod();
     }
 
+    @Test(groups = {"features", "gpdb"})
+    public void testGpdbUpgradeExtensionVersion2_0Scenario() throws Exception {
+
```
yimingli-vmware, line 39: Is this one testing a scenario that user is upgrading pxf from `1.0` to `2.0`?

automation/src/test/java/org/greenplum/pxf/automation/features/multibytedelimiter/MultibyteDelimiterTest.java
```diff
@@ -0,0 +1,651 @@
...
+                        "n16 int",
+                        "n17 int"},
+                protocol.getExternalTablePath(hdfs.getBasePath(), hdfsFilePath),
+                null);
+        exTable.setFormat("CUSTOM");
+        exTable.setFormatter("pxfdelimited_import");
+        exTable.setProfile(protocol.value() + ":csv");
+
+        encodedDataTable = new Table("data", null);
+        encodedDataTable.addRow(new String[]{"4", "tá sé seo le tástáil dea-"});
```
yimingli-vmware, line 194: what is encodedDataTable?

automation/src/test/java/org/greenplum/pxf/automation/features/extension/PxfExtensionTest.java
```diff
@@ -0,0 +1,135 @@
...
+        // create a regular external table
+        createReadablePxfTable("default", location, false);
+        // create an external table with the multibyte formatter
+        createReadablePxfTable("default", location_multi, true);
+        runTincTest("pxf.features.extension_tests.create_extension.runTest");
+    }
+
+    @Test(groups = {"pxfExtensionVersion2"})
+    public void testPxfCreateExtensionOldRPM() throws Exception {
```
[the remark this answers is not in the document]
ashuka24: This one is for the older PXF version. The results of this test should NOT contain `pxfdelimited_import`. This test is run only as part of the new upgrade job and is the only test in the `pxfExtensionVersion2` group.

automation/src/test/java/org/greenplum/pxf/automation/features/extension/PxfExtensionTest.java
ashuka24: The default version installed is dictated by the `pxf.control` file. With this PR, the default version installed when doing a new installation will be 2.1

automation/src/test/java/org/greenplum/pxf/automation/features/gpupgrade/GpupgradeTest.java
ashuka24: This test suite makes sure that after upgrading GPDB< we still have a working PXF. With the new default PXF extension as 2.1 the old test is now testing an upgrade of GPDB with the new PXF extension. This test was added to ensure that we could still successfully upgrade GPDB with the older PXF extension 2.0 and have a running PXF after.

automation/src/test/java/org/greenplum/pxf/automation/features/extension/PxfExtensionTest.java
```diff
@@ -0,0 +1,135 @@
...
+        location = hdfs.getWorkingDirectory() + "/upgrade-test-data.txt";
+        hdfs.writeTableToFile(location, smallData, ",");
+
+        location_multi = hdfs.getWorkingDirectory() + "/upgrade-test-data_multibyte.txt";
+        hdfs.writeTableToFile(location_multi, smallData, "停");
+    }
+
+    @Override
+    public void beforeMethod() throws Exception {
+        gpdb.setDb("pxfautomation_extension");
```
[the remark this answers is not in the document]
ashuka24: Before class sets the DB before the entire test suite is run. It is needed for the initial set up. 

Before method is run before every test. I added it here to ensure that we were connecting to the correct DB for each test.

automation/src/test/java/org/greenplum/pxf/automation/features/extension/PxfExtensionTest.java
```diff
@@ -0,0 +1,135 @@
...
+        gpdb.connectToDataBase("pxfautomation_extension");
```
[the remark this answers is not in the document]
ashuka24: line 27 is doing a `createDatabase` and the boolean is `ignoreFail`. Setting it to false there means that if we fail to create the database, we should fail instead of ignoring the failure. 

Here, we are connecting to an existing database. There is no second parameter to `connectToDatabase`

automation/src/test/java/org/greenplum/pxf/automation/features/extension/PxfExtensionTest.java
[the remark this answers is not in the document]
bradfordb-vmware: It doesn't look like we call `setDb` anywhere else in the test class besides `beforeClass` and `beforeMethod` so would the database ever not be `pxfautomation_extension`?

automation/src/test/java/org/greenplum/pxf/automation/features/multibytedelimiter/MultibyteDelimiterTest.java
```diff
@@ -0,0 +1,651 @@
...
+            "2299-11-28 05:46:40",
+            "101",
+            "1001",
+            "10001",
+            "10001",
+            "10001",
+            "10001",
+            "10001"};
+
+    private class CsvSpec  {
```
bradfordb-vmware, line 62: If you add lombok to `automation/pom.xml`, you can then use `@AllArgsConstructor` and `@Setter` to save some lines.

bradfordb-vmware (approved):

external-table/src/pxfdelimited_formatter.c
```diff
@@ -0,0 +1,719 @@
...
+	char *end = (char*)data;
+	StringInfo buf = makeStringInfo();
+	int index = 0;
+	int delimiter_len = strlen(myData->delimiter);
+	int two_quote_len = (myData->quote != NULL ? 2 : 0); // the last quote of this column and the first quote of next column
+
+	char* quote_hint_msg = "Please verify that columns in the data are properly quoted.";
+	char* col_hint_msg = "Please verify the number of columns in the table definition.";
+
+	if(myData->quote != NULL)
```
hpandeycodeit, line 431:
```suggestion
	if (myData->quote != NULL)
```

external-table/src/pxfdelimited_formatter.c
```diff
@@ -0,0 +1,719 @@
...
+	{
+		if(*data != *myData->quote || data[len-1] != *myData->quote)
```
hpandeycodeit, line 433:
```suggestion
		if (*data != *myData->quote || data[len-1] != *myData->quote)
```

external-table/src/pxfdelimited_formatter.c
```diff
@@ -0,0 +1,719 @@
...
+			myData->saw_delim = true;
+		}
+		int column_len = end - start;
+		if (column_len == 0)
+		{
+			myData->nulls[index] = true;
+		}
+		else
+		{
+			if(myData->escape == NULL)
```
hpandeycodeit, line 482:
```suggestion
			if (myData->escape == NULL)
```

external-table/src/pxfdelimited_formatter.c
```diff
@@ -0,0 +1,719 @@
...
+	 *  2. We have found the true eol ex: `" ";" ";" "\n`
+	 *
+	 *  As such, we need to make sure that we correctly find the eol marker and are not taking
+	 *  any value inside quotes as the eol
+	 *
+	 *  The two cases that we need to pay special attention to is option 1b and option 2.
+	 *  	in the former situation, `"\n` is in the beginning
+	 *  	in the latter situation, there must be a delimiter like `;` before `"\n`
+	 */
+	if(myData->quote != NULL &&
```
hpandeycodeit, line 638:
```suggestion
	if (myData->quote != NULL &&
```

external-table/src/pxfdelimited_formatter.c
```diff
@@ -0,0 +1,719 @@
...
+	   (line_border == data_buf + data_cur + 1 || memcmp(line_border - 1  - delimiter_len, myData->delimiter, delimiter_len) == 0) )
+	{
+		/*
+		 * Go through the data_buf and ensure that we have the correct number of quotes and handle
+		 */
+		char *real_line_border = find_whole_line(data_buf + data_cur, data_buf + data_len, myData);
+
+		// if we can't find a whole line by counting quote, we treat this part of data as bad data
+		if(real_line_border == NULL)
```
hpandeycodeit, line 647:
```suggestion
		if (real_line_border == NULL)
```

external-table/src/pxfdelimited_formatter.c
```diff
@@ -0,0 +1,719 @@
...
+		if (p == NULL || p > right_border - strlen(t))
+		{
+			// nothing found or the entire target is not within the bounds dictated
+			break;
+		}
+
+		int escape_count = 0;
+
+		// make sure that the value found is not an escaped representation of the given target
+		if(myData->escape != NULL)
```
hpandeycodeit, line 79:
```suggestion
		if (myData->escape != NULL)
```

external-table/src/pxfdelimited_formatter.c
```diff
@@ -0,0 +1,719 @@
...
+ * @param p where to start searching
+ * @param left_border where to stop searching
+ * @param val the char to count
+ * @return the number of continuous occurrences found of the given character
+ */
+static int
+count_preceding_occurrences_of_char(char *p, char *left_border, char val)
+{
+	int count = 0;
+	while(p >= left_border && *p == val)
```
hpandeycodeit, line 30:
```suggestion
	while (p >= left_border && *p == val)
```

external-table/src/pxfdelimited_formatter.c
```diff
@@ -0,0 +1,719 @@
...
+					errmsg("Missing quote in row head or tail"),
+					errhint("%s", quote_hint_msg)));
+		}
+
+		// exclude the first and the last quote
+		++start;
+		--len;
+	}
+
+	while ( (end - (char*)data) < len)
```
hpandeycodeit, line 445:
```suggestion
	while ((end - (char*)data) < len)
```

hpandeycodeit (approved):
nit.

external-table/src/pxfheaders.c
denalex: if the formatter is specified but profile is not, this will now not segfault but it will not error out either. so if it is specified for a F/A/R combo such as in JDBC profile for example, it will not fail here, but will fail when parsing returned results.

I think it is fair to require the new formatter to be used only when profile is specified and has `:text` and `:csv` in it. Which would require changing these checks a little bit.

automation/src/test/java/org/greenplum/pxf/automation/features/multibytedelimiter/MultibyteDelimiterTest.java
```diff
@@ -0,0 +1,651 @@
...
+        gpdb.createTableAndVerify(exTable);
+
+        // verify results
+        runTincTest("pxf.features.multibyte_delimiter.three_byte.runTest");
+    }
+
+    @Test(groups = {"gpdb", "hcfs", "security"})
+    public void readFourByteDelimiter() throws Exception {
+        CsvSpec fileSpec = new CsvSpec("\uD83D\uDE42");
+        CsvSpec tableSpec = fileSpec.cloneForFormatting();
```
denalex, line 238: a lot of tests have this line. I'd overload `runScenario` method to not have a `tableSpec` param and do this line inside that method since `tableSpec` is not used anywhere else in this method.

yimingli-vmware (approved):

denalex (approved):

PR #960, opened 2023-04-05
Add support year with more than 4 digits in 'date' or 'timestamp'

RomaZe:
Some users want to save dates that contain a year with more than 4 digits. The current `JdbcResolver` doesn't allow us to do this. It uses `java.sql.Date` or `java.sql.Timestamp` which cannot understand the date with more than 4 digits in the year. Moreover, the  `JdbcResolver` doesn't understand dates with `BC` (see example in the bottom).

This PR adds an opportunity to work with dates that contain the year with more than 4 digits. A special parameter `<mapping option="date_wide_range" property="jdbc.date.wideRange"/>` was added into the `pxf-profiles-default.xml` profile. It might be used as a parameter in the `LOCATION` (`date_wide_range=true`) or as a parameter in the `jdbc-site.xml` file (`jdbc.date.wideRange`). The default value is `false`. It means that the previous logic will be used and `JdbcResolver` will use `java.sql.Date` or `java.sql.Timestamp`.

Setting `jdbc.date.wideRange` to `true` or `date_wide_range` to `true` means:
1.  `JdbcResolver` starts to use `LocalDate`, `LocalDateTime`, `OffsetDateTime` classes. These classes can work with years that contain more than 4 digits.
2.  `JdbcResolver` understands  the difference between dates with `BC` and `AD`.

**How to use:** 

**In the `jdbc-site.xml` file**
```xml
<property>
    <name>jdbc.date.wideRange</name>
    <value>true</value>
    <description>Set the value to 'true' if a year might contain more than 4 digits in a 'date' or in a 'timestamp'</description>
</property>
```
**In the `LOCATION`**
```
--- WRITABLE TABLE 
CREATE WRITABLE EXTERNAL TABLE wide_year_write(
  birth_date date,
  birth_date_dad timestamp)
LOCATION ('pxf://wide_year_dist?PROFILE=Jdbc&JDBC_DRIVER=org.postgresql.Driver&DB_URL=jdbc:postgresql://10.10.10.2:5432/postgres&USER=ext_user&date_wide_range=true') 
FORMAT 'CUSTOM' (FORMATTER='pxfwritable_export');

--- READABLE TABLE 
CREATE EXTERNAL TABLE wide_year_read(
  birth_date date,
  birth_date_dad timestamp,
  birth_date_mom timestamp with time zone)
LOCATION ('pxf://wide_year_dist?PROFILE=Jdbc&JDBC_DRIVER=org.postgresql.Driver&DB_URL=jdbc:postgresql://10.10.10.2:5432/postgres&USER=ext_user&date_wide_range=true') 
```
**Example of dates, timestamps and timestamps with time zone that can be processed with the new parameter:**
date: `0010-12-11 BC`
date: `456789-12-11`
timestamp: `1234-10-19 10:11:15.456 BC`
timestamp: `123456-10-19 11:12:13`
timestamp with time zone: `12345-02-01 10:15:16+00` (only for READABLE EXTERNAL tables)
timestamp with time zone: `2000-02-01 04:15:16.123457+00 BC` (only for READABLE EXTERNAL tables)

### DIFF
--- a/server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
+++ b/server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
@@ -364,7 +364,8 @@ public class JdbcBasePlugin extends BasePlugin {
             poolQualifier = configuration.get(JDBC_POOL_QUALIFIER_PROPERTY_NAME);
         }
 
-        // Optional parameter. Get the flag whether the year might contain more than 4 digits in `date` or 'timestamp' or not
+        // Optional parameter to determine if the year might contain more than 4 digits in `date` or 'timestamp'.
+        // Populate from the context if it exists, otherwise pull from the configuration. The default value is false.
         String dateWideRange = context.getOption(JDBC_DATE_WIDE_RANGE);
         if (dateWideRange != null) {
             isDateWideRange = Boolean.parseBoolean(dateWideRange);
--- a/server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
+++ b/server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
@@ -96,7 +96,7 @@ public class JdbcBasePlugin extends BasePlugin {
     private static final String HIVE_URL_PREFIX = "jdbc:hive2://";
     private static final String HIVE_DEFAULT_DRIVER_CLASS = "org.apache.hive.jdbc.HiveDriver";
     private static final String MYSQL_DRIVER_PREFIX = "com.mysql.";
-    private static final String JDBC_DATE_WIDE_RANGE = "jdbc.date.wide-range";
+    private static final String JDBC_DATE_WIDE_RANGE = "jdbc.date.wideRange";
 
     private enum TransactionIsolation {
         READ_UNCOMMITTED(1),
@@ -365,13 +365,8 @@ public class JdbcBasePlugin extends BasePlugin {
         }
 
         // Optional parameter to determine if the year might contain more than 4 digits in `date` or 'timestamp'.
-        // Populate from the context if it exists, otherwise pull from the configuration. The default value is false.
-        String dateWideRange = context.getOption(JDBC_DATE_WIDE_RANGE);
-        if (dateWideRange != null) {
-            isDateWideRange = Boolean.parseBoolean(dateWideRange);
-        } else {
-            isDateWideRange = configuration.getBoolean(JDBC_DATE_WIDE_RANGE, false);
-        }
+        // The default value is false.
+        isDateWideRange = configuration.getBoolean(JDBC_DATE_WIDE_RANGE, false);
     }
 
     /**
--- a/server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
+++ b/server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
@@ -96,6 +96,7 @@ public class JdbcBasePlugin extends BasePlugin {
     private static final String HIVE_URL_PREFIX = "jdbc:hive2://";
     private static final String HIVE_DEFAULT_DRIVER_CLASS = "org.apache.hive.jdbc.HiveDriver";
     private static final String MYSQL_DRIVER_PREFIX = "com.mysql.";
+    private static final String JDBC_DATE_WIDE_RANGE = "jdbc.date.wide-range";
 
     private enum TransactionIsolation {
         READ_UNCOMMITTED(1),
@@ -163,6 +164,9 @@ public class JdbcBasePlugin extends BasePlugin {
 
     private final ConnectionManager connectionManager;
     private final SecureLogin secureLogin;
+
+    // Flag which is used when the year might contain more than 4 digits in `date` or 'timestamp'
+    protected boolean isDateWideRange;
 
     static {
         // Deprecated as of Oct 22, 2019 in version 5.9.2+
@@ -358,6 +362,14 @@ public class JdbcBasePlugin extends BasePlugin {
             // get the qualifier for connection pool, if configured. Might be used when connection session authorization is employed
             // to switch effective user once connection is established
             poolQualifier = configuration.get(JDBC_POOL_QUALIFIER_PROPERTY_NAME);
+        }
+
+        // Optional parameter. Get the flag whether the year might contain more than 4 digits in `date` or 'timestamp' or not
+        String dateWideRange = context.getOption(JDBC_DATE_WIDE_RANGE);
+        if (dateWideRange != null) {
+            isDateWideRange = Boolean.parseBoolean(dateWideRange);
+        } else {
+            isDateWideRange = configuration.getBoolean(JDBC_DATE_WIDE_RANGE, false);
         }
     }
 
--- a/server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
+++ b/server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
@@ -23,7 +23,9 @@ import org.greenplum.pxf.api.OneField;
 import org.greenplum.pxf.api.OneRow;
 import org.greenplum.pxf.api.io.DataType;
 import org.greenplum.pxf.api.model.Resolver;
+import org.greenplum.pxf.api.security.SecureLogin;
 import org.greenplum.pxf.api.utilities.ColumnDescriptor;
+import org.greenplum.pxf.plugins.jdbc.utils.ConnectionManager;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -55,6 +57,10 @@ import static java.time.format.DateTimeFormatter.ISO_OFFSET_TIME;
  * JDBC tables resolver
  */
 public class JdbcResolver extends JdbcBasePlugin implements Resolver {
+    /**
+     * LOCAL_DATE_GET_FORMATTER is used to format LocalDate to String.
+     * Examples: 2023-01-10 -> "2023-01-10 AD"; +12345-02-01 -> "12345-02-01 AD"; -0009-12-11 -> "0010-12-11 BC"
+     */
     private static final DateTimeFormatter LOCAL_DATE_GET_FORMATTER = (new DateTimeFormatterBuilder())
             .appendValue(ChronoField.YEAR_OF_ERA, 4, 9, SignStyle.NORMAL).appendLiteral("-")
             .appendValue(ChronoField.MONTH_OF_YEAR, 2).appendLiteral('-')
@@ -62,6 +68,11 @@ public class JdbcResolver extends JdbcBasePlugin implements Resolver {
             .appendPattern(" G")
             .toFormatter();
 
+    /**
+     * LOCAL_DATE_TIME_GET_FORMATTER is used to format LocalDateTime to String.
+     * Examples: 2018-10-19T10:11 -> "2018-10-19 10:11:00 AD"; +123456-10-19T11:12:13 -> "123456-10-19 11:12:13 AD";
+     * -1233-10-19T10:11:15.456 -> "1234-10-19 10:11:15.456 BC"
+     */
     private static final DateTimeFormatter LOCAL_DATE_TIME_GET_FORMATTER = (new DateTimeFormatterBuilder())
             .appendValue(ChronoField.YEAR_OF_ERA, 4, 9, SignStyle.NORMAL).appendLiteral("-")
             .appendValue(ChronoField.MONTH_OF_YEAR, 2).appendLiteral('-')
@@ -70,6 +81,12 @@ public class JdbcResolver extends JdbcBasePlugin implements Resolver {
             .appendPattern(" G")
             .toFormatter();
 
+
+    /**
+     * OFFSET_DATE_TIME_GET_FORMATTER is used to format OffsetDateTime to String.
+     * Examples: 1956-02-01T07:15:16Z -> "1956-02-01 07:15:16Z AD"; +12345-02-01T10:15:16Z -> "12345-02-01 10:15:16Z AD";
+     * -1999-02-01T04:15:16Z -> "2000-02-01 04:15:16Z BC"
+     */
     private static final DateTimeFormatter OFFSET_DATE_TIME_GET_FORMATTER = (new DateTimeFormatterBuilder())
             .appendValue(ChronoField.YEAR_OF_ERA, 4, 9, SignStyle.NORMAL).appendLiteral("-")
             .appendValue(ChronoField.MONTH_OF_YEAR, 2).appendLiteral('-')
@@ -78,6 +95,10 @@ public class JdbcResolver extends JdbcBasePlugin implements Resolver {
             .appendPattern(" G")
             .toFormatter();
 
+    /**
+     * LOCAL_DATE_SET_FORMATTER is used to format String to LocalDate.
+     * Examples: "1977-12-11" -> 1977-12-11; "456789-12-11" -> +456789-12-11; "0010-12-11 BC" -> -0009-12-11
+     */
     private static final DateTimeFormatter LOCAL_DATE_SET_FORMATTER = (new DateTimeFormatterBuilder())
             .appendValue(ChronoField.YEAR_OF_ERA, 1, 9, SignStyle.NORMAL).appendLiteral('-')
             .appendValue(ChronoField.MONTH_OF_YEAR, 2).appendLiteral('-')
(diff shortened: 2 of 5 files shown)


bradfordb-vmware:
Thank you for opening this PR -- the team will pick up reviewing this and provide feedback shortly.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -36,15 +36,64 @@
...
+            .appendValue(ChronoField.MONTH_OF_YEAR, 2).appendLiteral('-')
+            .appendValue(ChronoField.DAY_OF_MONTH, 2)
+            .optionalStart().appendPattern(" G").optionalEnd()
+            .toFormatter();
+
+    private static final DateTimeFormatter LOCAL_DATE_TIME_SET_FORMATTER = (new DateTimeFormatterBuilder())
+            .appendValue(ChronoField.YEAR_OF_ERA, 1, 9, SignStyle.NORMAL).appendLiteral('-')
+            .appendValue(ChronoField.MONTH_OF_YEAR, 1, 2, SignStyle.NORMAL).appendLiteral('-')
+            .appendValue(ChronoField.DAY_OF_MONTH, 1, 2, SignStyle.NORMAL)
+            .appendLiteral(" ")
```
ashuka24, line 92: for consistency with the other examples above can we have the follwoing: 
```suggestion
            .appendValue(ChronoField.DAY_OF_MONTH, 1, 2, SignStyle.NORMAL).appendLiteral(" ")

File: server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java [hunk 2]
```diff
@@ -36,15 +36,64 @@
...
+            .appendValue(ChronoField.DAY_OF_MONTH, 2).appendLiteral(" ")
+            .append(ISO_OFFSET_TIME)
+            .appendPattern(" G")
+            .toFormatter();
+
+    private static final DateTimeFormatter LOCAL_DATE_SET_FORMATTER = (new DateTimeFormatterBuilder())
+            .appendValue(ChronoField.YEAR_OF_ERA, 1, 9, SignStyle.NORMAL).appendLiteral('-')
+            .appendValue(ChronoField.MONTH_OF_YEAR, 2).appendLiteral('-')
+            .appendValue(ChronoField.DAY_OF_MONTH, 2)
+            .optionalStart().appendPattern(" G").optionalEnd()
```
ashuka24, line 85: can you explain why the era is optional when we set the date but not when we are getting the date?

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -36,15 +36,64 @@
...
 import java.util.Set;
 
+import static java.time.format.DateTimeFormatter.ISO_LOCAL_TIME;
+import static java.time.format.DateTimeFormatter.ISO_OFFSET_TIME;
+
 /**
  * JDBC tables resolver
  */
 public class JdbcResolver extends JdbcBasePlugin implements Resolver {
+    private static final DateTimeFormatter LOCAL_DATE_GET_FORMATTER = (new DateTimeFormatterBuilder())
```
denalex, line 58: would that be beneficial if we have a comment line for each of these formatters that illustrates an example of how a formatted value will look like ?

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -36,15 +36,64 @@
...
+    private static final DateTimeFormatter OFFSET_DATE_TIME_GET_FORMATTER = (new DateTimeFormatterBuilder())
+            .appendValue(ChronoField.YEAR_OF_ERA, 4, 9, SignStyle.NORMAL).appendLiteral("-")
+            .appendValue(ChronoField.MONTH_OF_YEAR, 2).appendLiteral('-')
+            .appendValue(ChronoField.DAY_OF_MONTH, 2).appendLiteral(" ")
+            .append(ISO_OFFSET_TIME)
+            .appendPattern(" G")
+            .toFormatter();
+
+    private static final DateTimeFormatter LOCAL_DATE_SET_FORMATTER = (new DateTimeFormatterBuilder())
+            .appendValue(ChronoField.YEAR_OF_ERA, 1, 9, SignStyle.NORMAL).appendLiteral('-')
```
denalex, line 82: why do we have GET and SET separate formatters ? Why `ChronoField.YEAR_OF_ERA` is  `4, 9` for GET and `1,9` for SET ?

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -311,19 +389,55 @@ public static void decodeOneRowToPreparedStatement(OneRow row, PreparedStatement
                     if (field.val == null) {
                         statement.setNull(i, Types.TIMESTAMP);
                     } else {
-                        statement.setTimestamp(i, (Timestamp) field.val);
+                        if (field.val instanceof LocalDateTime) {
+                            statement.setObject(i, (LocalDateTime) field.val);
```
denalex, line 393: I don't think we really need to cast to a specific type ( here to `(LocalDateTime)` ) for `setObject` call.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -311,19 +389,55 @@ public static void decodeOneRowToPreparedStatement(OneRow row, PreparedStatement
...
+                        }
                     }
                     break;
                 case DATE:
                     if (field.val == null) {
                         statement.setNull(i, Types.DATE);
                     } else {
-                        statement.setDate(i, (Date) field.val);
+                        if (field.val instanceof LocalDate) {
+                            statement.setObject(i, (LocalDate) field.val);
```
denalex, line 404: I don't think we really need to cast to a specific type ( here to `(LocalDate)` ) for `setObject` call.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -117,10 +166,31 @@ public List<OneField> getFields(OneRow row) throws SQLException {
...
+                        value = localDateTime != null ? localDateTime.format(LOCAL_DATE_TIME_GET_FORMATTER) : null;
+                    } else {
+                        value = result.getTimestamp(colName);
+                    }
+                    break;
+                case TIMESTAMP_WITH_TIME_ZONE:
+                    if (isDateWideRange) {
+                        OffsetDateTime offsetDateTime = result.getObject(colName, OffsetDateTime.class);
+                        value = offsetDateTime != null ? offsetDateTime.format(OFFSET_DATE_TIME_GET_FORMATTER) : null;
+                    } else {
```
denalex, line 188: I know there is no current implementation for it and maybe it is out of scope of this PR, but why would we not support TIMESTAMP_WITH_TIME_ZONE for non-wide range date ?

server/pxf-jdbc/src/test/java/org/greenplum/pxf/plugins/jdbc/JdbcResolverTest.java
```diff
@@ -0,0 +1,554 @@
...
+    @Test
+    void getFieldDateWithWideRangeTest() throws SQLException {
+        LocalDate localDate = LocalDate.of(1977, 12, 11);
+        boolean isDateWideRange = true;
+        columnDescriptors.add(new ColumnDescriptor("birth_date", DataType.DATE.getOID(), 0, "date", null));
+        context.setTupleDescription(columnDescriptors);
+        when(row.getData()).thenReturn(result);
+        when(result.getObject("birth_date", LocalDate.class)).thenReturn(localDate);
+
+        try (MockedStatic<SpringContext> springContextMockedStatic = mockStatic(SpringContext.class)) {
```
denalex, line 54: why bother with mocking Spring when there is a special constructor that simulates dependency injection on the base class: 
`JdbcBasePlugin(ConnectionManager connectionManager, SecureLogin secureLogin)`
We can add 2 constructors to `JdbcResolver` (one default empty, another with similar params) that will call `super` and then just use the `new JdbcResolver(mockConnectionManager, mockSecureLogin)` here.

server/pxf-jdbc/src/test/java/org/greenplum/pxf/plugins/jdbc/JdbcResolverTest.java
```diff
@@ -0,0 +1,554 @@
...
+            resolver.columns = context.getTupleDescription();
+            resolver.isDateWideRange = isDateWideRange;
+            List<OneField> oneFields = resolver.getFields(row);
+            assertEquals("1977-12-11 AD", oneFields.get(0).val);
+        }
+    }
+
+    @Test
+    void getFieldDateNullWithWideRangeTest() throws SQLException {
+        boolean isDateWideRange = true;
```
denalex, line 66: thanks for adding the test case. It would also be nice to have a few methods where `isDateWideRange` set to false to see what the expectations are and to prove there are no regressions after this PR is applied.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
```diff
@@ -359,6 +363,14 @@ public void afterPropertiesSet() {
             // to switch effective user once connection is established
             poolQualifier = configuration.get(JDBC_POOL_QUALIFIER_PROPERTY_NAME);
         }
+
+        // Optional parameter. Get the flag whether the year might contain more than 4 digits in `date` or 'timestamp' or not
```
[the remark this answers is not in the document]
RomaZe: Thank you. Took your option.

server/pxf-jdbc/src/test/java/org/greenplum/pxf/plugins/jdbc/JdbcResolverTest.java
```diff
@@ -0,0 +1,554 @@
...
+    @Mock
+    private ResultSet result;
+    RequestContext context = new RequestContext();
+    List<ColumnDescriptor> columnDescriptors = new ArrayList<>();
+    List<OneField> oneFieldList = new ArrayList<>();
+    private JdbcResolver resolver;
+
+    @Test
+    void getFieldDateWithWideRangeTest() throws SQLException {
+        LocalDate localDate = LocalDate.of(1977, 12, 11);
```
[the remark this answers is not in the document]
RomaZe: Refactored these unit tests: extracted 2 methods.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
RomaZe: Yes, sure! Modified.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
RomaZe: Depends on the ERA. The date with BC comes from the database as a string with "BC" suffix ("0010-12-11 BC"). The date with AD comes without any additional suffix ("2018-08-11"). That is the reason why the era is optional.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
```diff
@@ -359,6 +363,14 @@ public void afterPropertiesSet() {
...
             poolQualifier = configuration.get(JDBC_POOL_QUALIFIER_PROPERTY_NAME);
         }
+
+        // Optional parameter. Get the flag whether the year might contain more than 4 digits in `date` or 'timestamp' or not
+        String dateWideRange = context.getOption(JDBC_DATE_WIDE_RANGE);
+        if (dateWideRange != null) {
+            isDateWideRange = Boolean.parseBoolean(dateWideRange);
+        } else {
+            isDateWideRange = configuration.getBoolean(JDBC_DATE_WIDE_RANGE, false);
+        }
```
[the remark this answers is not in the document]
RomaZe: I added the parameter into `pxf-profiles-default.xml`. But if the customer not update the profile file it will never get this parameter from the configuration. That is the reason why I added the condition to check the parameter in the context first.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
RomaZe: I think it is a good idea. Added comment lines for all formatters.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
RomaZe: The LocalDate class represents a date without a time-zone in the ISO-8601 calendar system, such as 2007-12-03 (YYYY-MM-DD).
When we read date from an external table it comes as LocalDate class. So we don't need to transfer a date that contains less than 4 digits in year. The second reason is that if we try to insert date with 2 or less digits in the year as a date into the database we will receive incorrect date. For example:
```
select '0010-03-11 AD'::date;  -- correct result
 0010-03-11

select '10-03-11 AD'::date; -- wrong result
 2011-10-03
```

When we write data into an external table, the date comes as a string and we have to be ready to convert string to LocalDate  class even if the year contains less than 4 digits.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
RomaZe: Agree. Removed casting from the code.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
RomaZe: In this PR I didn't want to mix adding a new option and changing the behavior of the current logic. I think it is possible but I am not sure that all JDBC drivers can properly deal with timestamp with timezone.

server/pxf-jdbc/src/test/java/org/greenplum/pxf/plugins/jdbc/JdbcResolverTest.java
RomaZe: Added 2 additional constructors and modified  all tests in  `JdbcResolverTest` class.

server/pxf-service/src/main/resources/pxf-profiles-default.xml
```diff
@@ -43,6 +43,7 @@ under the License.
             <mapping option="batch_size" property="jdbc.statement.batchSize"/>
             <mapping option="fetch_size" property="jdbc.statement.fetchSize"/>
             <mapping option="query_timeout" property="jdbc.statement.queryTimeout"/>
+            <mapping option="date_wide_range" property="jdbc.date.wide-range"/>
```
ashuka24, line 46: for consistency can we make this camelcase 

```suggestion
            <mapping option="date_wide_range" property="jdbc.date.wideRange"/>

Comment (RomaZe) [reply server/pxf-service/src/main/resources/pxf-profiles-default.xml]: Agree, it looks better. Made the name in the camel case format

File: server/pxf-jdbc/src/test/java/org/greenplum/pxf/plugins/jdbc/JdbcResolverTest.java [hunk 14]
```diff
@@ -0,0 +1,354 @@
...
+        isDateWideRange = true;
+        LocalDate localDate = LocalDate.of(-1234567, 6, 1);
+        OneField oneField = getOneField(localDate, DataType.DATE.getOID(), "date");
+        // The year -1234567 is transferred to 1234568 BC: https://en.wikipedia.org/wiki/Astronomical_year_numbering
+        assertEquals("1234568-06-01 BC", oneField.val);
+    }
+
+    @Test
+    void getFieldDateWithEraTest() throws SQLException {
+        isDateWideRange = true;
```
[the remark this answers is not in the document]
RomaZe: I added some unit tests where `isDateWideRange = false` fro the `setFields(List<OneField> record)` method.
I am not sure it will be possible to do for `getFields(OneRow row)` method.  We sent the name of the column to a JDBC driver and the implementation of the specific JDBC driver is responsible for the return value or for generating error.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -36,15 +38,87 @@
...
 public class JdbcResolver extends JdbcBasePlugin implements Resolver {
+    /**
+     * LOCAL_DATE_GET_FORMATTER is used to format LocalDate to String.
+     * Examples: 2023-01-10 -> "2023-01-10 AD"; +12345-02-01 -> "12345-02-01 AD"; -0009-12-11 -> "0010-12-11 BC"
+     */
+    private static final DateTimeFormatter LOCAL_DATE_GET_FORMATTER = (new DateTimeFormatterBuilder())
+            .appendValue(ChronoField.YEAR_OF_ERA, 4, 9, SignStyle.NORMAL).appendLiteral("-")
+            .appendValue(ChronoField.MONTH_OF_YEAR, 2).appendLiteral('-')
+            .appendValue(ChronoField.DAY_OF_MONTH, 2)
+            .appendPattern(" G")
```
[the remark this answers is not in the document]
RomaZe: Yes, good idea! Added.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcBasePlugin.java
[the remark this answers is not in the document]
RomaZe: Yes, I think it will be enough to take the value of the parameter from the configuration. You are right. I removed the lines where I take the parameter from the context.

denalex (approved):

denalex:
@RomaZe - please update the PR description to reflect the change to the property `jdbc.date.wide-range` name.

ashuka24 (approved):
Thanks so much for the PR. Can we update the PR message to also say `wideRange` instead of `wide-range`

PR #965, opened 2023-04-11
Modifying Tests to work with FDW. 

hpandeycodeit:
Following Tests are either fixed (WorksWithFDW) or marked as SkipsForFDW/FailsWithFDW:

- HcfsGlobbingTest: WorksWithFDW
- HdfsReadableSequenceTest: WorksWithFDW
- FileAsRowTest  : WorksWithFDW mostly for GP7. 

> Only one test `testMultilineJsonFile` fails for GP6 so marked it as FailsWithFDW
> 
> This Test Fails with the error: 
>   ```ERROR:  set-valued function called in context that cannot accept a set``` 
> 

- ApiTest : SkipForFDW | group = features so this can be skipped?
- PluginTest : SkipForFDW | group = features so this can be skipped?
- SyntaxTest: SkipForFDW| group = features so this can be skipped?

- FailOverTest: FailsWithFDW   

> It fails with error:
> ```< ERROR:  PXF server error : Profile 'hdfs:text' already defines: [ACCESSOR, FRAGMENTER, RESOLVER]```
> 

This doesn't contain all the tests. This is a small subset. Rest of the tests are being worked on.


automation/src/test/java/org/greenplum/pxf/automation/features/hcfs/FileAsRowTest.java
```diff
@@ -131,7 +138,7 @@ private void runTestScenario(String name, String[] fields, String hdfsPath, Stri
         String tableName = "file_as_row_" + name;
 
         ProtocolEnum protocol = ProtocolUtils.getProtocol();
-        exTable = new ReadableExternalTable(tableName, fields, protocol.getExternalTablePath(hdfs.getBasePath(), locationPath), "CSV");
+        exTable = TableFactory.getPxfReadableCSVTable(tableName, fields, protocol.getExternalTablePath(hdfs.getBasePath(), locationPath), ",");
```
denalex, line 141: this might be an indication of a bigger problem, we need more details to see why it fails

automation/src/test/java/org/greenplum/pxf/automation/features/general/ApiTest.java
```diff
@@ -12,6 +13,7 @@
 /**
  * Test PXF API
  */
+@SkipForFDW
```
denalex, line 16: I am not sure we need to annotate classes that are fully consist of tests with only "features" group as they will not be run by CI right now and the annotation will not be exercised. We should also decide at some point what to do with them and if we bring them into "gpdb" group, then if they don't work with FDW we can annotate them as such.

automation/src/test/java/org/greenplum/pxf/automation/features/general/FailOverTest.java
```diff
@@ -8,6 +9,7 @@
 import java.io.File;
 
 /** Tests how failures are handled **/
+@FailsWithFDW
```
denalex, line 12: why does it fail ? If this is because profile is not defines, we should use dummy `test:text` profile that should work with FDW

automation/src/test/java/org/greenplum/pxf/automation/features/hcfs/HcfsGlobbingTest.java
```diff
@@ -58,7 +61,9 @@ public void testMatchASingleCharacterNotFromCharSetOrRange() throws Exception {
 
     @Test(groups = {"gpdb", "hcfs", "security"})
     public void testEscapeSpecialCharacters() throws Exception {
-        prepareTestScenario("escape_special_characters", "ab[c.d", null, null, null, "ab\\\\[c.d");
+        //FDW also adds an escape backslash from its side. If we are passing }\{bc, it will be }\\{bc in FDW. However, in external table it runs as expected.
```
denalex, line 64: which particular logic in FDW adds extra backslash ?

automation/src/test/java/org/greenplum/pxf/automation/features/hcfs/HcfsTextTest.java
```diff
@@ -44,7 +45,7 @@ private void runTestScenario(String name, String[] fields, String srcPath, Strin
 
         ProtocolEnum protocol = ProtocolUtils.getProtocol();
         String tableName = "hcfs_text_" + name;
-        exTable = new ReadableExternalTable(tableName, fields, protocol.getExternalTablePath(hdfs.getBasePath(), hdfsPath), "TEXT");
+        exTable = TableFactory.getPxfReadableTextTable(tableName, fields, protocol.getExternalTablePath(hdfs.getBasePath(), hdfsPath), delimiter);
```
denalex, line 48: I wonder where FDW requires `delimiter` option ?

automation/src/test/java/org/greenplum/pxf/automation/features/general/FailOverTest.java
hpandeycodeit: Using `test:text` profile isn't helping either. Getting this error: 

```
< ERROR:  PXF server error : test:text is not defined in pxf-profiles.xml
```

automation/src/test/java/org/greenplum/pxf/automation/features/hcfs/HcfsTextTest.java
hpandeycodeit: Found the issue. Fixed this.

automation/tincrepo/main/pxf/features/hcfs/globbing/match_string_from_string_set/expected/query04.ans
```diff
@@ -244,6 +244,9 @@ select * from hcfs_glob_match_string_from_string_set_10 order by name, num;
 -- m/CONTEXT:.*line.*/
 -- s/line \d* of //g
 --
+-- m/, file pxf:\/\/(.*)/
```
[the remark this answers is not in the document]
hpandeycodeit: with FDW, it is printing extra text like 
```
> CONTEXT:  External table hcfs_glob_match_string_from_string_set_11, file pxf://pxf_automation_data?PROFILE=*:text
254d256
< CONTEXT:  External table hcfs_glob_match_string_from_string_set_11
```

automation/src/test/java/org/greenplum/pxf/automation/features/hcfs/HcfsGlobbingTest.java
```diff
@@ -85,8 +90,11 @@ public void testMatchAStringFromStringSet() throws Exception {
         prepareTestScenario("match_string_from_string_set_4", "match_string_from_string_set_10", null, null, null, null, "}{ac,?}");
         // test ill-formed curly
         prepareTestScenario("match_string_from_string_set_4", "match_string_from_string_set_11", null, null, null, null, "}{bc");
+
+        //FDW also adds an escape backslash from its side. If we are passing }\{bc, it will be }\\{bc in FDW. However, in external table it runs as expected.
```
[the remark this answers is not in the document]
hpandeycodeit: Removed the duplicate comment. 

For the escaping: 

 External Table adds (escape) E to the location which escapes the backslash while FDW doesn't add E to the resource.

For e.g: FDW:
```
CREATE FOREIGN TABLE hcfs_glob_escape_special_characters (name text, num integer, dub double precision, longNum bigint, bool boolean) SERVER default_hdfs OPTIONS
(resource 'tmp/pxf_automation_data/4ab38346-079f-4a93-bd7f-fe3ab366182e/escape_special_characters/ab\\[c.d',format 'text',delimiter ',');
```

External Table:

```
CREATE EXTERNAL TABLE hcfs_glob_escape_special_characters (name text, num integer, dub double precision, longNum bigint, bool boolean) 
LOCATION (E'pxf://tmp/pxf_automation_data/51752ebe-8c1f-4788-b639-fb6344e2eff5/escape_special_characters/ab\\[c.d?PROFILE=hdfs:text')
FORMAT 'Text' ( DELIMITER ',')

```

So using the escapedGlob here for FDW.

automation/src/test/java/org/greenplum/pxf/automation/features/hcfs/FileAsRowTest.java
hpandeycodeit: Looks like this error is coming from postgres/server. 

GPDB version: 6.23

This works for external table : 

```
gpadmin=# \d+ e1
                      External table "public.e1"
  Column   | Type | Modifiers | Storage  | Stats target | Description 
-----------+------+-----------+----------+--------------+-------------
 json_blob | json |           | extended |              | 
Type: readable
Encoding: UTF8
Format type: csv
Format options: delimiter ',' null '' escape '"' quote '"'
External options: {}
External location: pxf://tmp/tweets-pp.json?PROFILE=HDFS:text:multi&file_as_row=true
Execute on: all segments

```

SELECT: 

```
gpadmin=# select json_array_elements(json_blob->'root') from e1 limit 1; 
                                               json_array_elements                                               
-----------------------------------------------------------------------------------------------------------------
 {                                                                                                              +
       "record":{                                                                                               +
         "created_at":"Fri Jun 07 22:45:02 +0000 2013",                                                         +
         "id":343136547115253761,                                                                               +
         "id_str":"343136547115253761",                                                                         +
         "text":"text1",                                                                                        +
         "source":"<a href=\"http:\/\/twitter.com\/download\/iphone\" rel=\"nofollow\">Twitter for iPhone<\/a>",+
         "user":{                                                                                               +
           "id":26643566,                                                                                       +
           "name":"SpreadButter",                                                                               +
           "screen_name":"SpreadButter",                                                                        +
           "location":"Austin, Texas"                                                                           +
         },                                                                                                     +
         "entities":{                                                                                           +
           "hashtags":[                                                                                         +
             "tweetCongress",                                                                                   +
             "IRS"                                                                                              +
           ]                                                                                                    +
         }                                                                                                      +
       }                                                                                                        +
     }
(1 row)


```


FDW Table: 

```
pxfautomation=# \d+ f1
                             Foreign table "public.f1"
  Column   | Type | Modifiers | FDW Options | Storage  | Stats target | Description 
-----------+------+-----------+-------------+----------+--------------+-------------
 json_blob | json |           |             | extended |              | 
Server: default_hdfs
FDW Options: (resource '/tmp/tweets-pp.json', format 'text:multi', delimiter ',', file_as_row 'true')

```

SELECT: 

```

pxfautomation=# select json_array_elements(json_blob->'root') from f1 limit 1; 
ERROR:  set-valued function called in context that cannot accept a set  (seg0 slice1 10.254.0.226:6000 pid=350985)
```

I will try to debug and see if we can find something.

automation/tincrepo/main/pxf/features/hcfs/globbing/match_string_from_string_set/expected/query04.ans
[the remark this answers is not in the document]
ashuka24: If that is the case, is this something we should be adding to the global init file?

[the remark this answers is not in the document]
hpandeycodeit: This is one case I have found so far for this, so may be have it in this file only for now.

automation/src/test/java/org/greenplum/pxf/automation/features/hcfs/HcfsGlobbingTest.java
```diff
@@ -85,8 +89,23 @@ public void testMatchAStringFromStringSet() throws Exception {
         prepareTestScenario("match_string_from_string_set_4", "match_string_from_string_set_10", null, null, null, null, "}{ac,?}");
         // test ill-formed curly
         prepareTestScenario("match_string_from_string_set_4", "match_string_from_string_set_11", null, null, null, null, "}{bc");
+
```
ashuka24, line 92: can we move this comment up to line 64 with the first introduction of the `escapeGlob`

automation/src/test/java/org/greenplum/pxf/automation/features/hcfs/HcfsTextTest.java
```diff
@@ -44,10 +46,20 @@ private void runTestScenario(String name, String[] fields, String srcPath, Strin
...
         exTable.setProfile(protocol.value() + ":text");
 
         if (delimiter != null) {
+            /* This sets delimiter 'OFF' which doesn't work for FDW. So Skipping it setting explicitly for FDW
```
ashuka24, line 53:
```suggestion
            /* This sets delimiter 'OFF' which doesn't work for FDW. So skip setting it explicitly for FDW

File: automation/src/test/java/org/greenplum/pxf/automation/features/hcfs/HcfsTextTest.java [hunk 10]
```diff
@@ -44,10 +46,20 @@ private void runTestScenario(String name, String[] fields, String srcPath, Strin
...
+             *
+             * CREATE FOREIGN TABLE hcfs_text_delimiter_off (text_blob text) SERVER default_hdfs
+             * OPTIONS (resource 'tmp/pxf_automation_data/b6afbc30-2a0e-40a8-9c4c-61a54be173e5/hcfs-text/no-delimiter',format 'text',delimiter 'OFF');
```
ashuka24, line 56:
```suggestion
             * OPTIONS (resource 'tmp/pxf_automation_data/b6afbc30-2a0e-40a8-9c4c-61a54be173e5/hcfs-text/no-delimiter', format 'text', delimiter 'OFF');

File: automation/src/test/java/org/greenplum/pxf/automation/features/hcfs/FileAsRowTest.java (hunk 1 above)
Comment (bradfordb-vmware) [reply automation/src/test/java/org/greenplum/pxf/automation/features/hcfs/FileAsRowTest.java:141]: I suspect the error is with the PXF FDW; I ran through the following simple test on GP6 and GP7 and the only difference I see is that `type` of `json_array_elements` changes from `normal` on GP6 to `func` on GP7.

```sql
-- on GP6
\df json_array_elements
--                                        List of functions
--    Schema   |        Name         | Result data type |      Argument data types       |  Type  
-- ------------+---------------------+------------------+--------------------------------+--------
--  pg_catalog | json_array_elements | SETOF json       | from_json json, OUT value json | normal
-- (1 row)

SELECT json_array_elements('[1,2,3]'::json);
--  json_array_elements 
-- ---------------------
--  1
--  2
--  3
-- (3 rows)

SELECT * FROM json_array_elements('[1,2,3]'::json);
--  value 
-- -------
--  1
--  2
--  3
-- (3 rows)


-- on GP7
\df json_array_elements
--                                       List of functions
--    Schema   |        Name         | Result data type |      Argument data types       | Type 
-- ------------+---------------------+------------------+--------------------------------+------
--  pg_catalog | json_array_elements | SETOF json       | from_json json, OUT value json | func
-- (1 row)

SELECT json_array_elements('[1,2,3]'::json);
--  json_array_elements 
-- ---------------------
--  1
--  2
--  3
-- (3 rows)

SELECT * FROM json_array_elements('[1,2,3]'::json);
--  value 
-- -------
--  1
--  2
--  3
-- (3 rows)
```

automation/src/test/java/org/greenplum/pxf/automation/features/hcfs/HcfsGlobbingTest.java
```diff
@@ -58,7 +61,8 @@ public void testMatchASingleCharacterNotFromCharSetOrRange() throws Exception {
 
     @Test(groups = {"gpdb", "hcfs", "security"})
     public void testEscapeSpecialCharacters() throws Exception {
-        prepareTestScenario("escape_special_characters", "ab[c.d", null, null, null, "ab\\\\[c.d");
+        String escapeGlob = FDWUtils.useFDW ? "ab\\[c.d" : "ab\\\\[c.d";
```
bradfordb-vmware, line 64: Should we move this into `prepareTestScenario` -- essentially if `FDWUtils.useFDW` is true, we want to to replace every `\\\\` with `\\`:

```java
// in prepareTestScenario
if (FDWUtils.useFDW) {
    escapeGlob = escapeGlob.replace("\\\\", "\\");
}
```

This also makes it a better place to document why we're changing the escape glob instead of at a random call site.

automation/src/test/java/org/greenplum/pxf/automation/features/hcfs/HcfsTextTest.java
```diff
@@ -44,10 +46,20 @@ private void runTestScenario(String name, String[] fields, String srcPath, Strin
...
+             * CREATE FOREIGN TABLE
+             * pxfautomation=# select * from hcfs_text_delimiter_off;
+             * ERROR:  using no delimiter is only supported for external tables
+             */
+            if(delimiter == "OFF" && FDWUtils.useFDW)
+                return;
```
bradfordb-vmware, line 62: This is an early return so the call to `runTincTest` doesn't happen -- do we want to skip running this scenario? If so, wouldn't it be better to throw `SkipException` here so that it's captured in the test report that we skipped the test? If we don't want to skip the scenario, then we need to fix this.

automation/src/test/java/org/greenplum/pxf/automation/features/hcfs/FileAsRowTest.java
```diff
@@ -63,6 +67,8 @@ public void testMultilineFile() throws Exception {
                 hdfsBasePath + multiLineTextFile, srcPaths);
     }
 
+    // This Test is Failing for GP6 FDW with the error :
```
ashuka24, line 70: We have a comment here that says the test is failing, but do not have a `@FailsWithFDW` tag. Is that expected?

automation/src/test/java/org/greenplum/pxf/automation/features/hcfs/HcfsGlobbingTest.java
```diff
@@ -122,6 +126,24 @@ private void prepareTestScenario(String path, String testName, String data1, Str
...
+        // CREATE FOREIGN TABLE hcfs_glob_escape_special_characters (name text, num integer, dub double precision, longNum bigint, bool boolean)
+        // SERVER default_hdfs OPTIONS
+        // (resource 'tmp/pxf_automation_data/4ab38346-079f-4a93-bd7f-fe3ab366182e/escape_special_characters/ab\\[c.d',format 'text',delimiter ',');
+        //
+        // External Table:
+        // CREATE EXTERNAL TABLE hcfs_glob_escape_special_characters (name text, num integer, dub double precision, longNum bigint, bool boolean)
+        // LOCATION (E'pxf://tmp/pxf_automation_data/51752ebe-8c1f-4788-b639-fb6344e2eff5/escape_special_characters/ab\\[c.d?PROFILE=hdfs:text')
+        // FORMAT 'Text' ( DELIMITER ',')
+        //
+        // So using the escaped glob for FDW.
```
ashuka24, line 141: For large comments like this, I would recommend using `/* */` syntax instead of `//` for every line. 

Can we also add a comment about what the glob then becomes? 

```suggestion
        /*
         * External Table adds (escape) E to the location which escapes the backslash while FDW doesn't add E to the resource.
         * For e.g:
         * FDW:
         *     CREATE FOREIGN TABLE hcfs_glob_escape_special_characters (name text, num integer, dub double precision, longNum bigint, bool boolean)
         *     SERVER default_hdfs OPTIONS
         *     (resource 'tmp/pxf_automation_data/4ab38346-079f-4a93-bd7f-fe3ab366182e/escape_special_characters/ab\\[c.d',format 'text',delimiter ',');
         *
         * The glob would be `ab\\[c.d`
         * 
         * External Table:
         *     CREATE EXTERNAL TABLE hcfs_glob_escape_special_characters (name text, num integer, dub double precision, longNum bigint, bool boolean)
         *     LOCATION (E'pxf://tmp/pxf_automation_data/51752ebe-8c1f-4788-b639-fb6344e2eff5/escape_special_characters/ab\\[c.d?PROFILE=hdfs:text')
         *     FORMAT 'Text' ( DELIMITER ',')
         * 
         * The glob would be `ab\\\\[c.d` as we escape the backslash
         * 
         * So using the escaped glob for FDW.
         */

File: automation/src/test/java/org/greenplum/pxf/automation/features/hcfs/FileAsRowTest.java [hunk 15]
```diff
@@ -1,14 +1,18 @@
 package org.greenplum.pxf.automation.features.hcfs;
 
+import annotations.FailsWithFDW;
```
ashuka24, line 3: This import doesn't look like it's being used now?

automation/src/test/java/org/greenplum/pxf/automation/features/hcfs/FileAsRowTest.java
hpandeycodeit: Yes, the test is only failing for GP6. I am still investigating it.

ashuka24 (approved):

yimingli-vmware (approved):

automation/src/test/java/org/greenplum/pxf/automation/features/hcfs/FileAsRowTest.java
```diff
@@ -131,11 +161,18 @@ private void runTestScenario(String name, String[] fields, String hdfsPath, Stri
...
 
-        runTincTest("pxf.features.hcfs.file_as_row." + name + ".runTest");
+        if(gpdb.getVersion() < 6 && name.equals("json"))
+        {
+            runTincTest("pxf.features.hcfs.file_as_row." + name + "_gp5.runTest");
+        }
+        else
+        {
+            runTincTest("pxf.features.hcfs.file_as_row." + name + ".runTest");
+        }
```
denalex, line 176: formatting
```suggestion
        if (gpdb.getVersion() < 6 && name.equals("json")) {
            runTincTest("pxf.features.hcfs.file_as_row." + name + "_gp5.runTest");
        } else {
            runTincTest("pxf.features.hcfs.file_as_row." + name + ".runTest");
        }
```

automation/src/test/java/org/greenplum/pxf/automation/features/hcfs/FileAsRowTest.java
denalex: is this smth that requires further investigation ? Do we need to change any docs if there are changes between GP6 and GP7 on how to use JSON profile ?

hpandeycodeit: @denalex 
So there is  no changes profile wise. And the error that is thrown here seems to be common in postgres but I am not sure why this error is only in case of FDW. Changing the query a bit made the tests pass.

automation/src/test/java/org/greenplum/pxf/automation/features/hcfs/FileAsRowTest.java
```diff
@@ -166,12 +166,10 @@ private void runTestScenario(String name, String[] fields, String hdfsPath, Stri
         exTable.setUserParameters(new String[]{"FILE_AS_ROW=true"});
         gpdb.createTableAndVerify(exTable);
 
-        if(gpdb.getVersion() < 6 && name.equals("json"))
-        {
+        if(gpdb.getVersion() < 6 && name.equals("json")) {
```
ashuka24, line 169:
```suggestion
        if (gpdb.getVersion() < 6 && name.equals("json")) {

File: automation/src/test/java/org/greenplum/pxf/automation/features/hcfs/HcfsGlobbingTest.java [hunk 18]
```diff
@@ -121,6 +125,31 @@ private void prepareTestScenario(String path, String testName, String data1, Str
         prepareTableData(path, data2, "2b");
         prepareTableData(path, data3, "3c");
         prepareTableData(path, data4, "4d");
+        /*
+         * External Table adds (escape) E to the location which escapes the backslash while FDW doesn't add E to the resource.
```
denalex, line 129: should we change our code that generates FDW create statement to add `E` to the option definition ? If we do that, will original test work ?

hpandeycodeit: Yes, it will work if we can add `E` to the Option definition. For now I have left a TODO note in the Test.  We can address it in the next PR.

PR #966, opened 2023-04-12
Updated locations of GP artifacts in CI pipelines

denalex:
This PR fixes 2 issues:
- GP7 beta artifact now comes from the "released" bucket, not the "published" one
- GP6-RHEL8 artifact comes from the Pivnet mirror bucket, not the "released" bucket

GP7 artifact will continue to come from the "released" bucket and not Pivnet mirror for now until after GP7 GA as we do not yet guarantee backward / forward compatibility of PXF with GP7 beta releases other than the ones a given PXF version has been built with.

### DIFF
--- a/concourse/pipelines/templates/dev_build_pipeline-tpl.yml
+++ b/concourse/pipelines/templates/dev_build_pipeline-tpl.yml
@@ -257,7 +257,7 @@ resources:
     url: ((ud-pipeline-bot-[[user]]-webhook))
 {% endif %}
 
-## ---------- Product Packages ----------
+## ---------- Greenplum Packages ----------
 {% set gp_ver = 5 %}
 - name: gpdb[[gp_ver]]_rhel7_rpm_latest-0
   type: gcs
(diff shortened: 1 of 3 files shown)


concourse/pipelines/templates/dev_build_pipeline-tpl.yml
```diff
@@ -289,9 +289,9 @@ resources:
   type: gcs
   icon: google-drive
   source:
-    bucket: ((ud/pxf/common/gpdb-concourse-resources-prod-bucket-name))
-    json_key: ((concourse-gcs-resources-service-account-key))
-    regexp: server/published/gpdb6/greenplum-db-(6.*)-rhel8-x86_64.rpm
+    bucket: ((ud/pxf/common/pivnet-artifacts-bucket-name))
+    json_key: ((ud/pxf/secrets/pxf-storage-service-account-key))
+    regexp: latest-0_gpdb[[gp_ver]]/greenplum-db-(6.*)-rhel8-x86_64.rpm
```
ashuka24, line 294: since we do 6.* in the regex, do we really need `[[gp_ver]]`

ashuka24 (approved):

concourse/pipelines/templates/dev_build_pipeline-tpl.yml
denalex: good point, I think the design here is that all resource definitions look similar, that's why [[gp_ver]] is used and is set to one version and then the next. The number in regex probably guards against name typos, to make sure we will really use the RPM for the version we think we need.

I'd like to add the number to other resources like this:
```
regexp: latest-0_gpdb[[gp_ver]]/greenplum-db-([[gp_ver]].*)- ...
```

bradfordb-vmware (approved):

ashuka24 (approved):

PR #967, opened 2023-04-14
FDW: Fix for skipping the dropped and correctly counting Projection Index 

hpandeycodeit:
When a column was dropped, The existing logic doesn't work properly : 

- It does not skip the dropped column thus causing the incorrect results. 
- The Projection Index was not getting calculated correctly. 

This PR fixes the issue. 
The above change can be validated with the AlterTableTest test. 

For AlterTableTest `dropColumnsPxfWritableExport` (profile :parquet) fails with the error `PXF server error : class java.io.DataInputStream cannot be cast to class java.lang.String (java.io.DataInputStream and java.lang.String are in module java.base of loader 'bootstrap') `  So skipping it for now as we are in process of fixing the FDW write separately.

### DIFF
--- a/automation/src/main/java/org/greenplum/pxf/automation/structures/tables/utils/TableFactory.java
+++ b/automation/src/main/java/org/greenplum/pxf/automation/structures/tables/utils/TableFactory.java
@@ -657,7 +657,7 @@ public abstract class TableFactory {
     }
 
     /**
-     * Generate a PXF External Writable or Foreign Table using JDBC profile.
+     * Prepares PXF Readable External or Foreign Table for custom data format.
      *
      * @param name name of the external table which will be generated
      * @param fields fields of the external table
--- a/automation/src/main/java/org/greenplum/pxf/automation/structures/tables/utils/TableFactory.java
+++ b/automation/src/main/java/org/greenplum/pxf/automation/structures/tables/utils/TableFactory.java
@@ -3,6 +3,7 @@ package org.greenplum.pxf.automation.structures.tables.utils;
 import java.util.ArrayList;
 import java.util.List;
 
+import org.apache.commons.lang.StringUtils;
 import org.greenplum.pxf.automation.enums.EnumPartitionType;
 
 import org.greenplum.pxf.automation.enums.EnumPxfDefaultProfiles;
@@ -653,6 +654,29 @@ public abstract class TableFactory {
     public static ExternalTable getPxfJdbcWritableTable(String name, String[] fields, String dataSourcePath, String server) {
         String customParameter = server != null ? "SERVER=" + server : null;
         return getPxfJdbcWritableTable(name, fields, dataSourcePath, null, null, null, customParameter);
+    }
+
+    /**
+     * Generate a PXF External Writable or Foreign Table using JDBC profile.
+     *
+     * @param name name of the external table which will be generated
+     * @param fields fields of the external table
+     * @param path for external table path
+     * @param dataFormat dataFormat for the external table
+     * @return PXF Readable External or Foreign table
+     */
+    public static ReadableExternalTable getPxfReadableCustomTable(String name,
+                                                                  String[] fields,
+                                                                  String path,
+                                                                  String dataFormat) {
+        ReadableExternalTable exTable = getReadableExternalOrForeignTable(name, fields, path, "CUSTOM");
+        exTable.setFormatter("pxfwritable_import");
+
+        if (StringUtils.isNotBlank(dataFormat)) {
+            exTable.setProfile(ProtocolUtils.getProtocol().value() + ":" + dataFormat);
+        }
+
+        return exTable;
     }
 
     // ============ FDW Adapter ============
--- a/automation/src/test/java/org/greenplum/pxf/automation/features/columnprojection/ColumnProjectionTest.java
+++ b/automation/src/test/java/org/greenplum/pxf/automation/features/columnprojection/ColumnProjectionTest.java
@@ -55,8 +55,8 @@ public class ColumnProjectionTest extends BaseFeature {
         // SELECT t0, colprojvalue FROM test_column_projection GROUP BY t0, colprojvalue HAVING AVG(a1) < 5 ORDER BY t0;
         // SELECT b.value, a.colprojvalue FROM test_column_projection a JOIN t0_values b ON a.t0 = b.key;
         if (gpdb.getVersion() >= 7) {
-            /** The below query is acting differently for external-table and FDW, As it is projecting for the FDW so the results are little different
-             * from the external table. The Call stack is different in case of external-table and FDW.
+            /** The below query (mentioned in above comment as well) is propagating for FDW but not for external-table, so a different test set for FDW.
+             * The Call stack is different in case of external-table and FDW.
 
              SELECT b.value, a.colprojvalue FROM test_column_projection a JOIN t0_values b ON a.t0 = b.key;
              value |  colprojvalue
--- a/automation/src/test/java/org/greenplum/pxf/automation/features/columnprojection/ColumnProjectionTest.java
+++ b/automation/src/test/java/org/greenplum/pxf/automation/features/columnprojection/ColumnProjectionTest.java
@@ -55,7 +55,7 @@ public class ColumnProjectionTest extends BaseFeature {
         // SELECT t0, colprojvalue FROM test_column_projection GROUP BY t0, colprojvalue HAVING AVG(a1) < 5 ORDER BY t0;
         // SELECT b.value, a.colprojvalue FROM test_column_projection a JOIN t0_values b ON a.t0 = b.key;
         if (gpdb.getVersion() >= 7) {
-            /** The below query (mentioned in above comment as well) is propagating for FDW but not for external-table,
+            /* The below query (mentioned in above comment as well) is propagating for FDW but not for external-table,
              *  so use a different test set for FDW.
              * The Call stack is different in case of external-table and FDW.
 
@@ -133,7 +133,7 @@ public class ColumnProjectionTest extends BaseFeature {
              Memory used:  128000kB
              Execution Time: 117.692 ms
              (16 rows)
-             **/
+             */
             if (FDWUtils.useFDW) {
                 runTincTest("pxf.features.columnprojection.checkColumnProjection_fdw.runTest");
             }
--- a/automation/src/test/java/org/greenplum/pxf/automation/features/columnprojection/ColumnProjectionTest.java
+++ b/automation/src/test/java/org/greenplum/pxf/automation/features/columnprojection/ColumnProjectionTest.java
@@ -54,25 +54,22 @@ public class ColumnProjectionTest extends BaseFeature {
         // TODO: revert when 2 queries with GP7 planner start propagating projection info to foreign scans
         // SELECT t0, colprojvalue FROM test_column_projection GROUP BY t0, colprojvalue HAVING AVG(a1) < 5 ORDER BY t0;
         // SELECT b.value, a.colprojvalue FROM test_column_projection a JOIN t0_values b ON a.t0 = b.key;
-        if (gpdb.getVersion() >= 7 && !FDWUtils.useFDW) {
-            runTincTest("pxf.features.columnprojection.checkColumnProjection_gp7.runTest");
-        } else if (FDWUtils.useFDW) {
-            /** For FDW ( irrespective of GP 6 or GP 7), the query below is projecting the columns so the results are different.
-                 SELECT * FROM test_column_projection ORDER BY t0;
-                 t0 | a1 | b2 |     colprojvalue
-                ----+----+----+-----------------------
-                 A  |  0 | t  | t0|a1|b2|colprojvalue
-                 B  |  1 | f  | t0|a1|b2|colprojvalue
-                 C  |  2 | t  | t0|a1|b2|colprojvalue
-                 D  |  3 | f  | t0|a1|b2|colprojvalue
-                 E  |  4 | t  | t0|a1|b2|colprojvalue
-                 F  |  5 | f  | t0|a1|b2|colprojvalue
-                 G  |  6 | t  | t0|a1|b2|colprojvalue
-                 H  |  7 | f  | t0|a1|b2|colprojvalue
-                 I  |  8 | t  | t0|a1|b2|colprojvalue
-                 J  |  9 | f  | t0|a1|b2|colprojvalue
+        if (gpdb.getVersion() >= 7) {
+            /** The below query is acting differently for external-table and FDW, As it is projecting for the FDW so the results are little different
+             * from the external table. The Call stack is different in case of external-table and FDW.
+
+             SELECT b.value, a.colprojvalue FROM test_column_projection a JOIN t0_values b ON a.t0 = b.key;
+             value |  colprojvalue
+             -------+-----------------
+             50 | t0|colprojvalue
+             (1 row)
              **/
-            runTincTest("pxf.features.columnprojection.checkColumnProjection_fdw.runTest");
+            if (FDWUtils.useFDW) {
+                runTincTest("pxf.features.columnprojection.checkColumnProjection_fdw.runTest");
+            }
+            else {
+                runTincTest("pxf.features.columnprojection.checkColumnProjection_gp7.runTest");
+            }
         } else {
             runTincTest("pxf.features.columnprojection.checkColumnProjection.runTest");
         }
--- a/automation/src/test/java/org/greenplum/pxf/automation/features/columnprojection/ColumnProjectionTest.java
+++ b/automation/src/test/java/org/greenplum/pxf/automation/features/columnprojection/ColumnProjectionTest.java
@@ -55,7 +55,8 @@ public class ColumnProjectionTest extends BaseFeature {
         // SELECT t0, colprojvalue FROM test_column_projection GROUP BY t0, colprojvalue HAVING AVG(a1) < 5 ORDER BY t0;
         // SELECT b.value, a.colprojvalue FROM test_column_projection a JOIN t0_values b ON a.t0 = b.key;
         if (gpdb.getVersion() >= 7) {
-            /** The below query (mentioned in above comment as well) is propagating for FDW but not for external-table, so a different test set for FDW.
+            /** The below query (mentioned in above comment as well) is propagating for FDW but not for external-table,
+             *  so use a different test set for FDW.
              * The Call stack is different in case of external-table and FDW.
 
              SELECT b.value, a.colprojvalue FROM test_column_projection a JOIN t0_values b ON a.t0 = b.key;
@@ -64,8 +65,8 @@ public class ColumnProjectionTest extends BaseFeature {
              50 | t0|colprojvalue
              (1 row)
 
-             There is the explain plan for the external-table and FDW for the same query, '
-             that might be the reason for one it's projecting and for other it's not.
+             Following are the explain plans for the external-table and FDW for the same query,
+             The different explain plans explains that for one it is projecting and for other it's not.
              External Table:
 
              pxfautomation=# \d+ e_test_column_projection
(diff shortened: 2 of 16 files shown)


fdw/pxf_deparse.c
```diff
@@ -33,18 +33,31 @@ deparseTargetList(Relation rel, Bitmapset *attrs_used, List **retrieved_attrs)
...
-			*retrieved_attrs = lappend_int(*retrieved_attrs, i);
+			// Dropped attributes count needs for proper indexing of the projected columns.
+			// For eg:
+			//  ---------------------------------------------
+			// |  col1  |  col2 (dropped)  |  col3  |  col4  |
+			//  ---------------------------------------------
+			//
+			// Let's assume that col1 and col4 are projected, the reported projected
+			// indices will be 0, 2. This is because we use 0-based indexing and because
+			// col2 was dropped, the indices for col3 and col4 get shifted by -1.
```
ashuka24, line 59: I think the ordering of the comments here are a little confusing and took me a little while to parse. 

```suggestion
			// Dropped attributes count needs for proper indexing of the projected columns.
			// For eg:
			//  ---------------------------------------------
			// |  col1  |  col2 (dropped)  |  col3  |  col4  |
			//  ---------------------------------------------
			//
			// We use 0-based indexing and since col2 was dropped, 
			// the indices for col3 and col4 get shifted by -1.
			// Let's assume that col1 and col4 are projected, the reported projected
			// indices will then be 0, 2.
			```

File: fdw/pxf_header.c [hunk 2]
```diff
@@ -282,8 +283,13 @@ AddTupleDescriptionToHttpHeader(CHURL_HEADERS headers, Relation rel)
 					break;
 			}
 		}
+		attrIx++;
 	}
 
+	/* Convert the number of attributes to a string */
+	pg_ltoa(attrIx, long_number);
```
ashuka24, line 290: Seems like it is possible to increment `attrIx` before we set this value, is that expected behavior?

fdw/pxf_header.c
```diff
@@ -180,29 +181,29 @@ AddTupleDescriptionToHttpHeader(CHURL_HEADERS headers, Relation rel)
...
-	churl_headers_append(headers, "X-GP-ATTRS", long_number);
-
 	/* Iterate attributes */
-	for (int i = 0; i < tuple->natts; ++i)
+	for (i = 0, attrIx = 0; i < tuple->natts; ++i)
 	{
 		Form_pg_attribute attr = TupleDescAttr(tuple, i);
 
+		// Ignore dropped attributes
+        if(attr->attisdropped)
```
denalex, line 190:
```suggestion
        if (attr->attisdropped)
```
also check alignment

fdw/pxf_header.c
```diff
@@ -282,8 +283,13 @@ AddTupleDescriptionToHttpHeader(CHURL_HEADERS headers, Relation rel)
...
 		}
+		attrIx++;
 	}
 
+	/* Convert the number of attributes to a string */
+	pg_ltoa(attrIx, long_number);
+	churl_headers_append(headers, "X-GP-ATTRS", long_number);
+
 	pfree(formatter.data);
 }
```
denalex, line 295: I assume we are porting to FDW a commit https://github.com/greenplum-db/pxf/commit/a29c6614c6f965fcd3d0ec0471f72f5c2d4466fd from the external table code. In this case, do we also need to worry about the logic for `AddProjectionDescHttpHeader` where the projection index is shifted by the number of dropped attributes before the current one ? I think you ported that logic into `deparse.c`, but I think it better be used only near where we send http headers (see the other comment above as well).

fdw/pxf_deparse.c
```diff
@@ -33,18 +33,31 @@ deparseTargetList(Relation rel, Bitmapset *attrs_used, List **retrieved_attrs)
...
+			*retrieved_attrs = lappend_int(*retrieved_attrs, i - droppedAttrsCount);
```
denalex, line 60: this results in having `retrieved_attrs` list contain adjusted attribute indices that are only relevant to PXF server-side and used for http headers. If some other logic gets a hold of this list, they can get incorrect indices. Should we not do it here and rater do it in the logic that is close to sending http headers ?

automation/tincrepo/main/pxf/features/general/alter/csv/expected/query01.ans
```diff
@@ -18,6 +18,9 @@
 -- m/CONTEXT:.*line.*/
 -- s/line \d* of //g
 --
+-- m/tmp\/pxf_automation_data\/(.*)/
+-- s/tmp\/pxf_automation_data\/(.*)/pxf:\/\/pxf_automation_data?PROFILE=*:text/g
+--
```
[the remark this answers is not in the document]
hpandeycodeit: FDW has this resource info return which is different from the external table case: 

```
< CONTEXT:  External table pxf_alter_csv_table, tmp/pxf_automation_data/f66f75c0-d5fb-4f36-8b36-a3b3f9c02ae9/alter-tests/csv/data.txt, column num: "1.0"
---
> CONTEXT:  External table pxf_alter_csv_table, pxf://pxf_automation_data?PROFILE=*:text
./tincrepo/main/pxf/features/general/alter/csv/output/query01_2023418223140453156.diff (END)

```

automation/tincrepo/main/pxf/features/general/alter/pxfwritable_import/without_column_projection/expected/query01.ans
```diff
@@ -8,6 +8,9 @@
 -- m/DETAIL/
 -- s/DETAIL/CONTEXT/
 --
+-- m/,.*line (.*)/
+-- s/,.*line (.*)//g
+--
```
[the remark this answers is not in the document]
hpandeycodeit: yes, basically this is how it comes in FDW: 

```
< CONTEXT:  External table pxf_alter_avro_table, line 1 of tmp/pxf_automation_data/ce45a178-6e9c-4752-a1c8-1f5779d5d744/alter-tests/avro/supported_primitive_types.avro

```

automation/tincrepo/main/pxf/features/general/alter/csv/expected/query01.ans
```diff
@@ -18,6 +18,9 @@
 -- m/CONTEXT:.*line.*/
 -- s/line \d* of //g
 --
+-- m/tmp\/pxf_automation_data\/(.*)/
```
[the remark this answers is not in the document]
hpandeycodeit: FDW has this resource info return which is different from the external table case:

< CONTEXT:  External table pxf_alter_csv_table, tmp/pxf_automation_data/f66f75c0-d5fb-4f36-8b36-a3b3f9c02ae9/alter-tests/csv/data.txt, column num: "1.0"
---
> CONTEXT:  External table pxf_alter_csv_table, pxf://pxf_automation_data?PROFILE=*:text
./tincrepo/main/pxf/features/general/alter/csv/output/query01_2023418223140453156.diff (END)

fdw/pxf_header.c
hpandeycodeit: yes, this is same as external table. 
https://github.com/greenplum-db/pxf/blob/086c06c88bdd6b29cb847cf2620e6c1bcecdd0a5/external-table/src/pxfheaders.c#L373

fdw/pxf_header.c
```diff
@@ -180,29 +181,29 @@ AddTupleDescriptionToHttpHeader(CHURL_HEADERS headers, Relation rel)
 	/* Get tuple description itself */
 	tuple = RelationGetDescr(rel);
 
-	/* Convert the number of attributes to a string */
-	pg_ltoa(tuple->natts, long_number);
-	churl_headers_append(headers, "X-GP-ATTRS", long_number);
-
 	/* Iterate attributes */
-	for (int i = 0; i < tuple->natts; ++i)
+	for (i = 0, attrIx = 0; i < tuple->natts; ++i)
```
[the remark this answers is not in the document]
hpandeycodeit: Since this is the similar logic as external table, I have added the same description here. Hopefully that helps.

automation/src/main/java/org/greenplum/pxf/automation/structures/tables/utils/TableFactory.java
```diff
@@ -655,6 +656,29 @@ public static ExternalTable getPxfJdbcWritableTable(String name, String[] fields
         return getPxfJdbcWritableTable(name, fields, dataSourcePath, null, null, null, customParameter);
     }
 
+    /**
+     * Generate a PXF External Writable or Foreign Table using JDBC profile.
+     *
```
yimingli-vmware, line 661: From code we are returning a readable external table, so should it be something like the following?
```suggestion
     * Generate a PXF External Readable or Foreign Table using JDBC profile.
```

automation/src/test/java/org/greenplum/pxf/automation/features/general/AlterTableTest.java
```diff
@@ -89,46 +90,41 @@ public void beforeClass() throws Exception {
...
         exTable.setPort(pxfPort);
-        exTable.setFormatter("pxfwritable_import");
-        exTable.setProfile(ProtocolUtils.getProtocol().value() + ":parquet");
-
         gpdb.createTableAndVerify(exTable);
 
         runTincTest("pxf.features.general.alter.pxfwritable_import.with_column_projection.runTest");
     }
 
+    @FailsWithFDW
```
denalex, line 102: can we comment here why it fails so we can fix it later ?

fdw/pxf_header.c
```diff
@@ -282,33 +294,57 @@ AddTupleDescriptionToHttpHeader(CHURL_HEADERS headers, Relation rel)
...
+		// For eg:
+		//  ---------------------------------------------
+		// |  col1  |  col2 (dropped)  |  col3  |  col4  |
+		//  ---------------------------------------------
+		//
+		// We use 0-based indexing and since col2 was dropped,
+		// the indices for col3 and col4 get shifted by -1.
+		// Let's assume that col1 and col4 are projected, the reported projected
+		// indices will then be 0, 2.
+		if(TupleDescAttr(tupdesc, i-1)->attisdropped)
```
denalex, line 331:
```suggestion
		if (TupleDescAttr(tupdesc, i-1)->attisdropped)
```

fdw/pxf_header.c
```diff
@@ -282,33 +294,57 @@ AddTupleDescriptionToHttpHeader(CHURL_HEADERS headers, Relation rel)
...
+AddProjectionDescHttpHeader(CHURL_HEADERS headers, List *retrieved_attrs, Relation rel)
 {
 	ListCell   *lc1 = NULL;
 	char		long_number[sizeof(int32) * 8];
+	TupleDesc	tupdesc = RelationGetDescr(rel);
+	int droppedCount = 0;
+	int headerCount = 0;
 
-	foreach(lc1, retrieved_attrs)
+	for (int i = 1; i <= tupdesc->natts; i++)
```
denalex, line 319: should we be iterating over `retrieved_attrs` rather than over `tupdesc->attrs`, otherwise what's the point of projection ? Maybe we should be checking TupleDesc if the attr is dropped, but I don't even know if it is even possible to get the dropped attribute in retrieved list. If my assumptions is correct, why is the test not failing for this change ?

automation/src/test/java/org/greenplum/pxf/automation/features/columnprojection/ColumnProjectionTest.java
```diff
@@ -53,8 +54,25 @@ public void checkColumnProjection() throws Exception {
         // TODO: revert when 2 queries with GP7 planner start propagating projection info to foreign scans
         // SELECT t0, colprojvalue FROM test_column_projection GROUP BY t0, colprojvalue HAVING AVG(a1) < 5 ORDER BY t0;
         // SELECT b.value, a.colprojvalue FROM test_column_projection a JOIN t0_values b ON a.t0 = b.key;
-        if (gpdb.getVersion() >= 7) {
+        if (gpdb.getVersion() >= 7 && !FDWUtils.useFDW) {
             runTincTest("pxf.features.columnprojection.checkColumnProjection_gp7.runTest");
+        } else if (FDWUtils.useFDW) {
+            /** For FDW ( irrespective of GP 6 or GP 7), the query below is projecting the columns so the results are different.
```
ashuka24, line 60: likewise, are the results of this different between planner and ORCA?

automation/src/test/java/org/greenplum/pxf/automation/features/columnprojection/ColumnProjectionTest.java
```diff
@@ -53,8 +54,25 @@ public void checkColumnProjection() throws Exception {
         // TODO: revert when 2 queries with GP7 planner start propagating projection info to foreign scans
```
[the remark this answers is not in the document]
hpandeycodeit: we are testing it with both ORCA and planner just that in planner it is not propagating. That's why the results are different for FDW and external-table and for GP6.

automation/src/test/java/org/greenplum/pxf/automation/features/columnprojection/ColumnProjectionTest.java
hpandeycodeit: Yes, same comment as above. It is propagating for ORCA but not for planner.

automation/src/test/java/org/greenplum/pxf/automation/features/columnprojection/ColumnProjectionTest.java
```diff
@@ -54,25 +54,91 @@ public void checkColumnProjection() throws Exception {
...
-                 C  |  2 | t  | t0|a1|b2|colprojvalue
-                 D  |  3 | f  | t0|a1|b2|colprojvalue
-                 E  |  4 | t  | t0|a1|b2|colprojvalue
-                 F  |  5 | f  | t0|a1|b2|colprojvalue
-                 G  |  6 | t  | t0|a1|b2|colprojvalue
-                 H  |  7 | f  | t0|a1|b2|colprojvalue
-                 I  |  8 | t  | t0|a1|b2|colprojvalue
-                 J  |  9 | f  | t0|a1|b2|colprojvalue
+        if (gpdb.getVersion() >= 7) {
+            /** The below query (mentioned in above comment as well) is propagating for FDW but not for external-table, so a different test set for FDW.
```
ashuka24, line 58:
```suggestion
            /* The below query (mentioned in above comment as well) is propagating for FDW but not for external-table, so use a different test set for FDW.

File: automation/src/test/java/org/greenplum/pxf/automation/features/columnprojection/ColumnProjectionTest.java [hunk 17]
```diff
@@ -54,25 +54,91 @@ public void checkColumnProjection() throws Exception {
...
+             * The Call stack is different in case of external-table and FDW.
+
+             SELECT b.value, a.colprojvalue FROM test_column_projection a JOIN t0_values b ON a.t0 = b.key;
+             value |  colprojvalue
+             -------+-----------------
+             50 | t0|colprojvalue
+             (1 row)
+
+             There is the explain plan for the external-table and FDW for the same query, '
```
ashuka24, line 67:
```suggestion
             The following is the explain plan for the external-table and FDW for the same query,

File: automation/src/test/java/org/greenplum/pxf/automation/features/columnprojection/ColumnProjectionTest.java [hunk 18]
```diff
@@ -54,25 +54,91 @@ public void checkColumnProjection() throws Exception {
...
+             that might be the reason for one it's projecting and for other it's not.
```
ashuka24, line 68: can we be specific here? "that might be the reason it is projecting for FDW and not for external-table

fdw/pxf_header.c
```diff
@@ -71,10 +72,12 @@ BuildHttpHeaders(CHURL_HEADERS headers,
...
 	}
 
-	if (retrieved_attrs != NULL)
+	// If projectionInfo is Null, it means there should be no projection.
+	// We can skip the whole logic of adding projection desc to headers.
+	if (projectionInfo != NULL)
 	{
 		/* add the list of attrs to the projection desc http headers */
-		AddProjectionDescHttpHeader(headers, retrieved_attrs, relation);
+		AddProjectionDescHttpHeader(headers, retrieved_attrs,  relation);
```
ashuka24, line 80: nit: extra space
```suggestion
		AddProjectionDescHttpHeader(headers, retrieved_attrs, relation);

File: automation/src/test/java/org/greenplum/pxf/automation/features/general/AlterTableTest.java [hunk 20]
```diff
@@ -178,10 +165,14 @@ public void dropAndAddColumsCsv() throws Exception {
                         "longNum bigint",
                         "bool boolean"
                 }, hdfsPath + "/csv/" + fileName, ",");
+        setParamsAndVerifyTable();
+
+        runTincTest("pxf.features.general.alter.csv.runTest");
+    }
+    private void setParamsAndVerifyTable() throws Exception
```
ashuka24, line 172: add new line between 171 and 172

fdw/pxf_header.c
```diff
@@ -282,34 +297,69 @@ AddTupleDescriptionToHttpHeader(CHURL_HEADERS headers, Relation rel)
...
-		AddProjectionIndexHeader(headers, attno - 1, long_number);
+		attrs_projected = bms_add_member(attrs_projected,
+							attno - FirstLowInvalidHeapAttributeNumber);
 	}
 
-	if (retrieved_attrs->length == 0)
-		return;
+	for (int i = 1; i <= tupdesc->natts; i++)
+	{
+		// Dropped attributes count needs for proper indexing of the projected columns.
```
ashuka24, line 335: can we use `/* */` for long comments

automation/src/test/java/org/greenplum/pxf/automation/features/columnprojection/ColumnProjectionTest.java
```diff
@@ -54,7 +55,91 @@ public void checkColumnProjection() throws Exception {
         // SELECT t0, colprojvalue FROM test_column_projection GROUP BY t0, colprojvalue HAVING AVG(a1) < 5 ORDER BY t0;
         // SELECT b.value, a.colprojvalue FROM test_column_projection a JOIN t0_values b ON a.t0 = b.key;
         if (gpdb.getVersion() >= 7) {
-            runTincTest("pxf.features.columnprojection.checkColumnProjection_gp7.runTest");
+            /** The below query (mentioned in above comment as well) is propagating for FDW but not for external-table,
```
ashuka24, line 58: i believe double ** is used for java docs, but here we are just writing a comment so we can just use one *
```suggestion
            /* The below query (mentioned in above comment as well) is propagating for FDW but not for external-table,

File: automation/src/test/java/org/greenplum/pxf/automation/features/general/AlterTableTest.java [hunk 23]
```diff
@@ -89,69 +90,55 @@ public void beforeClass() throws Exception {
...
+                PARQUET_TABLE_COLUMNS, hdfsPath + "/parquet/" + PARQUET_PRIMITIVE_TYPES, "parquet");
+        setParamsAndVerifyTable();
 
         // Create writable table
-        exTable = new WritableExternalTable(PXF_ALTER_WRITE_PARQUET_TABLE,
-                PARQUET_TABLE_COLUMNS, hdfsPath + "/parquet-write/" + PARQUET_WRITE_PRIMITIVES, "custom");
-        exTable.setHost(pxfHost);
-        exTable.setPort(pxfPort);
+        exTable = TableFactory.getPxfWritableTextTable(PXF_ALTER_WRITE_PARQUET_TABLE,
+                PARQUET_TABLE_COLUMNS, hdfsPath + "/parquet-write/" + PARQUET_WRITE_PRIMITIVES, null);
```
ashuka24, line 115: seems like the old test was setting the profile to be parquet. why are we using null now?

hpandeycodeit: We are still setting the `profile` to `parquet`. If you are referring to the last param here, that is the delimiter which is null in this case and this is same as before.

ashuka24 (approved):

yimingli-vmware (approved):

PR #975, opened 2023-05-18
Add Make variable to conditionally include check-creds

bradfordb-vmware:
When setting a Concourse pipeline, the set-pipeline command can take an optional flag named --check-creds that will validate credential variables against the credential manager. Lately this validation can take an extended period of time which can slow down the development cycle when iterating on CI pipelines. Rather than remove the flag altogether and risk missing an incorrect credential in a production pipeline, this commit adds a new Make varialble named CHECK_CREDS which defaults to true. If this variable is true, then the --check-creds flag to set-pipeline will be included.

### DIFF
--- a/concourse/Makefile
+++ b/concourse/Makefile
@@ -18,6 +18,7 @@ NUM_GPDB5_VERSIONS          ?= 10
 NUM_GPDB6_VERSIONS          ?=  9
 REDHAT_MAJOR_VERSION        ?= 7
 FLY_CMD                     ?= fly
+CHECK_CREDS                 ?= true
 TEMPLATE_CMD                 = ./template_tool
 FLY_OPTION_NON-INTERACTIVE   =
 SLACK                       ?= true
@@ -37,6 +38,11 @@ MINIO                       ?= false
 OEL7                        ?= false
 FILE                        ?= false
 
+SET_PIPELINE := set-pipeline
+ifeq ($(CHECK_CREDS), true)
+SET_PIPELINE += --check-creds
+endif
+
 .PHONY: build certification dev pr cloudbuild longevity
 build: set-build-pipeline
 certification: set-certification-pipeline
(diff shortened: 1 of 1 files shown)


ashuka24 (approved):

concourse/Makefile
```diff
@@ -52,12 +58,11 @@ set-build-pipeline:
 	@PIPELINE_FILE=$$(mktemp) && \
 	$(TEMPLATE_CMD) --template build_pipeline-tpl.yml --vars \
 		environment=prod \
-		slack_notification=$(SLACK) \
+		slack_notification=true \
```
denalex, line 61: I don't think this change was intended ?

concourse/Makefile
```diff
@@ -37,6 +38,11 @@ MINIO                       ?= false
 OEL7                        ?= false
 FILE                        ?= false
 
+SET_PIPELINE := set-pipeline
+ifeq ($(CHECK_CREDS), true)
```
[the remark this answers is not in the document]
bradfordb-vmware: I don't think so -- with the Slack notifications, we would updated the value of the Make variable `SLACK` which would then be used to set the value of the template variable `slack_notification` which was passed into `template_tool`. Here we using `CHECK_CREDS` to change what command Make runs when setting the pipeline.

concourse/Makefile
bradfordb-vmware: The change was intentional but not well communicated.

Setting the template variable `slack_notification` to false does the following items in the build pipeline:

1. Removes the `on_success` and `on_failure` steps that `put` to `slack-alert` in `Promote PXF-GP5 and PXF-GP6 Artifacts` job
1. Removes `notify-releng` step from `Publish_PXF-GP5_and_PXF-GP6_Artifacts_to_GP-RelEng`

I don't think either of these Slack notifications should ever be omitted from the build pipeline and they should not be dependent on what environment variables may or may not bet set in a developers environment.

We could remove `slack_notification` altogether from the build pipeline but then this would mean that any dev release pipeline would contain the Slack notifications. This may be OK since dev release pipelines don't get set frequently and when they do get set, it might not be a bad idea to validate that Slack messages still get sent properly. The other option (which is what I did here) is leave `slack_notification` in the build pipeline template but always set it to true when setting the production pipeline.

PR #977, opened 2023-05-23
docs - multibyte delimiter support for csv

lisakowen:
docs for https://github.com/greenplum-db/pxf/pull/956.

the `*:text` and `*:csv profiles` support now multi-byte and multiple character delimiters.  in this PR:
- add a topic "About Reading Data Containing Multi-Byte or Multi-Character Delimiters" to the hdfs page to describe the new feature.
- add a few example CREATE EXTERNAL TABLE calls, didn't create a step-by-step

notes:
- CREATE EXTERNAL TABLE supports DELIMITER/QUOTE/ESCAPE/NEWLINE.  this new formatter does as well, but it is based on the COPY framework.  COPY supports these as well.  i tried to call out what was different with this new formatter, not sure if i covered everything.
- it didn't seem to me that we needed any upgrade doc modifications, we already instruct the user to `ALTER EXTENSION pxf UPDATE` after they `pxf cluster register`.
- when the content is approved, i will add a parallel section or xref in the object store docs.

doc review site link (behind vpn):  https://docs-staging.vmware.com/en/VMware-Tanzu-Greenplum-Platform-Extension-Framework/mbdelim/tanzu-greenplum-platform-extension-framework/GUID-hdfs_text.html#about-reading-data-containing-multibyte-or-multicharacter-delimiters-7

### DIFF
--- a/docs/content/hdfs_text.html.md.erb
+++ b/docs/content/hdfs_text.html.md.erb
@@ -395,7 +395,7 @@ PXF recognizes the following formatter options when reading data that contains a
 | DELIMITER=\<delim_string\> | The single-byte or multi-byte delimiter string that separates columns. The string may be up to 32 bytes in length, and may not contain quote or escape characters. **Required** | None |
 | QUOTE=\<char\> | The single one-byte ASCII quotation character for all columns. | None |
 | ESCAPE=\<char\> | The single one-byte ASCII character used to escape special characters (for example, the `DELIM`, `QUOTE`,  or `NEWLINE` value, or the `ESCAPE` value itself). | None, or the `QUOTE` value if that is set |
-| NEWLINE=\<value\> | The end-of-line indicator that designates the end of a row. Preface the \<value\> with an `E` when the value is an escape sequence. Examples: `NEWLINE=E'\r'`, `NEWLINE 'aaa'`. | `\n` |
+| NEWLINE=\<bytecode\> | The end-of-line indicator that designates the end of a row. Valid values are `LF` (line feed), `CR` (carriage return), or `CRLF` (carriage return plus line feed. | `LF` |
 
 The following sections provide further information about, and examples for, specifying the delimiter, quote, escape, and new line options.
 
@@ -452,9 +452,17 @@ When PXF reads data that contains a multi-byte or multi-character delimiter, its
 
 ### <a id="about_newline"></a>About the NEWLINE Options
 
-Every line in the file must be terminated with the specified new line value.
+PXF requires that every line in the file be terminated with the same new line value.
 
-By default, PXF uses the `\n` (LF) character for the new line delimiter. When the line delimiter for the external file is also `\n`, you need not specify the `NEWLINE` formatter option. If the `NEWLINE` formatter option is provided and contains `\r` (CR), `\r\n` (CRLF), or a set of custom escape characters, you must also specify the `NEWLINE` option in the external table `LOCATION` URI, and set the value to `CR`, `CRLF` or the set of bytecode characters, respectively.
+By default, PXF uses the line feed character (`LF`) for the new line delimiter. When the new line delimiter for the external file is also a line feed, you need not specify the `NEWLINE` formatter option.
+
+If the `NEWLINE` formatter option is provided and contains `CR` or `CRLF`, you must also specify the same `NEWLINE` option in the external table `LOCATION` URI. For example, if the new line delimiter is `CRLF`, create the external table as follows:
+
+```
+CREATE READABLE EXTERNAL TABLE mbyte_newline_crlf (id int, city text, country text)
+  LOCATION ('pxf://multibyte_example_crlf?PROFILE=hdfs:csv&NEWLINE=CRLF')
+FORMAT 'CUSTOM' (FORMATTER='pxfdelimited_import', DELIMITER='¤', NEWLINE='CRLF');
+```
 
 ### <a id="mbd_examples"></a>Examples
 
--- a/docs/content/hdfs_text.html.md.erb
+++ b/docs/content/hdfs_text.html.md.erb
@@ -395,9 +395,9 @@ PXF recognizes the following formatter options when reading data that contains a
 | DELIMITER=\<delim_string\> | The single-byte or multi-byte delimiter string that separates columns. The string may be up to 32 bytes in length, and may not contain quote or escape characters. **Required** | None |
 | QUOTE=\<char\> | The single one-byte ASCII quotation character for all columns. | None |
 | ESCAPE=\<char\> | The single one-byte ASCII character used to escape special characters (for example, the `DELIM`, `QUOTE`,  or `NEWLINE` value, or the `ESCAPE` value itself). | None, or the `QUOTE` value if that is set |
-| NEWLINE=\<bytecode\> | The end-of-line indicator that designates the end of a row. Every line in the file must be terminated with the specified newline value. When you set the `NEWLINE` formatter option, you must also set `NEWLINE` in the `LOCATION` URI. Valid values are `LF` (line feed, `0x0A`), `CR` (carriage return, `0x0D`), or `CRLF` (carriage return plus line feed, `0x0D` `0x0A`). | `LF` |
+| NEWLINE=\<value\> | The end-of-line indicator that designates the end of a row. Preface the \<value\> with an `E` when the value is an escape sequence. Examples: `NEWLINE=E'\r'`, `NEWLINE 'aaa'`. | `\n` |
 
-The following sections provide further information about, and examples for, specifying the delimiter, quote, and escape options.
+The following sections provide further information about, and examples for, specifying the delimiter, quote, escape, and new line options.
 
 ### <a id="about_delim"></a>Specifying the Delimiter
 
@@ -449,6 +449,12 @@ When PXF reads data that contains a multi-byte or multi-character delimiter, its
 <sup>2</sup> All data columns must quoted when you specify a quote character.
 
 > **Note** PXF expects that there are no extraneous characters between the quote value and the delimiter value, nor between the quote value and the end-of-line value. Additionally, there must be no white space between delimiters and quotes.
+
+### <a id="about_newline"></a>About the NEWLINE Options
+
+Every line in the file must be terminated with the specified new line value.
+
+By default, PXF uses the `\n` (LF) character for the new line delimiter. When the line delimiter for the external file is also `\n`, you need not specify the `NEWLINE` formatter option. If the `NEWLINE` formatter option is provided and contains `\r` (CR), `\r\n` (CRLF), or a set of custom escape characters, you must also specify the `NEWLINE` option in the external table `LOCATION` URI, and set the value to `CR`, `CRLF` or the set of bytecode characters, respectively.
 
 ### <a id="mbd_examples"></a>Examples
 
--- a/docs/content/hdfs_text.html.md.erb
+++ b/docs/content/hdfs_text.html.md.erb
@@ -367,7 +367,7 @@ Perform the following procedure to create Greenplum Database writable external t
 
 ## <a id="multibyte_delim"></a>About Reading Data Containing Multi-Byte or Multi-Character Delimiters
 
-When data contains a multi-byte delimiter or multiple delimiter characters, you can use only a `*:text` or `*:csv` PXF profile to read the text. The syntax for creating a readable external table for such data follows:
+When data contains a multi-byte delimiter or multiple delimiter characters, you can use only a `*:text` or `*:csv` PXF profile to read the data. The syntax for creating a readable external table for such data follows:
 
 ``` sql
 CREATE EXTERNAL TABLE <table_name>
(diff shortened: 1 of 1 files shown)


docs/content/hdfs_text.html.md.erb
```diff
@@ -365,3 +365,97 @@ Perform the following procedure to create Greenplum Database writable external t
...
+FORMAT 'CUSTOM' (FORMATTER='pxfdelimited_import', DELIMITER='¤'); 
+```
+
+Because some characters have different byte representations in different encodings, when you specify the byte representation for the delimiter, you must represent it in the database encoding. For example, if the database encoding is `UTF8`, the file encoding is `LATIN1`, and the delimiter is the `¤` currency symbol, you must specify the `UTF8` byte representation for `¤` (`\xC2\xA4`):
+
+```
+CREATE READABLE EXTERNAL TABLE byterep_delim (id int, city text, state text)
+  LOCATION ('pxf://multibyte_example?PROFILE=hdfs:text')
+FORMAT 'CUSTOM' (FORMATTER='pxfdelimited_import', DELIMITER=E'\xC2\xA4') ENCODING 'LATIN1';
+```
```
denalex, line 410: it is not really clear when this needs to be done. I think it looks like a user needs to do `E'..'` style when the table encoding is different from the database encoding ? If this is the case we need to clearly state this.

docs/content/hdfs_text.html.md.erb
```diff
@@ -365,3 +365,97 @@ Perform the following procedure to create Greenplum Database writable external t
...
+| ESCAPE | The single one-byte ASCII character used to escape special characters (for example, the `DELIM`, `QUOTE`,  or `NEWLINE` value, or the `ESCAPE` value itself). | The `QUOTE` value. |
+| NEWLINE | The end-of-line indicator that designates the end of a row. Every line in the file must be terminated with the specified newline value. Valid values are `LF` (line feed, `0x0A`), `CRLF` (carriage return, `0x0D`), or `CR` (carriage return plus line feed, `0x0D` `0x0A`). | `LF` |
+
+The following sections provide further information about the delimiter, quote, and escape options.
+
+### <a id="about_delim"></a>Specifying the Delimiter
+
+You can directly specify the delimiter or provide its byte representation.
+
+When you directly specify the delimiter, Greenplum Database converts between the file encoding and the database encoding for you. The following example directly specifies the `¤` currency symbol for the delimiter:
```
denalex, line 396: `When you directly specify the delimiter, Greenplum Database converts between the file encoding and the database encoding for you.` -- I am not sure what implications of this are. Is it good or bad ?

docs/content/hdfs_text.html.md.erb
```diff
@@ -365,3 +365,97 @@ Perform the following procedure to create Greenplum Database writable external t
...
+CREATE READABLE EXTERNAL TABLE byterep_delim (id int, city text, state text)
+  LOCATION ('pxf://multibyte_example?PROFILE=hdfs:text')
+FORMAT 'CUSTOM' (FORMATTER='pxfdelimited_import', DELIMITER=E'\xC2\xA4') ENCODING 'LATIN1';
+```
+
+### <a id="about_quote"></a>Specifying the Quotation Character
+
+If you do not specify a quotation character for CSV data, PXF assumes that all columns are unquoted.
+
+When you specify a quotation character, PXF expects that all columns are quoted, and you must also specify an escape character. PXF reads any content between quote characters as-is, except for escaped characters.
```
denalex, line 416: `you must also specify an escape character.` - this is not what the example below shows, where no escape character is explicitly specified.

docs/content/hdfs_text.html.md.erb
```diff
@@ -365,3 +365,97 @@ Perform the following procedure to create Greenplum Database writable external t
...
+FORMAT 'CUSTOM' (FORMATTER='pxfdelimited_import', DELIMITER='¤', ESCAPE '\');
+```
+
+And with this external table definition, PXF reads the text field as:
+
+```
+"hello, my name is jane" she said. let's escape something \
+```
+
+When you specify an escape character and do not specify a quote character, PXF assumes that all columns are unquoted and escapes only the delimiter, newline, and escape itself.
```
denalex, line 460: in the example above `specify an escape character and do not specify a quote character` was true but all the columns were quoted, whereas the rest of the sentence says `PXF assumes that all columns are unquoted`

docs/content/hdfs_text.html.md.erb
```diff
@@ -365,3 +365,97 @@ Perform the following procedure to create Greenplum Database writable external t
...
+```
+
+Note the `FORMAT` line in the syntax block. While the syntax is similar to that of reading plain text or CSV, PXF requires a custom formatter to read data containing a multi-byte or multi-character delimiter. You must specify the `'CUSTOM'` format and the `pxfdelimited_import` formatter. You must also specify a delimiter in the formatter options.
+
+PXF recognizes the following formatter options:
+
+| Option Name  | Description | Default Value |
+|-------|---------------|---------------|
+| DELIMITER | The single-byte or multi-byte delimiter string that separates columns. The string may be up to 32 bytes in length, and may not contain quote or escape characters. **Required** | None |
+| QUOTE | When reading CSV data with a `*:csv` profile, the single one-byte ASCII quotation character for all columns. | `"` (double quote) |
```
denalex, line 386: what about `*:text` profile ?

also, what about null values ? For CSV it uses no value between delimiters, while for TEXT it uses `\N`. How does this formatter handle it ? Is there a way to specify a null value ?

docs/content/hdfs_text.html.md.erb
ashuka24: If you write `¤` or if you write `停` in the table definition, GPDB saves it internally in database encoding. 

Maybe it is better to say that Greenplum Database converts the option provided in the table definition into the database encoding

lisakowen:
hold off on reviewing, i am going to revise the docs a bit.

lisakowen:
@greenplum-db/gpdb-ud - ok, i pushed some changes to this topic and re-pushed the review site, please refresh.  i re-orged a bit, reframed some of the original content, and added an example section.  also added a level 2 topic about setting the table encoding (not specific to multibyte delimiter).

since setting the byte representation of the delimiter isn't required, i added it as a sub-section of the "setting the delimiter" section.  wondering if it is unnecessarily complicating matters, should i just remove it?

docs/content/hdfs_text.html.md.erb
```diff
@@ -365,3 +365,134 @@ Perform the following procedure to create Greenplum Database writable external t
...
     To query data from the newly-created HDFS directory named `pxfwritable_hdfs_textsimple2`, you can create a readable external Greenplum Database table as described above that references this HDFS directory and specifies `FORMAT 'CSV' (delimiter=':')`.
 
+## <a id="about_encoding"></a>About Setting the External Table Encoding
+
+When the external file encoding differs from the database encoding, you must set the external table `ENCODING` to match that of the data file, create the external table as follows.
+
+For example, if the database encoding is `UTF8`, the file encoding is `LATIN1`, and the delimiter is the `¤` currency symbol:
+
+```
+CREATE READABLE EXTERNAL TABLE byterep_delim (id int, city text, state text)
```
ashuka24, line 375: not sure if it was intended to have 2 create statements here, but the first table name is `byterep_delim` but the delimiter is not in the byte representation.

docs/content/hdfs_text.html.md.erb
```diff
@@ -365,3 +365,134 @@ Perform the following procedure to create Greenplum Database writable external t
 
     To query data from the newly-created HDFS directory named `pxfwritable_hdfs_textsimple2`, you can create a readable external Greenplum Database table as described above that references this HDFS directory and specifies `FORMAT 'CSV' (delimiter=':')`.
 
+## <a id="about_encoding"></a>About Setting the External Table Encoding
+
+When the external file encoding differs from the database encoding, you must set the external table `ENCODING` to match that of the data file, create the external table as follows.
```
ashuka24, line 370:
```suggestion
When the external file encoding differs from the database encoding, you must set the external table `ENCODING` to match that of the data file, create the external table as follows:

File: docs/content/hdfs_text.html.md.erb [hunk 8]
```diff
@@ -365,3 +365,134 @@ Perform the following procedure to create Greenplum Database writable external t
...
+
+PXF recognizes the following formatter options when reading data that contains a multi-byte or multi-character delimiter:
+
+| Option Name  | Value Description | Default Value |
+|-------|---------------|---------------|
+| DELIMITER=\<delim_string\> | The single-byte or multi-byte delimiter string that separates columns. The string may be up to 32 bytes in length, and may not contain quote or escape characters. **Required** | None |
+| QUOTE=\<char\> | The single one-byte ASCII quotation character for all columns. | None |
+| ESCAPE=\<char\> | The single one-byte ASCII character used to escape special characters (for example, the `DELIM`, `QUOTE`,  or `NEWLINE` value, or the `ESCAPE` value itself). | None, or the `QUOTE` value if that is set |
+| NEWLINE=\<bytecode\> | The end-of-line indicator that designates the end of a row. Every line in the file must be terminated with the specified newline value. Valid values are `LF` (line feed, `0x0A`), `CRLF` (carriage return, `0x0D`), or `CR` (carriage return plus line feed, `0x0D` `0x0A`). | `LF` |
```
ashuka24, line 403: can we add a note here that if they set newline in the format options, it must also be set in the location uri?

docs/content/hdfs_text.html.md.erb
```diff
@@ -365,3 +365,134 @@ Perform the following procedure to create Greenplum Database writable external t
...
+
+```
+CREATE READABLE EXTERNAL TABLE mybte_delim_quoted_escaped (sentence text, num int)
+  LOCATION ('pxf://multibyte_qe?PROFILE=hdfs:csv')
+FORMAT 'CUSTOM' (FORMATTER='pxfdelimited_import', DELIMITER='¤', QUOTE '"', ESCAPE '\');
+```
+
+Given this external table definition, PXF reads the `sentence` text field as:
+
+```
```
ashuka24, line 495: this may be more clear if you did a `SELECT sentence FROM mbyte_delim_quoted_escaped`

docs/content/hdfs_text.html.md.erb
```diff
@@ -365,3 +365,134 @@ Perform the following procedure to create Greenplum Database writable external t
...
+
+For example, if the database encoding is `UTF8`, the file encoding is `LATIN1`, and the delimiter is the `¤` currency symbol:
+
+```
+CREATE READABLE EXTERNAL TABLE byterep_delim (id int, city text, state text)
+  LOCATION ('pxf://multibyte_example?PROFILE=hdfs:csv')
+FORMAT 'CUSTOM' (FORMATTER='pxfdelimited_import', DELIMITER='¤') ENCODING 'LATIN1';
+CREATE EXTERNAL TABLE pxf_csv_latin1(location text, month text, num_orders int, total_sales float8)
+  LOCATION ('pxf://data/pxf_examples/pxf_hdfs_simple.txt?PROFILE=hdfs:csv')
+FORMAT 'CSV' ENCODING 'LATIN1';
```
ashuka24, line 380:
```suggestion
```

i think if we want to keep this, we should not have the line on 372

docs/content/hdfs_text.html.md.erb
lisakowen: nope, i didn't want to create statements.  i kept the second one.

docs/content/hdfs_text.html.md.erb
lisakowen: i didn't realize this.  is the syntax/behaviour the same as for the *:fixedwidth profile?  i.e. for that  profile, the docs state that the user sets the URI NEWLINE to CR/LF/CRLF and the formatter option (which is named line_delim for fixedwith and NEWLINE for this custom formatter) to something like `E'\n'`.

ashuka24: The behavior is the same. Instead of matching to what the fixedwidth profile has (`line_delim`) we intentionally decided to match with the existing CSV option `NEWLINE` which uses LF, CR, CRLF.

docs/content/hdfs_text.html.md.erb
```diff
@@ -365,3 +365,149 @@ Perform the following procedure to create Greenplum Database writable external t
...
+
+When PXF reads data that contains a multi-byte or multi-character delimiter, its behavior depends on the quote and escape character settings:
+
+| QUOTE Set? | ESCAPE Set? | PXF Behaviour |
+|-------------|--------------|----------|
+| No<sup>1</sup> | No | PXF reads the data as-is. |
+| Yes<sup>2</sup> | Yes (the default is the quote value). | PXF reads data between the quotes as-is except for escaped characters. |
+| Yes<sup>2</sup> | No (`ESCAPE 'OFF'`) | PXF reads the data between quote characters as-is. |
+| No<sup>1</sup> | Yes | PXF reads the data as-is and un-escapes only the delimiter, newline, and escape itself. |
+| Yes<sup>2</sup> | Yes | PXF reads the data between quote characters as-is and un-escapes only the quote and escape characters. |
```
yimingli-vmware, line 445:
```suggestion
| Yes<sup>2</sup> | Yes | PXF reads the data between quote characters as-is and un-escapes only the quote and escaped characters. |
```

What's the difference between line 445 and line 442?

yimingli-vmware (approved):

docs/content/hdfs_text.html.md.erb
ashuka24: for line 442, i think it is the situation where escape is not defined in the table definition. but we set a default value since quote is defined. i think this could be clarified a bit more to mitigate this confusion @lisakowen 

here in line 445, we are referring to the character set as quote and the character set as escape. so `the quote and escape characters` is actually the correct phrase.

docs/content/hdfs_text.html.md.erb
ashuka24: It is not required, a user can still use regular characters if there is a diff table encoding. However, if they choose to do byte representation, they must do it in the server encoding.

ashuka24 (approved):

docs/content/hdfs_text.html.md.erb
lisakowen: decided that this feature is for csv profiles only.

lisakowen:
thank you for reviewing @ashuka24 and @yimingli-vmware .
i am merging this PR.  will separately address any further comments.

PR #978, opened 2023-06-02
Added erroring out logic for decimal overflow for ORC

yimingli-vmware:
In this PR, an erroring out logic will be provided for ORC when the precision of the decimal value exceeds the maximum supported precision of ORC, which is 38. The erroring out logic for Parquet also got refactored.

- Since in Apache ORC code, it limits the maximum precision of DECIMAL type to be 38, it is not possible for ORC to process a decimal whose precision is greater than 38.
- The erroring out logic for PXF ORC is pretty similar to the PXF Parquet one, so I created a dedicated class `DecimalUtilities` to handle the decimal parsing and erroring out for ORC and Parquet. The only difference of the previous decimal parsing behavior between ORC and Parquet was that ORC doesn't enforce precision and scale. For example, if we want to write a decimal value `123456789012345678901234567890.12345` to a Greenplum `NUMERIC` column, the integer digit count is 30 and precision is 35:
  - when using ORC profile, data will be stored as `DECIMAL(38,10)`, so the maximum integer digit count is 28. Since ORC does not enforce precision and scale, the value will be stored as-is `123456789012345678901234567890.12345`.
  - when using Parquet profile, data will be stored as `DECIMAL(38,18)`, so the maximum integer digit count is 28. Since Parquet enforces precision and scale, a `NULL` data will be returned.
- Now in this PR, PXF enforces precision and scale when writing decimal with ORC profile with `error` or `round` decimal overflow option, and restores backwards compatibility with `ignore` option.


server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCSchemaBuilder.java
```diff
@@ -120,11 +121,21 @@ private static TypeDescription orcTypeFromGreenplumType(ColumnDescriptor columnD
...
             if (precision != null) {
+                // In ORC code, no getter for ORC's MAX_PRECISION but there's a getter for ORC's precision.
+                // By default, ORC precision is set to a hardcoded value 38, which is same as MAX_PRECISION's value
+                // and at this point, typeDescription.getPrecision() is still the default precision 38
+                int maxPrecision = typeDescription.getPrecision();
+                // if precision provided in the table DDL exceeds the max precision, ORC will throw an exception
+                // here we provide a PXF error message containing more details
+                if (precision > maxPrecision) {
+                    throw new UnsupportedTypeException(String.format("Column %s is defined as NUMERIC with precision %d " +
+                            "which exceeds maximum supported precision %d of ORC.", columnName, precision, maxPrecision));
```
denalex, line 137:
```suggestion
                            "which exceeds maximum supported precision %d for an ORC field.", columnName, precision, maxPrecision));
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedMappingFunctions.java
```diff
@@ -549,23 +548,22 @@ private static void initWriteFunctionsMap() {
             }
         });
         writeFunctionsMap.put(TypeDescription.Category.DECIMAL, (columnVector, row, val) -> {
-            // also there is Decimal and Decimal64 column vectors, see TypeUtils.createColumn
-            HiveDecimal convertedValue = HiveDecimal.create((String) val);
-            if (convertedValue == null) {
-                // converted value can be null if the original value exceeds precision and cannot be rounded
+            if (val == null) {
```
denalex, line 551: null check happens somewhere else, you do not need to check val for null here

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedMappingFunctions.java
```diff
@@ -549,23 +548,22 @@ private static void initWriteFunctionsMap() {
...
+                // val can be null if the original value exceeds precision and cannot be rounded
                 // Hive just stores NULL as the value, let's do the same
                 columnVector.isNull[row] = true;
                 columnVector.noNulls = false;
-                LOG.warn("Ignored numeric value {} as it exceeds ORC precision and cannot be rounded.", val);
             } else {
+                // also there is Decimal and Decimal64 column vectors, see TypeUtils.createColumn
+                HiveDecimal convertedValue = HiveDecimal.create((String) val);
```
denalex, line 558: so how are we guaranteed now that this will not produce `null` ? Also this conversion already happened during decimal processing, can we re-use the value ?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedResolver.java
```diff
@@ -244,13 +251,22 @@ public OneRow setFieldsForBatch(List<List<OneField>> records) {
             // fill up column vectors for the columns of the given row with record values using mapping functions
             for (OneField field : record) {
                 ColumnVector columnVector = vectorizedRowBatch.cols[columnIndex];
-                if (field.val == null) {
+                Object val = field.val;
```
denalex, line 254: why do we need this, let's stick with `field.val`

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedResolver.java
```diff
@@ -244,13 +251,22 @@ public OneRow setFieldsForBatch(List<List<OneField>> records) {
...
                 } else {
-                    writeFunctions[columnIndex].accept(columnVector, rowIndex, field.val);
+                    if (columnVector.type.name().equals("DECIMAL")) {
+                        BigDecimal convertedValue = decimalUtilities.parseDecimalString((String) val, orcSchema.getPrecision(), orcSchema.getScale(), orcSchema.getFieldNames().get(columnIndex));
+                        if (convertedValue == null) {
+                            val = null;
+                        } else {
+                            val = convertedValue.toString();
+                        }
+                    }
```
denalex, line 268: this whole block above needs refactoring:
- all this custom logic needs to go back into the writeFunction for `DECIMAL`
- we do not want to flip back/forth between BigDecimal and String

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/DecimalUtilities.java
```diff
@@ -0,0 +1,226 @@
...
+    private boolean isDecimalOverflowOptionRound;
+    private boolean isPrecisionOverflowWarningLogged;
+    private boolean isIntegerDigitCountOverflowWarningLogged;
+    private boolean isScaleOverflowWarningLogged;
+
+    /**
+     *
+     * @param profile can be ORC or Parquet
+     */
+    public DecimalUtilities(String profile) {
```
denalex, line 29: ideally this utility should not know about profiles or data formats

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/DecimalUtilities.java
```diff
@@ -0,0 +1,226 @@
...
+         (https://www.postgresql.org/docs/9.3/datatype-numeric.html#DATATYPE-NUMERIC-TABLE)
+         so we parse the value into Parquet with BigDecimal instead of HiveDecimal when precision is greater than Hive's max precision 38
+         */
+
+        if (StringUtils.equalsIgnoreCase(profile, "parquet")) {
+            // TODO: provide a logic of a parsing incoming decimal string with BigDecimal when its precision is greater than 38 for Parquet.
+            //  If the precision is smaller than or equal to 38, still parse it with HiveDecimal
+            return parseDecimalStringWithHiveDecimal(value, precision, scale, true, columnName);
+        }
+        return parseDecimalStringWithHiveDecimal(value, precision, scale, false, columnName);
```
denalex, line 89: the current logic does the same thing in if and else branches other that the value of the enforcement parameter. Maybe this param needs to be the constructor param instead of profile.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/DecimalUtilities.java
```diff
@@ -0,0 +1,226 @@
...
+         HiveDecimal.create will return the decimal value as-is.
+         For example, 123456.123456 can fit in DECIMAL(38,18) without any data loss.
+         */
+
+        // HiveDecimal.create will return a decimal value which can fit in DECIMAL(38)
+        // also there is Decimal and Decimal64 column vectors for ORC, see TypeUtils.createColumn
+        HiveDecimal hiveDecimal = HiveDecimal.create(value);
+        if (hiveDecimal == null) {
+            if (isDecimalOverflowOptionError || isDecimalOverflowOptionRound) {
+                throw new UnsupportedTypeException(String.format("The value %s for the NUMERIC column %s using %s profile exceeds maximum precision %d.",
```
denalex, line 134: we can take profile info out from this message

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/DecimalUtilities.java
```diff
@@ -0,0 +1,226 @@
...
+
+            if (!isScaleOverflowWarningLogged) {
+                LOG.warn("There are rows where for the NUMERIC column {} using {} profile the values exceed {} " +
+                                "and have been rounded off. Enable TRACE log level for row-level details.",
+                        profile, columnName, limitationForGettingAccurateValue);
+                isScaleOverflowWarningLogged = true;
+            }
+        }
+        return hiveDecimal.bigDecimalValue();
+    }
```
denalex, line 225: this ends up being a long method with logic lost among huge comments.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/DecimalUtilities.java
```diff
@@ -0,0 +1,226 @@
...
+
+        (3) (integer digit count <= precision - scale) && (total digits of the value <= precision defined in schema)
+        HiveDecimal.enforcePrecisionScale will return exactly the decimal value as-is.
+        For example, 123456.123456 can fit in DECIMAL
+        (38,18) without any data loss.
+
+        Current logic for ORC doesn't call HiveDecimal.enforcePrecisionScale but Parquet does. That will bring in the
+        inconsistency in the error messages in the third check.
+         */
+        String limitationForGettingAccurateValue = String.format("maximum precision %s", precision);
```
denalex, line 174: this method is called for each tuple and column, needs to be very efficient, we cannot allocate and process strings for potential errors that might not happen.

server/pxf-service/src/templates/templates/pxf-site.xml
```diff
@@ -89,7 +100,7 @@
         <value>round</value>
         <description>
             Specifies behavior of the PXF Parquet profiles when writing NUMERIC data that exceeds the maximum precision of 38. Valid values are error, round, and ignore. The default value is round.
-            If set to error, an error will be reported and the transaction will fail, potentially leaving an incomplete dataset in the external system.
+            If set to error, PXF will try to write the value without precision loss, or report an error, potentially leaving an incomplete dataset in the external system.
```
denalex, line 103: wouldn't our parquet handler error out rather than trying to write data ?

server/pxf-service/src/templates/templates/pxf-site.xml
```diff
@@ -75,6 +75,17 @@
         </description>
     </property>
 
+    <property>
+        <name>pxf.orc.write.decimal.overflow</name>
+        <value>round</value>
+        <description>
+            Specifies behavior of the PXF ORC profiles when writing NUMERIC data that exceeds the maximum precision of 38. Valid values are error, round and ignore. The default value is round.
+            If set to error, PXF will try to write the value without precision loss, or report an error, potentially leaving an incomplete dataset in the external system.
```
denalex, line 83: wouldn't we error out ?

server/pxf-service/src/templates/templates/pxf-site.xml
```diff
@@ -75,6 +75,17 @@
...
+            If set to round, PXF will try to round and write the value, or report an error if the value cannot be rounded, potentially leaving an incomplete dataset in the external system.
+            If set to ignore, PXF will write NULL or rounded value instead of the value (as in previous PXF versions).
```
denalex, line 85: I don't understand what will actually happen.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/DecimalUtilities.java
```diff
@@ -0,0 +1,226 @@
...
+                    isIntegerDigitCountOverflowWarningLogged = true;
+                }
+                return null;
+            }
+
+            limitationForGettingAccurateValue = String.format("maximum scale %s", scale);
+        }
+
+        // At this point, the integer digit count must less than or equal to (precision - scale) for Parquet,
+        // or less than or equal to the precision for ORC, but the total digits may still greater than precision.
```
ashuka24, line 204:
```suggestion
        // or less than or equal to the precision for ORC, but the total digits may still greater be than the precision value.

File: server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/DecimalUtilities.java [hunk 15]
```diff
@@ -0,0 +1,226 @@
...
+        // Here is to check whether the value has been rounded off. If the decimal overflow option is set to 'error',
```
ashuka24, line 205:
```suggestion
        // Here, check whether the value has been rounded off. If the decimal overflow option is set to 'error',

File: server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCSchemaBuilder.java (hunk 1 above)
Comment (ashuka24) [inline server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCSchemaBuilder.java:137]: this limitation is a limitation of ORC itself and not related to PXF profiles. Maybe we could make this a little more clear here: 
```suggestion
                            "which exceeds the maximum supported precision value (%d) for ORC files.", columnName, precision, maxPrecision));
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/DecimalUtilities.java
bradfordb-vmware: Looking at how the string `limitationForGettingAccurateValue` is used, I think you can replace it with a boolean `isPrecisionLimitationForAccurateValue` with a default value of `true`. On line 200, you can then set it to `false` since we've gotten through the precision check OK. Then after the `if`-statement of line 207, you can create the string since we now know that we are in an "error" case where we will want to print it.

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedResolverWriteTest.java
```diff
@@ -148,17 +210,68 @@ public void testExceedingDefaultPrecisionNoRounding() {
...
+
+        String decimalString = "123456789012345678901234567890123456789";
+        records = new ArrayList<>(1);
+        List<OneField> record = getRecord(0, -1);
+        // reset the decimal value to a higher unsupported (>38) precision
+        record.set(14, new OneField(DataType.NUMERIC.getOID(), decimalString));
+        records.add(record);
+
+        Exception e = assertThrows(UnsupportedTypeException.class, () -> resolver.setFieldsForBatch(records));
+        assertEquals(String.format("The value %s for the NUMERIC column %s exceeds maximum precision 38.",
```
hpandeycodeit, line 272: since it's hardcoded: 

```suggestion
        assertEquals(String.format("The value %s for the NUMERIC column col14 exceeds maximum precision 38.",
```

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedResolverWriteTest.java
```diff
@@ -148,17 +210,68 @@ public void testExceedingDefaultPrecisionNoRounding() {
...
+        when(mockWriterOptions.getSchema()).thenReturn(getSchemaForAllColumns());
+        when(mockWriterOptions.getUseUTCTimestamp()).thenReturn(true);
+        context.setMetadata(mockWriterOptions);
+        configuration.set("pxf.orc.write.decimal.overflow", "round");
+        context.setConfiguration(configuration);
+
+        resolver.setRequestContext(context);
+        resolver.afterPropertiesSet();
+
+        String decimalString = "123456789012345678901234567890123456789";
```
hpandeycodeit, line 264: Since this value is used in many places here, it will be better to have it assigned to a variable.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetFileAccessor.java
```diff
@@ -555,9 +555,14 @@ private Type getTypeForColumnDescriptor(ColumnDescriptor columnDescriptor) {
...
                 // this error will be thrown no matter what decimal overflow option is
+                // TODO: replace HiveDecimal.MAX_PRECISION with some other precision
+                //  For Parquet, there is no precision limitation for decimal
+                //  https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#decimal
+                //  For postgres, although there is a precision limitation, the precision is very large
+                //  https://www.postgresql.org/docs/15/datatype-numeric.html
                 if (precision > HiveDecimal.MAX_PRECISION) {
                     throw new UnsupportedTypeException(String.format("Column %s is defined as NUMERIC with precision %d " +
-                            "which exceeds maximum supported precision %d.", columnName, precision, HiveDecimal.MAX_PRECISION));
+                            "which exceeds maximum supported precision %d of Parquet.", columnName, precision, HiveDecimal.MAX_PRECISION));
```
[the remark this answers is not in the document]
yimingli-vmware: no profile info in the message any more

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCSchemaBuilder.java
yimingli-vmware: no profile info in error messages any more

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/DecimalUtilities.java
```diff
@@ -0,0 +1,233 @@
+package org.greenplum.pxf.plugins.hdfs.utilities;
```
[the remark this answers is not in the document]
yimingli-vmware: `DecimalUtilitiesTest.java` and `DecimalOverflowOptionTest.java` are added

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/ParquetFileAccessor.java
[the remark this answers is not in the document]
yimingli-vmware: profile info has been removed

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedResolver.java
```diff
@@ -474,4 +500,25 @@ private DataType getArrayDataType(TypeDescription typeDescription) {
...
+    public DecimalOverflowOption parseDecimalOverflowOption(Configuration configuration) {
+        String decimalOverflowOption = configuration.get(PXF_ORC_WRITE_DECIMAL_OVERFLOW_PROPERTY_NAME, DecimalOverflowOption.DECIMAL_OVERFLOW_ERROR_WITH_ENFORCING.getDecimalOverflowOption()).toLowerCase();
+        switch (decimalOverflowOption) {
+            case PXF_ORC_WRITE_DECIMAL_OVERFLOW_OPTION_ERROR:
+                return DecimalOverflowOption.DECIMAL_OVERFLOW_ERROR_WITHOUT_ENFORCING;
+            case PXF_ORC_WRITE_DECIMAL_OVERFLOW_OPTION_ROUND:
+                return DecimalOverflowOption.DECIMAL_OVERFLOW_ROUND_WITHOUT_ENFORCING;
+            case PXF_ORC_WRITE_DECIMAL_OVERFLOW_OPTION_IGNORE:
+               return DecimalOverflowOption.DECIMAL_OVERFLOW_IGNORE_WITHOUT_ENFORCING;
+            default:
```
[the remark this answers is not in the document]
yimingli-vmware: `configuration.get(propertyName, defaultValue)` makes sure that if no such `propertyName` is in the configuration list, then `dafaultValue` will be used.

server/pxf-service/src/templates/templates/pxf-site.xml
yimingli-vmware: If the value can be written without any precision overflow/loss, the value can still be stored when using 'error' option.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/DecimalOverflowOption.java
```diff
@@ -0,0 +1,100 @@
...
+    DecimalOverflowOption(String decimalOverflowValue, boolean wasEnforcedPrecisionAndScale) {
+        this.decimalOverflowOption = decimalOverflowValue;
+        this.wasEnforcedPrecisionAndScale = wasEnforcedPrecisionAndScale;
+    }
+
+    /**
+     * Parse the configuration value of the configurationPropertyName defined in 'pxf-site.xml' into a DecimalOverflowOption enum object
+     *
+     * @param configuration               contains all the configuration properties of the current server
+     * @param configurationPropertyName   is the decimal overflow configuration property going to be parsed
```
ashuka24, line 43:
```suggestion
     * @param configurationPropertyName   name of decimal overflow configuration property to be parsed

File: server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/DecimalUtilities.java [hunk 22]
```diff
@@ -1,117 +1,60 @@
...
 import org.greenplum.pxf.api.error.UnsupportedTypeException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.math.BigDecimal;
 
+/**
+ * DecimalUtilities is used for parsing decimal values for PXF Parquet profile and PXF ORC profile.
+ * Parsing behaviors are depending on the decimal overflow option values.
+ * Warning logs for different types of overflows will be logged only once if overflows happen.
```
ashuka24, line 13: nit
```suggestion
 * Warning logs for different types of overflow will be logged only once if overflow happen.

File: server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/DecimalUtilities.java [hunk 23]
```diff
@@ -1,117 +1,60 @@
...
+    private final DecimalOverflowOption decimalOverflowOption;
     private boolean isPrecisionOverflowWarningLogged;
     private boolean isIntegerDigitCountOverflowWarningLogged;
     private boolean isScaleOverflowWarningLogged;
 
     /**
+     * Construct a DecimalUtilities object with a DecimalOverflowOption object
      *
-     * @param profile can be ORC or Parquet
+     * @param decimalOverflowOption is parsed by DecimalOverflowOption.parseDecimalOverflowOption
```
ashuka24, line 26: this java doc seems a bit redundant and doesn't provide any additional information. can we make this a little more clear

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/DecimalUtilities.java
```diff
@@ -127,35 +70,39 @@ By default, NUMERIC will be stored as DECIMAL(38,18) in Parquet's schema, or DEC
...
-        At this point, the integer digit count must less than or equal to the precision, but the total digits may still greater than the precision.
+        At this point, the integer digit count must less than or equal to the precision, but the total digits may still
+        be greater than the precision value.
         HiveDecimal.enforcePrecisionScale has different behaviors given the integer digit count and the max integer digit count (precision - scale).
 
         (1) integer digit count > precision - scale
-        HiveDecimal.enforcePrecisionScale returns NULL. To make the behavior consistent with Hive's behavior
-        when storing on a Parquet/ORC-backed table, we store the value as null when the decimal overflow option is set to 'ignore'.
+        HiveDecimal.enforcePrecisionScale returns NULL. In previous version of PXF, writing data with ORC profile
+        did not enforce precision and scale whereas Parquet did. For backward compatibility,
```
ashuka24, line 99:
```suggestion
        did not enforce precision and scale whereas the PXF Parquet profile did. For backward compatibility,

File: server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/DecimalUtilities.java [hunk 25]
```diff
@@ -171,56 +118,67 @@ For example, the column will be stored as DECIMAL(38,18). The integer digit coun
...
+             */
+            return decimalOverflowOption.wasEnforcedPrecisionAndScale() ? null : hiveDecimal;
         }
 
-        // At this point, the integer digit count must less than or equal to (precision - scale) for Parquet,
-        // or less than or equal to the precision for ORC, but the total digits may still greater than precision.
-        // Here is to check whether the value has been rounded off. If the decimal overflow option is set to 'error',
-        // an exception will be thrown.
+        /*
+        At this point, the integer digit count must less than or equal to (precision - scale),
```
ashuka24, line 152:
```suggestion
        At this point, the integer digit count must be less than or equal to (precision - scale),

File: server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/utilities/DecimalUtilitiesTest.java [hunk 26]
```diff
@@ -0,0 +1,151 @@
...
+        assertEquals(expectedHiveDecimal, hiveDecimal);
+    }
+
+    @Test
+    public void testParseDecimalOverflowsScaleOptionIgnoreWithoutEnforcing_decimalPartFailedToBorrowDigits() {
+        decimalUtilities = new DecimalUtilities(DecimalOverflowOption.DECIMAL_OVERFLOW_IGNORE_WITHOUT_ENFORCING);
+        String decimalString = "12345678901234567890.12345678901234567890";
+
+        HiveDecimal hiveDecimal = decimalUtilities.parseDecimalStringWithHiveDecimal(decimalString, precision, scale, columnName);
+        HiveDecimal expectedHiveDecimal = HiveDecimalWritable.enforcePrecisionScale(new HiveDecimalWritable(decimalString), precision, scale).getHiveDecimal();
```
ashuka24, line 148: i'm a little confused by this test. it looks like we are not enforcing but our expected value is being enforced?

yimingli-vmware: I think what I'm trying to show that if `integer digit count == precision - scale`, when decimal part overflows, it cannot borrow digit slots from integer part. Then the parsing behavior should be the same as parsing with precision and precision enforced.  I'll add a comment to it.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedMappingFunctions.java
```diff
@@ -512,51 +515,54 @@ private static void initWriteFunctionsMap() {
...
                 ((BytesColumnVector) columnVector).setRef(row, ((ByteBuffer) val).array(), 0, ((ByteBuffer) val).limit());
             }
         });
-        writeFunctionsMap.put(TypeDescription.Category.DECIMAL, (columnVector, row, val) -> {
-            if (val == null) {
-                // val can be null if the original value exceeds precision and cannot be rounded
+        writeFunctionsMap.put(TypeDescription.Category.DECIMAL, (columnVector, row, val, configuration, columnName) -> {
+            DecimalOverflowOption decimalOverflowOption = DecimalOverflowOption.parseDecimalOverflowOption(configuration, PXF_ORC_WRITE_DECIMAL_OVERFLOW_PROPERTY_NAME, false);
+            DecimalUtilities decimalUtilities = new DecimalUtilities(decimalOverflowOption);
+            // also there are Decimal and Decimal64 column vectors, see TypeUtils.createColumn
```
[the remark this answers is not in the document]
yimingli-vmware: This is an old comment. from `TypeUtils.createColumn`, if the precision of incoming value is not greater than MAX_DECIMAL64_PRECISION 18, apache orc will create a Decimal64ColumnVector. But we definitely have values whose precision is greater than 18, so we should use the more generic DecimalColumnVector. I've updated the comments.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedMappingFunctions.java
```diff
@@ -553,7 +553,9 @@ private static void initWriteFunctionsMap() {
         writeFunctionsMap.put(TypeDescription.Category.DECIMAL, (columnVector, row, val, configuration, columnName) -> {
             DecimalOverflowOption decimalOverflowOption = DecimalOverflowOption.parseDecimalOverflowOption(configuration, PXF_ORC_WRITE_DECIMAL_OVERFLOW_PROPERTY_NAME, false);
             DecimalUtilities decimalUtilities = new DecimalUtilities(decimalOverflowOption);
-            // also there are Decimal and Decimal64 column vectors, see TypeUtils.createColumn
+            // Also there are Decimal and Decimal64 column vectors, see TypeUtils.createColumn.
+            // If the precision of incoming value is not greater than MAX_DECIMAL64_PRECISION 18, in apache orc's TypeUtils.createColumn, it will create a Decimal64ColumnVector
```
ashuka24, line 557: nit
```suggestion
            // Both Decimal and Decimal64 column vectors are possible here (see TypeUtils.createColumn).
            // If the precision of incoming value is not greater than MAX_DECIMAL64_PRECISION 18, Apache ORC's TypeUtils.createColumn will create a Decimal64ColumnVector.

**Review by ashuka24 — APPROVED** [review]

File: server/pxf-service/src/templates/templates/pxf-site.xml [hunk 29]
```diff
@@ -89,9 +100,9 @@
         <value>round</value>
         <description>
             Specifies behavior of the PXF Parquet profiles when writing NUMERIC data that exceeds the maximum precision of 38. Valid values are error, round, and ignore. The default value is round.
-            If set to error, an error will be reported and the transaction will fail, potentially leaving an incomplete dataset in the external system.
+            If set to error, PXF will try to write the value without precision loss, or report an error if there is a decimal overflow, potentially leaving an incomplete dataset in the external system.
             If set to round, PXF will try to round and write the value, or report an error if the value cannot be rounded, potentially leaving an incomplete dataset in the external system.
-            If set to ignore, PXF will write NULL instead of the value (as in previous PXF versions).
+            If set to ignore, PXF will write NULL instead of the value if there is a decimal overflow and the value cannot be rounded (as in previous PXF versions).
```
denalex, line 105: for ease of comparison with the previous string (assuming my correction is correct)
```suggestion
            If set to ignore, PXF will try to round and write the value, or write NULL if the value cannot be rounded, potentially leaving an incomplete dataset in the external system.(as in previous PXF versions).
```

server/pxf-service/src/templates/templates/pxf-site.xml
```diff
@@ -89,9 +100,9 @@
         <value>round</value>
         <description>
             Specifies behavior of the PXF Parquet profiles when writing NUMERIC data that exceeds the maximum precision of 38. Valid values are error, round, and ignore. The default value is round.
-            If set to error, an error will be reported and the transaction will fail, potentially leaving an incomplete dataset in the external system.
+            If set to error, PXF will try to write the value without precision loss, or report an error if there is a decimal overflow, potentially leaving an incomplete dataset in the external system.
```
denalex, line 103: how can it `write the value without precision loss` if we stated the max precision has been exceeded ?

server/pxf-service/src/templates/templates/pxf-site.xml
```diff
@@ -75,6 +75,17 @@
...
     </property>
 
+    <property>
+        <name>pxf.orc.write.decimal.overflow</name>
+        <value>round</value>
+        <description>
+            Specifies behavior of the PXF ORC profiles when writing NUMERIC data that exceeds the maximum precision of 38. Valid values are error, round and ignore. The default value is round.
+            If set to error, PXF will try to write the value without precision loss, or report an error if there is a decimal overflow, potentially leaving an incomplete dataset in the external system.
+            If set to round, PXF will try to round and write the value, or report an error if the value cannot be rounded, potentially leaving an incomplete dataset in the external system.
+            If set to ignore, PXF will round and write the value if the integer digits does not exceed the maximum precision of 38; otherwise PXF will write NULL instead of the value (as in previous PXF versions).
```
denalex, line 85: we need to review these descriptions, they are not easily understood (if at all) after a few reading attempts :)

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedResolver.java
```diff
@@ -250,7 +251,7 @@ public OneRow setFieldsForBatch(List<List<OneField>> records) {
                     }
                     columnVector.isNull[rowIndex] = true;
                 } else {
-                    writeFunctions[columnIndex].accept(columnVector, rowIndex, field.val);
+                    writeFunctions[columnIndex].accept(columnVector, rowIndex, field.val, configuration, orcSchema.getFieldNames().get(columnIndex));
```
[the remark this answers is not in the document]
yimingli-vmware: before this step, orcSchema is generated and the name of the schema field is generated based on Greenplum column name, so they are always the same.

server/pxf-api/src/main/java/org/greenplum/pxf/api/function/PentaConsumer.java
```diff
@@ -12,29 +12,29 @@
  * <p>This is a <a href="package-summary.html">functional interface</a>
  * whose functional method is {@link #accept(Object, Object, Object, Object, Object)}.
  *
- * @param <T> the type of the first argument to the operation
- * @param <U> the type of the second argument to the operation
- * @param <V> the type of the third argument to the operation
- * @param <S> the type of the fourth argument to the operation
- * @param <P> the type of the fifth argument to the operation
+ * @param <P> the type of the first argument to the operation
```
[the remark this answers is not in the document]
yimingli-vmware: Alex he suggested the arguments are better to be in alphabetical order

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/DecimalUtilities.java
```diff
@@ -8,73 +8,154 @@
...
+     */
+    private HiveDecimal parseDecimalStringWithHiveDecimalOnError(String value, int precision, int scale, String columnName) {
+        BigDecimal bigDecimal = new BigDecimal(value);
+        // precision of the bigDecimal cannot be above designated precision
+        if (bigDecimal.precision() > precision) {
+            throw new UnsupportedTypeException(String.format("The value %s for the NUMERIC column %s exceeds the maximum supported precision %d.",
+                    value, columnName, precision));
+        }
+        // integer digit count of the bigDecimal cannot be above designated precision and scale
+        if (bigDecimal.precision() - bigDecimal.scale() > precision - scale) { // integer count of value must also be within allowed integer count
```
denalex, line 87: the comment at the end of the line seems not necessary as there is a similar comment above the line

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/DecimalUtilities.java
```diff
@@ -8,73 +8,154 @@
...
+                    value, columnName));
+        }
+        return hiveDecimal;
+    }
 
-        // HiveDecimal.create will return a decimal value which can fit in DECIMAL(38)
+    /**
+     * Parse the incoming decimal string into a decimal number for PXF Parquet or ORC profile when the decimal overflow option is 'round'.
+     * PXF should error out on the overflow case 1 and 2, or return a rounded HiveDecimal which can fit within dictated precision and scale.
+     * Previous decimal parsing logic enforced precision and scale only for the PXF Parquet profile and not for the PXF ORC profile.
```
denalex, line 109:
```suggestion
     * Previous (before release 6.7.0) decimal parsing logic enforced precision and scale only for the PXF Parquet profile and not for the PXF ORC profile.
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/DecimalUtilities.java
```diff
@@ -8,73 +8,154 @@
...
+        if (hiveDecimalEnforcedPrecisionAndScale == null) {
+            throw new UnsupportedTypeException(String.format("The value %s for the NUMERIC column %s exceeds the maximum supported precision and scale (%d,%d).",
+                    value, columnName, precision, scale));
+        }
+        return hiveDecimalEnforcedPrecisionAndScale;
+    }
 
+    /**
+     * Parse the incoming decimal string into a decimal number for PXF Parquet or ORC profile when the decimal overflow option is 'ignore'.
+     * PXF should error out on the overflow case 1, or return a NULL/rounded HiveDecimal for backwards compatibilities.
```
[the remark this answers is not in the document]
denalex: this utility is not supposed to know anything about the profiles, so the logic is based on incoming parameters only

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/DecimalUtilities.java
```diff
@@ -86,97 +167,30 @@ By default, NUMERIC will be stored as DECIMAL(38,18) in PXF Parquet's schema, or
...
-            }
-
-            LOG.trace("The value {} for the NUMERIC column {} exceeds the maximum supported scale {} and has been rounded off.",
-                    value, columnName, scale);
-
-            if (!isScaleOverflowWarningLogged) {
-                LOG.warn("There are rows where for the NUMERIC column {} the values exceed the maximum supported scale {} " +
-                        "and have been rounded off. Enable TRACE log level for row-level details.", columnName, scale);
-                isScaleOverflowWarningLogged = true;
+        // Do we need to also enforce scale?
```
denalex, line 170: is this still a question ?

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/utilities/DecimalOverflowOptionTest.java
```diff
@@ -1,134 +1,27 @@
...
-        assertDecimalOverflowOption(DecimalOverflowOption.DECIMAL_OVERFLOW_IGNORE_WITHOUT_ENFORCING, orcDecimalOverflowOption);
-
-        configuration.set(PXF_ORC_WRITE_DECIMAL_OVERFLOW_PROPERTY_NAME, "IGNORE");
-        orcDecimalOverflowOption = DecimalOverflowOption.parseDecimalOverflowOption(configuration, PXF_ORC_WRITE_DECIMAL_OVERFLOW_PROPERTY_NAME, false);
-        assertDecimalOverflowOption(DecimalOverflowOption.DECIMAL_OVERFLOW_IGNORE_WITHOUT_ENFORCING, orcDecimalOverflowOption);
-
-        configuration.set(PXF_PARQUET_WRITE_DECIMAL_OVERFLOW_PROPERTY_NAME, "ignore");
-        DecimalOverflowOption parquetDecimalOverflowOption = DecimalOverflowOption.parseDecimalOverflowOption(configuration, PXF_PARQUET_WRITE_DECIMAL_OVERFLOW_PROPERTY_NAME, true);
-        assertEquals(DecimalOverflowOption.DECIMAL_OVERFLOW_IGNORE, parquetDecimalOverflowOption);
+    public void testDecimalOverflowOptionValidness() {
```
denalex, line 10:
```suggestion
    public void testDecimalOverflowOptionValidity() {
```
I never heard `validness`, even though it is a proper dictionary word with the similar meaning, but I think `validity` is more frequently used :)

server/pxf-api/src/main/java/org/greenplum/pxf/api/function/PentaFunction.java
```diff
@@ -8,43 +8,43 @@
...
+public interface PentaFunction<P, S, T, U, V, W> {
     /**
      * Applies this function to the given arguments.
      *
-     * @param t the first function argument
-     * @param u the second function argument
-     * @param v the third function argument
-     * @param s the fourth function argument
-     * @param p the fifth input argument
+     * @param p the first input argument
```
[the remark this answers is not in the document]
yimingli-vmware: yes

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/DecimalUtilities.java
yimingli-vmware: no, it's a redundant comment

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/orc/ORCVectorizedMappingFunctions.java
```diff
@@ -481,16 +478,16 @@ public static OneField[] getNullResultSet(int oid, int size) {
...
         TypeDescription.Category columnTypeCategory = typeDescription.getCategory();
-        PentaConsumer<ColumnVector, Integer, Object, Configuration, String> writeFunction;
+        PentaConsumer<String, ColumnVector, Integer, Object, DecimalUtilities> writeFunction;
         // if timestamps need to be written in local timezone (and not in UTC), get a special function not in the map
         if (columnTypeCategory.equals(TypeDescription.Category.TIMESTAMP) && !timestampsInUTC) {
             writeFunction = timestampInLocalWriteFunction;
         } else if (columnTypeCategory.equals(TypeDescription.Category.LIST)) {
             // pass along the underlying category of the list
             TypeDescription childTypeDescription = typeDescription.getChildren().get(0);
-            if (childTypeDescription.getCategory().equals(TypeDescription.Category.TIMESTAMP) && !timestampsInUTC) {
```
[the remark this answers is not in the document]
yimingli-vmware: in the `main` branch it is `childTypeDescription.equals(TypeDescription.Category.TIMESTAMP)` and intellij tells me it's better to compare between `Category`. But since we don't have any errors about this line for now, maybe it's better not to change this line in this PR

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/utilities/DecimalUtilities.java
```diff
@@ -8,8 +8,8 @@
 import java.math.BigDecimal;
 
 /**
- * DecimalUtilities is used for parsing decimal values for PXF Parquet and ORC profiles.
- * Parsing behavior is dependent on the decimal overflow option values.
+ * DecimalUtilities is used for parsing decimal values according to the decimal schema and the decimal overflow options.
```
ashuka24, line 11: what do you mean by decimal schema?

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/utilities/DecimalUtilitiesTest.java
```diff
@@ -105,20 +133,32 @@ public void testParseDecimalIntegerDigitCountOverflowsPrecisionMinusScaleOptionI
...
-        decimalUtilities = new DecimalUtilities(DecimalOverflowOption.ROUND, true);
+        HiveDecimal expectedHiveDecimal = HiveDecimalWritable.enforcePrecisionScale(new HiveDecimalWritable(decimalStringOverflowsScale), precision, scale).getHiveDecimal();
 
+        // results should be the same no matter enforcePrecisionAndScaleOnIgnore is true or false
+        decimalUtilities = new DecimalUtilities(DecimalOverflowOption.ROUND, true);
         HiveDecimal hiveDecimal = decimalUtilities.parseDecimalStringWithHiveDecimal(decimalStringOverflowsScale, precision, scale, columnName);
-        HiveDecimal expectedHiveDecimal = HiveDecimalWritable.enforcePrecisionScale(new HiveDecimalWritable(decimalStringOverflowsScale), precision, scale).getHiveDecimal();
+        assertEquals(expectedHiveDecimal, hiveDecimal);
+
+        // results should be the same no matter enforcePrecisionAndScaleOnIgnore is true or false
```
ashuka24, line 159: i don't think this comment needs to be here since you have it on line 154

server/pxf-api/src/main/java/org/greenplum/pxf/api/function/PentaFunction.java
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * @param <P> The first argument type
+ * @param <S> The second argument type
+ * @param <T> The third argument type
+ * @param <U> The fourth argument type
+ * @param <V> The fifth argument type
+ * @param <W> The result type
+ * @see Function
+ */
+public interface PentaFunction<P, S, T, U, V, W> {
```
ashuka24, line 19: this has `W` here, but below it looks like we use `X` below. Should this be `X` as well?

yimingli-vmware: the code in line 45 is to return a composed function which calls 2 functions in sequential order. The first function is `andThen`, which I think uses interface `PentaFunction<P, S, T, U, V, X>`, `X` is the return type of this call. The second function is `after`, which uses `W apply(P p, S s, T t, U u, V v)` and matches the definition of the interface `PentaFunction<P, S, T, U, V, W>` in line 19

denalex (approved):

PR #979, opened 2023-06-09
Add logging to the LineBreakAccessor for the write

hpandeycodeit:
Add Log.trace logging to the LineBreakAccessor for the write. 
This will help in investigation in case we need to investigate any issues.

### DIFF
--- a/server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/LineBreakAccessor.java
+++ b/server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/LineBreakAccessor.java
@@ -52,6 +52,7 @@ public class LineBreakAccessor extends HdfsSplittableDataAccessor {
     private FSDataOutputStream fsdos;
     private FileSystem fs;
     private Path file;
+    private int iterationCount;
 
     /**
      * Constructs a LineBreakAccessor.
@@ -132,19 +133,19 @@ public class LineBreakAccessor extends HdfsSplittableDataAccessor {
             InputStream inputStream = (InputStream) onerow.getData();
             final byte[] buffer = new byte[bufferSize];
 
-            // Log copy statistics every 32 MB
-            final int logIntervalBytes = 32 * 1024 * 1024;
-
             // The logic below is copied from IOUtils.copyLarge to add logging
             long count = 0;
             int n;
             while (-1 != (n = inputStream.read(buffer))) {
                 dos.write(buffer, 0, n);
                 count += n;
-                if (LOG.isDebugEnabled() && count % logIntervalBytes == 0) {
-                    // Log this message after every 32 MB
-                    LOG.debug("wrote {} additional bytes, total so far {} (buffer size {})", n, count, bufferSize);
+                if (LOG.isDebugEnabled() && iterationCount == 100) {
+                    // Log this message after every 100th Iteration
+                    LOG.debug("wrote {} bytes, total so far {} (buffer size {})", n, count, bufferSize);
+                    iterationCount = 0;
                 }
+
+                iterationCount++;
             }
 
             LOG.debug("Wrote {} bytes to outputStream using a buffer of size {}", count, bufferSize);
--- a/server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/LineBreakAccessor.java
+++ b/server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/LineBreakAccessor.java
@@ -135,17 +135,16 @@ public class LineBreakAccessor extends HdfsSplittableDataAccessor {
             // The logic below is copied from IOUtils.copyLarge to add logging
             long count = 0;
             int n;
-            int iterationCount = 0;
+            long iterationCount = 0;
             long iterationByteCount = 0;
 
             while (-1 != (n = inputStream.read(buffer))) {
                 dos.write(buffer, 0, n);
                 count += n;
                 iterationByteCount += n;
-                if (LOG.isDebugEnabled() && iterationCount == 100) {
+                if (LOG.isDebugEnabled() && (iterationCount % 100 == 0)) {
                     // Log this message after every 100th Iteration
                     LOG.debug("wrote {} bytes, total so far {} (buffer size {})", iterationByteCount, count, bufferSize);
-                    iterationCount = 0;
                     iterationByteCount = 0;
                 }
 
--- a/server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/LineBreakAccessor.java
+++ b/server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/LineBreakAccessor.java
@@ -53,6 +53,7 @@ public class LineBreakAccessor extends HdfsSplittableDataAccessor {
     private FileSystem fs;
     private Path file;
     private int iterationCount;
+    private long totalBytesCount;
 
     /**
      * Constructs a LineBreakAccessor.
@@ -141,10 +142,11 @@ public class LineBreakAccessor extends HdfsSplittableDataAccessor {
                 count += n;
                 if (LOG.isDebugEnabled() && iterationCount == 100) {
                     // Log this message after every 100th Iteration
-                    LOG.debug("wrote {} bytes, total so far {} (buffer size {})", n, count, bufferSize);
+                    LOG.debug("wrote {} bytes, total so far {} (buffer size {})", totalBytesCount, count, bufferSize);
                     iterationCount = 0;
+                    totalBytesCount = 0;
                 }
-
+                totalBytesCount +=n;
                 iterationCount++;
             }
 
--- a/server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/LineBreakAccessor.java
+++ b/server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/LineBreakAccessor.java
@@ -133,18 +133,18 @@ public class LineBreakAccessor extends HdfsSplittableDataAccessor {
             final byte[] buffer = new byte[bufferSize];
 
             // The logic below is copied from IOUtils.copyLarge to add logging
-            long count = 0;
+            long totalByteCount = 0;
             int n;
             long iterationCount = 0;
             long iterationByteCount = 0;
 
             while (-1 != (n = inputStream.read(buffer))) {
                 dos.write(buffer, 0, n);
-                count += n;
+                totalByteCount += n;
                 iterationByteCount += n;
                 if (LOG.isDebugEnabled() && (iterationCount % 100 == 0)) {
                     // Log this message after every 100th Iteration
-                    LOG.debug("wrote {} bytes, total so far {} (buffer size {})", iterationByteCount, count, bufferSize);
+                    LOG.debug("wrote {} bytes, total so far {} (buffer size {})", iterationByteCount, totalByteCount, bufferSize);
                     iterationByteCount = 0;
                 }
 
--- a/server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/LineBreakAccessor.java
+++ b/server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/LineBreakAccessor.java
@@ -132,9 +132,8 @@ public class LineBreakAccessor extends HdfsSplittableDataAccessor {
             InputStream inputStream = (InputStream) onerow.getData();
             final byte[] buffer = new byte[bufferSize];
 
-            // We want to log it about every 5 secs.
-            // We have seen the upload speed of 8MB/sec, so it is like 1000 buffers per sec or 5000 buffers per 5 sec
-            final int frequency = 5000 * bufferSize;
+            // Log copy statistics every 32 MB
+            final int logIntervalBytes = 32 * 1024 * 1024;
 
             // The logic below is copied from IOUtils.copyLarge to add logging
             long count = 0;
@@ -142,8 +141,8 @@ public class LineBreakAccessor extends HdfsSplittableDataAccessor {
             while (-1 != (n = inputStream.read(buffer))) {
                 dos.write(buffer, 0, n);
                 count += n;
-                if (LOG.isDebugEnabled() && n % frequency == 0) {
-                    // Log this message after every frequency*8K (default) bytes
+                if (LOG.isDebugEnabled() && count % logIntervalBytes == 0) {
+                    // Log this message after every 32 MB
                     LOG.debug("wrote {} additional bytes, total so far {} (buffer size {})", n, count, bufferSize);
                 }
             }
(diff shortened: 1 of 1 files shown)


server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/LineBreakAccessor.java
```diff
@@ -129,7 +128,25 @@ public boolean writeNextObject(OneRow onerow) throws IOException {
...
+            int n, byteCount;
+            while (EOF != (n = inputStream.read(buffer))) {
+                dos.write(buffer, 0, n);
+                byteCount = n;
+                if(byteCount % 2*DEFAULT_BUFFER_SIZE == 0){
+                    // Log this message after every 2*8K Bytes
+                    LOG.trace("Successfully Wrote {} bytes to outputStream using a buffer of size {}", byteCount, bufferSize);
+                }
+                count += n;
+            }
```
denalex, line 148: great, but let's re-arrange a bit, get rid of new variable `byteCount`, adjust logged message, do calculations only if `LOG.trace` is enabled and choose a window that will log about every 5 secs. We have seen the upload speed of 8MB/sec, so like 1000 buffers per sec or 5000 buffers per 5 secs. Otherwise the log will be littered with entries. Or we can introduce a property that can manage the frequency.
```suggestion
            InputStream inputStream = (InputStream) onerow.getData();
            final byte[] buffer = new byte[bufferSize];
            final int frequency = 5000 * bufferSize;

            // the logic below is copied from IOUtils.copyLarge to add logging
            long count = 0;            
            int n;
            while (-1 != (n = inputStream.read(buffer))) {
                dos.write(buffer, 0, n);
                count += n;
                if (LOG.isTraceEnabled() && n % frequency == 0) {
                    // log this message after every frequency*8K (default) bytes
                    LOG.trace("wrote {} additional bytes, total so far {} (buffer size {})", n, count, bufferSize);
                }
            }
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/LineBreakAccessor.java
```diff
@@ -129,7 +128,23 @@ public boolean writeNextObject(OneRow onerow) throws IOException {
             // match same buffer size to read data from the input stream. The
             // buffer size can be configured externally.
             int bufferSize = configuration.getInt("io.file.buffer.size", DEFAULT_BUFFER_SIZE);
-            long count = IOUtils.copyLarge((InputStream) onerow.getData(), dos, new byte[bufferSize]);
+
+            InputStream inputStream = (InputStream) onerow.getData();
+            final byte[] buffer = new byte[bufferSize];
+            final int frequency = 5000 * bufferSize;
```
ashuka24, line 134: can we add a comment about why we use the value 5000 so that we do not have a floating magic number around? 

is this a value that we would want to be able to toggle somehow?

yimingli-vmware (approved):

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/LineBreakAccessor.java
denalex: Also, since we will be printing not for every row, maybe using DEBUG level here will be more appropriate

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/LineBreakAccessor.java
ashuka24: Sure? we use 5 seconds because we have seen an upload speed of 8 MB/sec. What happens if the upload speed is significantly better or significantly worse, would we want to modify the frequency at which we are logging?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/LineBreakAccessor.java
```diff
@@ -129,7 +128,26 @@ public boolean writeNextObject(OneRow onerow) throws IOException {
...
+            // We have seen the upload speed of 8MB/sec, so it is like 1000 buffers per sec or 5000 buffers per 5 sec
+            final int frequency = 5000 * bufferSize;
+
+            // The logic below is copied from IOUtils.copyLarge to add logging
+            long count = 0;
+            int n;
+            while (-1 != (n = inputStream.read(buffer))) {
+                dos.write(buffer, 0, n);
+                count += n;
+                if (LOG.isDebugEnabled() && n % frequency == 0) {
```
bradfordb-vmware, line 145: I think this logic is wrong:
Since `n` is the number of bytes read into the buffer, it should always be less than or equal to `bufferSize` and `frequency` is a multiple of `bufferSize` so `n` will always be less than `frequency`. The only value of `n` for which `n % frequency == 0` is `n = 0`.

I _think_ you want to check `count % frequency == 0`.

bradfordb-vmware (changes requested):

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/LineBreakAccessor.java
```diff
@@ -129,7 +128,25 @@ public boolean writeNextObject(OneRow onerow) throws IOException {
             // match same buffer size to read data from the input stream. The
             // buffer size can be configured externally.
             int bufferSize = configuration.getInt("io.file.buffer.size", DEFAULT_BUFFER_SIZE);
-            long count = IOUtils.copyLarge((InputStream) onerow.getData(), dos, new byte[bufferSize]);
+
+            InputStream inputStream = (InputStream) onerow.getData();
+            final byte[] buffer = new byte[bufferSize];
+
+            // Log copy statistics every 32 MB
+            final int logIntervalBytes = 32 * 1024 * 1024;
```
denalex, line 136: this should be a class-level constant

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/LineBreakAccessor.java
```diff
@@ -129,7 +128,25 @@ public boolean writeNextObject(OneRow onerow) throws IOException {
...
+
+            // The logic below is copied from IOUtils.copyLarge to add logging
+            long count = 0;
+            int n;
+            while (-1 != (n = inputStream.read(buffer))) {
+                dos.write(buffer, 0, n);
+                count += n;
+                if (LOG.isDebugEnabled() && count % logIntervalBytes == 0) {
```
denalex, line 144: `count % logIntervalBytes == 0` still assumes we read the full buffer at a time, need more robust logic here then

also the message below `wrote {} additional bytes` with parameter value `n` is incorrect since the loop will execute many times, so `the number of additional bytes` will be way larger than `n`.

denalex (changes requested):

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/LineBreakAccessor.java
```diff
@@ -129,7 +129,25 @@ public boolean writeNextObject(OneRow onerow) throws IOException {
...
+
+            // The logic below is copied from IOUtils.copyLarge to add logging
+            long count = 0;
+            int n;
+            while (-1 != (n = inputStream.read(buffer))) {
+                dos.write(buffer, 0, n);
+                count += n;
+                if (LOG.isDebugEnabled() && iterationCount == 100) {
+                    // Log this message after every 100th Iteration
+                    LOG.debug("wrote {} bytes, total so far {} (buffer size {})", n, count, bufferSize);
```
denalex, line 144: this is not correct, `wrote {} bytes` with parameter n -- the logic will write more than `n` bytes between the iterations.

denalex (changes requested):

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/LineBreakAccessor.java
hpandeycodeit: Introduced another variable that will calculate the bytes in between iterations and update here in the error message.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/LineBreakAccessor.java
```diff
@@ -53,6 +52,8 @@ public class LineBreakAccessor extends HdfsSplittableDataAccessor {
     private FSDataOutputStream fsdos;
     private FileSystem fs;
     private Path file;
+    private int iterationCount;
+    private long totalBytesCount;
```
denalex, line 56: neither of these have to be class fields, just local vars

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/LineBreakAccessor.java
```diff
@@ -129,7 +130,26 @@ public boolean writeNextObject(OneRow onerow) throws IOException {
...
+
+            // The logic below is copied from IOUtils.copyLarge to add logging
+            long count = 0;
+            int n;
+            while (-1 != (n = inputStream.read(buffer))) {
+                dos.write(buffer, 0, n);
+                count += n;
+                if (LOG.isDebugEnabled() && iterationCount == 100) {
+                    // Log this message after every 100th Iteration
+                    LOG.debug("wrote {} bytes, total so far {} (buffer size {})", totalBytesCount, count, bufferSize);
```
denalex, line 145:
```suggestion
                    LOG.debug("wrote {} bytes, total so far {} (buffer size {})", iterationByteCount, count, bufferSize);
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/LineBreakAccessor.java
```diff
@@ -129,7 +130,26 @@ public boolean writeNextObject(OneRow onerow) throws IOException {
...
+                    iterationCount = 0;
+                    totalBytesCount = 0;
```
denalex, line 147:
```suggestion
                    iterationByteCount = 0;
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/LineBreakAccessor.java
```diff
@@ -129,7 +130,26 @@ public boolean writeNextObject(OneRow onerow) throws IOException {
...
+                }
+                totalBytesCount +=n;
```
denalex, line 149:
```suggestion
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/LineBreakAccessor.java
```diff
@@ -129,7 +130,26 @@ public boolean writeNextObject(OneRow onerow) throws IOException {
...
+
+            InputStream inputStream = (InputStream) onerow.getData();
+            final byte[] buffer = new byte[bufferSize];
+
+            // The logic below is copied from IOUtils.copyLarge to add logging
+            long count = 0;
+            int n;
+            while (-1 != (n = inputStream.read(buffer))) {
+                dos.write(buffer, 0, n);
+                count += n;
```
denalex, line 142:
```suggestion
                count += n;
                iterationByteCount += n;
```

denalex (commented):
let's rename `totalBytesCount` into `iterationByteCount` to avoid confusion and make these local vars

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/LineBreakAccessor.java
```diff
@@ -129,7 +128,30 @@ public boolean writeNextObject(OneRow onerow) throws IOException {
...
+                iterationByteCount += n;
+                if (LOG.isDebugEnabled() && iterationCount == 100) {
+                    // Log this message after every 100th Iteration
+                    LOG.debug("wrote {} bytes, total so far {} (buffer size {})", iterationByteCount, count, bufferSize);
+                    iterationCount = 0;
+                    iterationByteCount = 0;
+                }
+
+                iterationCount++;
```
denalex, line 152: if debug is not enabled, `iterationCount` will never be zeroed and can go above 2B for large dataset, let's make it a `long`. Also, maybe change `&& iterationCount == 100` to `&& (iterationCount % 100 == 0)` and remove zeroing out the iteration count for uniformity.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/LineBreakAccessor.java
```diff
@@ -129,7 +128,29 @@ public boolean writeNextObject(OneRow onerow) throws IOException {
             // match same buffer size to read data from the input stream. The
             // buffer size can be configured externally.
             int bufferSize = configuration.getInt("io.file.buffer.size", DEFAULT_BUFFER_SIZE);
-            long count = IOUtils.copyLarge((InputStream) onerow.getData(), dos, new byte[bufferSize]);
+
+            InputStream inputStream = (InputStream) onerow.getData();
+            final byte[] buffer = new byte[bufferSize];
+
+            // The logic below is copied from IOUtils.copyLarge to add logging
+            long count = 0;
```
ashuka24, line 136: It looks like `count` is just the total accumulated number of bytes for the input stream. Could we update the name of the variable to reflect that?

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/LineBreakAccessor.java
```diff
@@ -129,9 +128,31 @@ public boolean writeNextObject(OneRow onerow) throws IOException {
...
+                if (LOG.isDebugEnabled() && (iterationCount % 100 == 0)) {
+                    // Log this message after every 100th Iteration
+                    LOG.debug("wrote {} bytes, total so far {} (buffer size {})", iterationByteCount, totalByteCount, bufferSize);
+                    iterationByteCount = 0;
+                }
+
+                iterationCount++;
+            }
+
+            LOG.debug("Wrote {} bytes to outputStream using a buffer of size {}", iterationCount, bufferSize);
```
denalex, line 154:
```suggestion
            LOG.debug("Wrote {} bytes to outputStream using a buffer of size {}", totalByteCount, bufferSize);
```

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/LineBreakAccessor.java
```diff
@@ -129,9 +128,31 @@ public boolean writeNextObject(OneRow onerow) throws IOException {
...
+            return iterationCount > 0;
```
denalex, line 155: while this is correct, should we keep it in the spirit of the original code that checked the number of bytes written (vs number of iterations ?)
```suggestion
            return totalByteCount > 0;
```

denalex (approved):

ashuka24 (approved):

PR #984, opened 2023-06-15
Enable writing data in JSON format using *:json profiles

denalex:
This PR introduces the following changes:
 * (NEW): enabled support for writing data from Greenplum into Json files
 * (NEW): enabled support for reading array types from Json files into Greenplum
 * (CHANGE): now considers binary data (from Greenplum `BYTEA` and `BYTEA[]`) to be stored in Json as a Base64 encoded string
 * (CHANGE) now preserves hight precision of numbers mapped into Greenplum NUMERIC types during reading

It is now possible to export data from Greenplum using PXF `*:json` profiles. Before this, PXF `*:json` profiles were only supporting reading data into Greenplum.

PXF supports writing JSON data in 2 layouts. By default the data will be written in JSONL format, where each database tuple is written on a separate line and is represented by a Json object:
```
 {..tuple1..}
 {..tuple2..}
 ```
The file will have the `.jsonl` extension to denote that it is not a valid parsable Json by itself. However, each row will be a valid Json object representing a database tuple. It will be possible to read these files back using PXF Json profile without any special options provided in the corresponding readable table's LOCATION URL.

Alternatively, if the `ROOT` option is provided with a non-empty value in the table's LOCATION URL the data will be written as a valid JSON object with the root level attribute having the name given by the option value. The value of the attribute will be an array of JSON objects, each one representing a database tuple. For example, if the option `ROOT=records` is provided, the data in each file will look like:
```
{"records":[
{..tuple1..}
,{..tuple2..}
 ...
]}
```
The file will have the `.json` extension to denote that it is a valid parsable Json by itself. To read these files back using PXF Json profile a user will have to define the `IDENTIFIER` option in the corresponding readable table's LOCATION URL.

The writing use case supports file compression. A user can specify the `COMPRESSION_CODEC` option in the writable external table LOCATION URL, like in other profiles. When the files are compressed they will have extensions with that compression suffix, e.g. `.jsonl.gz` for GZIP compressed files in JSONL layout and `.json.gz` for GZIP compressed files in the object layout.

In either layout a tuple is a valid JSON object, e.g.:
```
{"id":1,"name":"James Bond","code":"007","years":[2008,2012,2015,2021]}
```

### CODE REVIEW

File: server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonResolver.java [hunk 1]
```diff
@@ -213,7 +226,8 @@ private void addFieldFromJsonNode(ColumnDescriptorCache columnMetadata, JsonNode
                 oneField.val = val.asBoolean();
                 break;
             case BYTEA:
-                oneField.val = val.asText().getBytes();
+                // we assume that this is binary data that is stored as Base64 encoded in Json
+                oneField.val = val.binaryValue();
```
ashuka24, line 230: why did we not make this assumption previously/why is this change needed now? Will it affect current users?

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonAccessor.java
```diff
@@ -117,7 +236,50 @@ public boolean writeNextObject(OneRow onerow) {
...
+        } catch (IOException e) {
+            // remember that the exception was caught and rethrow it to propagate upwards
+            caughtException = true;
+            throw e;
+        } finally {
+            if (jsonGenerator != null) {
+                try {
+                    jsonGenerator.close();
+                } catch (IOException e) {
+                    // generator close failed, but if there was a more important exception caught before, supress this one
```
ashuka24, line 259:
```suggestion
                    // generator close failed, but if there was a more important exception caught before, suppress this one

File: server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonAccessor.java [hunk 3]
```diff
@@ -106,9 +199,35 @@ public boolean openForWrite() {
      * @return true if the write succeeded
      * @throws Exception writing to the resource failed
      */
+    @SuppressWarnings("unchecked")
```
ashuka24, line 202: can you explain why this warning is needed now?

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonAccessor.java
```diff
@@ -89,14 +165,31 @@ protected Object getReader(JobConf conf, InputSplit split) throws IOException {
...
+    public boolean openForWrite() throws IOException {
+        boolean result = super.openForWrite();
+        // this should not really happen, but complying with the interface, if the operation returned false and
+        // there was no exception, there is nothing else to do here, just propagate the result to caller
+        if (!result) {
+            return false;
+        }
+
+        // setup Json machinery, allow for use of UTF8 encoding only
+        jsonGenerator = jsonFactory.createGenerator((OutputStream) dos, JsonEncoding.UTF8);
```
ashuka24, line 183: if i understand this correctly, this means that we cannot write to files in other encodings? aka, if we set the table encoding in GPDB, it will not honor that encoding? do we error out in that situation?

automation/src/test/java/org/greenplum/pxf/automation/features/json/JsonWriteTest.java
```diff
@@ -0,0 +1,171 @@
...
+        gpdbPrimitiveTypesTable = new Table("gpdb_" + PRIMITIVE_TYPES, PRIMITIVE_TYPES_FIELDS);
+        gpdbPrimitiveTypesTable.setDistributionFields(new String[]{"id"});
+        gpdb.createTableAndVerify(gpdbPrimitiveTypesTable);
+        gpdb.copyFromFile(gpdbPrimitiveTypesTable, new File(resourcePath + PRIMITIVE_TYPES + ".csv"), ",", true);
+    }
+
+    @Test(groups = {"gpdb", "security", "hcfs"})
+    public void writePrimitiveTypesObject() throws Exception {
+        runScenario(PRIMITIVE_TYPES + "_object", PRIMITIVE_TYPES_FIELDS, gpdbPrimitiveTypesTable,
+                null, new String[]{"IDENTIFIER=id"}, JSON_EXTENSION_ASSERTER);
```
ashuka24, line 80: can we also add tests for using `LAYOUT=object` and `KEY=<value>`?

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonAccessor.java
```diff
@@ -19,32 +19,76 @@
...
+ * For the 'object' layout, data in each file will look like:
+ * <pre>
+ * {"records":[
+ * {..tuple1..}
+ * ,{..tuple2..}
+ * ...
+ * ]}
+ * </pre>
+ * The file itself will be a valid parsable Json. The value of the top-level property can be customized from the default
+ * value of "records" by setting an external table option "KEY".
```
ashuka24, line 62: Do we throw an error if we provide a `KEY` option but no `LAYOUT` option? Do we give a warning or do we just ignore the option altogether?

automation/src/test/java/org/greenplum/pxf/automation/features/json/JsonWriteTest.java
denalex: we can, but the reading back logic will not care what KEY was as it relies on IDENTIFIER which is a part of the tuple, whereas KEY is the attr name of the wrapper element. I will still add the test for completeness.

automation/tincrepo/main/pxf/features/hdfs/readable/json/array_as_text/expected/query01.ans
```diff
@@ -3,26 +3,26 @@
...
-  1 | []      | {}      | [2,null,-2.3,2.234567890123457]  | [true,null,false] | ["hello",null,"wor\"ld"] | [["a","b"],null,[1,2]] | [{"a":1},null,{"a":2}] | {"level":0,"key":"a","data":{"level":1,"key":"b","data":{"level":2,"key":"y","data":["wow",null]}}}
-  2 | []      | {}      | [3,-3.3,null,3.234567890123457]  | [true,false,null] | ["hello","wor\"ld",null] | [["a","b"],[1,2],null] | [{"a":1},{"a":2},null] | {"level":0,"key":"a","data":{"level":1,"key":"b","data":{"level":2,"key":"z","data":["wow",null]}}}
+----+---------+---------+------------------------------------------+-------------------+--------------------------+------------------------+------------------------+-----------------------------------------------------------------------------------------------------
+  0 | []      | {}      | [null,1,-1.3,1.234567890123456789012345] | [null,true,false] | [null,"hello","wor\"ld"] | [null,["a","b"],[1,2]] | [null,{"a":1},{"a":2}] | {"level":0,"key":"a","data":{"level":1,"key":"b","data":{"level":2,"key":"x","data":["wow",null]}}}
```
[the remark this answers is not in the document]
denalex: I switched to `USE_BIGDECIMAL_FOR_FLOATS` option for Json READING to be able to read numeric types without precision loss.

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonAccessor.java
denalex: We are ignoring the KEY option for `rows` layout. If LAYOUT option is not provided, the layout defaults to `object` and we still check for KEY option in this case.

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonAccessor.java
denalex: good point, according to standards, Json must be encoded in UTF8. We can analyze whether Table encoding is UTF8 or if it is not specified whether the DB encoding is UTF8 and error out if the resulting encoding is not UTF8.

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonAccessor.java
denalex: because we are doing `(List<OneField>) onerow.getData()` cast and Java cannot verify and enforce at compile time the generics info ( that a list will be a `List<OneField>` and not a `List<Foo>` ). So we have to say that unchecked cast is ok, as we're sure of the actual type.

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonResolver.java
denalex: a standard way to store binary data in Json is by using base64 encoding. The previous read logic did not do that properly, so we need to correct it. It will impact existing users but hopefully in a positive way. I doubt any of them were reading binary data created by other systems (we did not support Json write before). If they were to attempt to read binary data, which is most likely was Base64 encoded, we would get garbage in GP. We should Release Note this, though.

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonUtilities.java
```diff
@@ -0,0 +1,198 @@
...
+            case BIGINT:
+                jsonGenerator.writeNumber((long) value);
+                break;
+            case REAL:
+                jsonGenerator.writeNumber((float) value);
+                break;
+            case FLOAT8:
+                jsonGenerator.writeNumber((double) value);
+                break;
+            // TODO: maybe later use ISO8601 timestamp representation (2012-04-23T18:25:43.511Z) but for now use GP format
```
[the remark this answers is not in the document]
denalex: There is no standard Timestamp type in Json, instead application developers can decide how to store timestamp info in Json that suits interoperability of their systems. Sometimes developers use longs for efficient transfers and fast Java serde. 

Bust most often timestamps are stored as strings. There are multiple formats that can be used to represent a timestamp as a string, one of them being ISO8601 that uses a letter `T` between date and time portions as well as the final letter `Z` to denote UTC.

We do not know what downstream consumers of GP Json write will be, and our read logic does not support ISO8601. We use strings, but we assume the format of the string is the Postgres format, which is different from ISO8601. Supporting ISO (perhaps as an extra option) might be a future feature.

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonUtilities.java
```diff
@@ -0,0 +1,198 @@
...
+        if (val == null) {
+            return null;
+        }
+        // split array value into elements and unescape each element
+        String[] splits = pgUtilities.splitArray(val);
+        List<Object> data = new ArrayList<>(splits.length);
+        for (String split : splits) {
+            try {
+                // convert each element to a corresponding Java object
+                data.add(decodeString(split, primitiveType));
```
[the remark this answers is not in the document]
denalex: I thought about it, but there are other utilities that have similar functions. They all are 80% similar, but the differences did not immediately reveal a good generalization pattern. We'll need to look again, for now I just wanted a quick way to get things working that is similar to what we have elsewhere in the the code.

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonAccessor.java
yimingli-vmware: So logically we should check `LAYOUT` first? If `LAYOUT` is `object`, then we check `KEY`; if `LAYOUT` is `rows`, we just skip checking `KEY`. Is my understanding correct?

denalex: yes, correct, that is exactly what lines 142-146 are doing.

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonUtilities.java
```diff
@@ -0,0 +1,198 @@
...
+            return;
+        }
+        // first check if we are dealing with arrays
+        if (dataType.isArrayType()) {
+            jsonGenerator.writeStartArray();
+            DataType elementDataType = dataType.getTypeElem();
+            // split array into elements and write the value for each element
+            List<Object> elementValues = parsePostgresArray((String) value, elementDataType);
+            for (Object elementValue : elementValues) {
+                writeValue(jsonGenerator, elementDataType, elementValue);
```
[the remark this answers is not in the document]
denalex: no, the function can only handle single-dimensional arrays as `DataType dataType` is Greenplum datatype, so it will be, for example an `INTARRAY` and the first recursive invocation will pass `INT` for the element type. So on the second invocation it will only expect to deal with the primitive types, not arrays.

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonAccessor.java
```diff
@@ -19,32 +19,79 @@
...
+ * <p>
+ * For the 'rows' layout, data in each file will look like:
+ * <pre>
+ * {..tuple1..}
+ * {..tuple2..}
+ * </pre>
+ * The file itself will not be a valid parsable Json, but each row will be a valid Json object representing a database tuple.
```
bradfordb-vmware, line 72: On an initial read through the JavaDoc, it seems like the configuration for reading and writing JSON uses different parameter names and have defaults that are opposite to each other. When reading JSON, the `*:json` profile by default assumes one JSON object per line in the file while writing assumes a muli-object JSON array (a multi-line JSON value). If a user creates a writable table using the JSON profile with no configuration parameters in the `LOCATION` URL, inserts some rows, creates a readable table using the JSON profile with no configuration parameters, and tries to read I think it will fail.

I think we could align the profile to use a consistent set of configuration parameters and matching defaults: If no `IDENTIFIER` is given, then both reading and writing assume one JSON object per line. For writing JSON, if the user provides a value for `IDENTIFIER` that that indicates that they want a single multi-element JSON array.

bradfordb-vmware: > I don't think we should have users provide IDENTIFIER options for write, as it is not relevant for that use case.

I'm not sure that I agree with this statement. I think that `LAYOUT=object` and `KEY=<some-value>` could be handled by `IDENTIFIER=<some-value>` and `LAYOUT=rows` could be handled by no value for `IDENTIFIER`.

As for the default, it looks like there is an existing convention of [JSON Lines format text file](https://jsonlines.org/) for storing one JSON object per line in a file. As two data points, both [`jq`](https://jqlang.github.io/jq/) and Apache Spark have no trouble parsing files with one JSON object per line. Additionally, Spark writes JSON as one object per line.

ashuka24 (approved):

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonAccessor.java
denalex: the changes that reflect these ideas have been made

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonAccessor.java
denalex: the check enforcing UTF8 encoding has been implemented

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/JsonUtilitiesTest.java
```diff
@@ -0,0 +1,299 @@
...
+
+public class JsonUtilitiesTest {
+
+    private static final JsonFactory jsonFactory = new JsonFactory();
+    private JsonGenerator generator;
+    private StringWriter writer;
+    private JsonUtilities jsonUtilities;
+
+    @BeforeEach
+    private void before() throws IOException {
```
yimingli-vmware, line 43: All the tests here are writing tests. Do we need `context.setRequestType(RequestContext.RequestType.WRITE_BRIDGE);` here?

yimingli-vmware (approved):

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/HdfsSplittableDataAccessor.java
```diff
@@ -67,8 +68,10 @@ public void afterPropertiesSet() {
         // Check if the underlying configuration is for HDFS
         hcfsType = HcfsType.getHcfsType(context);
 
-        // Parse fileSplit from context
-        fileSplit = HdfsUtilities.parseFileSplit(context.getDataSource(), context.getFragmentMetadata());
+        // Parse fileSplit from context for reading only
+        if (context.getRequestType() == RequestContext.RequestType.READ_BRIDGE) {
```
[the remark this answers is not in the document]
denalex: `fileSplit` is only needed for the reading case as it is a unit of reading. For writing, we read data from the Tomcat input stream and write to the external system as a single file.

server/pxf-hdfs/src/main/java/org/greenplum/pxf/plugins/hdfs/LineBreakAccessor.java
```diff
@@ -61,10 +63,24 @@ public LineBreakAccessor() {
...
+    protected LineBreakAccessor(InputFormat<?, ?> inputFormat) {
+        super(inputFormat);
+    }
+
     @Override
     public void afterPropertiesSet() {
         super.afterPropertiesSet();
-        ((TextInputFormat) inputFormat).configure(jobConf);
+        if (context.getRequestType() == RequestContext.RequestType.WRITE_BRIDGE) {
+            return; // early return for write use case
```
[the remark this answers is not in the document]
denalex: because the logic below that line is only applicable to reading use case.

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/JsonUtilitiesTest.java
denalex: yes, all the tests are for writing, but se are not using `RequestContext` at this level so no need to set any properties. these tests are at the level of JsonGenerator that writes to String and then we compare what is has written.

ashuka24 (approved):

server/pxf-json/src/test/resources/test-write-object-three.json
```diff
@@ -0,0 +1,5 @@
+{"records":[
+{"id":1,"color":"red"}
+,{"id":2,"color":"yellow"}
```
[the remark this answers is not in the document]
denalex: yes, unfortunately the generator automatically places a comma before writing the next array element, so the only way for us to break a line for a new tuple is at the end of the old tuple and before the comma.

PR #986, opened 2023-06-16
Update cloudbuild maven version

yimingli-vmware:
Currently we are using maven 3.6.3 in our docker images, but it's no longer available. In this PR we upgrade the maven version to 3.9.2.

### DIFF
--- a/concourse/docker/pxf-dev-base/cloudbuild.yaml
+++ b/concourse/docker/pxf-dev-base/cloudbuild.yaml
@@ -182,7 +182,7 @@ steps:
 
 substitutions:
   _GO_VERSION: '1.19.6' # default values
-  _MAVEN_VERSION: '3.9.2' #default values
+  _MAVEN_VERSION: '3.9.2' # default values
 
 # Push images to Cloud Build to Container Registry
 images:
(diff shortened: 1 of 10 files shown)


hpandeycodeit (approved):

concourse/docker/pxf-dev-base/gpdb6/centos7/Dockerfile
```diff
@@ -24,7 +24,7 @@ RUN useradd -s /sbin/nologin -d /opt/minio minio \
     && chmod +x /opt/minio/bin/minio \
     && chown -R minio:minio /opt/minio
 
-ARG MAVEN_VERSION=3.6.3
+ARG MAVEN_VERSION
```
denalex, line 27: why did we remove the default version from the argument ? would it make sense to update it to 3.9.2 ? I believe we can still override it when running docker build command, if needed

concourse/docker/pxf-dev-base/cloudbuild.yaml
```diff
@@ -177,6 +182,7 @@ steps:
 
 substitutions:
   _GO_VERSION: '1.19.6' # default values
+  _MAVEN_VERSION: '3.9.2' #default values
```
ashuka24, line 185:
```suggestion
  _MAVEN_VERSION: '3.9.2' # default values

ashuka24 (approved):

PR #987, opened 2023-06-16
Add Escape E to the Options in FDW

hpandeycodeit:
Added escape`E` to the resource in FDW Options.

### DIFF
--- a/automation/src/main/java/org/greenplum/pxf/automation/structures/tables/pxf/ForeignTable.java
+++ b/automation/src/main/java/org/greenplum/pxf/automation/structures/tables/pxf/ForeignTable.java
@@ -47,7 +47,6 @@ public class ForeignTable extends WritableExternalTable {
         // foreign tables do not have locations, parameters go into options
         // path (resource option for FDW) should always be present
         StringJoiner joiner = new StringJoiner(",", " OPTIONS (", ")");
-        // Adding Escape E here in the resource just like external table in the URI.
         appendOption(joiner,"resource ", getPath(), !getPath().startsWith("E"));
 
         String formatOption = getFormatOption();
--- a/automation/src/main/java/org/greenplum/pxf/automation/structures/tables/pxf/ForeignTable.java
+++ b/automation/src/main/java/org/greenplum/pxf/automation/structures/tables/pxf/ForeignTable.java
@@ -47,8 +47,8 @@ public class ForeignTable extends WritableExternalTable {
         // foreign tables do not have locations, parameters go into options
         // path (resource option for FDW) should always be present
         StringJoiner joiner = new StringJoiner(",", " OPTIONS (", ")");
-        // TODO Add Escape E to the URI like external table It will help cases like HcfsGlobbingTest
-        appendOption(joiner,"resource", getPath(), true);
+        // Adding Escape E here in the resource just like external table in the URI.
+        appendOption(joiner,"resource ", "E'"+getPath()+"'", false);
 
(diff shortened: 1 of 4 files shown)


ashuka24 (approved):

automation/src/main/java/org/greenplum/pxf/automation/structures/tables/pxf/ForeignTable.java
```diff
@@ -47,8 +47,8 @@ protected String createOptions() {
         // foreign tables do not have locations, parameters go into options
         // path (resource option for FDW) should always be present
         StringJoiner joiner = new StringJoiner(",", " OPTIONS (", ")");
-        // TODO Add Escape E to the URI like external table It will help cases like HcfsGlobbingTest
-        appendOption(joiner,"resource", getPath(), true);
+        // Adding Escape E here in the resource just like external table in the URI.
+        appendOption(joiner,"resource ", "E'"+getPath()+"'", false);
```
denalex, line 51: instead of doing this unconditionally, can we use the existing pattern and do here:
```suggestion
        appendOption(joiner,"resource ", getPath(), !getPath().startsWith("E"));
```
and just use `E'whatever'` in the actual test where we need to escape the path ?

automation/src/test/java/org/greenplum/pxf/automation/features/hcfs/HcfsGlobbingTest.java
```diff
@@ -128,10 +129,16 @@ private void prepareTestScenario(String path, String testName, String data1, Str
         ProtocolEnum protocol = ProtocolUtils.getProtocol();
 
         // Create GPDB external table directed to the HDFS file
+
+        String updatedPath = protocol.getExternalTablePath(hdfs.getBasePath(), hdfs.getWorkingDirectory()) + "/" + path + "/" + glob;
+        if (FDWUtils.useFDW) {
```
ashuka24, line 134: just to double check my understanding. It is not possible for the path to already contain `E'` when we are in the fdw case?

ashuka24 (approved):

automation/src/main/java/org/greenplum/pxf/automation/structures/tables/pxf/ForeignTable.java
```diff
@@ -47,8 +47,8 @@ protected String createOptions() {
         // foreign tables do not have locations, parameters go into options
         // path (resource option for FDW) should always be present
         StringJoiner joiner = new StringJoiner(",", " OPTIONS (", ")");
-        // TODO Add Escape E to the URI like external table It will help cases like HcfsGlobbingTest
-        appendOption(joiner,"resource", getPath(), true);
+        // Adding Escape E here in the resource just like external table in the URI.
```
denalex, line 50: comment is no longer relevant

automation/src/test/java/org/greenplum/pxf/automation/features/hcfs/HcfsGlobbingTest.java
denalex: not in the test cases, `E'..'` is added by ExternalTable when constructing LOCATION clause, unnecessarily. I'd prefer we escape the string when we really know it needs escaping, otherwise we are not testing the same values as end users would be using. While changing `ExternalTable` is outside the scope, we can do the right thing here for the `ForeignTable`.

denalex (approved):

PR #988, opened 2023-06-17
Few more tests updated to worked with FDW

hpandeycodeit:
This PR modified these tests for FDW: 

- JsonTest
- JdbcHiveTest
- OrcReadTest
- ParquetTest

**OrcReadTest**: All the tests except  orcReadStringsContainingNullByte are working fine. This test orcReadStringsContainingNullByte  is failing because:

```
FDW fails for the data that have NUL. This behaviour is different from external-table but same as GPDB Heap
FDW Failure: invalid byte sequence for encoding "UTF8": 0x00

GPDB also throws the same error when copying the data with NUL

 postgres=# copy test from '/Users/pandeyhi/Documents/bad_data.txt' ;
ERROR: invalid byte sequence for encoding "UTF8": 0x00

Do we need to do some changes to make sure the external-table behaves the same way as GPDB/FDW?

```

**JsonTest**:

All Tests except the following are failing. Reason, extra quote getting appended in the output for FDW:

```
  JsonTest.malformedRecord
  JsonTest.malformedRecordWithCsvWireFormat
  JsonTest.malformedRecordWithRejectLimit
  JsonTest.mismatchedTypes
  JsonTest.mismatchedTypesWithRejectLimit
```

These tests failing because the returned error has extra quotes "

```
For e.g.:
""[truncated 73 chars]; line: 3, column: 22]'. invalid JSON record
Instead of
"[truncated 73 chars]; line: 3, column: 22]'. invalid JSON record
This is getting appended in toCsvField method of GreenplumCSV.java
And is happening only for the error cases. Adding escape "\" doesn't help here as well.

```
May be add a separate output file for these ? but for now skipped these.

**ParquetReadTest**:  Working fine 
**JdbcHiveTest**: Working fine

### DIFF
--- a/automation/src/main/java/org/greenplum/pxf/automation/structures/tables/utils/TableFactory.java
+++ b/automation/src/main/java/org/greenplum/pxf/automation/structures/tables/utils/TableFactory.java
@@ -229,6 +229,27 @@ public abstract class TableFactory {
     }
 
     /**
+     * Prepares PXF Readable External or Foreign Table for TEXT data, using TEXT format and "<protocol>:text" profile.
+     *
+     * @param name name of the table
+     * @param fields fields of the table
+     * @param path for external table path
+     * @param format format used in the external data
+     * @return PXF Readable External or Foreign table
+     */
+    public static ReadableExternalTable getPxfReadableJsonTable(String name,
+                                                                String[] fields,
+                                                                String path,
+                                                                String format) {
+        ReadableExternalTable exTable = getReadableExternalOrForeignTable(name, fields, path, format);
+        if (StringUtils.equals(format, "custom")) {
+            exTable.setFormatter("pxfwritable_import");
+        }
+        exTable.setProfile(ProtocolUtils.getProtocol().value() + ":json");
+        return exTable;
+    }
+
+    /**
      * Prepares PXF Readable External or Foreign Table for CSV data, using CSV format and "<protocol>:csv" profile.
      *
      * @param name name of the table
--- a/automation/src/main/java/org/greenplum/pxf/automation/structures/tables/utils/TableFactory.java
+++ b/automation/src/main/java/org/greenplum/pxf/automation/structures/tables/utils/TableFactory.java
@@ -229,7 +229,7 @@ public abstract class TableFactory {
     }
 
     /**
-     * Prepares PXF Readable External or Foreign Table for TEXT data, using TEXT format and "<protocol>:text" profile.
+     * Prepares PXF Readable External or Foreign Table for Json data, using custom format or CSV format and "<protocol>:json" profile.
      *
      * @param name name of the table
      * @param fields fields of the table
--- a/automation/src/test/java/org/greenplum/pxf/automation/features/orc/OrcReadTest.java
+++ b/automation/src/test/java/org/greenplum/pxf/automation/features/orc/OrcReadTest.java
@@ -1,12 +1,14 @@
 package org.greenplum.pxf.automation.features.orc;
 
+import annotations.FailsWithFDW;
+import annotations.WorksWithFDW;
 import org.greenplum.pxf.automation.features.BaseFeature;
-import org.greenplum.pxf.automation.structures.tables.pxf.ReadableExternalTable;
 import org.greenplum.pxf.automation.structures.tables.utils.TableFactory;
 import org.greenplum.pxf.automation.utils.system.ProtocolEnum;
 import org.greenplum.pxf.automation.utils.system.ProtocolUtils;
 import org.testng.annotations.Test;
 
+@WorksWithFDW
 public class OrcReadTest extends BaseFeature {
 
     private static final String ORC_PRIMITIVE_TYPES = "orc_types.orc";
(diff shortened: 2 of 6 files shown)


automation/src/test/java/org/greenplum/pxf/automation/features/json/JsonTest.java
```diff
@@ -249,6 +253,16 @@ public void exceedsMaxSize() throws Exception {
      *
      * @throws Exception if test fails to run
      */
+    // Some of these tests failing because the returned error has extra quotes "
+    // For e.g.:
+    // ""[truncated 73 chars]; line: 3, column: 22]'. invalid JSON record
+    //  Instead of
+    // "[truncated 73 chars]; line: 3, column: 22]'. invalid JSON record
+    // This is getting appended in toCsvField method of GreenplumCSV.java
```
ashuka24, line 261: It seems like this function has the ability to skip appending quotes if it is not needed. Is there something we can do here to avoid the extra quotes?

automation/src/test/java/org/greenplum/pxf/automation/features/orc/OrcReadTest.java
```diff
@@ -155,6 +157,18 @@ public void orcReadMultiDimensionalLists() throws Exception {
         runTincTest("pxf.features.orc.read.multidim_list_types.runTest");
     }
 
+    /*
+     * FDW fails for the data that have NUL. This behaviour is different from external-table but same as GPDB Heap
```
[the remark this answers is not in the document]
hpandeycodeit: `NUL` bytes in the data. 

For eg: This is the data which contains `NUL` byte in the string ( from record 4 onward) 

```
Column vector types: 0:LONG, 1:BYTES, 2:BYTES
[1, "simple string", "hello"]
[2, "simple string with space", "hello world"]
[3, "simple string with double quote", "hello "world""]
[4, "NUL-byte in middle of string", "helloworld"]
[5, "NUL-byte at the beginning of string", "hello world"]
[6, "NUL-byte at the end of string", "hello world"]
```

External table reads till `NUL` and skips the rest:

```
 4 | NUL-byte in middle of string        | hello
 5 | NUL-byte at the beginning of string | 
 6 | NUL-byte at the end of string       | hello world
```

automation/src/test/java/org/greenplum/pxf/automation/features/json/JsonTest.java
hpandeycodeit: I couldn't find a way to skip these appending quotes.

automation/src/test/java/org/greenplum/pxf/automation/features/orc/OrcReadTest.java
```diff
@@ -155,6 +157,18 @@ public void orcReadMultiDimensionalLists() throws Exception {
...
+     * FDW Failure: invalid byte sequence for encoding "UTF8": 0x00
+     *
+     * GPDB also throws the same error when copying the data with NUL
+     *
+     * postgres=# copy test from '/Users/pandeyhi/Documents/bad_data.txt' ;
+     * ERROR:  invalid byte sequence for encoding "UTF8": 0x00
+     * TODO Do we need to do some changes to make sure the external-table behaves the same way as GPDB/FDW?
```
ashuka24, line 168: was there there a discussion on this? what was the resolution?

hpandeycodeit: Not yet. I mean there was one on Slack but there is no resolution yet, that's why it's in TODO.

automation/src/test/java/org/greenplum/pxf/automation/features/json/JsonTest.java
```diff
@@ -249,6 +253,16 @@ public void exceedsMaxSize() throws Exception {
...
+    // And is happening only for the error cases. Adding escape "\" doesn't help here as well.
+
+    // TODO: May be create a separate test suite for all these failing tests
```
yimingli-vmware, line 264: Are these failing tests failing for the same reason?

yimingli-vmware (approved):

automation/src/test/java/org/greenplum/pxf/automation/features/json/JsonTest.java
hpandeycodeit: yes

automation/src/test/java/org/greenplum/pxf/automation/features/orc/OrcReadTest.java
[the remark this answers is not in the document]
hpandeycodeit: Here is the NUL byte data looks like : 

```
{"id": 4, "context": "NUL-byte in middle of string", "value": "hello\u0000world"}

```

automation/src/main/java/org/greenplum/pxf/automation/structures/tables/utils/TableFactory.java
```diff
@@ -228,6 +228,27 @@ public static ReadableExternalTable getPxfReadableTextTable(String name,
         return exTable;
     }
 
+    /**
+     * Prepares PXF Readable External or Foreign Table for TEXT data, using TEXT format and "<protocol>:text" profile.
```
ashuka24, line 232: comment here says `:text` profile, but actual code says json

automation/src/test/java/org/greenplum/pxf/automation/features/orc/OrcReadTest.java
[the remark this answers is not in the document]
ashuka24:
```suggestion
     * FDW fails for the data that contain a NUL-byte (i.e. \u000). This behaviour is different from external-table but same as GPDB Heap

File: automation/src/test/java/org/greenplum/pxf/automation/features/orc/OrcReadTest.java [hunk 6]
```diff
@@ -155,6 +157,18 @@ public void orcReadMultiDimensionalLists() throws Exception {
         runTincTest("pxf.features.orc.read.multidim_list_types.runTest");
     }
 
+    /*
+     * FDW fails for the data that have NUL. This behaviour is different from external-table but same as GPDB Heap
+     * FDW Failure: invalid byte sequence for encoding "UTF8": 0x00
+     *
+     * GPDB also throws the same error when copying the data with NUL
```
ashuka24, line 164:
```suggestion
     * GPDB also throws the same error when copying the data containing a NUL-byte

bradfordb-vmware (approved):

ashuka24 (approved):

PR #991, opened 2023-07-05
docs - jdbc connector supports years with >4 chars

lisakowen:
docs for https://github.com/greenplum-db/pxf/pull/960.

treating this property similar to that of query timeout; both do not currently have commented out entries in jdbc-site.xml.  add a new section for the jdbc.date.wideRange property to jdbc config topic, add external table option DATE_WIDE_RANGE to the jdbc "using" topic.


lisakowen:
@RomaZe please review if interested.

denalex (approved):

ashuka24 (approved):

PR #995, opened 2023-07-10
Inclusive Terminology changes 

hpandeycodeit:
Following terms are changed/renamed: 

Master -> Coordinator
Slave -> Secondary 
whitelist -> allowlist


denalex (commented):
I would expect the majority of changes in `pxf-cli` and its tests. Will that be a separate PR ?

automation/src/main/java/org/greenplum/pxf/automation/components/common/ShellSystemObject.java
```diff
@@ -57,7 +57,7 @@ public class ShellSystemObject extends BaseSystemObject {
             "GPHOME",
             "GPHD_ROOT",
             "GPDATA",
-            "MASTER_DATA_DIRECTORY",
+            "COORDINATOR_DATA_DIRECTORY",
```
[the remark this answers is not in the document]
hpandeycodeit: No, It won't work. It reads this from `default.xml`  for MASTER_DATA_DIRECTORY and fails with Null pointer. Reverted it.

hpandeycodeit:
> I would expect the majority of changes in `pxf-cli` and its tests. Will that be a separate PR ?

@denalex I haven't done changes in pxf-cli since there is no Coordinator host in GP5/GP6. Not sure if we can make differentiate it at version level on the CLI?

automation/src/main/java/org/greenplum/pxf/automation/components/cluster/installer/nodes/SegmentNode.java
```diff
@@ -0,0 +1,8 @@
+package org.greenplum.pxf.automation.components.cluster.installer.nodes;
+
+/**
+ * Represents a single StandBy Node in a cluster
```
ashuka24, line 4:
```suggestion
 * Represents a single Segment Node in a cluster

File: cli/cmd/cluster.go [hunk 3]
```diff
@@ -82,6 +82,22 @@ func handlePlurality(num int) string {
 	return "s"
 }
 
+func getHostFromVersion() string {
+
+    connection := dbconn.NewDBConnFromEnvironment("postgres")
```
bradfordb-vmware, line 87: Do we need this method? The `DBConn` struct already has the `Version` field and we open a connection in `doSetup` and store the connection on the returned `ClusterData` struct. You can add a method `func (c *ClusterData) getHost string` that returns either `master` or `coordinator` whenever we need it.

cli/cmd/cluster.go
```diff
@@ -39,16 +39,16 @@ func createCobraCommand(use string, short string, cmd *command) *cobra.Command {
...
+	initCmd     = createCobraCommand("init", "(deprecated) Install PXF extension under $GPHOME on %s, standby master, and all segment hosts", &InitCommand)
+	startCmd    = createCobraCommand("start", "Start the PXF server instances on %s, standby master, and all segment hosts", &StartCommand)
+	stopCmd     = createCobraCommand("stop", "Stop the PXF server instances on %s, standby master, and all segment hosts", &StopCommand)
+	statusCmd   = createCobraCommand("status", "Get status of PXF servers on %s, standby master, and all segment hosts", &StatusCommand)
+	syncCmd     = createCobraCommand("sync", "Sync PXF configs from %s to standby master and all segment hosts. Use --delete to delete extraneous remote files", &SyncCommand)
+	resetCmd    = createCobraCommand("reset", "(deprecated) No operation %.s", &ResetCommand)
+	registerCmd = createCobraCommand("register", "Install PXF extension under $GPHOME on %s, standby master, and all segment hosts", &RegisterCommand)
+	restartCmd  = createCobraCommand("restart", "Restart the PXF server on %s, standby master, and all segment hosts", &RestartCommand)
```
bradfordb-vmware, line 49: You also need to replace `standby master`.

bradfordb-vmware (changes requested):

cli/cmd/cluster.go
hpandeycodeit: I have modified this function a bit as per recommendation but we also need `     connection := dbconn.NewDBConnFromEnvironment("postgres") ` as `clusterData` object isn't available to format `Short` in `createCobraCommand`

cli/cmd/cluster.go
```diff
@@ -24,9 +24,14 @@ func createCobraCommand(use string, short string, cmd *command) *cobra.Command {
...
 	}
+
+	// creating a clusterData object to call getHost
+	var emptyClusterData *ClusterData
+	hostName := emptyClusterData.getHost()
+
 	return &cobra.Command{
 		Use:   use,
-		Short: short,
+		Short: fmt.Sprintf( short, hostName, hostName),
```
yimingli-vmware, line 34:
```suggestion
		Short: fmt.Sprintf(short, hostName, hostName),
```
why we have 2 `hostName` here?

cli/cmd/cluster.go
```diff
@@ -39,16 +44,16 @@ func createCobraCommand(use string, short string, cmd *command) *cobra.Command {
...
+	stopCmd     = createCobraCommand("stop", "Stop the PXF server instances on %s, standby %s, and all segment hosts", &StopCommand)
+	statusCmd   = createCobraCommand("status", "Get status of PXF servers on %s, standby %s, and all segment hosts", &StatusCommand)
+	syncCmd     = createCobraCommand("sync", "Sync PXF configs from %s to standby %s and all segment hosts. Use --delete to delete extraneous remote files", &SyncCommand)
+	resetCmd    = createCobraCommand("reset", "(deprecated) No operation %.s %.s", &ResetCommand)
+	registerCmd = createCobraCommand("register", "Install PXF extension under $GPHOME on %s, standby %s, and all segment hosts", &RegisterCommand)
+	restartCmd  = createCobraCommand("restart", "Restart the PXF server on %s, standby %s, and all segment hosts", &RestartCommand)
+	prepareCmd  = createCobraCommand("prepare", "Prepares a new base directory specified by the $PXF_BASE environment variable %.s %.s", &PrepareCommand)
+	migrateCmd  = createCobraCommand("migrate", "Migrates configurations from older installations of PXF %.s %.s", &MigrateCommand)
```
yimingli-vmware, line 56: I think `hostName` would replace these `%s`, but why in this format it's not here?

concourse/scripts/cli/test_sync.sh
```diff
@@ -7,6 +7,7 @@ source "${dir}/common.sh"
 cluster_description="$(get_cluster_description)"
 cluster_sync_description="$(get_cluster_sync_description)"
 num_hosts="${#all_cluster_hosts[@]}"
+get_host_from_version="$(get_host_from_version)"
```
yimingli-vmware, line 10: Maybe `hostName` would be a better name?

concourse/scripts/cli/common.sh
```diff
@@ -142,22 +142,38 @@ has_standby_coordinator() {
...
+  local cluster_description="$(get_host_from_version) host"
   if has_standby_coordinator; then
-    cluster_description+=", standby master host,"
+    cluster_description+=", standby $(get_host_from_version) host,"
   fi
   local num_segment_hosts
   num_segment_hosts="$(grep -c -P 'sdw\d+' hostfile_all)"
   cluster_description+=" and ${num_segment_hosts} segment hosts..."
 
+
```
yimingli-vmware, line 169: extra line

concourse/scripts/cli/common.sh
```diff
@@ -142,22 +142,38 @@ has_standby_coordinator() {
...
+      else
+        mainHostName="coordinator"
+  fi
+
+  echo "${mainHostName}"
+}
+
 get_cluster_description() {
-  local cluster_description="master host"
+  local cluster_description="$(get_host_from_version) host"
```
yimingli-vmware, line 161: we use this function twice, let's create a variable and reuse it.

bradfordb-vmware (approved):

automation/src/main/java/org/greenplum/pxf/automation/components/cluster/installer/nodes/MasterNode.java
```diff
@@ -1,8 +0,0 @@
-package org.greenplum.pxf.automation.components.cluster.installer.nodes;
-
-/**
- * Represents Master Node for a cluster
- */
-public class MasterNode extends Node {
-
```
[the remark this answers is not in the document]
hpandeycodeit: This is renamed to Coordinator , so won't break anything.

cli/cmd/cluster.go
hpandeycodeit: This is changed. Not formatting this field not.

cli/cmd/cluster.go
hpandeycodeit: This is changed, replaced master to coordinator.

concourse/scripts/cli/test_sync.sh
hpandeycodeit: this is changed to original code.

concourse/scripts/cli/common.sh
hpandeycodeit: This is reverted to original code.

concourse/scripts/cli/common.sh
hpandeycodeit: This is reverted to original code.

ashuka24 (approved):

PR #1000, opened 2023-07-14
fix rpmrebuild-rocky8 container image

bradfordb-vmware:
There is a compatibility issue between the version of coreutils (v8.30-15) included in rockylinux:8 and the version of Concourse (v5.7.1) used for PXF's CI. This causes the command `ls <some-dir>` to fail with the error message

```sh
ls: cannot access '<some-dir>': Operation not permitted
```

The same comand in the same container image running on Concourse 7.8.3 works successfully.

The Greenplum release engineering team previsouly encountered this same issue and resolved it by downgrading coreutils-single to v8.30-8. This version is no longer available from upstream repos for downgrading in the rpmrebuild images. Instead, this commit switches the base image from upstream rockylinux:8 to gpdb6-rocky8-build which is maintained by the Greenplum release engineering team and includes the older version of coreutils-single. For consistency, this commit also changes upstream centos:7 to gpdb6-centos7-build.

### DIFF
--- a/concourse/docker/rpmrebuild/cloudbuild.yaml
+++ b/concourse/docker/rpmrebuild/cloudbuild.yaml
@@ -8,7 +8,7 @@ steps:
   id: rpmrebuild-centos7-image
   args:
     - 'build'
-    - '--build-arg=BASE_IMAGE=centos:7'
+    - '--build-arg=BASE_IMAGE=${_BASE_IMAGE_REPOSITORY}/gpdb6-centos7-build:latest'
     - '--tag=gcr.io/$PROJECT_ID/rpmrebuild-centos7:latest'
     - '-f'
     - 'concourse/docker/rpmrebuild/centos/Dockerfile'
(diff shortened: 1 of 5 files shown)


yimingli-vmware (approved):

ashuka24 (commented):
Can we also update the diagram in the README.md file to reflect the new base image? 

Also, do we need to update triggers anywhere so that rebuilding these images will happen when the upstream image gets rebuilt?

hpandeycodeit (approved):

bradfordb-vmware:
> Can we also update the diagram in the README.md file to reflect the new base image?
Good catch -- I've pushed up a commit that updates the container image diagram with the latest sets of changes.

> Also, do we need to update triggers anywhere so that rebuilding these images will happen when the upstream image gets rebuilt?
I've added gpdb6-centos7-build and gpdb6-rocky8-build to the set of upstream images that will trigger a rebuild of the rpmrebuild images.

concourse/docker/rpmrebuild/cloudbuild.yaml
```diff
@@ -8,7 +8,7 @@ steps:
   id: rpmrebuild-centos7-image
   args:
     - 'build'
-    - '--build-arg=BASE_IMAGE=centos:7'
+    - '--build-arg=BASE_IMAGE=gcr.io/data-gpdb-public-images/gpdb6-centos7-build:latest'
```
ashuka24, line 11: If we already set the `BASE_IMAGE` in the dockerfile itself, why do we also need it here as a `build-arg`?

ashuka24 (approved):

concourse/docker/rpmrebuild/cloudbuild.yaml
bradfordb-vmware: In `pxf-dev-base`, we use the substitution variable `$_BASE_IMAGE_REPOSITORY`; I've checked and the same substitution variable is available in the Cloud Build trigger for the rpmrebuild images. I've updated to match pxf-dev-base.

PR #1012, opened 2023-08-14
Adding CLI Test Job for GP7 for dev pipeline

hpandeycodeit:
- In GP7, gpscp is replaced by gpsync. Introduced a variable (GP_SCP_CMD) to specify the appropriate command based on GP version.
- Replaced the hardcoded pxf home (/usr/local/pxf-gp6) with the variable PXF_HOME.
- Added variables to the Makefile (GP7_CLI) and dev pipeline template (gp7_cli), when set to true, this will introduce PXF_CLI tests for GP7.
- Ran yamllint and shellcheck to confirm there are no issues with the new changes.
- For ccp, use PLATFORM value (rocky8-gpdb7) as the corresponding VM Image comes preinstalled with all GPDB7 runtime dependencies.

Note: See Jira details for dev-pipeline supporting this PR.

### DIFF
--- a/concourse/Makefile
+++ b/concourse/Makefile
@@ -37,6 +37,7 @@ GS                          ?= false
 MINIO                       ?= false
 OEL7                        ?= false
 FILE                        ?= false
+GP7_CLI                     ?= false
 
 SET_PIPELINE := set-pipeline
 ifeq ($(CHECK_CREDS), true)
@@ -116,6 +117,7 @@ set-dev-build-pipeline:
 		gs=$(GS) \
 		minio=$(MINIO) \
 		oel7=$(OEL7) \
+		gp7_cli=$(GP7_CLI) \
 		dev_pipeline=true \
 		user=$(USER) \
 		branch=$(BRANCH) \
--- a/concourse/pipelines/templates/dev_build_pipeline-tpl.yml
+++ b/concourse/pipelines/templates/dev_build_pipeline-tpl.yml
@@ -83,7 +83,7 @@ resource_types:
     password: ((ud/pxf/secrets/pxf-cloudbuild-service-account-key))
 {% endif %}
 
-{% if multinode or gp7_cli %}
+{% if multinode %}
 - name: terraform-0.14.10
   type: registry-image
   source:
@@ -617,7 +617,7 @@ jobs:
 {% endfor %}
 {% set gp_ver = None %}
 
-## ----------GP7 CLI tests ----------\
+## ----------GP7 PXF CLI tests START-----------
 
 {% if gp7_cli %}
 {% set gp_ver = 7 %}
@@ -647,7 +647,7 @@ jobs:
           generate_random_name: true
           terraform_source: ccp_src/google/
           vars:
-            PLATFORM: rocky8-gpdb7
+            PLATFORM: rocky8-gpdb[[gp_ver]]
             number_of_nodes: ((number_of_gpdb_nodes))
             extra_nodes: 1
             segments_per_host: 4
@@ -666,7 +666,7 @@ jobs:
           AWS_DEFAULT_REGION: ((ud/common/aws-region))
           BUCKET_PATH: ((tf-bucket-path))
           BUCKET_NAME: ((ud/pxf/common/tf-bucket-name))
-          PLATFORM: rocky8-gpdb7
+          PLATFORM: rocky8-gpdb[[gp_ver]]
           CLOUD_PROVIDER: google
           GPDB_RPM: true
   - task: Initialize Greenplum
@@ -683,15 +683,14 @@ jobs:
       PXF_JVM_OPTS: ((pxf-jvm-opts))
   - task: Test PXF CLI
     on_success:
-      in_parallel:
-        steps:
-        - *destroy_gpdb_cluster
+      <<: *destroy_gpdb_cluster
     image: gpdb[[gp_ver]]-pxf-dev-rocky8-image
     file: pxf_src/concourse/tasks/test_pxf_cli.yml
 {% if slack_notification %}
     <<: *slack_alert
 {% endif %}
 {% endif %}
+## ----------GP7 PXF CLI tests END ----------
 
 ## ---------- Extended Tests ----------
 {% set gp_ver = 6 %}
--- a/concourse/pipelines/templates/dev_build_pipeline-tpl.yml
+++ b/concourse/pipelines/templates/dev_build_pipeline-tpl.yml
@@ -617,7 +617,7 @@ jobs:
 {% endfor %}
 {% set gp_ver = None %}
 
-## ----------GP7 PXF CLI tests START-----------
+## ---------- GP7 PXF tests ----------
 
 {% if gp7_cli %}
 {% set gp_ver = 7 %}
@@ -689,8 +689,8 @@ jobs:
 {% if slack_notification %}
     <<: *slack_alert
 {% endif %}
-{% endif %}
-## ----------GP7 PXF CLI tests END ----------
+{% endif %} # if gp7_cli ends here
+{% set gp_ver = None %}
 
 ## ---------- Extended Tests ----------
 {% set gp_ver = 6 %}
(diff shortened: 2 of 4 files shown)


concourse/scripts/install_pxf.bash
```diff
@@ -12,6 +12,14 @@ if [[ ${PXF_COMPONENT} == "true" ]]; then
 else
 	PXF_HOME=${GPHOME}/pxf
 fi
+
+# gpscp command is replaced with gpsync in GP7
+if [ "${GP_VER}" -lt 7 ]; then
```
denalex, line 17:
```suggestion
if [ ${GP_VER} -lt 7 ]; then
```
otherwise you're comparing a string to a number ?

denalex (commented):
Also should we change the title of the PR / commit message from "Adding CLI Test Job for GP7" to "Adding CLI Test Job for GP7 for dev pipeline" ?

concourse/scripts/install_pxf.bash
bradfordb-vmware: According to `man 1 test`, `INTEGER1 -lt INTEGER2` tests if `INTEGER1` is less than `INTEGER2`. If you omit the double-quotes, then shellcheck warns about globbing and word splitting. The quotes don't cause bash to treat the value as an string.

concourse/pipelines/templates/dev_build_pipeline-tpl.yml
```diff
@@ -83,7 +83,7 @@ resource_types:
     password: ((ud/pxf/secrets/pxf-cloudbuild-service-account-key))
 {% endif %}
 
-{% if multinode %}
+{% if multinode or gp7_cli %}
```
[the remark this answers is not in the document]
bradfordb-vmware: The CCP resource (`name: terraform`) is still using the `terraform` resource type so its still only the IPA cluster that uses Terraform 0.14.10.

concourse/pipelines/templates/dev_build_pipeline-tpl.yml
```diff
@@ -618,6 +617,82 @@ jobs:
...
+    - do:
+      - put: terraform_gpdb
+        resource: terraform
+        params:
+          action: create
+          delete_on_failure: true
+          generate_random_name: true
+          terraform_source: ccp_src/google/
+          vars:
+            PLATFORM: rocky8-gpdb7
```
ashuka24, line 650: in all the other locations, we only use the general platform `PLATFORM: rocky8`, is there a reason we are now specifying the gpdb version as well?

concourse/pipelines/templates/dev_build_pipeline-tpl.yml
```diff
@@ -618,6 +617,82 @@ jobs:
...
+    on_success:
+      in_parallel:
+        steps:
+        - *destroy_gpdb_cluster
+    image: gpdb[[gp_ver]]-pxf-dev-rocky8-image
+    file: pxf_src/concourse/tasks/test_pxf_cli.yml
+{% if slack_notification %}
+    <<: *slack_alert
+{% endif %}
+{% endif %}
```
ashuka24, line 694: the beginning of this if is relatively far up, can we comment that this is the end if for the `if gp7-cli`

concourse/pipelines/templates/dev_build_pipeline-tpl.yml
hpandeycodeit: `rocky8-gpdb7` is provided by releng. It contains the needed runtime dependencies for GP7. Most recently this was updated to include python3 dependencies, which are no longer provided with GP7 rpm.

concourse/pipelines/templates/dev_build_pipeline-tpl.yml
```diff
@@ -618,6 +617,82 @@ jobs:
...
+      PXF_JVM_OPTS: ((pxf-jvm-opts))
+  - task: Test PXF CLI
+    on_success:
+      in_parallel:
+        steps:
+        - *destroy_gpdb_cluster
+    image: gpdb[[gp_ver]]-pxf-dev-rocky8-image
+    file: pxf_src/concourse/tasks/test_pxf_cli.yml
+{% if slack_notification %}
+    <<: *slack_alert
```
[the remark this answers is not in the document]
hpandeycodeit: Following existing slack alert patterns which are present in the majority of test jobs. There is slight inconsistency in non-extended tests usage of the slack alerts.

concourse/scripts/install_pxf.bash
hpandeycodeit: @denalex for your reference, quotes are fine here. We are concerned about the shellcheck issue that Bradford mentioned (shellcheck warns about globbing and word splitting)
Also, we have confirmed quoting has no impact on the numeric comparison. Please refer to this link : https://tldp.org/LDP/abs/html/comparison-ops.html  as it has examples of quoted numeric variables.

concourse/pipelines/templates/dev_build_pipeline-tpl.yml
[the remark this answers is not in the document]
edespino: @denalex nice catch and @bradfordb-vmware good input. Unlike unused resources, Concourse does not warn/error with unused resource_types (i.e. `terraform-0.14.10`) exist in a pipeline. The `terraform-0.14.10` resource_type is only used in the `terraform_ipa_hadoop` which is only enabled with the `MULTINODE` (pipeline: `multinode`) configuration.

concourse/Makefile
```diff
@@ -116,6 +117,7 @@ set-dev-build-pipeline:
 		gs=$(GS) \
 		minio=$(MINIO) \
 		oel7=$(OEL7) \
+		gp7_cli=$(GP7_CLI) \
```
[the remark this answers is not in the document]
edespino: @denalex Thank you for your observation and question - The current pattern we use for the dev test configuration options doesn't provide an indication that an option is only applicable to a specific GP version. The CLI test is explicitly being extracted from within the *IMPERS* GP 6 jobs for GP7 use. We are attempting to provide developer friendly configuration option which indicates it is only applicable to a specific GP version (GP7_CLI). I would like to give this enhanced pattern a try with the intent that we can get some developer feedback during actual use and determine if it is a worthwhile to introduce in other applicable scenarios.

concourse/pipelines/templates/dev_build_pipeline-tpl.yml
```diff
@@ -618,6 +617,81 @@ jobs:
...
+  - task: Test PXF CLI
+    on_success:
+      <<: *destroy_gpdb_cluster
+    image: gpdb[[gp_ver]]-pxf-dev-rocky8-image
+    file: pxf_src/concourse/tasks/test_pxf_cli.yml
+{% if slack_notification %}
+    <<: *slack_alert
+{% endif %}
+{% endif %}
+## ----------GP7 PXF CLI tests END ----------
```
edespino, line 693: We need to unset gp_ver at the end of the new GP 7 CLI block.

`{% set gp_ver = None %}`

concourse/Makefile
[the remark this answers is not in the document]
hpandeycodeit: As discussed during the meeting, we are keeping it  as `GP7_CLI`.

concourse/pipelines/templates/dev_build_pipeline-tpl.yml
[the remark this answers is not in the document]
hpandeycodeit: Thanks for pointing it out! I have tested it in the pipeline and this resource type `terraform-0.14.10` is not needed for this job. Updated the template file as well.

edespino (approved):
+1

yimingli-vmware (approved):

concourse/pipelines/templates/dev_build_pipeline-tpl.yml
```diff
@@ -618,6 +617,81 @@ jobs:
 {% endfor %}
 {% set gp_ver = None %}
 
+## ---------- GP7 PXF tests ----------
```
ashuka24, line 620: the only test in this section is the CLI and we have other GP7 related test jobs above. should we update this comment?

ashuka24 (approved):

concourse/scripts/install_pxf.bash
denalex: then at least put 7 in quotes ?

Issue #909, opened 2022-12-12
PXF - gs:parquet - Can't get Master Kerberos principal for use as renewer


mpospis:
Hello,
When we try to query external table using `gs:parquet` profile after querying another external table using `Hive` profile with Kerberos authentication, the statement fails by reporting the following PXF server error:
```
[08000] ERROR: PXF server error : Can't get Master Kerberos principal for use as renewer
```

However after restarting PXF, a query selecting data from the external table with `gs:parquet` profile finishes successfully, but only until there's an external table with `Hive` profile queried. Attached [pxf-service.log](https://github.com/greenplum-db/pxf/files/10210691/pxf-service.log) file captures DEBUG output with the sequence of events as follows:
1. start PXF
2. query external table `adhoc.ext_test_gs` pointing to 'folder' in GS using `gs:parquet` profile - finishes successfully
3. query external table `adhoc.ext_test_hive` pointing to Hive table in a cluster with Kerberos auth enabled using `Hive` profile - finishes successfully
4. query external table `adhoc.ext_test_gs` pointing to 'folder' in GS using `gs:parquet` profile - fails with `PXF server error : Can't get Master Kerberos principal for use as renewer`

The question is if this is a bug or misconfiguration issue?

Many thanks!

_Greenplum Version:_ **6.22.2**
_PXF version:_ **6.4.2**

**gs-site.xml:**
```
<?xml version="1.0" encoding="UTF-8"?>
<configuration>
    <property>
        <name>google.cloud.auth.service.account.enable</name>
        <value>true</value>
    </property>
    <property>
        <name>google.cloud.auth.service.account.json.keyfile</name>
        <value>/home/gpadmin/keys/*redacted*</value>
    </property>
    <property>
        <name>fs.AbstractFileSystem.gs.impl</name>
        <value>com.google.cloud.hadoop.fs.gcs.GoogleHadoopFS</value>
    </property>
</configuration>
```
**pxf-env.sh:**
```
#!/bin/bash

##############################################################################
# This file contains PXF properties that can be specified by users           #
# to customize their deployments. This file is sourced by PXF Server control #
# scripts upon initialization, start and stop of the PXF Server.             #
#                                                                            #
# To update a property, uncomment the line and provide a new value.          #
##############################################################################

# Path to JAVA
# export JAVA_HOME=/opt/jdk11

# Path to Log directory
# export PXF_LOGDIR="${PXF_BASE}/logs"

# Path to Run directory
[...]

PR #1014, opened 2023-08-18
Update dataproc script to optionally create a kerberized cluster

ashuka24:
This PR adds functionality to the developer script to ease creation of a kerberized dataproc cluster.


dev/Dataproc-with-Kerberos.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Environment SetUp
+
+1. Run the modified `dataproc-cluster.bash` script with following options to create a local Dataproc cluster with Kerberos authentication
+
+    ```sh
+    KERBERIZED=true IMAGE_VERSION=1.5-debian10 ./dataproc-cluster.bash --create 'core:hadoop.security.auth_to_local=RULE:[1:$1] RULE:[2:$1] DEFAULT,hdfs:dfs.client.use.datanode.hostname=true'
+    ```
+
+1. SSH into cluster code (e.g., `gcloud compute ssh "${DATAPROC_CLUSTER_NAME}-m --zone=${ZONE}"`) and created a PXF service principal named `${USER}`
```
denalex, line 13:
```suggestion
1. SSH into cluster code (e.g., `gcloud compute ssh "${DATAPROC_CLUSTER_NAME}-m --zone=${ZONE}"`) and create a PXF service principal named `${USER}`
```

dev/dataproc-cluster.bash
```diff
@@ -90,10 +96,19 @@ function create_firewall_rule() {
     # HDFS NameNode         9870 http://${CLUSTER_NAME}-m:9870
     #
     # https://cloud.google.com/dataproc/docs/concepts/accessing/cluster-web-interfaces
+    local allow_list="tcp:8020,tcp:8088,tcp:9083,tcp:9866,tcp:9870,tcp:10000"
+    if [[ "${KERBERIZED}" == true ]]; then
+    # Kerberos Service  Port
+    # KDC               88
+    # Admin server      750
+      echo "Include KDC and admin server ports to allow list..."
```
denalex, line 104:
```suggestion
      echo "Adding KDC and admin server ports to allow list..."
```

dev/Dataproc-with-Kerberos.md
```diff
@@ -0,0 +1,137 @@
+# Local Dataproc Cluster with Kerberos Authentication
+
+This developer note will guide you through the process of creating a Google Cloud Dataproc cluster with Kerberos authentication enabled.
+
+## Environment SetUp
+
+1. Run the modified `dataproc-cluster.bash` script with following options to create a local Dataproc cluster with Kerberos authentication
+
+    ```sh
+    KERBERIZED=true IMAGE_VERSION=1.5-debian10 ./dataproc-cluster.bash --create 'core:hadoop.security.auth_to_local=RULE:[1:$1] RULE:[2:$1] DEFAULT,hdfs:dfs.client.use.datanode.hostname=true'
```
denalex, line 10: can the extra options be hardcoded and applied automatically if KERBERIZED is set to `true` ?

dev/Dataproc-with-Kerberos.md
```diff
@@ -0,0 +1,137 @@
...
+    sudo kadmin.local -q "add_principal -nokey ${USER}"
+    sudo kadmin.local -q "ktadd -k pxf.service.keytab ${USER}"
+    sudo chown "${USER}:" ~/pxf.service.keytab
+    chmod 0600 ~/pxf.service.keytab
+    sudo addgroup "${USER}" hdfs
+    sudo addgroup "${USER}" hadoop
+
+    # verify the keytab
+    klist -ekt pxf.service.keytab
+    ```
```
denalex, line 25: can these steps above be automatically applied inside the script if KERBERIZED is set to `true` ?

dev/dataproc-cluster.bash
bradfordb-vmware: Ditto to Alex's comment -- it's not clear that if `Include ...` is an imperative to the user to do something or a statement about what the script is doing.

dev/dataproc-cluster.bash
```diff
@@ -15,6 +15,7 @@ NETWORK="${NETWORK:-default}"
 # valid choices for this can be found
 # https://cloud.google.com/dataproc/docs/concepts/versioning/dataproc-version-clusters
 IMAGE_VERSION="${IMAGE_VERSION:-2.0-debian10}"
```
bradfordb-vmware, line 17: The script seems to set a default of `2.0-debian10` but the write-up directs us to use `1.5-debian10`. Do we know why we need 1.5 and not 2.0? Looking at the Google Cloud link, `1.5-debian10` is no longer supported.

dev/dataproc-cluster.bash
```diff
@@ -204,6 +230,49 @@ Now do the following:
 EOF
 }
 
+function print_user_instructions_for_kerberos_create() {
+    cat <<EOF
+This cluster has Kerberos Authentication enabled, please check the Dataproc-with-Kerberos README to finish
+setting up the cluster and your local environment:
+
+    1. Create a PXF service principal in the newly generated cluster and copy it to the local working directory.
+      Steps to create and verify the principal and keytab can be found in the Dataproc-with-Kerberos.md file.
```
bradfordb-vmware, line 239: Instead of having a separate README for the kerberized dataproc, should we just inline the extra steps here? Also, since there may be several steps, should we also print these next steps to a file under the `dataproc_env_files/` directory? Something like `cat <<EOF | tee dataproc_env_files/NEXT_STEPS.md`?

dev/Dataproc-with-Kerberos.md
bradfordb-vmware: Looking at the bash function, `create_dataproc_cluster` I think we can trivially prepend/append this to the local variable `extra_hadoop_config`.  The only tricky bit is that it may prevent a developer from customizing the value of `auth_to_local` but I think the likelihood of a developer needing to customize that value is very low.

dev/Dataproc-with-Kerberos.md
```diff
@@ -0,0 +1,137 @@
...
+
+    #export HADOOP_HOME=<path/to/hadoop>
+    #export HIVE_HOME=<path/to/hive>
+    "${HADOOP_HOME}/bin/hdfs" dfs -ls /
+    "${HIVE_HOME}/bin/beeline" -u "jdbc:hive2://${DATAPROC_CLUSTER_NAME}-m.c.data-gpdb-ud.internal:10000/default;principal=hive/${DATAPROC_CLUSTER_NAME}-m.c.data-gpdb-ud.internal@C.DATA-GPDB-UD.INTERNAL"
+    ```
+
+    **NOTE:** Java 8 does not like/support the [directives `include` or `includedir`][0]; rather than attempt to automate editing the system's `/etc/krb5.conf` or provide manual steps for editing it (which would require also removing the config when destroying the cluster), this guide takes a more conservative approach of using an alternate location for the Kerberos config (e.g., `KRB5_CONFIG` and `-Djava.security.krb5.conf` above).
+
+## Hive Setup
```
bradfordb-vmware, line 55: I think the rest of this README was specific to staging data in the dataproc cluster to investigate #909. Do we need this?

ashuka24: I think it is helpful as a test run to make sure that your local system is properly set up to connect to the kerberized cluster

dev/Dataproc-with-Kerberos.md
ashuka24: I added new functionality that does exactly this.

dev/dataproc-cluster.bash
ashuka24: All the steps have been added in as functions and I believe a separate file is no longer necessary

dev/dataproc-cluster.bash
ashuka24: Tested with 2.0 and it works. I will be removing the writeup override

bradfordb-vmware (approved):

dev/Dataproc-with-Kerberos.md
```diff
@@ -0,0 +1,117 @@
...
+    klist
+    export HADOOP_OPTS="-Djava.security.krb5.conf=${PWD}/dataproc_env_files/krb5.conf"
+
+    #export HADOOP_HOME=<path/to/hadoop>
+    #export HIVE_HOME=<path/to/hive>
+    "${HADOOP_HOME}/bin/hdfs" dfs -ls /
+    "${HIVE_HOME}/bin/beeline" -u "jdbc:hive2://${DATAPROC_CLUSTER_NAME}-m.c.data-gpdb-ud.internal:10000/default;principal=hive/${DATAPROC_CLUSTER_NAME}-m.c.data-gpdb-ud.internal@C.DATA-GPDB-UD.INTERNAL"
+    ```
+
+    **NOTE:** Java 8 does not like/support the [directives `include` or `includedir`][0]; rather than attempt to automate editing the system's `/etc/krb5.conf` or provide manual steps for editing it (which would require also removing the config when destroying the cluster), this guide takes a more conservative approach of using an alternate location for the Kerberos config (e.g., `KRB5_CONFIG` and `-Djava.security.krb5.conf` above).
```
hpandeycodeit, line 33: Since Java8 doesn't support certain directives, can we specify which java version is recommended here?

hpandeycodeit (approved):

yimingli-vmware (approved):

dev/Dataproc-with-Kerberos.md
ashuka24: As this section is for setting up the user's local environment. I think it is fine to leave this here. Singlecluster on local machines work only with Java 8 but if you are connecting directly to the cluster, Java 11 may be ok. This is really dependent on the user's environment

PR #1018, opened 2023-08-24
Add pxf.service.kerberos.ticket-renew-window option to pxf-site.xml

ashuka24:
PXF currently renews the kerberos ticket only when at least 80% of the ticket lifespan has passed. 

This PR introduces a new property `pxf.service.kerberos.ticket-renew-window` that allows users to change the threshold at which we renew the ticket. The default value is 0.8f to maintain backwards compatibility. This value can be changed by adding the property into the pxf-site.xml file and can be different for each server config. For example, to renew the ticket after 30% of the ticket lifespan has passed, set the following: 
```
    <property>
        <name>pxf.service.kerberos.ticket-renew-window</name>
        <value>0.3</value>
    </property>
```

### DIFF
--- a/server/pxf-api/src/main/java/org/apache/hadoop/security/LoginSession.java
+++ b/server/pxf-api/src/main/java/org/apache/hadoop/security/LoginSession.java
@@ -21,11 +21,13 @@ public class LoginSession {
     private String principalName;
     private String keytabPath;
     private long kerberosMinMillisBeforeRelogin;
+    private float kerberosTicketRenewWindow;
 
     // derived fields stored to be re-used for subsequent requests
     private UserGroupInformation loginUser;
     private Subject subject;
     private User user;
+
 
     /**
      * Creates a new session object with a config directory
@@ -33,7 +35,7 @@ public class LoginSession {
      * @param configDirectory server configuration directory
      */
     public LoginSession(String configDirectory) {
-        this(configDirectory, null, null, null, null, 0L);
+        this(configDirectory, null, null, null, null, 0L, 0f);
     }
 
     /**
@@ -43,7 +45,7 @@ public class LoginSession {
      * @param loginUser       UserGroupInformation for the given principal after login to Kerberos was performed
      */
     public LoginSession(String configDirectory, UserGroupInformation loginUser) {
-        this(configDirectory, null, null, loginUser, null, 0L);
+        this(configDirectory, null, null, loginUser, null, 0L, 0f);
     }
 
     /**
@@ -54,7 +56,7 @@ public class LoginSession {
      * @param keytabPath      full path to a keytab file for the principal
      */
     public LoginSession(String configDirectory, String principalName, String keytabPath, long kerberosMinMillisBeforeRelogin) {
-        this(configDirectory, principalName, keytabPath, null, null, kerberosMinMillisBeforeRelogin);
+        this(configDirectory, principalName, keytabPath, null, null, kerberosMinMillisBeforeRelogin, 0f);
     }
 
     /**
@@ -66,9 +68,10 @@ public class LoginSession {
      * @param loginUser                      UserGroupInformation for the given principal after login to Kerberos was performed
      * @param subject                        the subject
      * @param kerberosMinMillisBeforeRelogin the number of milliseconds before re-login
+     * @param kerberosTicketRenewWindow the percentage of the ticket lifespan
      */
     public LoginSession(String configDirectory, String principalName, String keytabPath, UserGroupInformation loginUser,
-                        Subject subject, long kerberosMinMillisBeforeRelogin) {
+                        Subject subject, long kerberosMinMillisBeforeRelogin, float kerberosTicketRenewWindow) {
         this.configDirectory = configDirectory;
         this.principalName = principalName;
         this.keytabPath = keytabPath;
@@ -78,6 +81,7 @@ public class LoginSession {
             this.user = subject.getPrincipals(User.class).iterator().next();
         }
         this.kerberosMinMillisBeforeRelogin = kerberosMinMillisBeforeRelogin;
+        this.kerberosTicketRenewWindow = kerberosTicketRenewWindow;
     }
 
     /**
@@ -134,6 +138,10 @@ public class LoginSession {
         return principalName;
     }
 
+    public float getKerberosTicketRenewWindow() {
+        return kerberosTicketRenewWindow;
+    }
+
     @Override
     public boolean equals(Object o) {
         if (this == o) return true;
@@ -143,12 +151,13 @@ public class LoginSession {
         return Objects.equals(configDirectory, that.configDirectory) &&
                 Objects.equals(principalName, that.principalName) &&
                 Objects.equals(keytabPath, that.keytabPath) &&
-                kerberosMinMillisBeforeRelogin == that.kerberosMinMillisBeforeRelogin;
+                kerberosMinMillisBeforeRelogin == that.kerberosMinMillisBeforeRelogin &&
+                kerberosTicketRenewWindow == that.kerberosTicketRenewWindow;
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(configDirectory, principalName, keytabPath, kerberosMinMillisBeforeRelogin);
+        return Objects.hash(configDirectory, principalName, keytabPath, kerberosMinMillisBeforeRelogin, kerberosTicketRenewWindow);
     }
 
     @Override
@@ -158,6 +167,7 @@ public class LoginSession {
                 .append("principal", principalName)
                 .append("keytab", keytabPath)
                 .append("kerberosMinMillisBeforeRelogin", kerberosMinMillisBeforeRelogin)
+                .append("kerberosTicketRenewWindow", kerberosTicketRenewWindow)
                 .toString();
     }
 }
--- a/server/pxf-api/src/main/java/org/apache/hadoop/security/LoginSession.java
+++ b/server/pxf-api/src/main/java/org/apache/hadoop/security/LoginSession.java
@@ -50,9 +50,9 @@ public class LoginSession {
     /**
      * Creates a new session object.
      *
-     * @param configDirectory server configuration directory
-     * @param principalName   Kerberos principal name to use to obtain tokens
-     * @param keytabPath      full path to a keytab file for the principal
+     * @param configDirectory           server configuration directory
+     * @param principalName             Kerberos principal name to use to obtain tokens
+     * @param keytabPath                full path to a keytab file for the principal
      * @param kerberosTicketRenewWindow the percentage of the ticket lifespan
      */
     public LoginSession(String configDirectory, String principalName, String keytabPath, long kerberosMinMillisBeforeRelogin, float kerberosTicketRenewWindow) {
@@ -68,7 +68,7 @@ public class LoginSession {
      * @param loginUser                      UserGroupInformation for the given principal after login to Kerberos was performed
      * @param subject                        the subject
      * @param kerberosMinMillisBeforeRelogin the number of milliseconds before re-login
-     * @param kerberosTicketRenewWindow the percentage of the ticket lifespan
+     * @param kerberosTicketRenewWindow      the percentage of the ticket lifespan
      */
     public LoginSession(String configDirectory, String principalName, String keytabPath, UserGroupInformation loginUser,
                         Subject subject, long kerberosMinMillisBeforeRelogin, float kerberosTicketRenewWindow) {
--- a/server/pxf-api/src/main/java/org/apache/hadoop/security/LoginSession.java
+++ b/server/pxf-api/src/main/java/org/apache/hadoop/security/LoginSession.java
@@ -28,7 +28,6 @@ public class LoginSession {
     private Subject subject;
     private User user;
 
-
     /**
      * Creates a new session object with a config directory
      *
--- a/server/pxf-api/src/main/java/org/apache/hadoop/security/LoginSession.java
+++ b/server/pxf-api/src/main/java/org/apache/hadoop/security/LoginSession.java
@@ -53,6 +53,7 @@ public class LoginSession {
      * @param configDirectory server configuration directory
      * @param principalName   Kerberos principal name to use to obtain tokens
      * @param keytabPath      full path to a keytab file for the principal
+     * @param kerberosTicketRenewWindow the percentage of the ticket lifespan
      */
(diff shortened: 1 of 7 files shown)


server/pxf-api/src/main/java/org/apache/hadoop/security/PxfUserGroupInformation.java
```diff
@@ -129,7 +126,7 @@ public synchronized LoginSession loginUserFromKeytab(Configuration configuration
      * @throws IOException           when an IO error occurs
      * @throws KerberosAuthException on a failure
      */
-    public void reloginFromKeytab(String serverName, LoginSession loginSession) throws KerberosAuthException {
+    public void reloginFromKeytab(String serverName, LoginSession loginSession, float ticket_renew_threshold) throws KerberosAuthException {
```
bradfordb-vmware, line 129:
```suggestion
    public void reloginFromKeytab(String serverName, LoginSession loginSession, float ticketRenewThreshold) throws KerberosAuthException {
```

server/pxf-api/src/main/java/org/apache/hadoop/security/PxfUserGroupInformation.java
```diff
@@ -237,10 +234,10 @@ private void fixKerberosTicketOrder(Subject subject) {
         LOG.warn("Warning, no kerberos tickets found while attempting to renew ticket");
     }
 
-    private long getRefreshTime(KerberosTicket tgt) {
+    private long getRefreshTime(KerberosTicket tgt, float ticket_renew_threshold) {
```
bradfordb-vmware, line 237:
```suggestion
    private long getRefreshTime(KerberosTicket tgt, float ticketRenewThreshold) {
```

server/pxf-api/src/main/java/org/apache/hadoop/security/PxfUserGroupInformation.java
```diff
@@ -40,10 +40,7 @@ public class PxfUserGroupInformation {
     private static final String MUST_FIRST_LOGIN_FROM_KEYTAB = "loginUserFromKeyTab must be done first";
     private static final String OS_LOGIN_MODULE_NAME = getOSLoginModuleName();
 
-    /**
-     * Percentage of the ticket window to use before we renew ticket.
-     */
-    private static final float TICKET_RENEW_WINDOW = 0.80f;
+    public static final String PROPERTY_TICKET_RENEW_THRESHOLD = "pxf.service.kerberos.ticket-renew-threshold";
```
bradfordb-vmware, line 43: In `SecureLogin`, it looks like we use a prefix of `CONFIG_KEY_` -- can we use the same here?
**Edit:** Follow-up question -- there a reason this key name is defined here and not in `SecureLogin` where we use it?

server/pxf-service/src/templates/templates/pxf-site.xml
```diff
@@ -26,6 +26,16 @@
             Set to true to enable, false to disable.
         </description>
     </property>
+    <!--
+    <property>
+        <name>pxf.service.kerberos.ticket-renew-threshold</name>
```
bradfordb-vmware, line 31: Should we include `percent` or `percentage` in the property name?

server/pxf-api/src/main/java/org/apache/hadoop/security/PxfUserGroupInformation.java
denalex, line 43: I prefer we keep the concept as "ticket renew window" since this is lifted from here: https://github.com/apache/hadoop/blob/trunk/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/security/UserGroupInformation.java#L112

The property name can be `PROPERTY_` or `CONFIG_KEY_` .. TICKET_RENEW_WINDOW and the actual property can have a name `pxf.service.kerberos.ticket-renew-window"

server/pxf-api/src/main/java/org/apache/hadoop/security/PxfUserGroupInformation.java
denalex, line 129: we can pass it around or maybe we can bake it into the `LoginSession` object since we might need it later for the refresh thread. Arguably we can do that later when we actually need it.

server/pxf-api/src/main/java/org/greenplum/pxf/api/security/SecureLogin.java
```diff
@@ -65,6 +67,8 @@ public class SecureLogin {
     public static final String CONFIG_KEY_SERVICE_USER_IMPERSONATION = "pxf.service.user.impersonation";
     public static final String CONFIG_KEY_SERVICE_USER_NAME = "pxf.service.user.name";
 
+    public static final float DEFAULT_TICKET_RENEW_THRESHOLD = 0.8f;
```
denalex, line 70:
```suggestion
    public static final float DEFAULT_TICKET_RENEW_WINDOW = 0.8f;
```
see previous comments as to why

server/pxf-api/src/main/java/org/greenplum/pxf/api/security/SecureLogin.java
```diff
@@ -130,7 +134,11 @@ public UserGroupInformation getLoginUser(String serverName, String configDirecto
 
         // try to relogin to keep the TGT token from expiring, if it still has a long validity, it will be a no-op
         if (Utilities.isSecurityEnabled(configuration)) {
-            pxfUserGroupInformation.reloginFromKeytab(serverName, loginSession);
+
+            // the percentage of the ticket window to use before we renew ticket
+            float ticket_renew_threshold = configuration.getFloat(PROPERTY_TICKET_RENEW_THRESHOLD, DEFAULT_TICKET_RENEW_THRESHOLD);
```
denalex, line 139: we need to do a check here that the number is between 0 and 1, error out if not

server/pxf-api/src/test/java/org/apache/hadoop/security/PxfUserGroupInformationTest.java
```diff
@@ -15,6 +15,8 @@
 import javax.security.auth.login.LoginException;
 import java.util.Date;
 
+import static org.apache.hadoop.security.PxfUserGroupInformation.PROPERTY_TICKET_RENEW_THRESHOLD;
+import static org.greenplum.pxf.api.security.SecureLogin.DEFAULT_TICKET_RENEW_THRESHOLD;
```
denalex, line 19: this cross-package import does not feel right. In general, I am trying to avoid importing production code constants in test cases.

server/pxf-service/src/templates/templates/pxf-site.xml
ashuka24: As per Alex's comment below, I will change the name to ticket-renew-window to match with the hadoop project for consistency.

server/pxf-api/src/test/java/org/apache/hadoop/security/PxfUserGroupInformationTest.java
```diff
@@ -145,14 +145,70 @@ public void testLoginFromKeytabMinMillisFromDefault() throws Exception {
...
+        // assert that the login session was created with properly wired up ugi/subject/user/loginContext
+        assertEquals(0.2f, session.getKerberosTicketRenewWindow()); // will pick from configuration
+        assertEquals("/path/to/keytab", session.getKeytabPath());
+        assertEquals("principal/some.host.com@EXAMPLE.COM", session.getPrincipalName());
+        assertEquals(ugi, session.getLoginUser()); // UGI equality only compares enclosed subjects
+        assertNotSame(ugi, session.getLoginUser()); // UGI equality only compares enclosed subjects
+        assertSame(subject, session.getSubject());
+        assertSame(user, session.getUser());
+        assertSame(mockLoginContext, session.getUser().getLogin());
+        assertEquals(UserGroupInformation.AuthenticationMethod.KERBEROS, session.getLoginUser().getAuthenticationMethod());
```
denalex, line 164: this long block now repeats in 4 places, maybe extract into a helper function ?

server/pxf-service/src/templates/templates/pxf-site.xml
```diff
@@ -26,6 +26,16 @@
...
+    <!--
+    <property>
+        <name>pxf.service.kerberos.ticket-renew-window</name>
+        <value>0.8</value>
+        <description>
+            By default, PXF refreshes the Kerberos ticket when 80% (0.8) of its lifespan has elapsed.
+            Specifies the window, as a float, at which PXF will actively refresh the ticket.
+        </description>
+    </property>
+    -->
```
denalex, line 38: should we have this uncommented ? There's no harm in leaving the property active. We comment out properties only if the default value will cause some detrimental effect to the otherwise normal flow of execution.

server/pxf-api/src/main/java/org/apache/hadoop/security/PxfUserGroupInformation.java
```diff
@@ -43,7 +44,8 @@ public class PxfUserGroupInformation {
     /**
      * Percentage of the ticket window to use before we renew ticket.
      */
-    private static final float TICKET_RENEW_WINDOW = 0.80f;
+    public static final String CONFIG_KEY_TICKET_RENEW_WINDOW = "pxf.service.kerberos.ticket-renew-window";
```
yimingli-vmware, line 47: Now we have `TICKET_RENEW_WINDOW_DEFAULT` as the default, will `TICKET_RENEW_WINDOW` or  `TICKET_RENEW_WINDOW_CONFIG` be a more consistent one?

server/pxf-api/src/main/java/org/apache/hadoop/security/PxfUserGroupInformation.java
```diff
@@ -197,7 +200,7 @@ public void reloginFromKeytab(String serverName, LoginSession loginSession) thro
 
     public long getKerberosMinMillisBeforeRelogin(String serverName, Configuration configuration) {
         try {
-            return 1000L * configuration.getLong(HADOOP_KERBEROS_MIN_SECONDS_BEFORE_RELOGIN, 60L);
+            return 1000L * configuration.getLong(HADOOP_KERBEROS_MIN_SECONDS_BEFORE_RELOGIN, HADOOP_KERBEROS_MIN_SECONDS_BEFORE_RELOGIN_DEFAULT);
```
yimingli-vmware, line 203: Is `HADOOP_KERBEROS_MIN_SECONDS_BEFORE_RELOGIN_DEFAULT` = 60L ?

server/pxf-api/src/main/java/org/apache/hadoop/security/PxfUserGroupInformation.java
```diff
@@ -206,6 +209,18 @@ public long getKerberosMinMillisBeforeRelogin(String serverName, Configuration c
         }
     }
 
+    public float getKerberosTicketRenewWindow(String serverName, Configuration configuration) {
+        float ticketRenewWindow = configuration.getFloat(CONFIG_KEY_TICKET_RENEW_WINDOW, TICKET_RENEW_WINDOW_DEFAULT);
+        if (ticketRenewWindow < 0f || ticketRenewWindow > 1f) {
```
bradfordb-vmware, line 214: Looking at how `ticketRenewWindow` is used in `getRefreshTime`, does it make sense to allow for a value of `0`? I don't think a value of `0` would break PXF or cause the service to loop getting new tickets but it will mean that every request to PXF results in getting a new one.

yimingli-vmware (approved):

bradfordb-vmware (approved):

server/pxf-api/src/main/java/org/apache/hadoop/security/PxfUserGroupInformation.java
ashuka24: yes

server/pxf-api/src/main/java/org/apache/hadoop/security/PxfUserGroupInformation.java
ashuka24: I used `CONFIG_KEY_TICKET_RENEW_WINDOW` because we already have that in places. I used `TICKET_RENEW_WINDOW_DEFAULT` to match `HADOOP_KERBEROS_MIN_SECONDS_BEFORE_RELOGIN_DEFAULT` but we do use `DEFAULT_<name>` in other locations of PXF, so I can update that here if needed.

server/pxf-api/src/main/java/org/apache/hadoop/security/PxfUserGroupInformation.java
ashuka24: It's not wrong per say, and maybe someone wants that exact behavior. But maybe we can give a warning when they set it to let them know what it means?

denalex: I am ok with setting it to DEFAULT (0.8) if the parameter is not explicitly provided by the user.

denalex (approved):

server/pxf-api/src/main/java/org/apache/hadoop/security/LoginSession.java
```diff
@@ -66,9 +68,10 @@ public LoginSession(String configDirectory, String principalName, String keytabP
      * @param loginUser                      UserGroupInformation for the given principal after login to Kerberos was performed
      * @param subject                        the subject
      * @param kerberosMinMillisBeforeRelogin the number of milliseconds before re-login
+     * @param kerberosTicketRenewWindow the percentage of the ticket lifespan
```
hpandeycodeit, line 71: The description here needs alignment as other params above.

server/pxf-api/src/main/java/org/apache/hadoop/security/PxfUserGroupInformation.java
```diff
@@ -43,7 +44,8 @@ public class PxfUserGroupInformation {
     /**
      * Percentage of the ticket window to use before we renew ticket.
      */
-    private static final float TICKET_RENEW_WINDOW = 0.80f;
+    public static final String CONFIG_KEY_TICKET_RENEW_WINDOW = "pxf.service.kerberos.ticket-renew-window";
+    public static final float DEFAULT_TICKET_RENEW_WINDOW = 0.8f;
```
hpandeycodeit, line 48: Is this in seconds ?

ashuka24: No, as the comment indicates, it is a percentage value. Here 0.8f means 80%. The ticket will renew when at least 80% of its lifespan has been used

PR #1003, opened 2023-07-18
Add filter pushdown support for numeric, varchar and bpchar

chanynprc:
Add filter pushdown support for numeric, varchar and bpchar

PR #1019, opened 2023-08-24
Add pushdown of NUMERIC and handling of CHAR and VARCHAR predicates for JDBC profile

denalex:
This PR is based on the original PR https://github.com/greenplum-db/pxf/pull/1003 by @chanynprc. It adds:

* propagating predicate pushdown for `NUMERIC` type from Greenplum to PXF
* support for pushdown of `CHAR`, `VARCHAR` and `NUMERIC` types in JDBC profile
* support for pushdown of `NUMERIC` type in S3-SELECT logic
* fixes to rounding of `NUMERIC` values for predicate pushdown of profiles using SARG (ORC, Hive)
* additional unit and automation tests to support updated logic
* enhancement of `system:filter` profile to return data with NULLs for testing of `IS NULL` and `IS NOT NULL` operators
* removal of test Fragmenter / Accessor / Resolver with the logic duplicate of `system:filter` profile


regression/scripts/substitute.bash
```diff
@@ -34,12 +42,17 @@ _host_is_local() {
...
-	base64 < /dev/urandom | tr -cd '[:alnum:]' | head -c "${length}"
+	if [ "$(uname)" == "Darwin" ]; then
+		# base64 on MacOS fails the command below, so instead we use the openssl library to get a random number
+		openssl rand -hex "${length}"
+	else
+		base64 < /dev/urandom | tr -cd '[:alnum:]' | head -c "${length}"
+	fi
 }
 
 _gen_uuid() {
```
bradfordb-vmware, line 53: On macOS, you can get a UUID with `uuidgen` which comes with the base system (`/usr/bin/uuidgen`) and on Linux, you can get a UUID by reading `/proc/sys/kernel/random/uuid` so something like

```sh
       if [[ $(uname) = Darwin ]]; then
               uuidgen
       else
               tr '[:lower:]' '[:upper:]' < /proc/sys/kernel/random/uuid
       fi
```

automation/tincrepo/main/pxf/features/filterpushdown/checkFilterPushDown/expected/query01.ans
```diff
@@ -1,183 +1,1517 @@
...
+ D  |  3 | f  |      | DD  | DD | a0c25s1dCo6
+ E  |  4 | t  | 4.41 |     | EE | a0c25s1dCo6
+ F  |  5 | f  | 5.51 | FF  |    | a0c25s1dCo6
+ G  |  6 | t  | 6.61 | GG  | GG | a0c25s1dCo6
+ H  |  7 | f  | 7.71 | HH  | HH | a0c25s1dCo6
+ I  |  8 | t  | 8.81 | II  | II | a0c25s1dCo6
+ J  |  9 | f  | 9.91 | JJ  | JJ | a0c25s1dCo6
+(8 rows)
+
+SELECT * FROM test_filter WHERE  t0 LIKE     'C%' ORDER BY t0, a1;
```
ashuka24, line 191: a good majority of these sql statements have extraneous spaces. is that to be expected? 

```suggestion
SELECT * FROM test_filter WHERE  t0 LIKE 'C%' ORDER BY t0, a1;
```

bradfordb-vmware (approved):

server/pxf-hdfs/src/test/java/org/greenplum/pxf/plugins/hdfs/filter/SearchArgumentBuilderTest.java
```diff
@@ -148,6 +151,75 @@ public void testNotInteger() throws Exception {
...
+        String filterString = "a7c1700s39d123456789.01234567890123456789012345678o5";
+
+        SearchArgument.Builder filterBuilder = helper(filterString, columnDescriptors);
+        assertNotNull(filterBuilder);
+        // make sure we are not losing precision as before
+        assertEquals("leaf-0 = (EQUALS col-numeric 123456789.01234567890123456789012345678), expr = leaf-0", filterBuilder.build().toString());
+    }
+
+    @Test
+    public void testNumericFilterScaleOverflow() throws Exception {
```
ashuka24, line 205: for this case, and the one below, do we want to give different hints/error messages when its scale overflow vs integer part overflow?

PR #1020, opened 2023-08-25
Add MultiNode Test Job for GP7 for dev pipeline

hpandeycodeit:
- Added variables to the Makefile (GP7_MULTINODE) and dev pipeline template (gp7_multinode), when set to true, this will introduce MultiNode tests for GP7.
- In pxf_common.bash, `grep "JAVA_HOME"` command returs two occurrences of JAVA_HOME (for Rocky8) and the script to fails. To address this issue, I have modified the script to grep for a single occurrence of JAVA_HOME: `grep "export JAVA_HOME"`
- Replaced the hard coded `centos` user with the variable (DEFAULT_CCP_USER) which contains the user based on the OS.
- `ssh-keyscan` doesn't work in Rocky8. Replaced it with awk command to set the localhost in `/home/gpadmin/.ssh/known_hosts`.
- In GP7, `gpscp` is replaced by `gpsync`. Introduced a variable (GP_SCP_CMD) to specify the appropriate command based on GP version.
- Added `gp6_python_libs` to set the `PYTHONPATH` in test_pxf_multinode.bash (same as test.bash). This is needed to run the Tinc Test.
- The command used in the script to set the locale `sudo localedef -c -i ru_RU -f CP1251 ru_RU.CP1251` fails for Rocky8. The package `glibc-locale-source` is needed to successfully set the locale and was missing from ccp image for Rocky8. Added the `yum install` command in the script to install this package if missing from the vms.
- Ran `yamllint` and `shellcheck` to confirm there are no issues with the new changes.

### DIFF
--- a/concourse/pipelines/templates/dev_build_pipeline-tpl.yml
+++ b/concourse/pipelines/templates/dev_build_pipeline-tpl.yml
@@ -73,7 +73,7 @@ anchors:
 ## ======================================================================
 resource_types:
 
-{% if multinode or multinode_no_impersonation or file or gp7_cli %}
+{% if multinode or multinode_no_impersonation or file or gp7_cli or gp7_multinode %}
 - name: terraform
   type: registry-image
   source:
@@ -83,7 +83,7 @@ resource_types:
     password: ((ud/pxf/secrets/pxf-cloudbuild-service-account-key))
 {% endif %}
 
-{% if multinode %}
+{% if multinode or gp7_multinode %}
 - name: terraform-0.14.10
   type: registry-image
   source:
@@ -169,7 +169,7 @@ resources:
     branch: ((pxf-git-branch))
     uri: ((ud/pxf/common/git-remote))
 
-{% if multinode or multinode_no_impersonation or file or gp7_cli %}
+{% if multinode or multinode_no_impersonation or file or gp7_cli or gp7_multinode %}
 - name: ccp_src
   type: git
   source:
@@ -236,7 +236,7 @@ resources:
 {% endif %}
 {% set gp_ver = None %}
 
-{% if multinode or multinode_no_impersonation or ambari or file or gp7_cli %}
+{% if multinode or multinode_no_impersonation or ambari or file or gp7_cli or gp7_multinode %}
 - name: ccp-7-image
   type: registry-image
   icon: docker
@@ -350,7 +350,7 @@ resources:
 {% endfor %}
 
 ## ---------- Auxiliary Resources ----------
-{% if multinode or multinode_no_impersonation or file or gp7_cli %}
+{% if multinode or multinode_no_impersonation or file or gp7_cli or gp7_multinode %}
 - name: terraform
   type: terraform
   source:
@@ -368,7 +368,7 @@ resources:
       bucket_path: ((tf-bucket-path))
 {% endif %}
 
-{% if multinode %}
+{% if multinode or gp7_multinode %}
 - name: terraform_ipa_hadoop
   type: terraform-0.14.10
   source:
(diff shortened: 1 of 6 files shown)


concourse/pipelines/templates/dev_build_pipeline-tpl.yml
```diff
@@ -1431,3 +1431,150 @@ jobs:
...
+      in_parallel:
+        steps:
+        - *destroy_dataproc_1
+        - *destroy_dataproc_2
+        - *destroy_gpdb_cluster
+        - *destroy_hadoop_cluster
+{% if slack_notification %}
+    <<: *slack_alert
+{% endif %}
+{% endif %} # if gp7_multinode ends here
```
ashuka24, line 1580: let's unset the `gp_ver` at the end of this block as well

concourse/scripts/pxf_common.bash
```diff
@@ -870,6 +870,8 @@ function setup_minio() {
 function set_default_ccp_user() {
     metadata_file="cluster_env_files/terraform/metadata"
 
+    # TODO: Remove the jq installation from here once available in the base image.
+    #       PR https://github.com/pivotal/gp-image-baking/pull/379
```
denalex, line 874: this is a link to a private repo, cannot be mentioned in the OSS project

concourse/scripts/test_pxf_multinode.bash
```diff
@@ -86,6 +86,8 @@ function update_pghba_conf() {
 
 function add_testing_encoding() {
 	# install new encoding and restart Greenplum so that the new encoding is picked up by Greenplum
+	# TODO: Remove the glibc-locale-source installation from here once available in CCP
+	#       PR https://github.com/pivotal/gp-cloud-image-baking/pull/150
```
denalex, line 90: this is a link to a private repo, cannot be mentioned in the OSS project

concourse/scripts/pxf_common.bash
```diff
@@ -9,6 +9,7 @@ PROXY_USER=${PROXY_USER:-pxfuser}
 PROTOCOL=${PROTOCOL:-}
 GOOGLE_PROJECT_ID=${GOOGLE_PROJECT_ID:-data-gpdb-ud}
 PXF_SRC=$(find /tmp/build -name pxf_src -type d)
+DEFAULT_CCP_USER=""
```
denalex, line 12: why `DEFAULT_CCP_USER` and not just `CCP_USER` or `CCP_OS_USER`. Default assumes it can be changed as an input parameter and if the parameter is absent, the default value will be used. But I don't think we take this as a parameter, do we ?

hpandeycodeit: Value of `.ami_default_user` is set into the `DEFAULT_CCP_USER` so the variable has `DEFAULT` in it. But sure, this can be changed to `CCP_OS_USER`

concourse/scripts/pxf_common.bash
```diff
@@ -864,3 +866,23 @@ function setup_minio() {
 	# export minio credentials as access environment variables
 	export ACCESS_KEY_ID=${MINIO_ACCESS_KEY} SECRET_ACCESS_KEY=${MINIO_SECRET_KEY}
 }
+
+function set_default_ccp_user() {
+    metadata_file="cluster_env_files/terraform/metadata"
+
+    # TODO: Remove the jq installation from here once available in the base image.
+    # Check if jq is installed
+    if ! command -v jq >/dev/null 2>&1; then
```
bradfordb-vmware, line 875: Since this is bash, we can use a short-hand here:

```bash
if ! command jq &> /dev/null; then
```

Also, there is no need for `-v` since we're redirecting into `/dev/null`.

concourse/scripts/pxf_common.bash
```diff
@@ -864,3 +866,23 @@ function setup_minio() {
...
+    fi
+
+    # Check if the metadata file exists
+    if [ -e "$metadata_file" ]; then
+        # shellcheck disable=SC2034
+        DEFAULT_CCP_USER=$(jq -r '.ami_default_user' "$metadata_file")
+    else
+        echo "The $metadata_file file does not exist."
+        exit 2
+    fi
```
bradfordb-vmware, line 887: I would re-write this as a guard clause since we are bailing-out if the metadata file doesn't exist.

```suggestion
    # Check if the metadata file exists
    if [ ! -e "$metadata_file" ]; then
        echo "The $metadata_file file does not exist."
        exit 2
    fi

    # shellcheck disable=SC2034
    DEFAULT_CCP_USER=$(jq -r '.ami_default_user' "$metadata_file")
```

bradfordb-vmware (approved):

edespino (approved):
+1

denalex (approved):

ashuka24 (approved):

concourse/scripts/pxf_common.bash
```diff
@@ -872,7 +872,7 @@ function set_ccp_os_user() {
 
     # TODO: Remove the jq installation from here once available in the base image.
     # Check if jq is installed
-    if ! command jq &> /dev/null; then
+    if ! rpm -q jq &> /dev/null; then
```
bradfordb-vmware, line 875: Any reason to not use the bash builtin `type`?

hpandeycodeit: The `command` was hanging for `RHEL7` after I removed `-v` from it. So I changed it to `rpm` and it works for both platforms. Not sure why removal of `-v` was causing problems in one vs the other.

concourse/pipelines/templates/dev_build_pipeline-tpl.yml
```diff
@@ -1431,3 +1431,151 @@ jobs:
...
+        vars:
+          gcp_project: ((ud/pxf/common/ipa-google-project-id))
+    - task: Generate Multinode Hadoop Cluster
+      file: pxf_src/concourse/tasks/install_multinode_hadoop.yml
+      image: gpdb[[gp_ver]]-pxf-dev-rocky8-image
+      params:
+        ANSIBLE_VAR_gcp_storage_bucket: ((ud/pxf/common/build-resources-bucket-name))
+        ANSIBLE_VAR_ipa_password: ((ud/pxf/secrets/ipa-password))
+        ANSIBLE_VAR_ssl_store_password: ((ud/pxf/secrets/ssl-store-password))
+  - task: Setup PXF
```
yimingli-vmware, line 1543: I'm comparing this block with our existing multi-node tests. In our existing multi-node pipeline we have a task to `setup pxf 5 latest` and then `upgrade to pxf 6` but here we only have `setup pxf`.  Which PXF version we are setting in the block? Will we also have some logic like `setup PXF 6` and then `upgrade to PXF 7`?

yimingli-vmware (approved):

concourse/pipelines/templates/dev_build_pipeline-tpl.yml
hpandeycodeit: Since we aren't setting `PXF_VERSION` the pxf version is 6 here. For now it's just PXF 6 because that is what is supported in GP7. When PXF 7 will come then we might add an upgrade job here.

PR #1022, opened 2023-08-28
Replace TINC with Go-based SQL test query runner

bradfordb-vmware:


automation/pxf_regress/main.go
```diff
@@ -0,0 +1,581 @@
...
+		fmt.Println("(using postmaster on Unix socket, default port)")
+	}
+}
+
+// Issue a command via psql, connecting to the specified database
+func psqlCommandOutput(database string, query string, a ...any) string {
+	// Generate the query with inseration of sprintf arguments
+	queryFormatted := fmt.Sprintf(query, a...)
+
+	// Now escape any shell double-quote metacharacters
```
denalex, line 371: seems we are adding a backslash if the character is either \ or " or $ or ` , correct ? if so, can we state as such ?

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,581 @@
...
+	}
+
+	status(" %8.0f ms", elapsedMillisecs(time.Now(), startTime))
+	statusEnd()
+}
+
+// count bytes in file
+func fileSize(name string) int64 {
+	fileInfo, err := os.Stat(name)
+	if err != nil {
```
denalex, line 429: can we use file existence function ?

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,581 @@
...
+		fmt.Fprintf(os.Stderr, "could not open file %q for reading: %s\n", name, err.Error())
+		return -1
+	}
+	return fileInfo.Size()
+}
+
+func clusterHealthy() bool {
+	line := psqlCommandOutput("postgres", "SELECT * FROM gp_segment_configuration WHERE status = 'd' OR preferred_role != role;")
```
denalex, line 437: are we going to run this before every test ? if yes, isn't this too heavy ?

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,581 @@
...
+	fmt.Fprintf(os.Stderr, "\n==================================\n")
+	fmt.Fprintf(os.Stderr, " Cluster validation failed.\n%s", line)
+	fmt.Fprintf(os.Stderr, "==================================\n")
+
+	return false
+}
+
+// Check the actual result file for the given test agains the expected results
+//
+// Returns true if different (failure), false if they mathc.
```
denalex, line 451:
```suggestion
// Returns true if different (failure), false if they match.
```

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,581 @@
...
+	var prettyDiffOptsBuilder strings.Builder
+	diffOptsBuilder.WriteString(basicDiffOpts)
+	prettyDiffOptsBuilder.WriteString(prettyDiffOpts)
+
+	if initFile != "" {
+		diffOptsBuilder.WriteString(" --gpd_init " + initFile)
+		prettyDiffOptsBuilder.WriteString(" --gpd_init " + initFile)
+	}
+
+	// add auto generated init file if it is generated
```
denalex, line 469: who and how will generate this ?

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,581 @@
...
+	// Name to use for temporary diff file
+	diff := fmt.Sprintf("%s.diff", resultsFile)
+
+	// Add init file arguments if provided via commandline
+	var diffOptsBuilder strings.Builder
+	var prettyDiffOptsBuilder strings.Builder
+	diffOptsBuilder.WriteString(basicDiffOpts)
+	prettyDiffOptsBuilder.WriteString(prettyDiffOpts)
+
+	if initFile != "" {
```
denalex, line 464: is this for global init file or smth else ?

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,581 @@
...
+	generatedInitFile := fmt.Sprintf("%s.initfile", resultsFile)
+	if fileExists(generatedInitFile) {
+		diffOptsBuilder.WriteString(" --gpd_init " + generatedInitFile)
+		prettyDiffOptsBuilder.WriteString(" --gpd_init " + generatedInitFile)
+	}
+
+	// ok run the diff
+	// pg_regress compares the expected test output to the actual test output
+	// tinc compares the actual test output to the expected test output
+	// I would expect diff A B == diff B A and it is except for HcfsGlobbingTest#testMatchAStringFromStringSet
```
denalex, line 479: is the order important for writing match/subs ? Are match/subs applied to the expected file or actual file or both before comparison ?

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,581 @@
...
+			if exiterr.ExitCode() > 1 {
+				fatalf("diff command failed: %s\n", err.Error())
+			}
+			return exiterr.ExitCode()
+		}
+		fatalf("diff command failed: %s\n", err.Error())
+	}
+
+	// if err is nil, the command exited with a zero exit status
+	return 0
```
denalex, line 521: can we use an early return to get rid of 1 level of indentation ?

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,581 @@
...
+	sqlDir := path.Join(testDir, "sql")
+	if !directoryExists(sqlDir) {
+		return nil, fmt.Errorf("directory %s does not exist", sqlDir)
+	}
+
+	sqlFiles, err := filepath.Glob(path.Join(sqlDir, "*.sql"))
+	if err != nil {
+		return nil, err
+	}
+
```
denalex, line 572: are we stripping .sql extension from the file names ? can we state as such ?

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,581 @@
...
+	n := len(".sql")
+	tests := make([]string, len(sqlFiles))
+	for i, sqlFilePath := range sqlFiles {
+		sqlFileBase := path.Base(sqlFilePath)
+		tests[i] = sqlFileBase[0 : len(sqlFileBase)-n]
+	}
+
```
denalex, line 579: should we order the tests ? can they be run not in alphanumeric order ?

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,581 @@
...
+	err := flagSet.Parse(args[1:])
+	if err != nil {
+		return err
+	}
+
+	if initFile == "" {
+		if initFile, err = findOtherFile("global_init_file"); err != nil {
+			return err
+		}
+	}
```
denalex, line 120: seem like a given init file will overwrite a global init file, is it good ?

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,581 @@
...
+	// Emit nice-looking summary message
+	summary := results.Summary()
+	n := len(summary)
+
+	fmt.Println()
+	fmt.Println(strings.Repeat("=", n))
+	fmt.Println(summary)
+	fmt.Println(strings.Repeat("=", n))
+	fmt.Println()
+	fmt.Println()
```
denalex, line 159: extract into `printSummary` function ?

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,581 @@
...
+	} else {
+		return fmt.Sprintf(" %d of %d tests failed. ", t.failCount, t.successCount+t.failCount)
+	}
+
+}
+
+func init() {
+	stdout = bufio.NewWriter(os.Stdout)
+}
+
```
denalex, line 71: Can we document the API that this utility implements - order and meaning of the arguments and environment values it expects or reads ?

automation/src/main/java/org/greenplum/pxf/automation/components/regress/Regress.java
```diff
@@ -0,0 +1,72 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Arrays;
+import java.util.Objects;
+import java.util.StringJoiner;
+import java.util.stream.Collectors;
+
+public class Regress extends ShellSystemObject {
```
denalex, line 18: can we Javadoc this please to explain what this is for

automation/src/main/java/org/greenplum/pxf/automation/components/regress/Regress.java
```diff
@@ -0,0 +1,72 @@
...
+    public void init() throws Exception {
+        ReportUtils.startLevel(report, getClass(), "init");
+        regressRunner = new File("pxf_regress/pxf_regress").getAbsolutePath();
+        super.init();
+        runCommand("source $GPHOME/greenplum_path.sh");
+        runCommand("cd " + new File(regressTestFolder).getAbsolutePath());
+        ReportUtils.stopLevel(report);
+    }
+
+    public void runRegressTest(final String tincTestModule) throws IOException, ShellCommandErrorException {
```
denalex, line 32:
```suggestion
    public void runRegressTest(final String regressTestModule) throws IOException, ShellCommandErrorException {
```
RIP tinc :)

automation/src/test/java/org/greenplum/pxf/automation/BaseTestParent.java
```diff
@@ -224,9 +231,14 @@ public void runAfterMethod() throws Exception {
      */
     protected void runTincTest(String tincTest) throws Exception {
         try {
-            tinc.runTest(tincTest);
+            if (TincUtils.useTinc){
+                tinc.runTest(tincTest);
+            } else {
+                regress.runRegressTest(tincTest);
```
denalex, line 237:
```suggestion
                regress.runTest(tincTest);
```
let's change the function name for symmetry

automation/src/test/java/org/greenplum/pxf/automation/BaseTestParent.java
```diff
@@ -224,9 +231,14 @@ public void runAfterMethod() throws Exception {
...
+            }
         } catch (Exception e) {
-            throw new Exception("Tinc Failure (" + e.getMessage() + ")");
+            String message = TincUtils.useTinc ? "Tinc Failure (" + e.getMessage() + ")" : "Regress Failure (" + e.getMessage() + ")";
+            throw new Exception(message);
```
denalex, line 241:
```suggestion
            throw new Exception(String.format("%s Failure (%s)", TincUtils.useTinc ? "Tinc" : "Regress", e.getMessage()));
```

automation/pxf_regress/global_init_file
```diff
@@ -0,0 +1,82 @@
+# Global matchsubs
```
ashuka24, line 1: this file looks like an exact copy of the one present in `automation/tinc/main/tinctest/lib/`. 

Will we be expected to keep both in sync? Is there a way that we can just have 1 `global_init_file`?

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,581 @@
...
+// gpdiff.pl options for comparing actual results file with expected
+//
+// Unlinke postgres/pg_regress, PXF includes
+//
+//	-w ignore all white space
+//	-B ignore changes lines are all blank
+//
+// TODO: rather than match/sub DETAIL (GP5) for CONTEXT (see global_init_file), should we add "-I DETAIL:" and "-I CONTEXT:"
+// TODO: rathern than having to add start_ignore/end_ignore, should we add "-I HINT:"
```
ashuka24, line 32:
```suggestion
// TODO: rather than having to add start_ignore/end_ignore, should we add "-I HINT:"

File: automation/pxf_regress/main.go [hunk 20]
```diff
@@ -0,0 +1,581 @@
...
+func psqlTest(testname string) (*exec.Cmd, string, string) {
+	inFile := fmt.Sprintf("%s/sql/%s.sql", testDir, testname)
+	outFile := fmt.Sprintf("%s/results/%s.out", testDir, testname)
+	expectFile := fmt.Sprintf("%s/expected/%s.ans", testDir, testname)
+
+	var psqlCmd strings.Builder
+
+	psqlCmd.WriteString(fmt.Sprintf("psql %s -X -a -f %s > %s 2>&1", dbName, inFile, outFile))
+
+	os.Setenv("PGAPPNAME", "pxf_regress/"+testname)
```
ashuka24, line 94: should this be `PG_APP_NAME` for consistency with other environment variables?

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,581 @@
...
+	f, err := os.Stat(dir)
+	if err != nil {
+		return false
+	}
+
+	return f.IsDir()
+}
+
+func makeDirectory(dir string) {
+	if err := os.Mkdir(dir, 0777); err != nil {
```
ashuka24, line 194: if I understand correctly, we are making a directory with 0777 permissions, which is read, write and execute permissions to all users. Is there a reason we are giving 0777 instead of a more restrictive set of permissions?

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,581 @@
...
+		fatalf("%s: could not create directory %q: %s", progName, dir, err.Error())
+	}
+}
+
+// Spawn an external command to execute the given shell command; don't wait for it
+//
+// Returns a reference to the command
+func spawnProcess(cmdLine string) *exec.Cmd {
+	// Have the shell "exec" the command to save useless processes per test run
```
ashuka24, line 203: if the processes are useless, why do we have to save them?

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,581 @@
...
+	// Trim off program name and keep just directory
+	cmdPath := path.Join(path.Dir(progName), target)
+
+	// Ensure that the file exists and is a regular file
+	fileInfo, err := os.Stat(cmdPath)
+	if err != nil {
+		return "", fmt.Errorf("check for %q failed: %s", cmdPath, err.Error())
+	}
+
+	if !fileInfo.Mode().IsRegular() {
```
ashuka24, line 242: what does it mean for a file to be "regular"

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,581 @@
...
+		return "", fmt.Errorf("check for %q failed: not a regular file", cmdPath)
+	}
+
+	// ensure that file is both  executable and readable
```
ashuka24, line 246:
```suggestion
	// ensure that file is both executable and readable

File: automation/pxf_regress/main.go [hunk 25]
```diff
@@ -0,0 +1,581 @@
...
+func findOtherFile(target string) (string, error) {
+	filePath := path.Join(path.Dir(progName), "global_init_file")
+
+	fileInfo, err := os.Stat(filePath)
+	if err != nil {
+		return "", fmt.Errorf("check for %q failed: %s", filePath, err.Error())
+	}
+
+	if !fileInfo.Mode().IsRegular() {
+		return "", fmt.Errorf("check for %q failed: not a regular file", filePath)
```
ashuka24, line 268: this looks like a duplicate of lines 238-243, can we pull these checks out?

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,581 @@
...
+
+// helper-types
+
+type testResults struct {
+	successCount int
+	failCount    int
+}
+
+func (t *testResults) Summary() string {
+	if t.failCount == 0 {
```
ashuka24, line 60: My understanding is that it's possible to have skipped tests. Do we need to include that count here?

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,581 @@
...
+}
+
+// Prepare environment variables for running regression tests
+func initializeEnvironment() {
+	// Set default application name. The test function may choose to
+	// override this, but if it doesn't, we have something useful in place.
+	os.Setenv("PGAPPNAME", "pxf_regress")
+
+	// Set translation-related settings to English; otherwise psql will
+	// produce translated messages and produce diffs.  (XXX If we ever support
```
ashuka24, line 321: we should use TODO instead of XXX

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,581 @@
...
+	os.Unsetenv("LC_ALL")
+	os.Unsetenv("LC_MESSAGES=C")
+
+	os.Unsetenv("PGCLIENTENCODING")
+
+	// Set timezone and datestyle for datetime-related tests
+	//
+	// Unlike postgres/pg_regress, PXF's existing expected test outputs
+	// have date and timestamps formatted with "ISO, MDY" instead of
+	// "Postgres, MDY". Additionally, we do not explicityly set a time zone
```
ashuka24, line 334:
```suggestion
	// "Postgres, MDY". Additionally, we do not explicitly set a time zone

File: automation/pxf_regress/main.go [hunk 29]
```diff
@@ -0,0 +1,581 @@
...
+	} else if pgHost == "" && pgPort != "" {
+		fmt.Printf("(using postmaster on Unix socket, port %s)\n", pgPort)
+	} else {
+		fmt.Println("(using postmaster on Unix socket, default port)")
+	}
+}
+
+// Issue a command via psql, connecting to the specified database
+func psqlCommandOutput(database string, query string, a ...any) string {
+	// Generate the query with inseration of sprintf arguments
```
ashuka24, line 368:
```suggestion
	// Generate the query with insertion of sprintf arguments

File: automation/pxf_regress/gpdiff.pl [hunk 30]
```diff
@@ -0,0 +1,548 @@
+#!/usr/bin/env perl
```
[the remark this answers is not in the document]
bradfordb-vmware: This isn't copied from GPDB -- it is a copy of `gpdiff.pl` from `automation/tinc/main/gpdiff.pl` so that `pxf_regress` is using the same utility as tinc for comparing the actual test output with the expected test output.

automation/pxf_regress/main.go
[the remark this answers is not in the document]
bradfordb-vmware: Yes. The current set of tinc tests invoke `gpdiff.pl` with this exact set of options. For this PR, I wanted to minimize the difference in behavior of `tinc` and `pxf_regress`. These `TODOs` are potential performance improvements once we are confident in the behavior of the new tool.

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,581 @@
...
+
+	if results.failCount != 0 {
+		os.Exit(1)
+	}
+
+	return nil
+}
+
+func fileExists(file string) bool {
+	f, err := os.Open(file)
```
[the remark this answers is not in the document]
bradfordb-vmware: Short-anwer: this function was ported from `pg_regress` and that is how it is written in `pg_regress`.

It turns out that we don't need this function -- it was called in one place and would always return false so we we just remove it.

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,581 @@
...
+	if err != nil && errors.Is(err, fs.ErrNotExist) {
```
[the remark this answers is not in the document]
bradfordb-vmware: From the godoc for the os package:

> This function predates errors.Is. It only supports errors returned by the os package. New code should use errors.Is(err, fs.ErrNotExist).

automation/pxf_regress/main.go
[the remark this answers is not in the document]
bradfordb-vmware: This function is ported/copied (including comments) from `pg_regress`. The high-level idea is that without `exec`, you'd have a process tree that looks like:

```
pxf_regress
└── sh
    └── cmd
```

Since `exec` replaces the shell with the given command, you end up with a process tree of

```
pxf_regress
└── cmd
```

For `pg_regress` which runs several tests in parallel, this can help reduce the number of processes that are created. For PXF's use case, this "optimization" might not be providing much value.

The original goal when porting/copying was to minimize the conceptual differences between `pg_regress` and `pxf_regress`. I can remove the use of `exec` (and the comment) since `pxf_regress` probably doesn't get much benefit from it.

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,581 @@
...
+	// Unlike postgres/pg_regress, PXF's existing expected test outputs
+	// have date and timestamps formatted with "ISO, MDY" instead of
+	// "Postgres, MDY". Additionally, we do not explicityly set a time zone
+	// for pxf_regress because this can cause pxf_regress and the PXF
+	// service to be in different time zones, resulting in spurious diff;
+	// instead we let pxf_regress and the PXF service use the environments
+	// set time zone.
+	os.Setenv("PGDATESTYLE", "ISO, MDY")
+
+	// Likewise, set intervalstyle to ensure consistent reuslts. this is a bit
```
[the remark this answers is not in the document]
bradfordb-vmware: This is taken from `pg_regress` -- this is for setting the output format of the `interval` type; there are [four supported styles](https://www.postgresql.org/docs/current/datatype-datetime.html#DATATYPE-INTERVAL-OUTPUT):

1. sql_standard
2. postgres
3. postgres_verbose
4. iso_8601

Setting this is similar to setting the `datestyle` format -- remove spurious test failures that result from different contributors using different locales/database settings.

automation/pxf_regress/main.go
bradfordb-vmware: Not really. The file `regression.diffs` is created at the start of the test (in `openResultFiles`). If a test fails (i.e., has a difference between actual and expected test output), then its diff is appended to `regression.diffs`. So `regression.diffs` will always exists. If its size is 0, then that means no test failed.

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,581 @@
...
+
+	// Create the diffs file as empty
+	diffFileName = fmt.Sprintf("%s/regression.diffs", testDir)
+	f, err = os.Create(diffFileName)
+	if err != nil {
+		fatalf("could not open file %q for writing: %s", logFileName, err.Error())
+	}
+	diffFile = f
+	// We don't keep the diffs file open continuously
+	diffFile.Close()
```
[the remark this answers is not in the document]
bradfordb-vmware: This is for the file `regression.diffs` that tests will append their diffs to if the test fails.

automation/pxf_regress/main.go
bradfordb-vmware: Yes -- this matches what is done in GPDB's version of `pg_regress`. However, I am a little skeptical at its utility (in GPDB's `pg_regress`) since I don't think it will actually result in a failure of if the cluster is ever unhealthy. I included it for the sake of trying to match the behavior of `pg_regress`. If we are OK deviating, then I have no objection to removing it.

automation/pxf_regress/main.go
bradfordb-vmware: We don't so I've removed this.

automation/pxf_regress/main.go
bradfordb-vmware: Both. If the user of `pxf_regress` calls it with `-init-file=some_init_file`, then we will use that in place of `global_init_file`. If the user doesn't provide an init file, then we will use `global_init_file` in the same directory as `pxf_regress`.

automation/pxf_regress/main.go
bradfordb-vmware: Don't know -- match/subs are handled `gpdiff.pl` which is what we are using with `tinc` too.

automation/pxf_regress/main.go
bradfordb-vmware: This is how the function would look if we did that:

```go
func runDiff(cmd string, filename string) int {
	c := exec.Command("sh", "-c", cmd)
	err := c.Run()
	// if err is nil, the command exited with a zero exit status
	if err == nil {
		return 0
	}
	// if the command starts but does not complete successfully,
	// the error is of type *ExitError
	if exiterr := (&exec.ExitError{}); errors.As(err, &exiterr) {
		if exiterr.ExitCode() > 1 {
			fatalf("diff command failed: %s\n", err.Error())
		}
		return exiterr.ExitCode()
	}
	fatalf("diff command failed: %s\n", err.Error())
}
```

Unfortunately the compiler now complains about a missing `return` at the end of the function even though `fatalf` doesn't return. So we'd have to add a fake `return 0` at the end of the function to make the compiler happy. I'm not sure if the benefit of removing a single-level of indentation for a short function is worth finding a way to make the compiler happy.

automation/pxf_regress/main.go
bradfordb-vmware: Good catch -- I just assumed that `filepath.Glob` would return in lexicographic order but there is nothing in its documentation comment that says this. I've added an explicit sort to ensure a consistent ordering of tests.

automation/src/main/java/org/greenplum/pxf/automation/components/regress/Regress.java
bradfordb-vmware: I intentionally named this `tincTestModule` because the argument as we currently give it is a Python module name (e.g., `pxf.smoke.small_data.runTest`). pxf_regress works with directories so we need to convert the tinc Python module name into a directory name. This logic is contained in this class and will be dropped when we fully remove `tinc` in a subsequent PR.

automation/pxf_regress/main.go
bradfordb-vmware: It may be possible to have skipped tests in `pg_regress` but `pxf_regress` does not have support for skipping/ignoring tests.

automation/pxf_regress/main.go
bradfordb-vmware: No. This is a [Postgres Environment Variable](https://www.postgresql.org/docs/current/libpq-envars.html).

automation/pxf_regress/main.go
bradfordb-vmware: The `global_init_file` contains common match/subs that we want to use for all test cases. Nominally, we'll just want to use this one by default. The `-init-file` is there if a user wants to use a different set of match/subs without making changes to `global_init_file`. We could have the argument of `-init-file` be appended as an additional set of match/subs but then a user cannot disable a match/sub in `global_init_file` without editing it.

Maybe a better name is `default_match_subs` instead of `global_init_file`?

automation/pxf_regress/main.go
bradfordb-vmware: That's mostly correct, but the permission bits are before umask is applied. On both macOS and Linux, the default umask is `022` so the above call to `Mkdir` will end up creating a directory with 0755.

automation/pxf_regress/main.go
bradfordb-vmware: A regular file is probably what you normally think of as a file -- a text, an executable, an image, etc. In UNIX, everything is a file (directories, pipes, sockets, etc) so we need to differentiate "regular" files from other files that are not.

automation/pxf_regress/main.go
bradfordb-vmware: This is copied verbatim from `pg_regress` source.

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,581 @@
...
+
+	return string(output)
+}
+
+func runSingleTest(test string) {
+	if !clusterHealthy() {
+		return
+	}
+
+	status("test %-28s ...", test)
```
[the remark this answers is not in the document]
bradfordb-vmware: It pads up to the given width; if the string is longer than the desired width, no padding is added.

automation/pxf_regress/main.go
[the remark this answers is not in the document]
bradfordb-vmware: Nope. But I verified that if we reverse the oder of the diff in tinc, this test fails too.

yimingli-vmware (approved):

automation/src/main/java/org/greenplum/pxf/automation/components/regress/Regress.java
```diff
@@ -29,21 +26,39 @@ public void init() throws Exception {
...
+     */
+    public void runTest(final String tincTestModule) throws IOException, ShellCommandErrorException {
         ReportUtils.startLevel(report, getClass(), "Run test: " + tincTestModule);
         String testPath = mapToRegressDir(tincTestModule);
         ReportUtils.report(report, getClass(), "test path: " + testPath);
 
         setCommandTimeout(_10_MINUTES);
         StringJoiner commandToRun = new StringJoiner(" ");
 
+        commandToRun.add("PGDATABASE=pxfautomation");
```
denalex, line 43: can we avoid hardcoding the database name here ? I recall we might be using another database with different locale as part of the test suite, wonder how would it work in that case

automation/src/main/java/org/greenplum/pxf/automation/components/regress/Regress.java
```diff
@@ -29,21 +26,39 @@ public void init() throws Exception {
...
         commandToRun.add(regressRunner);
         commandToRun.add(testPath);
```
denalex, line 45: should we log the command we're trying to run (for potential troubleshooting) ?

automation/src/main/java/org/greenplum/pxf/automation/components/regress/Regress.java
```diff
@@ -52,20 +67,6 @@ public String mapToRegressDir(String tincTestPath) {
         );
     }
 
-    public String listSqlQueries(String testPath) {
```
denalex: did we not use this method before ? It is removed but I do not see any corresponding usage changes

automation/src/test/java/org/greenplum/pxf/automation/BaseTestParent.java
```diff
@@ -224,9 +231,14 @@ public void runAfterMethod() throws Exception {
      */
     protected void runTincTest(String tincTest) throws Exception {
```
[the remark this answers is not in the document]
denalex: +1

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,427 @@
...
+		fatalf("%s: missing required argument: test-directory", progName)
+	}
+
+	f, err := findOtherFile("gpdiff.pl", true)
+	if err != nil {
+		fatalf("%s: cannot find gpdiff.pl: %s", progName, err.Error())
+	}
+	gpdiffProg = f
+
+	f, err = findOtherFile("global_init_file", false)
```
denalex, line 71:
```suggestion
	f, err = findFile("global_init_file", false)
```
should we simplify the function name  now that "executability" requirement is a parameter ?

automation/src/test/java/org/greenplum/pxf/automation/BaseTestParent.java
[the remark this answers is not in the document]
bradfordb-vmware: Generally I would agree with such a change but that would introduce a lot of diffs to this PR and I don't want to distract from the more important aspects of this change.

I was planning to rename this to either `runTest` or `runAndCompareSQL` or `runSQLTest` on a follow-up PR when we remove tinc.

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,427 @@
...
+func init() {
+	stdout = bufio.NewWriter(os.Stdout)
+}
+
+func main() {
+	pathName, err := os.Executable()
+	if err != nil {
+		fatalf(err.Error())
+	}
+	progName = pathName
```
denalex, line 59: why can we not have assigned it in the first place:
```
progName, err := os.Executable()
``` 
same questions for other variables below assigned only after the error check

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,427 @@
...
+
+	if len(os.Args) < 2 {
+		fatalf("%s: missing required argument: test-directory", progName)
+	}
+
+	f, err := findOtherFile("gpdiff.pl", true)
+	if err != nil {
+		fatalf("%s: cannot find gpdiff.pl: %s", progName, err.Error())
```
denalex, line 67: Is there a case where we call `findOtherFile` and do not want to fail on error ? If not, then maybe we can check for error and fail inside the function other that returning the error and having the caller perform similar checks.

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,427 @@
...
+	if err != nil {
+		fatalf("%s: cannot find 'global_init_file': %s", progName, err.Error())
+	}
+	initFile = f
+
+	testDir = os.Args[1]
+	tests, err := listTestQueries(testDir)
+	if err != nil {
+		fatalf("%s: cannot list test queries in %s: %s", progName, testDir, err.Error())
+	}
```
denalex, line 81: this would be good to check first when we look at the number of arguments

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,427 @@
...
+
```
denalex, line 82: this whole section above can be moved to `validateArguments()` helper method not to distract from the main logic

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,427 @@
...
+	openResultFiles()
+	initializeEnvironment()
+
+	// Ready to run the tests
+	header("running regression queries")
```
denalex, line 87:
```suggestion
	printHeader("running regression queries")
```
for consistency

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,427 @@
...
+	if regressionSummary != nil {
+		regressionSummary.Close()
+	}
+
+	// Emit nice-looking summary message
+	printSummary(stdout, successCount, failCount)
+
+	if failCount > 0 {
+		fmt.Printf("The differences that caused some tests to fail can be viewed in the\nfile %q. A copy of\nthe test summary that you see above is saved in the file\n%q.\n\n", regressionDiffsName, regressionSummaryName)
+		os.Exit(1)
```
denalex, line 102: can these 2 lines be replaced by `fatalf` ?

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,427 @@
...
+	}
+
+	os.Remove(regressionDiffsName)
+	os.Remove(regressionSummaryName)
```
denalex, line 106:
```suggestion
	deleteResultsFiles()
```
can we move this to a helper function for symmetry with `openResultFiles()`

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,427 @@
...
+	}
+	initFile = f
+
+	testDir = os.Args[1]
+	tests, err := listTestQueries(testDir)
+	if err != nil {
+		fatalf("%s: cannot list test queries in %s: %s", progName, testDir, err.Error())
+	}
+
+	openResultFiles()
```
denalex, line 83:
```suggestion
	createResultFiles()
```
I think we create the files, not necessarily keeping them open ?

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,427 @@
...
+	return filePath, nil
+}
+
+func IsReadOwner(mode fs.FileMode) bool {
+	return mode&0400 != 0
+}
+
+func IsExecOwner(mode fs.FileMode) bool {
+	return mode&0100 != 0
+}
```
denalex, line 194: can these 2 functions be just inlined if used in a single place ?

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,427 @@
...
+	}
+	regressionDiffs = f
+	// We don't keep the diffs file open continuously
+	regressionDiffs.Close()
+
+	// Also create the results directory if not present
+	resultsDir := fmt.Sprintf("%s/output", testDir)
+	if !directoryExists(resultsDir) {
+		makeDirectory(resultsDir)
+	}
```
denalex, line 220: can these be combined into a single function `createDirectory()` that will internally check if exists or not ?

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,427 @@
...
+		fmt.Printf("(using postmaster on %s, default port)\n", pgHost)
+	} else if pgHost == "" && pgPort != "" {
+		fmt.Printf("(using postmaster on Unix socket, port %s)\n", pgPort)
+	} else {
+		fmt.Println("(using postmaster on Unix socket, default port)")
+	}
+}
+
+func runSingleTest(test string) {
+	status("test %-28s ...", test)
```
denalex, line 264:
```suggestion
	printStatus("test %-28s ...", test)
```

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,427 @@
...
+	startTime := time.Now()
+	resultFile, expectFile := runPsql(test)
+
+	// Compare actual results file with expected
+	diff := resultsDiffer(resultFile, expectFile)
+	if diff {
```
denalex, line 270:
```suggestion	
	if resultsDiffer(resultFile, expectFile) {
```

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,427 @@
...
+		status("FAILED")
+		failCount += 1
+	} else {
+		status("ok")
+		successCount += 1
+	}
+
+	status(" %8.0f ms", elapsedMillisecs(time.Now(), startTime))
+	statusEnd()
```
denalex, line 279:
```suggestion
	printStatusEnd()
```

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,427 @@
...
+// Returns true if different (failure), false if they match.
+// In the true case, the diff is appended to the diffs file.
+func resultsDiffer(resultsFile string, expectFile string) bool {
+	diffOpts := baseDiffOpts
+	if initFile != "" {
+		diffOpts = append(diffOpts, "--gpd_init", initFile)
+	}
+	diffOpts = append(diffOpts, resultsFile, expectFile)
+
+	cmd := exec.Command(gpdiffProg, diffOpts...)
```
denalex, line 293: we should try to log the command we run to compare files

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,427 @@
...
+	if err != nil {
+		fatalf("%s: could not open file to write %q: %s", progName, regressionDiffsName, err.Error())
+	}
+
+	// append to summary 'results.diff'
+	summaryDiff, err := os.OpenFile(regressionDiffsName, os.O_APPEND|os.O_WRONLY, 0644)
+	if err != nil {
+		fatalf("%s: could not open file to write %q: %s", progName, regressionDiffsName, err.Error())
+	}
+	defer summaryDiff.Close()
```
denalex, line 325: what does this do ?

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,427 @@
...
+	if err != nil {
+		fatalf("%s: unable to create file %s: %s", progName, outputFile, err.Error())
+	}
+	defer f.Close()
+
+	cmd := exec.Command("psql", "-X", "-a", "-f", inputSql)
+	cmd.Env = append(os.Environ(), "PGAPPNAME=pxf_regress/"+testname)
+	cmd.Stdout = f
+	cmd.Stderr = f
+	err = cmd.Run()
```
denalex, line 125: would be good to log the command being run

automation/pxf_regress/main.go
bradfordb-vmware: > defers the execution of a function until the surrounding function returns; the deferred call's arguments are evaluated immediately, but the function call is not executed until the surrounding function returns.
>
> [A Tour of Go - Defer](https://go.dev/tour/flowcontrol/12)

In this case, this closes the summary diff file when the function `resultsDiffer` returns.

automation/pxf_regress/main.go
bradfordb-vmware: Since this tool will primarily be run by automation, I'm inclined to remove `header`, `status`, and `statusEnd` altogether and just write detailed log messages to stdout, instead of overly complicating this with "human-readable" output on stdout and creating a separate `pxf_regress.log` for detail logging. The majority of the time, the output of this tool will never be viewed by a human. As an added bonus, this means that all of the log messages from pxf_regress will be captured in automation's logs -- currently, tinc log messages are in a separate file.

automation/pxf_regress/main.go
bradfordb-vmware: They can be but I thought that having a helper function named `IsRedOwner` was easier to understand than a random bitwise AND (`mode&0400`). I suppose a short comment before the `if`-statement would suffice.

automation/pxf_regress/main.go
bradfordb-vmware: The call to `listTestQueries` has been moved to right after we check the length of args.

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,581 @@
...
+	if err != nil && errors.Is(err, fs.ErrNotExist) {
+		return false
+	}
+	f.Close()
+	return true
+}
+
+func directoryExists(dir string) bool {
+	f, err := os.Stat(dir)
+	if err != nil {
```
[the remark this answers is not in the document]
bradfordb-vmware: I've un-nested this code and hopefully the error handling is cleaner.

automation/src/test/java/org/greenplum/pxf/automation/BaseTestParent.java
```diff
@@ -90,7 +93,11 @@ public final void doInit() throws Exception {
             File localDataTempFolder = new File(dataTempFolder);
             localDataTempFolder.mkdirs();
             // Initialize Tinc System Object
-            tinc = (Tinc) systemManager.getSystemObject("tinc");
+            if (TincUtils.useTinc) {
+                tinc = (Tinc) systemManager.getSystemObject("tinc");
+            } else {
+                regress = (Regress) systemManager.getSystemObject("regress");
```
hpandeycodeit, line 99:
```suggestion
                // Initialize Regress System Object
                regress = (Regress) systemManager.getSystemObject("regress");
```

automation/src/test/java/org/greenplum/pxf/automation/features/hdfs/HdfsReadableTextTest.java
```diff
@@ -590,6 +591,9 @@ public void errorTable() throws Exception {
         dataTable.addRow(new String[]{"10", "ten - I love you!"});
 
         hdfs.writeTableToFile(hdfsFilePath, dataTable, ",");
+        if (protocol != ProtocolEnum.HDFS) {
+            sleep(10000);
```
hpandeycodeit, line 595: What is the need of `sleep` function here?

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,381 @@
...
+}
+
+func directoryExists(dir string) bool {
+	f, err := os.Stat(dir)
+	if err == nil {
+		return f.IsDir()
+	}
+
+	if !errors.Is(err, fs.ErrNotExist) {
+		logger.Fatalf("could not stat directory %q: %s", dir, err.Error())
```
hpandeycodeit, line 136: Here the error means this dir doesn't exist ?

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,381 @@
...
+
+	return false
+}
+
+func createDirectory(directory string) {
+	if directoryExists(directory) {
+		return
+	}
+
+	if err := os.Mkdir(directory, 0777); err != nil {
```
hpandeycodeit, line 147: can we put a comment here on what `0777` means or have it in a variable?

automation/pxf_regress/main.go
bradfordb-vmware: I have re-written this to un-nest and still have a `return` at the end of the function.

automation/pxf_regress/main.go
bradfordb-vmware: After discussing offline, I've removed the flag for setting a different global_init_file since we don't currently have a use-case for it. If/when we find a use for it, it can be added back easily.

automation/pxf_regress/main.go
bradfordb-vmware: This is an error of Go where I'm not well-versed -- the `:=` assignment operator is generally used for declaring and assigning variables at the same time:

```go
var foo string
foo = "Hello, World!"

foo := "Hello, World!"
```

So I think doing `progName, err := os.Executable()` would introduce a new local variable that shadows the global instead of assigning. It would probably have been clearer to do

```go
var err error
progName, err = os.Executable()
```

but this snippet of code has been refactored so its probably moot now.

automation/pxf_regress/main.go
bradfordb-vmware: I've updated our functions to not return and error, since we would just `fatal`-out in the caller. Now the callee will exit if it detects an error.

automation/pxf_regress/main.go
bradfordb-vmware: I have removed the functions by inlining them (w/ appropriate comments).

automation/pxf_regress/main.go
bradfordb-vmware: We called `os.Stat` which should return info about a directory but its returned and error and the error is not that the directory doesn't exist but some other error.

automation/src/test/java/org/greenplum/pxf/automation/features/hdfs/HdfsReadableTextTest.java
bradfordb-vmware: When running the 'file:*' tests in the pipeline, we had a test that consistently failed with missing data and it was always missing the last two or three lines of data. What was happening is that automation (running in a CI container) was generating some test data and writing it to a file in a directory that was mounted via NFS. As soon as the data was written, it ran the SQL to query that data via PXF. I'm assuming it took some time for the data to go from the CI container, to the NFS server, and then out to the GPDB segments. With `tinc`, since it is using reflection and is written in python maybe it was slow enough that the data was always written by time the SQL ran.

This sleep adds a little delay to ensure all of the generated data is synced before we try to query it.

automation/pxf_regress/main.go
```diff
@@ -318,239 +195,115 @@ func initializeEnvironment() {
...
-	statusEnd()
-}
-
-// count bytes in file
-func fileSize(name string) int64 {
-	fileInfo, err := os.Stat(name)
-	if err != nil {
-		fmt.Fprintf(os.Stderr, "could not open file %q for reading: %s\n", name, err.Error())
-		return -1
+		outcome = "ok"
```
ashuka24, line 249: why is failed in caps, but ok is in lower case, can we keep it consistent?

automation/src/test/resources/sut/MultiHadoopMultiNodesCluster.xml
```diff
@@ -108,6 +108,7 @@
     <regress>
         <class>org.greenplum.pxf.automation.components.regress.Regress</class>
         <regressTestFolder>tincrepo/main</regressTestFolder>
+        <dbName>pxfautomation</dbName>
```
ashuka24, line 111: why is this needed now?

bradfordb-vmware: For the tinc tests, if you look at the `runTest.py`, there is an annotation of `@db_name` that tinc used to determine which database to connect to and run the SQL against. For every test, the value given was `pxfautomation`. For the new SQL runner, we need to set the name of the database to connect to. Originally this was done via a flag (e.g., `-db-name=pxfautomation`) but that required extra code for setting up and parsing the flags and since we only had the one flag (database name), I decided that for now it would simpler to just use the existing Postgres environment variables for configuring `psql`'s connection.

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,369 @@
...
+
+// Prepare environment variables for running regression tests
+func initializeEnvironment() {
+	// Set default application name. The test function may choose to
+	// override this, but if it doesn't, we have something useful in place.
+	os.Setenv("PGAPPNAME", "pxf_regress")
+
+	// Set translation-related settings to English; otherwise psql will
+	// produce translated messages and produce diffs.
+	os.Unsetenv("LANGUAGE")
```
hpandeycodeit, line 199: May be move all the `Unsetenv` before `Setenv` for better readability.

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,369 @@
...
+	os.Unsetenv("LC_ALL")
+	os.Setenv("LC_MESSAGES", "C")
+
+	os.Unsetenv("PGCLIENTENCODING")
```
hpandeycodeit, line 203: Where these variables are set?

automation/pxf_regress/main.go
```diff
@@ -391,37 +337,33 @@ func listTestQueries(testDir string) ([]string, error) {
...
+	microsecond := t.Nanosecond() / nanosecondsPerMicrosecond
 	formatted := fmt.Sprintf("%04d%02d%02d%02d%02d%02d%06d",
 		t.Year(), t.Month(), t.Day(),
 		t.Hour(), t.Minute(), t.Second(), microsecond)
 
 	return formatted
 }
+
+func deleteResultsFiles() {
+	os.Remove(regressionDiffsName)
```
denalex, line 368: I guess we're ignoring the error if the file to be deleted does not exist ?

denalex (approved):

automation/pxf_regress/main.go
bradfordb-vmware: This variable isn't set anywhere, we're unsetting/removing (if it is set in some user's environment) from this processes environment so that spurious diffs aren't generated.

automation/pxf_regress/main.go
```diff
@@ -114,11 +88,12 @@ func runPsql(testname string) (string, string) {
 
 	f, err := os.Create(outputFile)
 	if err != nil {
-		fatalf("%s: unable to create file %s: %s", progName, outputFile, err.Error())
+		logger.Fatalf("unable to create file %s: %s", outputFile, err.Error())
 	}
 	defer f.Close()
 
 	cmd := exec.Command("psql", "-X", "-a", "-f", inputSql)
+	logger.Printf("running %q", cmd.String())
```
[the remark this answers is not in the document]
bradfordb-vmware: Looking at the code for `func (c *Cmd) String() string`, only `Path` and `Args` are included in the returned string. [os/exec/exec.go](https://github.com/golang/go/blob/go1.21.1/src/os/exec/exec.go#L454-L471)

automation/pxf_regress/main.go
```diff
@@ -127,29 +102,35 @@ func runPsql(testname string) (string, string) {
 	if err != nil {
 		exiterr := &exec.ExitError{}
 		if !errors.As(err, &exiterr) {
```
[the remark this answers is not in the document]
bradfordb-vmware: Comments added to both `if`-statements

hpandeycodeit (approved):

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,370 @@
...
+	defer summaryDiff.Close()
+
+	diffHeader := fmt.Sprintf("diff %s %s %s\n", strings.Join(diffOpts, " "), resultsFile, expectFile)
+	summaryDiff.Write([]byte(diffHeader))
+	summaryDiff.Write(diffOutput)
+
+	return true
+}
+
+func elapsedMillisecs(x time.Time, y time.Time) float64 {
```
ashuka24, line 310: why do we need to do this calculation ourselves? 

additionally, could we put both this helper function as well as the `formatTimestamp` helper functions together at the end of the file instead of interspersed with the non-helper functions?

automation/src/test/resources/sut/MultiHadoopIPAMultiNodesCluster.xml
```diff
@@ -120,6 +120,11 @@
         <tincFolder>tinc/main</tincFolder>
         <tincTestsFolder>tincrepo/main</tincTestsFolder>
     </tinc>
+    <regress>
```
ashuka24, line 123: nit: consistency, some of the sut files we have new lines between each section, but not all of them

automation/Makefile
```diff
@@ -123,6 +124,9 @@ clean: clean-logs
 
 clean-logs:
 	@rm -rf automation_logs/* tincrepo/main/log/* run-results/*
+	@find tincrepo/main/pxf/. -type d -name results -exec rm -rf {} +
+	@find tincrepo/main/pxf/. -type f -name regression.diffs -delete
```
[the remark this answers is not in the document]
bradfordb-vmware: The `+` is part of the `-exec` action; from `man 1 find` you can do either `-exec command ;` or `-exec command +`

> This variant of the -exec action runs the specified command on the selected files, but the command line is built by appending each selected file name at the  end; the total number of invocations of the command will be much less than the number of matched files.

The `-delete` action doesn't need either `;` or `+`.

automation/pxf_regress/main.go
```diff
@@ -0,0 +1,370 @@
...
+
+	// Emit nice-looking summary message
+	summary := getSummary(successCount, failCount)
+	logger.Println(summary)
+
+	if failCount > 0 {
+		logger.Fatalf("the differences that caused some tests to fail can be viewed in the file %q", regressionDiffsName)
+	}
+
+	deleteResultsFiles()
```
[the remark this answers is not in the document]
bradfordb-vmware: `logger.Fatalf` doesn't return -- it exits the program. So if `failCount` is non-zero, we exit the program in the body of the previous `if`-statement. We only delete the results file if `failCount == 0`.

automation/src/test/resources/sut/MultiHadoopIPAMultiNodesCluster.xml
bradfordb-vmware: Generally, when adding the new `<regress></regress>` section, I've matched the existing convention for newlines between sections. I'd prefer to not standardize across the files since that could be large meaningless diff.

ashuka24 (approved):

PR #1025, opened 2023-09-01
pxf-build:Updates Promotion job to run only once per commit

hpandeycodeit:
- Added a new backward compatibility (no-op) job for GP7 RHEL8

Supporting cleanup and consistency changes:
- Renamed Promotion Job to "Promote PXF for GP5_GP6_GP7 Artifacts"
- Renamed Publish job to "Publish_PXF-GP5_GP6_GP7_Artifacts_to_GP-RelEng"
- Introduced `old_name` to the renamed jobs to retain build history.

### DIFF
--- a/concourse/pipelines/templates/build_pipeline-tpl.yml
+++ b/concourse/pipelines/templates/build_pipeline-tpl.yml
@@ -26,6 +26,9 @@ groups:
 {% for gp_ver in range(5, 7) %}
   - Test PXF-GP[[gp_ver]] Backward Compatibility on RHEL7
 {% endfor %}
+{% set gp_ver = 7 %}
+  - Test PXF-GP[[gp_ver]] Backward Compatibility on RHEL8
+{% set gp_ver = None %}
 {% set gp_ver = 6 %}
   - Test PXF-GP[[gp_ver]] Backward Compatibility on Ubuntu18
   - Test PXF-GP[[gp_ver]] Backward Compatibility on RHEL8
@@ -46,8 +49,8 @@ groups:
   - Test PXF-GP[[gp_ver]]-HDP2-Upgrade-Extension on RHEL7
 {% set gp_ver = None %}
   - Compatibility Gate for PXF-GP
-  - Promote PXF-GP5 and PXF-GP6 Artifacts
-  - Publish_PXF-GP5_and_PXF-GP6_Artifacts_to_GP-RelEng
+  - Promote PXF for GP5_GP6_GP7 Artifacts
+  - Publish_PXF-GP5_GP6_GP7_Artifacts_to_GP-RelEng
 
 {% set gp_ver = None %}
 {% for gp_ver in range(5, 7) %}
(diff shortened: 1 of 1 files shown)


denalex (commented):
so what was the reason for the promotion job being run 2 times ? it's hard to see from the diff changes

concourse/pipelines/templates/build_pipeline-tpl.yml
```diff
@@ -1712,47 +1718,55 @@ jobs:
...
+      passed: [Compatibility Gate for PXF-GP]
+      trigger: true
+    - get: pxf_tarball
+      passed: [Compatibility Gate for PXF-GP]
+      resource: pxf_gp[[gp_ver]]_tarball_rhel8
+{% set gp_ver = None %}
+
 ## ---------- Artifact Promotion Gate ----------
-- name: Promote PXF-GP5 and PXF-GP6 Artifacts
+- name: Promote PXF for GP5_GP6_GP7 Artifacts
```
bradfordb-vmware, line 1736: This is changing the name of the CI job which means we will loose previous build history. Do we want to add `old_name` ([Concourse docs](https://concourse-ci.org/jobs.html#schema.job.old_name)) to preserve that history?

concourse/pipelines/templates/build_pipeline-tpl.yml
```diff
@@ -1712,47 +1718,55 @@ jobs:
...
   plan:
   - in_parallel:
     - get: pxf_src
+      trigger: true
```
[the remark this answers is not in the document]
bradfordb-vmware: > the previous trigger was on the tarball, not the source, so we had a chance to override the tarball (if needed in a critical case to patch the tarball) and have it promoted without requiring the code change

I may not be fully understanding you but I don't think that would have ever worked -- we had `passed` conditions on `pxf_src` and all of the tar archives meaning that we would have needed an archive and a PXF commit that passed a common set of jobs. This PR isn't changing the required set of passed conditions (aside from the new GP7-RHEL8 compatibility job).

concourse/pipelines/templates/build_pipeline-tpl.yml
```diff
@@ -1769,12 +1783,12 @@ jobs:
     <<: *slack_alert
 {% endif %}
 
-- name: Publish_PXF-GP5_and_PXF-GP6_Artifacts_to_GP-RelEng
+- name: Publish_PXF-GP5_GP6_GP7_Artifacts_to_GP-RelEng
```
bradfordb-vmware, line 1786: Same comment about `old_name` here; retaining build history here is even more important than in the promotion job.

concourse/pipelines/templates/build_pipeline-tpl.yml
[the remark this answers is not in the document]
hpandeycodeit: As discussed, the previous implementation does not provided the desired effect of being able to introduce an externally modified tarball artifact. We are removing that mechanism for now and moving the trigger to the `pxf_src` resource. This is consistent with the other `pxf_src` resource usage.

hpandeycodeit:
> he promotion job being run 2 times ? it's hard to see from the diff changes

The GP7 artifact had a trigger to run the Promote job and the combined Backward compatibility jobs had a trigger as well. GP7 artifact was triggering the promote job separately followed by the combined Backward compatibility job triggers, causing Promote job to trigger twice.

concourse/pipelines/templates/build_pipeline-tpl.yml
hpandeycodeit: @bradfordb-vmware Thanks for pointing it out! As recommended, we have added the `old_name` to preserve the build history. I have tested this in dev environment and it is functioning as expected.

edespino (approved):
+1 - worked with @hpandeycodeit on the original change plus PR feedback (retain build history).

denalex (approved):

PR #1039, opened 2023-10-11
Remove tinc test framework

bradfordb-vmware:


automation/src/test/resources/sut/MultiHadoopIPAMultiNodesCluster.xml
```diff
@@ -112,17 +112,9 @@
...
-        <userName>gpadmin</userName>
-        <gphome>/usr/local/greenplum-db-devel</gphome>
-        <tincFolder>tinc/main</tincFolder>
-        <tincTestsFolder>tincrepo/main</tincTestsFolder>
-    </tinc>
     <regress>
         <class>org.greenplum.pxf.automation.components.regress.Regress</class>
-        <regressTestFolder>tincrepo/main</regressTestFolder>
+        <regressTestFolder>sqltestrepo</regressTestFolder>
         <dbName>pxfautomation</dbName>
```
denalex, line 118: interesting that we do not need host/userName/gphome anymore

concourse/docker/pxf-dev-base/gpdb5/centos7/Dockerfile
```diff
@@ -49,8 +49,6 @@ RUN ssh-keygen -t rsa -N "" -f /root/.ssh/id_rsa \
     && echo -e "password\npassword" | passwd gpadmin 2> /dev/null \
     && { ssh-keyscan localhost; ssh-keyscan 0.0.0.0; } >> /home/gpadmin/.ssh/known_hosts \
     && chown -R gpadmin:gpadmin /home/gpadmin/.ssh \
-    # install psi
-    && /usr/bin/pip install psi --no-cache-dir \
```
denalex, line 53: will these removals break testing for PXF5 and previous versions  of PXF6 should we need to patch a previous release ?

concourse/docker/pxf-dev-base/gpdb6/ubuntu18.04/Dockerfile
```diff
@@ -13,9 +13,6 @@ RUN mkdir -p /tmp/pxf_src/ && cd /tmp \
     && echo "${GO_SHA256SUM} /tmp/go.tgz" | sha256sum --check \
     && rm -rf /usr/local/go && mkdir /opt/pxf && tar -C /opt/pxf -xzf go.tgz && rm go.tgz
 
-# install dependencies that are missing on the base images
```
denalex, line 16: do we want to leave this line in place ?

denalex (approved):

automation/Makefile
```diff
@@ -123,11 +122,11 @@ clean: clean-logs
...
-	@rm -rf automation_logs/* tincrepo/main/log/* run-results/*
-	@find tincrepo/main/pxf/. -type d -name results -exec rm -rf {} +
-	@find tincrepo/main/pxf/. -type f -name regression.diffs -delete
-	@find tincrepo/main/pxf/. -type d -name "output" -exec rm -rf {} +
-	@find tincrepo/main/pxf/. -type f -name "*segment_logs" -exec rm -rf {} +
+	@rm -rf automation_logs/* run-results/*
+	@find sqltestrepo -type d -name results -exec rm -rf {} +
+	@find sqltestrepo -type f -name regression.diffs -delete
+	@find sqltestrepo -type d -name "output" -exec rm -rf {} +
+	@find sqltestrepo -type f -name "*segment_logs" -exec rm -rf {} +
```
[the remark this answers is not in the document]
bradfordb-vmware: I have a slight preference for `sqlrepo` over `testrepo` because of the similarity to `src/test`. If there are no other concerns or suggestions, I'll switch to `sqlrepo`.

automation/src/test/resources/sut/MultiHadoopIPAMultiNodesCluster.xml
bradfordb-vmware: The reason we don't need them is:

1. We run tests as `gpadmin` OS user so when we execute `psql` (via `pxf_regress`), we default to connecting as `gpadmin`
2. We set `PGHOST` to `cdw` when running tests against a multinode GPDB cluster; for local cluster, we don't set `PGHOST` and we default to connecting on localhost
3. We require `GPHOME` to be set in the environment and read it from there

We can add fields for host, user, and GPHOME to the `Regress` component but then there's the potential for confusion about which value takes precedence.

yimingli-vmware (approved):

PR #1041, opened 2023-10-12
Add retries to Ansible tasks that fail frequently

bradfordb-vmware:
Looking at the previous 100 builds of the job "Test PXF-GP6-HDP2-SECURE-MULTI-IMPERS on RHEL7" in the "pxf-build" pipeline, there were 35 non-successful builds. Of those 35, 19 of the failures were durng the "Generate Multinode Hadoop Cluster" step. This commit adds retries to the Ansible tasks that most frequently fail.

### DIFF
--- a/concourse/ansible/ipa-multinode-hadoop/tasks/hdfs.yml
+++ b/concourse/ansible/ipa-multinode-hadoop/tasks/hdfs.yml
@@ -74,14 +74,28 @@
 
 # providing the absolute path for ipa-getkeytab since due to some reason it was not recognizing the
 # ipa-getkeytab and without && it is not reporting error in case the principal already exists.
-- name: Create HDFS service principal and export keytab
+- name: Create HDFS service principal
   ansible.builtin.shell:
     cmd: |
-      echo {{ ipa_password }} | kinit admin@{{ ansible_domain | upper }}  &&
+      echo {{ ipa_password }} | kinit admin@{{ ansible_domain | upper }} &&
       ipa service-add hdfs/{{ ansible_fqdn }}@{{ ansible_domain | upper }} &&
+      kdestroy
+  register: result
+  until: result.rc == 0
+  retries: 2
+  delay: 15
+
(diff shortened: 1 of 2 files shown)


concourse/ansible/ipa-multinode-hadoop/tasks/hdfs.yml
```diff
@@ -74,14 +74,28 @@
...
-      echo {{ ipa_password }} | kinit admin@{{ ansible_domain | upper }}  &&
+      echo {{ ipa_password }} | kinit admin@{{ ansible_domain | upper }} &&
       ipa service-add hdfs/{{ ansible_fqdn }}@{{ ansible_domain | upper }} &&
+      kdestroy
+  register: result
+  until: result.rc == 0
+  retries: 2
+  delay: 15
+
+- name: Export keytab for HDFS service principal
```
ashuka24, line 88: was there an issue with this being one task? is there a reason it was separated out

ashuka24 (approved):
I did a test run of this in the new concourse instance since it seems like we were hitting the error more often there

image

https://ci.ud.gpdb.pivotal.io/teams/main/pipelines/dev-axue-add-retries-ipa/jobs/test_pxf-gp6-hdp2-secure-multi-impers_on_rhel7/builds/3

Also, it seems like `Run IPA client installer` can also sometimes be flakey, do we want to add retries there as well?

concourse/ansible/ipa-multinode-hadoop/tasks/hdfs.yml
bradfordb-vmware: I split them up so that each task was idempotent -- if the two commands where in one task and adding the service succeeded but exporting the keytab failed, retrying the task would likely fail (can't create a service that already exists).

denalex (approved):

yimingli-vmware (approved):

hpandeycodeit (approved):

PR #1043, opened 2023-10-19
Templating Pipeline Jobs

hpandeycodeit:
- Added build and test template files.
- Added a dictionary that contains the configuration to generate the jobs.
- Renamed jobs to reflect the platform.
- The build and test jobs ( sanity test) jobs are now generated through the templates.

Pipeline: https://ci.ud.gpdb.pivotal.io/teams/main/pipelines/dev-pandeyhi-job_templates

### DIFF
--- a/concourse/pipelines/templates/dev_build_pipeline-tpl.yml
+++ b/concourse/pipelines/templates/dev_build_pipeline-tpl.yml
@@ -377,43 +377,38 @@ jobs:
 ## the following platform_gp_combinations dictionary definition:
 ##
 ##   gp_ver: GPDB Major Version
-##   execution_os: OS platform in which build and test jobs are executed
-##   os_ver: OS Major version
 ##   build_os: OS Platform in which the build was performed
+##   test_os: OS platform in which test jobs are executed
+##   os_ver: OS Major version
 ##
 ## ----------------------------------------------------------------------
 {% set platform_gp_combinations = [
-    {'gp_ver': '5', 'execution_os': 'centos', 'os_ver': '7',  'build_os': 'centos'},
-    {'gp_ver': '6', 'execution_os': 'centos', 'os_ver': '7',  'build_os': 'centos'},
-    {'gp_ver': '6', 'execution_os': 'rocky',  'os_ver': '8',  'build_os': 'rocky'},
-    {'gp_ver': '6', 'execution_os': 'oel',    'os_ver': '7',  'build_os': 'centos'},
-    {'gp_ver': '6', 'execution_os': 'ubuntu', 'os_ver': '18', 'build_os': 'ubuntu'},
-    {'gp_ver': '7', 'execution_os': 'rocky',  'os_ver': '8',  'build_os': 'rocky'}] %}
+    {'gp_ver': '5', 'build_os': 'centos', 'test_os': 'centos', 'os_ver': '7'},
+    {'gp_ver': '6', 'build_os': 'centos', 'test_os': 'centos', 'os_ver': '7'},
+    {'gp_ver': '6', 'build_os': 'rocky', 'test_os': 'rocky', 'os_ver': '8'},
+    {'gp_ver': '6', 'build_os': 'centos', 'test_os': 'oel', 'os_ver': '7'},
+    {'gp_ver': '6', 'build_os': 'ubuntu', 'test_os': 'ubuntu', 'os_ver': '18'},
+    {'gp_ver': '7', 'build_os': 'rocky', 'test_os': 'rocky', 'os_ver': '8'}] %}
 
 {% for config in platform_gp_combinations %}
 
-    {% set gp_ver = config.gp_ver %}
-    {% set platform = config.execution_os %}
-    {% set os_ver = config.os_ver %}
-    {% set build_platform = config.build_os %}
+    {% set test_platform = config.test_os ~ config.os_ver %}
+    {% set build_platform = config.build_os ~ config.os_ver %}
 
-    {% set platform_ver = platform ~ os_ver %}
-    {% set build_platform_ver = build_platform ~ os_ver %}
-
-    {% if platform in ['ubuntu'] %}
+    {% if config.test_os in ['ubuntu'] %}
       {% set pkg_type = 'deb' %}
       {% set gpdb_os_platform = 'ubuntu' %}
-      {% set pxf_tarball_filename = 'dist/pxf-gp' ~ gp_ver ~ '-*-ubuntu' ~ os_ver ~ '.04' ~ '.tar.gz' %}
+      {% set pxf_tarball_filename = 'dist/pxf-gp' ~ config.gp_ver ~ '-*-ubuntu' ~ config.os_ver ~ '.04' ~ '.tar.gz' %}
       {% set pxf_fdw_tarball_filename = '' %}{# Doesn't apply for ubuntu platform #}
-    {% else %}
+    {% else %} {# Else section applies to centos, rocky and oel platforms #}
       {% set pkg_type = 'rpm' %}
       {% set gpdb_os_platform = 'rhel' %}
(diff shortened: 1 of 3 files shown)


concourse/pipelines/templates/jobs/test-tpl.yml
edespino, line 1: As there are many test jobs for which this template does not apply, consider changing the name of this template to reflect the type of test (e.g. sanity-test-tpl.yml). I have always thought of the first test column as a sanity test. I will leave it to you to identify the proper name.

concourse/pipelines/templates/jobs/test-tpl.yml
```diff
@@ -0,0 +1,36 @@
+## Template for generating test jobs in the dev and build pipeline
```
edespino, line 1: If you do update the this test template name, consider updating line one here (again to easily allow an engineer to identify the use of the template.

edespino (changes requested):
The only change I see is the renaming of the test template file and comment in the corresponding test template. It will help to clearly identify what series of tests the test template applies to. Thanks.

edespino (approved):
Upon further discussion, at this point there is no need to rename the test files

edespino (approved):

concourse/pipelines/templates/dev_build_pipeline-tpl.yml
```diff
@@ -369,228 +369,116 @@ resources:
...
-      SECRET_ACCESS_KEY: ((tf-machine-secret-access-key))
+## ======================================================================
+##                    Build and Sanity Test Swimlanes
+## ----------------------------------------------------------------------
+##
+## The creation of the PXF Build and Sanity test jobs are driven by
+## the following platform_gp_combinations dictionary definition:
+##
+##   gp_ver: GPDB Major Version
+##   execution_os: OS platform in which build and test jobs are executed
```
denalex, line 380:
```suggestion
##   test_os: OS platform in which test jobs are executed
```
it seems this is only for testing, no ? build is done in` build_os` ?

concourse/pipelines/templates/dev_build_pipeline-tpl.yml
```diff
@@ -369,228 +369,116 @@ resources:
...
+##   os_ver: OS Major version
+##   build_os: OS Platform in which the build was performed
+##
+## ----------------------------------------------------------------------
+{% set platform_gp_combinations = [
+    {'gp_ver': '5', 'execution_os': 'centos', 'os_ver': '7',  'build_os': 'centos'},
```
denalex, line 386:
```suggestion
    {'gp_ver': '5', 'build_os': 'centos', 'test_os': 'centos', 'os_ver': '7'},
```
can we have this logical order ?

concourse/pipelines/templates/dev_build_pipeline-tpl.yml
```diff
@@ -369,228 +369,116 @@ resources:
...
+
+{% for config in platform_gp_combinations %}
+
+    {% set gp_ver = config.gp_ver %}
+    {% set platform = config.execution_os %}
+    {% set os_ver = config.os_ver %}
+    {% set build_platform = config.build_os %}
+
+    {% set platform_ver = platform ~ os_ver %}
+    {% set build_platform_ver = build_platform ~ os_ver %}
```
denalex, line 401: I am confused with conventions, especially `platform` and `platform_ver`.

It started nicely with `os` meaning the name of OS and `os_ver` meaning the version of the OS. What is the `platform` then ? I could understand if `platform` would be `os` + `os_ver`, but then there is `platform_ver`. Too complex for a developer to understand which one to use.

concourse/pipelines/templates/dev_build_pipeline-tpl.yml
```diff
@@ -369,228 +369,116 @@ resources:
...
+
+    {% if platform in ['ubuntu'] %}
+      {% set pkg_type = 'deb' %}
+      {% set gpdb_os_platform = 'ubuntu' %}
+      {% set pxf_tarball_filename = 'dist/pxf-gp' ~ gp_ver ~ '-*-ubuntu' ~ os_ver ~ '.04' ~ '.tar.gz' %}
+      {% set pxf_fdw_tarball_filename = '' %}{# Doesn't apply for ubuntu platform #}
+    {% else %}
```
denalex, line 408: does this section also apply to OEL ?

concourse/pipelines/templates/dev_build_pipeline-tpl.yml
```diff
@@ -369,228 +369,116 @@ resources:
...
+      {% set pkg_type = 'rpm' %}
+      {% set gpdb_os_platform = 'rhel' %}
```
denalex, line 410: what is gpdb_os_platform ? the OS the GPDB RPM is built for ?

concourse/pipelines/templates/dev_build_pipeline-tpl.yml
```diff
@@ -369,228 +369,116 @@ resources:
...
+    {'gp_ver': '5', 'execution_os': 'centos', 'os_ver': '7',  'build_os': 'centos'},
+    {'gp_ver': '6', 'execution_os': 'centos', 'os_ver': '7',  'build_os': 'centos'},
+    {'gp_ver': '6', 'execution_os': 'rocky',  'os_ver': '8',  'build_os': 'rocky'},
+    {'gp_ver': '6', 'execution_os': 'oel',    'os_ver': '7',  'build_os': 'centos'},
+    {'gp_ver': '6', 'execution_os': 'ubuntu', 'os_ver': '18', 'build_os': 'ubuntu'},
+    {'gp_ver': '7', 'execution_os': 'rocky',  'os_ver': '8',  'build_os': 'rocky'}] %}
+
+{% for config in platform_gp_combinations %}
+
+    {% set gp_ver = config.gp_ver %}
```
bradfordb-vmware, line 395: Is there a reason for doing `{% set gp_ver = config.gp_ver %}`? Could we not just use `config.gp_ver`?

concourse/pipelines/templates/dev_build_pipeline-tpl.yml
bradfordb-vmware: It might be clearer if we wrote

```
{% elif platform in ['centos', 'oel', 'rocky'] %}
```

so that it is clear that it does apply to OEL. However, we'd probably want to add in an `{% else %}` which raises an error message for unknown values of `platform`; it looks like we can add a `raise` function to the template tool ([StackOverflow](https://stackoverflow.com/a/29262304)). This would end up looking something like

```
{% if platform in ['ubuntu'] %}
...
{% elif platform in ['centos', 'oel', 'rocky'] %}
...
{% else %}
{{ raise("unknown platform") }}
{% endif %}
```

Or we just add a comment `# also applies to OEL`

concourse/pipelines/templates/jobs/build-tpl.yml
```diff
@@ -0,0 +1,25 @@
+## Template for generating build jobs in the dev and build pipeline
```
bradfordb-vmware, line 1: Can we add a doc comment listing what template variables are expected here?

concourse/pipelines/templates/jobs/test-tpl.yml
bradfordb-vmware, line 1: Can we add a doc comment listing what template variables are expected here?

concourse/pipelines/templates/dev_build_pipeline-tpl.yml
hpandeycodeit: yes. Like `rhel` or `ubuntu`

concourse/pipelines/templates/dev_build_pipeline-tpl.yml
hpandeycodeit: For now I have added a comment here. Once we move the logic/matrix to the template_tool, then we can create a similar function which can throw an error.

bradfordb-vmware (approved):

PR #1024, opened 2023-08-31
Install go to /opt/pxf/go in the docker images

ashuka24:
This PR changes the install location in the Docker images from /usr/local/go to /opt/pxf/go. 
It also adds a sha25sum check to make sure we are downloading the correct binary.

It also updates the concourse dev pipeline to include `PXF_DEV_IMAGE_TAG` as an option to make for faster testing with different images.

PR #1044, opened 2023-10-20
Upgrade Go toolchain to 1.21.3

bradfordb-vmware:
Go 1.21 introduce a small change to the numbering of releases (see [1] for more details); a practical consequence of this change is that go.mod can now use full version for the go directive (go 1.21.3 instead of just go 1.21).

In addition to upgrading the Go toolchain to the latest release, this commit also upgrades a few key dependencies:

* ginkgo is upgrade to ginkgo/v2
* gomega is updated to its latest release
* gp-common-go-libs is update to its latest release

Modules that are no longer needed were removed by running go mod tidy.

[1]: https://go.dev/doc/go1.21


concourse/docker/pxf-dev-base/gpdb5/centos7/Dockerfile
```diff
@@ -9,7 +9,7 @@ ARG GO_SHA256SUM
 RUN mkdir -p /tmp/pxf_src/ && cd /tmp \
     && wget -O go.tgz -q https://go.dev/dl/go${GO_VERSION}.linux-amd64.tar.gz \
     && echo "${GO_SHA256SUM} /tmp/go.tgz" | sha256sum --check \
-    && rm -rf /usr/local/go && mkdir /opt/pxf && tar -C /opt/pxf -xzf go.tgz && rm go.tgz
+    && rm -rf /usr/local/go && tar -C /usr/local -xzf go.tgz && rm go.tgz
```
yimingli-vmware, line 13: why we want to use the existing `/usr/local` instead of creating a new `/opt/pxf`?

yimingli-vmware (approved):

concourse/docker/pxf-dev-base/gpdb5/centos7/Dockerfile
bradfordb-vmware: We originally were installing the Go toolchain into `/usr/local/go` but then we started having a weird issue with it being corrupted on Rocky Linux 8 containers running in Concourse 5.7.1. We moved it to `/opt/pxf/go` in #1024 to unblock the pipeline and the 6.8.0 release of PXF. Since we've upgraded to Concourse v7, I'm moving it back to `/usr/local/go`.

I have built dev images and run a dev pipeline using these dev images.

denalex (approved):

ashuka24 (approved):

PR #1045, opened 2023-10-23
Minor fix: Add OEL test job only when the oel7 param set to true

hpandeycodeit:
Minor fix: Set the OEL test job only when `oel7` param set to true.


edespino (approved):
Thank you for addressing this swiftly.

PR #1046, opened 2023-10-27
Dev pipeline updates - OEL7 test job & build job references.

edespino:
This PR contains two critical changes for the dev pipeline.

- Add OEL7 basic test job only if OEL7 configuration is specified.

- Recently, the names of the build jobs have been modified to accurately represent the operating system on which PXF is built. This update has also been carried through to multiple test jobs that have resource-based constraints, which now reference the newly renamed build jobs.

### DIFF
--- a/concourse/pipelines/templates/dev_build_pipeline-tpl.yml
+++ b/concourse/pipelines/templates/dev_build_pipeline-tpl.yml
@@ -381,6 +381,9 @@ jobs:
 ##   test_os: OS platform in which test jobs are executed
 ##   os_ver: OS Major version
 ##
+## NOTE: The oel7 platform only includes a test job. The inclusion of
+## the oel7 test job is controlled by the Makefile OEL7 (disbled by
+## default) parameter.
 ## ----------------------------------------------------------------------
 {% set platform_gp_combinations = [
     {'gp_ver': '5', 'build_os': 'centos', 'test_os': 'centos', 'os_ver': '7'},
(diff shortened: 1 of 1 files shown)


concourse/pipelines/templates/dev_build_pipeline-tpl.yml
```diff
@@ -412,7 +412,9 @@ jobs:
       {% include 'jobs/build-tpl.yml' %}
     {% endif %}
 
-    {% include 'jobs/test-tpl.yml' %}
+    {% if config.test_os != 'oel' or oel7 %}
```
denalex, line 415: why `config.test_os != 'oel'` ? If test_os is OEL why don't we want to test on it ?

denalex: Bradford explained it that for OEL to have a test job in dev pipeline, OEL variable needs to be set explicitly. So no change is needed.

denalex (approved):

concourse/pipelines/templates/dev_build_pipeline-tpl.yml
edespino: @denalex The `if` clause has two possible conditions (`or`) under which it can be executed. First condition, we need to add the test job for all configurations except `oel7`.  Second condition, only add the test job if `oel7` is set. It is set from the Makefile and not always enabled.

edespino: I will add a note about `oel7` that the test configuration is enabled via the Makefile OEL7 variable.

PR #1047, opened 2023-10-30
Fix pxf register command

bradfordb-vmware:
The pxf register command assumes that $GPHOME/share/postgresql/extension is writable by the user running the command (usually gpadmin). The GPDB installation docs include a step for changing the ownership of the GPDB installation directory but if this step is skipped, then pxf register will fail. As it is currently implemented, the step of copying the control file into $GPHOME fails but does not cause the pxf utility to exit with a non-zero value. This in turn causes the pxf cluster register command to not recognize the pxf register command failure and not print the stderr output of the command. This makes it seem like the register command has succeeded until the user attempts to call CREATE EXTENSION pxf in psql.

This commit replaces the usage of cp and chmod with install and explicitly adds a failure case if the install fails.

### DIFF
--- a/concourse/scripts/cli/test_reset_init.sh
+++ b/concourse/scripts/cli/test_reset_init.sh
@@ -27,6 +27,16 @@ remove_extension_files() {
 	"
 }
 
+make_gphome_extension_dir_readonly() {
+  local host=$1
+  ssh "${host}" "chmod u-w ${GPHOME}/share/postgresql/extension/"
+}
+
+make_gphome_extension_dir_readwrite() {
+  local host=$1
+  ssh "${host}" "chmod u+w ${GPHOME}/share/postgresql/extension/"
+}
+
 # === Test "pxf cluster reset" - does nothing on the remote nodes, PXF can be running, no state changes ======
 successful_reset_message=\
 "*****************************************************************************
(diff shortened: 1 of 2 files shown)


server/pxf-service/src/scripts/pxf
```diff
@@ -260,8 +260,7 @@ function installExtensions() {
             echoGreen "Installing Greenplum External Table PXF Extension into '${GPHOME}'"
 
             local target_control_file="${GPHOME}/share/postgresql/extension/pxf.control"
-            cp -Lv "${parent_script_dir}/gpextable/pxf.control" "${target_control_file}"
-            chmod 644 "${target_control_file}"
+            install --verbose --mode=0644 "${parent_script_dir}/gpextable/pxf.control" "${target_control_file}" || fail "cannot install pxf.control to '${target_control_file}'"
```
denalex, line 263: `cp -L` was guaranteed to follow symlinks, are we losing this ability with the `install` command ?

concourse/scripts/cli/test_reset_init.sh
```diff
@@ -27,6 +27,16 @@ remove_extension_files() {
 	"
 }
 
+make_gphome_readonly() {
```
denalex, line 30:
```suggestion
make_gphome_extensions_dir_readonly() {
```

server/pxf-service/src/scripts/pxf
bradfordb-vmware: According to [this StackOverflow answer](https://stackoverflow.com/a/39295609), `install` always dereferences symlinks so `install` is equivalent to `cp -L`.

ashuka24 (approved):

concourse/scripts/cli/test_reset_init.sh
```diff
@@ -119,5 +129,41 @@ test_init_extension_exists_succeeds() {
...
+  # when : "pxf cluster register" command is run
+  local result status
+  result="$(pxf cluster register 2>&1)"
+  status=$?
+  # then : it fails
+  assert_not_equals "${status}" 0 "pxf cluster register (GPHOME/share/postgresql/extension not writable) should fail"
+  # and : prints an useful error message
+  assert_equals "${expected_register_error_message}" "${result}" "an error message should be printed when pxf cluster register fails"
+
+  for host in "${all_cluster_hosts[@]}"; do
```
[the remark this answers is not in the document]
bradfordb-vmware: At the start of the test, we make `$GPHOME/share/postgresql/extension` read-only and this is restoring that directory to its expected state.

yimingli-vmware (approved):

denalex (approved):

PR #1052, opened 2023-11-15
Modularized the development pipeline template

denalex:
The goal of this PR is to extract all YAML definitions from the development pipeline template into their own modules such that YAML definitions can be re-used between dev and prod pipeline templates.

Once all the YAML definitions are in their own files and are parameterized, the dev pipeline template was updated with mostly Jinja2 macros and includes to create testing swimlanes.


denalex:
This PR is closed in favor of https://github.com/greenplum-db/pxf/pull/1052 and a separate PR to enable Rocky9 support across dev/prod pipleines.

PR #1053, opened 2023-11-16
Refactored the development pipeline template with macros and sub-templates

denalex:
The goal of this PR is to extract all YAML definitions from the development pipeline template into their own modules such that YAML definitions can be re-used between dev and prod pipeline templates.

Once all the YAML definitions are in their own files and are parameterized, the dev pipeline template was updated with mostly Jinja2 macros and includes to create testing swimlanes.

This refactor does not include changes required to support Rocky9 operating system.


concourse/README.md
```diff
@@ -95,7 +95,7 @@ To deploy dev pipeline against gpdb 5X_STABLE and 6X_STABLE branches, use:
 make -C "${HOME}/workspace/pxf/concourse" dev
 ```
 
-To deploy multi-node dev pipeline, you can specify either the `MULTINODE` or
+To deploy multi-node dev pipeline, you can specify either the `MULTINODE`, `MULTINODE_EL8` or
```
xinzweb, line 98:
```suggestion
To deploy multi-node dev pipeline, you can specify either the `MULTINODE` for EL7, `MULTINODE_EL8` for EL8 or
```

concourse/pipelines/templates/anchors/destroy-cluster-tpl.yml
```diff
@@ -0,0 +1,26 @@
+{# Template for an anchor that destroys a cluster.
+ #
+ # Expects a 'cluster_type' variable to be set to either 'gpdb' or 'hadoop'
+ #}
+
+{% if cluster_type == 'gpdb' %}
+    {% set put_suffix           = 'gpdb' %}
+    {% set env_name_file_suffix = '-gpdb' %}
```
xinzweb, line 8: Do we have to use two variables? they are very alike, like `env_name_file_suffix` is just `-` + `put_suffix`.

concourse/pipelines/templates/anchors/destroy-cluster-tpl.yml
```diff
@@ -0,0 +1,26 @@
...
+    {% set terraform_source     = 'ccp_src/google/' %}
+{% else %}
+    {% set put_suffix           = 'ipa-hadoop' %}
```
xinzweb, line 11: similar here.

concourse/pipelines/templates/anchors/destroy-cluster-tpl.yml
```diff
@@ -0,0 +1,26 @@
...
+{% endif %}
+
+- &destroy_[[cluster_type]]_cluster
+  put: terraform-[[put_suffix]]
+{% if cluster_type == 'gpdb' %}
+  resource: terraform
+{% endif %}
+  params:
+    action: destroy
+    env_name_file: terraform[[env_name_file_suffix]]/name
```
xinzweb, line 23:
```suggestion
    env_name_file: terraform-[[put_suffix]]/name
```

something like that, maybe a better variable name.

concourse/pipelines/templates/anchors/destroy-dataproc-tpl.yml
```diff
@@ -0,0 +1,25 @@
+{# Template for an anchor that destroys a dataproc cluster.
+ #
+ # Expects a 'cluster_index' variable to be set
+ #}
+
+{% set google_prefix = '' %}
+- &destroy_dataproc_[[cluster_index]]
+  task: cleanup-dataproc-[[cluster_index]]
+{% if cluster_index == 2 %}
```
xinzweb, line 9: Seems like 2 is a special case?

What's valid range of the values for the `cluster_index`? What if we have 3, do we ever have 3?

xinzweb: Saw the example below.

This is anomaly caused by naming of the secrets.
Anyway we can create clone of the secrets with index prefix? e.g. 
`1-ccp-ci-service-account-key` and `2-ccp-ci-service-account-key`

We can keep the old key for now for backward compatibility.

Once we switch to the template, we can remove the old keys, thoughts?

concourse/pipelines/templates/dev_build_pipeline-tpl.yml
```diff
@@ -1,1213 +1,262 @@
...
+        {% include 'jobs/build-tpl.yml' %}
     {% endif %}
-
-    {# OEL doesn't have a separate build job #}
-    {% if config.test_os != 'oel' %}
-      {% include 'jobs/build-tpl.yml' %}
+{% endcall %}
+
+{# include a non-FDW test job for all OS other than OEL (unless requested in the Makefile) #}
+{% call(x) macros.for_each_config(build_test_pxf_combinations) %}
```
xinzweb, line 165: another design is that we can put the following variables also in your big table.

the table will be super wide, but you don't have to special case the configuration here, and it's less complex to understand the logic, what do you think?

concourse/pipelines/templates/dev_build_pipeline-tpl.yml
```diff
@@ -1,1213 +1,262 @@
...
-{% set gp_ver = 7 %}
-{% set passed = '[build_pxf-gp7_on_rocky8]' %}
-{% include 'jobs/multinode-tpl.yml'%}
-{% set passed = None %}
-{% if slack_notification %}
-    <<: *slack_alert
-{% endif %}
-{% endif %}{# if gp7_multinode ends here #}
-{% set gp_ver = None %}
+{% call(x) macros.for_each_config(build_test_pxf_combinations) %}
```
xinzweb, line 242: similar to the single node tests, to see if we can put these variables to the big table.

concourse/pipelines/templates/resources/github-repo-tpl.yml
```diff
@@ -0,0 +1,18 @@
+{# Template for a Github repository resource definition.
+ #
+ # Expects a 'repo_name" variable set to 'pxf' or 'ccp' value
```
xinzweb, line 3:
```suggestion
 # Expects a 'repo_name' variable set to 'pxf' or 'ccp' value
```

concourse/pipelines/templates/resources/terraform-tpl.yml
```diff
@@ -0,0 +1,31 @@
...
+      GOOGLE_CREDENTIALS: ((ud/pxf/secrets/[[credential_key_prefix]]ccp-ci-service-account-key))
+    storage:
+      access_key_id: ((tf-machine-access-key-id))
+      bucket: ((ud/pxf/common/tf-bucket-name))
+      bucket_path: (([[bucket_path_key]]))
+      region_name: ((ud/common/aws-region))
+      secret_access_key: ((tf-machine-secret-access-key))
+    vars:
+{% if is_ipa %}
+      gcp_project: ((ud/pxf/common/ipa-google-project-id))
```
xinzweb, line 28: you already have variable `credential_key_prefix`, want to reuse that with a better name?
```suggestion
      gcp_project: ((ud/pxf/common/[[credential_key_prefix]]google-project-id))
```

xinzweb (commented):
Great refactoring.

I learned a lot through this PR review.

I would say, it's already good at its current state.
Refactoring can always be improved.
At this moment, it's LGTM.

Feel free to reach me if you need further clarification. Thanks.

concourse/pipelines/templates/anchors/destroy-dataproc-tpl.yml
denalex: Valid indices are 1 and 2. unfortunately there is a lot of asymmetry in naming here as cluster 2 came later and we create it in a separate google project id with `kerberos-` name prefix. While it would be nice to have a better convention here, it is quite involved and is beyond the scope of this PR, IMO.

concourse/pipelines/templates/resources/terraform-tpl.yml
denalex: we could have removed `{% if is_ipa %}` and used the prefix, but the keys are different (`gcp_project` vs `project_id`). I can still put it here as you suggest, but I'm afraid readability will be worse when comparing 2 branches of the if statement.

concourse/pipelines/templates/anchors/destroy-dataproc-tpl.yml
```diff
@@ -0,0 +1,25 @@
...
+      path: pxf_src/concourse/scripts/cleanup_dataproc_cluster.bash
+    inputs:
+    - name: pxf_src
+    - name: dataproc_env_files
+    platform: linux
+  image: ccp-7-image
+  params:
+    GOOGLE_CREDENTIALS: ((ud/pxf/secrets/[[google_prefix]]ccp-ci-service-account-key))
+    GOOGLE_PROJECT_ID: ((ud/pxf/common/[[google_prefix]]google-project-id))
+    GOOGLE_ZONE: ((ud/pxf/common/[[google_prefix]]google-zone))
```
ashuka24, line 25: end new line at end of file

concourse/pipelines/templates/jobs/test-cli-tpl.yml
```diff
@@ -0,0 +1,82 @@
+{# Template for a PXF basic test job definition.
+ #
+ # Expects a dictionary with the name 'x' to be set as defined in macros.j2 file and these additional fields:
+ #
+ # x.passed     - an array of upstream job names that the artifacts need to pass before this job
```
ashuka24, line 5: this is not a required additional field but an optional one. can we combine this with line 8?

concourse/pipelines/templates/jobs/test-multinode-tpl.yml
```diff
@@ -0,0 +1,228 @@
...
+      in_parallel:
+        steps:
+        - *destroy_dataproc_1
+        - *destroy_dataproc_2
+        - *destroy_gpdb_cluster
+{% if x.use_impers == 'true' %}
+        - *destroy_hadoop_cluster
+{% endif %}
+{% if slack_notification %}
+    <<: *slack_alert
```
ashuka24, line 227: the slack alert has an `on_success` that seems to override the existing on success when set to true. can we just remove the `on success` part of the slack alert?

concourse/pipelines/templates/macros/macros.j2
```diff
@@ -0,0 +1,108 @@
...
+  - 'test_fdw' : whether to test PXF FDW extension in this configuration
+
+  Additional derived attributes added to the "x" context dictionary:
+  - 'build_platform'     : the name and version of the OS for building PXF artifacts
+  - 'test_platform'      : the name and version of the OS for testing  PXF artifacts
+  - 'pkg_type'           : the type of the PXF artifact package (rpm or deb)
+  - 'pkg_arch'           : the machine architecture of the package (x86_64 or amd64)
+  - 'release_platform'   : the platform used in PXF  artifact file names (el7, el8, el9, ubuntu18.04)
+  - 'greenplum_platform' : the platform used in GPDB artifact file names (rhel7, rhel8, el9, ubuntu18.04)
+  - 'pxf_tarball_infix'  : the symbol separating the platform in the PXF artifact names
```
ashuka24, line 22: since it is the symbol separating the platform in the pxf artifact names, can we call this `pxf_platform_infix`?

concourse/pipelines/templates/resources/greenplum-package-tpl.yml
```diff
@@ -0,0 +1,15 @@
...
+{# include Greenplum package only if both build and test platform are the same to avoid duplicates #}
+{% if x.build_platform == x.test_platform %}
+- name: gpdb[[x.gp_ver]]-[[x.release_platform]]-[[x.pkg_type]]-latest-0
+  type: gcs
+  icon: google-drive
+  source:
+    bucket: ((ud/pxf/common/pivnet-artifacts-bucket-name))
+    json_key: ((ud/pxf/secrets/pxf-storage-service-account-key))
+    regexp: latest-0_gpdb[[x.gp_ver]]/greenplum-db-([[x.gp_ver]].*)-[[x.greenplum_platform]]-[[x.pkg_arch]].[[x.pkg_type]]
+{% endif %}
```
ashuka24, line 15: nit: new line at end of file.

this seems to be true for quite a few files in this pr

concourse/pipelines/templates/resources/slack-notification-tpl.yml
```diff
@@ -0,0 +1,10 @@
+{# Template for a Slack notification resource definition.
+ #
+ # Expects a 'url_key" variable to be set
```
ashuka24, line 3:
```suggestion
 # Expects a 'url_key' variable to be set

File: concourse/README.md [hunk 16]
```diff
@@ -95,8 +95,8 @@ To deploy dev pipeline against gpdb 5X_STABLE and 6X_STABLE branches, use:
 make -C "${HOME}/workspace/pxf/concourse" dev
 ```
 
-To deploy multi-node dev pipeline, you can specify either the `MULTINODE` or
-`MULTINODE_NO_IMPERSONATION`, which will also run CLI tests:
+To deploy multi-node dev pipeline, you can specify either the `MULTINODE` for EL7, `MULTINODE_EL8` for EL8 or
```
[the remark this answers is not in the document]
denalex: makes sense

concourse/pipelines/templates/dev_build_pipeline-tpl.yml
denalex: The matrix is a collection of "swimlane" definitions, with swimlane having multiple jobs. One swimlane might have FDW and non-FDW jobs, so `use_fdw` variable controls, per-job, whether we are rendering one vs. the other. Same for impersonation / targets / distros. So I am not not sure how we can define them once for the swimlane where they need to vary for various jobs inside that swimlane.

concourse/pipelines/templates/jobs/test-cli-tpl.yml
denalex: I will add "optional" description. But I don't think we should combine -- the line here is our API doc (similar to a Javadoc), while line 8 explains what happens in line 9 where we set local variable based on the value coming via the "API"

concourse/pipelines/templates/dev_build_pipeline-tpl.yml
xinzweb: Thank you for the explanation. Let's keep the current design.

concourse/pipelines/templates/anchors/destroy-dataproc-tpl.yml
xinzweb: Yes, let's have a separate iteration to smooth out the inconsistency naming. For this PR, let's keep it for now.

concourse/pipelines/templates/macros/macros.j2
```diff
@@ -0,0 +1,108 @@
...
+        {% endif %}
+
+        {# adjust Greenplum release platform due to historical inconsistencies #}
+        {% if x.release_platform == 'el7' %}
+            {% do x.update({'greenplum_platform': 'rhel7'}) %}
+        {% elif x.release_platform == 'el8' and x.gp_ver < '7' %}
+            {% do x.update({'greenplum_platform': 'rhel8'}) %}
+        {% else %}
+            {% do x.update({'greenplum_platform': x.release_platform}) %}
+        {% endif %}
```
bradfordb-vmware, line 51: The way this conditional is currently written it makes it seem like there is something special about RHEL7, but that's only true indirectly. If `release_platform == 'el7'`, this implies that `gp_ver < 7`. I think logic is _ if GP 5 or GP 6 and release platform is Red Hat, then we should use `rhel` for the Greenplum platform_ -- the historical inconsistencies we are correcting for are Greenplum inconsistencies not Red Hat. Could we write this as

```jinja
{% if x.gp_ver in ['5', '6'] and x.release_platform.startswith('el') %}
    {% do x.update({'greenplum_platform': 'rh' ~ x.greenplum_platform %}
{% endif %}
```

concourse/pipelines/templates/resources/greenplum-package-tpl.yml
```diff
@@ -0,0 +1,15 @@
+{# Template for a Greenplum package resource definition.
+ #
+ # Expects a dictionary with the name 'x' to be set as defined in macros.j2 file
+ #}
+
+{# include Greenplum package only if both build and test platform are the same to avoid duplicates #}
```
bradfordb-vmware, line 6: Can we call out an example where build and test platform would not be the same?

concourse/pipelines/templates/resources/pxf-artifact-tpl.yml
```diff
@@ -0,0 +1,29 @@
+{# Template for a PXF build artifact resource definition. 
+ #
+ # Expects a dictionary with the name 'x' to be set as defined in macros.j2 file  
+ #}
+  
+{# include PXF artifact only if both build and test platform are the same to avoid duplicates #}
```
bradfordb-vmware, line 6: Can we call out an example where build and test platform would not be the same?

concourse/pipelines/templates/resources/pxf-dependencies-tpl.yml
```diff
@@ -0,0 +1,16 @@
+{# Template for a PXF dependencies resource definition.
+ #
+ # Expects a 'dependency_type' variable to be set with the values 'build' or 'automation'
+ #}
+
+{% if dependency_type not in ['build','automation'] %}
+    raise 'Invalid value of dependency_type variable: ' ~ dependency_type
+{% endif %}
```
bradfordb-vmware, line 8: There are other templates where the input is expected to be an enum (e.g., github-repo-tpl) but we don't do validation. Do we want to validate all enum inputs?

concourse/pipelines/templates/resources/registry-image-tpl.yml
```diff
@@ -0,0 +1,26 @@
...
+ #
+ # image_name       - name of the resource type
+ # image_repository - repository
+ # image_tag        - tag (optional)
+ # resource_type    - whether this is a resource type definition (true / false)
+ #}
+
+- name: [[image_name]]
+  type: registry-image
+{# add icon unless this renders for 'gcs' which is a resource_type instead of a resource #}
```
bradfordb-vmware, line 13:
```suggestion
{# add icon unless this is for a resource_type instead of a resource #}
```

concourse/pipelines/templates/jobs/build-tpl.yml
```diff
@@ -1,32 +1,30 @@
...
-      file: [[pxf_tarball_filename]]
-{% if config.gp_ver == '6'  and build_platform == 'rocky8' %}
-  - put: pxf_fdw_gp[[config.gp_ver]]_tarball_[[gpdb_os_platform]][[config.os_ver]]
+      file: dist/[[x.pxf_tarball_filename_regex]]
+{# PXF FDW Build Artifact for GP6 only (since PXF FDW extension is not packaged into pxf-gp6 tarball) #}
+{% if x.gp_ver == '6' and x.test_fdw %}
+  - put: pxf-fdw-gp[[x.gp_ver]]-tarball-[[x.release_platform]]
     params:
-      file: [[pxf_fdw_tarball_filename]]
+      file: dist/[[x.pxf_fdw_tarball_filename_regex]]
```
yimingli-vmware, line 29: why now we need this `dist` directory?

yimingli-vmware (approved):

concourse/pipelines/templates/resources/pxf-dependencies-tpl.yml
denalex: while this is what I wanted to do, it actually does not work (just renders `raise ...`), so I will remove this for now.

concourse/pipelines/templates/anchors/slack-alert-tpl.yml
```diff
@@ -0,0 +1,18 @@
+{# Template for an anchor that sends a Slack alert.
+ #
+ # Expects a 'use_slack' variable to be set to true to render the context
+ #}
+
+{% if use_slack %}
```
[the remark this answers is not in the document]
denalex: let's keep it here since there are no usages of this conditional in the dev pipeline and having it here makes it self-contained such that it determines itself where it needs to render the content or not

concourse/pipelines/templates/jobs/build-tpl.yml
denalex: we are (and were) picking the tarball from the `dist/` directory, before it was baked into the file name, but now it is more explicit and the filename does not have it.

bradfordb-vmware (approved):

PR #1058, opened 2023-11-29
Add RHEL9 Support

ashuka24:
This PR adds build and test jobs to both dev and pxf-build pipelines.

It also adds a GP6 Multinode job for RHEL9 which is configurable for the dev pipeline using MULTINODE_EL9

In order to provide support for EL9, we also have to update the crypto policies in the remote images and removed the use of the xterm terminal in order to run automation.

### DIFF
--- a/automation/src/main/java/org/greenplum/pxf/automation/components/common/cli/PivotalSshRsa.java
+++ b/automation/src/main/java/org/greenplum/pxf/automation/components/common/cli/PivotalSshRsa.java
@@ -8,7 +8,7 @@ import ch.ethz.ssh2.Connection;
 
 /**
  * Extends {@link SSH} to connect using private key if exists instead of password connection.
- * 
+ *
  */
 public class PivotalSshRsa extends SSH {
 
@@ -46,7 +46,7 @@ public class PivotalSshRsa extends SSH {
 			try {
 				if (privateKeyFile != null && privateKeyFile.isFile()) {
 					System.out.println("Connecting using Private Key");
-					
+
 					// connect using private key
 					isAuthenticated = conn.authenticateWithPublicKey(username, privateKeyFile, "");
 				} else {
@@ -65,11 +65,7 @@ public class PivotalSshRsa extends SSH {
 		/* Create a session */
 		sess = conn.openSession();
 
-		if (xtermTerminal) {
-			sess.requestPTY("xterm", 80, 24, 640, 480, null);
-		} else {
-			sess.requestPTY("dumb", 200, 50, 0, 0, null);
-		}
+		sess.requestDumbPTY();
 
 		sess.startShell();
 
--- a/automation/src/main/java/org/greenplum/pxf/automation/components/common/cli/PivotalSshRsa.java
+++ b/automation/src/main/java/org/greenplum/pxf/automation/components/common/cli/PivotalSshRsa.java
@@ -65,6 +65,13 @@ public class PivotalSshRsa extends SSH {
 		/* Create a session */
 		sess = conn.openSession();
 
+		/*
+		 PXF CLI uses colored output that was causing extraneous output when using xterm terminal in automation
+		 tests on RHEL9. This does not happen when using a dumb terminal. This is ok because PXF CLI is currently
+		 not terminal aware and will output color regardless of which terminal is in use.
+
+		 Should the PXF CLI become terminal-aware, it may then be necessary to use different terminals.
+		 */
 		sess.requestDumbPTY();
 
 		sess.startShell();
--- a/automation/src/main/java/org/greenplum/pxf/automation/components/common/cli/PivotalSshRsa.java
+++ b/automation/src/main/java/org/greenplum/pxf/automation/components/common/cli/PivotalSshRsa.java
@@ -66,11 +66,9 @@ public class PivotalSshRsa extends SSH {
 		sess = conn.openSession();
 
 		/*
-		 PXF CLI uses colored output that was causing extraneous output when using xterm terminal in automation
-		 tests on RHEL9. This does not happen when using a dumb terminal. This is ok because PXF CLI is currently
-		 not terminal aware and will output color regardless of which terminal is in use.
-
-		 Should the PXF CLI become terminal-aware, it may then be necessary to use different terminals.
+		 There are commands in automation that are terminal-aware and return colored output. This was
+		 causing extraneous output when using xterm terminal in automation tests on RHEL9, causing tests to fail.
+		 This does not happen when using a "dumb" PTY as the commands will not emit ANSI escape sequences.
 		 */
 		sess.requestDumbPTY();
 
--- a/concourse/README.md
+++ b/concourse/README.md
@@ -95,7 +95,7 @@ To deploy dev pipeline against gpdb 5X_STABLE and 6X_STABLE branches, use:
 make -C "${HOME}/workspace/pxf/concourse" dev
 ```
 
-To deploy multi-node dev pipeline, you can specify either the `MULTINODE_EL7` for EL7, `MULTINODE_EL8` for EL8 or
+To deploy multi-node dev pipeline, you can specify either the `MULTINODE_EL7` for EL7, `MULTINODE_EL8` for EL8, `MULTINODE_EL9` for EL9 or
 `MULTINODE_NO_IMPERSONATION` for EL7, which will also run CLI tests:
 
 ```shell
--- a/concourse/README.md
+++ b/concourse/README.md
@@ -95,8 +95,11 @@ To deploy dev pipeline against gpdb 5X_STABLE and 6X_STABLE branches, use:
 make -C "${HOME}/workspace/pxf/concourse" dev
 ```
 
-To deploy multi-node dev pipeline, you can specify either the `MULTINODE_EL7` for EL7, `MULTINODE_EL8` for EL8, `MULTINODE_EL9` for EL9 or
-`MULTINODE_NO_IMPERSONATION` for EL7, which will also run CLI tests:
+To deploy multi-node dev pipeline, you can specify the following options
+* `MULTINODE_EL7=<true|false>` for EL7
+* `MULTINODE_EL8=<true|false>` for EL8
+* `MULTINODE_EL9=<true|false>` for EL9
+* `MULTINODE_NO_IMPERSONATION=<true|false>` for EL7, which will also run CLI tests
 
 ```shell
 MULTINODE_EL7=true make -C "${HOME}/workspace/pxf/concourse" dev
--- a/concourse/pipelines/templates/build_pipeline-tpl.yml
+++ b/concourse/pipelines/templates/build_pipeline-tpl.yml
@@ -329,6 +329,8 @@ jobs:
 
 ## ---------- Compatibility Testing ----------
 {% call(x) macros.for_each_config(build_test_pxf_combinations) %}
+    {# rocky9 is a new platform and there are no previous versions of gpdb to test against.
+     # this conditional can be removed once there are at least gp_num_versions of rocky9 gpdb builds #}
     {% if x.os_ver == '9' %}
         {% do x.update({'gp_num_versions': 1}) %}
     {% else %}
--- a/concourse/pipelines/templates/build_pipeline-tpl.yml
+++ b/concourse/pipelines/templates/build_pipeline-tpl.yml
@@ -51,9 +51,11 @@
     {'gp_ver': '5', 'build_os': 'centos', 'test_os': 'centos', 'os_ver': '7',     'test_fdw': false, 'test_file': false, 'test_cli': false, 'test_multi': false, 'test_features': [], 'generate_release_tarball': true},
     {'gp_ver': '6', 'build_os': 'centos', 'test_os': 'centos', 'os_ver': '7',     'test_fdw': false, 'test_file': true , 'test_cli': false, 'test_multi': true , 'test_features': supported_features, 'generate_release_tarball': true},
     {'gp_ver': '6', 'build_os': 'rocky' , 'test_os': 'rocky' , 'os_ver': '8',     'test_fdw': true , 'test_file': false, 'test_cli': false, 'test_multi': false, 'test_features': [], 'generate_release_tarball': false},
+    {'gp_ver': '6', 'build_os': 'rocky' , 'test_os': 'rocky' , 'os_ver': '9',     'test_fdw': false, 'test_file': false, 'test_cli': false, 'test_multi': true , 'test_features': [], 'generate_release_tarball': false},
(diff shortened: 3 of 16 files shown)


xinzweb (approved):
LGTM.

Feel free to separate the commit for the USER and BRANCH in a separate PR.

.github/workflows/create-release-on-tag.yml
```diff
@@ -22,13 +22,13 @@ jobs:
...
           body: |
-            ## 6.8.0 (09/25/2023)
+            ## 6.9.0 (12/04/2023)
 
             ### Enhancements:
 
             - [#1013](https://github.com/greenplum-db/pxf/pull/1013) Bumped Azure Storage dependency to 5.5.0
             - [#1018](https://github.com/greenplum-db/pxf/pull/1018) Add pxf.service.kerberos.ticket-renew-window option to pxf-site.xml
-            - [#1019](https://github.com/greenplum-db/pxf/pull/1019) Add pushdown of NUMERIC and handling of CHAR and VARCHAR predicates for JDBC profile
+            - [#1058](https://github.com/greenplum-db/pxf/pull/1058) Add RHEL9 Support
```
denalex, line 31: this should be a part of `Bump version to 6.9.0` PR, not this one.

concourse/pipelines/templates/build_pipeline-tpl.yml
```diff
@@ -322,7 +329,11 @@ jobs:
 
 ## ---------- Compatibility Testing ----------
 {% call(x) macros.for_each_config(build_test_pxf_combinations) %}
-    {% do x.update({'gp_num_versions': gp_num_versions[x.gp_ver]}) %}
+    {% if x.os_ver == '9' %}
```
denalex, line 332: should we put a note here when we need to change this and to what logic ?

concourse/scripts/build.bash
```diff
@@ -97,7 +108,8 @@ function package_pxf_fdw() {
 install_gpdb
 # installation of GPDB from RPM/DEB doesn't ensure that the installation location will match the version
 # given in the gpdb_package, so set the GPHOME after installation
-GPHOME=$(find /usr/local/ -name "greenplum-db-${GPDB_VERSION}*")
+# In case we are testing a dev version (i.e: 6.25.3+dev.6.54a3437), GPDB_VERSION%%+* will remove any extra string from + onwards
```
denalex, line 111: should this be reversed once the RHEL9 releases since we will nt be testing with dev versions anymore ?

denalex (changes requested):
please remove the release notes stuff from this PR and maybe provide a summary in PR description what needs to be changed once GPDB RHEL9 releases.

otherwise looks good.

.github/workflows/create-release-on-tag.yml
ashuka24: That was my mistake. I committed on the wrong branch. I have removed it from this PR.

concourse/scripts/build.bash
ashuka24: Can we guarantee this? If releases have to be aligned, there is a chance that we have to build against dev versions in the future.

automation/src/main/java/org/greenplum/pxf/automation/components/common/cli/PivotalSshRsa.java
```diff
@@ -65,11 +65,14 @@ public void connect() throws IOException {
...
-		} else {
-			sess.requestPTY("dumb", 200, 50, 0, 0, null);
-		}
+		/*
+		 PXF CLI uses colored output that was causing extraneous output when using xterm terminal in automation
+		 tests on RHEL9. This does not happen when using a dumb terminal. This is ok because PXF CLI is currently
+		 not terminal aware and will output color regardless of which terminal is in use.
+
+		 Should the PXF CLI become terminal-aware, it may then be necessary to use different terminals.
+		 */
```
bradfordb-vmware, line 74: This comment is not correct -- we don't currently test PXF CLI with automation. The color codes were coming from other commands that were terminal-aware. By switching to a "dumb" PTY, those commands don't emit the ANSI escape sequences.

concourse/README.md
```diff
@@ -95,7 +95,7 @@ To deploy dev pipeline against gpdb 5X_STABLE and 6X_STABLE branches, use:
 make -C "${HOME}/workspace/pxf/concourse" dev
 ```
 
-To deploy multi-node dev pipeline, you can specify either the `MULTINODE_EL7` for EL7, `MULTINODE_EL8` for EL8 or
+To deploy multi-node dev pipeline, you can specify either the `MULTINODE_EL7` for EL7, `MULTINODE_EL8` for EL8, `MULTINODE_EL9` for EL9 or
```
bradfordb-vmware, line 98: Since we have four options now, I'd switch to a bulleted list:

```
To deploy multi-node dev pipeline, you can specify

* `MULTINODE_EL7=<true|false>` for EL7
* `MULTINODE_EL8=<true|false>` for EL8
* `MULTINODE_EL9=<true|false>` for EL9
* `MULTINODE_NO_IMPERSONATION=<true|false>` for EL7

File: concourse/pipelines/templates/jobs/test-multinode-tpl.yml [hunk 6]
```diff
@@ -147,7 +150,16 @@
             gcp_project: ((ud/pxf/common/ipa-google-project-id))
       - task: generate-multinode-hadoop-cluster
         file: pxf_src/concourse/tasks/install_multinode_hadoop.yml
+    {% if x.os_ver == '9' %}
+    # OpenSSH SCP is deprecated in RHEL9 and it would affect the SCP commands that run in the
+    # install_multinode_hadoop.yml. See line 80 and 81 where we try to SCP down the hadoop and hive conf files.
```
bradfordb-vmware, line 155: Could we replace `scp` with `rsync` in `install_multinode_hadoop` and avoid the need for special casing here?

concourse/pipelines/templates/resources/greenplum-rocky9-package-tpl.yml
```diff
@@ -0,0 +1,24 @@
...
+    regexp: maintenance/release-candidates/gpdb6/greenplum-db-server-(6.*)-rhel9-x86_64.rpm
+
+{% elif x.os_ver == '9' and x.gp_ver == '7' %}
+- name: gpdb7-el9-rpm-latest-0
+  type: gcs
+  icon: google-drive
+  source:
+    bucket: ((ud/pxf/common/gpdb-concourse-resources-prod-bucket-name))
+    json_key: ((concourse-gcs-resources-service-account-key))
+    regexp: server/release-candidates/gpdb7/greenplum-db-server-(7.*)-el9-x86_64.rpm
```
bradfordb-vmware, line 23: It looks like the only difference between GP6 and GP7 is `maintenance` vs `server` -- could we have a variable `path_prefix` that we set depending on `gp_ver` and then just have one resource definition?

concourse/scripts/build.bash
```diff
@@ -11,14 +11,25 @@ GPDB_VERSION=$(<"${GPDB_PKG_DIR}/version")
...
+        # extract the OS major version
+        os_major_version=$(cat /etc/os-release | grep VERSION_ID | tr -dc '0-9.'| cut -d \. -f1)
+
+        # there is only a GPDB server RPM available for RHEL9
+        if [[ $os_major_version -eq "9" ]]; then
+            gpdb_file_name="greenplum-db-server"
+        else
+            # if no os major version is found, we will just fall into the default case here
+            gpdb_file_name="greenplum-db"
+        fi
```
bradfordb-vmware, line 23:
```suggestion
        . /etc/os-release
        
        case "${VERSION_ID}" in
        9.*) gpdb_file_name="greenplum-db-server" ;;
        7.*|8.*) gpdb_file_name="greenplum-db" ;;
        *)
            echo >&2 "unknown value for VERSION_ID: ${VERSION_ID}"
            exit 1
            ;;
        esac
```

concourse/scripts/build.bash
```diff
@@ -11,14 +11,25 @@ GPDB_VERSION=$(<"${GPDB_PKG_DIR}/version")
...
+
         # For GP7 and above, a new rhel8 & rocky8 distro identifier
         # (el8) has been introduced.
         if [[ ${GPDB_VERSION%%.*} -ge 7 ]]; then
             DISTRO_MATCHING_PATTERN="el"
         else
             DISTRO_MATCHING_PATTERN="r"
         fi
-        pkg_file=$(find "${GPDB_PKG_DIR}" -name "greenplum-db-${GPDB_VERSION}-${DISTRO_MATCHING_PATTERN}*-x86_64.rpm")
+        pkg_file=$(find "${GPDB_PKG_DIR}" -name "${gpdb_file_name}-${GPDB_VERSION}-${DISTRO_MATCHING_PATTERN}*-x86_64.rpm")
```
bradfordb-vmware, line 32: It turns out that `find` has `-regex` option for matching file names if we wanted to match `greenplum-db-x.y.z` and `greenplum-db-server-x.y.z`, we could do

```sh
find . -regextype egrep -regex './greenplum-db(-server)?.*.rpm'
```

Or we should just do

```sh
find . -name 'greenplum-db-server-${GPDB_VERSION}-${DISTRO_MATCHING_PATTERN}*-x86_64.rpm' -o -name 'greenplum-db-${GPDB_VERSION}-${DISTRO_MATCHING_PATTERN}*-x86_64.rpm'
```

concourse/scripts/build.bash
bradfordb-vmware: I don't know if we can guarantee it but if that becomes the norm, then I'd want to explore a less "hacky" approach. But that would require us to understand more deeply

* Why can't our CI scripts just use the `/usr/local/greenplum-db` symbolic link?
* Why is there a mismatch between GPDB's version inside of the RPM versus the version in the filename?

I'm in favor of leaving a `TODO` to revert this once GPDB EL9 support is GA across 6 and 7.

concourse/scripts/pxf_common.bash
```diff
@@ -250,9 +252,8 @@ function install_gpdb_package() {
 
 	# create symlink to allow pgregress to run (hardcoded to look for /usr/local/greenplum-db-devel/psql)
 	rm -rf /usr/local/greenplum-db-devel
-	# get version from the package file name
-	: "${pkg_file#*greenplum-db-}"
-	version=${_%%-*}
+	# in case of dev builds, get simplified version from the version file
```
bradfordb-vmware, line 255: Can we use a bash regex match instead of reading the version from the Concourse resource created `version` file?

concourse/scripts/test_pxf_multinode.bash
```diff
@@ -95,6 +95,25 @@ function add_testing_encoding() {
...
+	# update-crypto-policies is only available for Rocky9
+	if grep -i el9 /etc/os-release; then
+		ssh "${SSH_OPTS[@]}" gpadmin@cdw "
+			source ${GPHOME}/greenplum_path.sh &&
+			gpssh -f ~gpadmin/hostfile_all -v -u ${CCP_OS_USER} -s -e 'sudo update-crypto-policies --set DEFAULT:SHA1 && sudo systemctl restart sshd'
+			"
+		if [ $? -eq 0 ]; then
+			echo "Crypto policy updated and SSH restarted successfully."
+		else
+			echo "Failed to update crypto policy or restart SSH. Please check the logs for more information."
```
bradfordb-vmware, line 110: If we fail to update the crypto policy, should we bail? It seems like this error message would just get lost in all the other CI output. Would we even know/think to search for this?
```suggestion
			echo >&2 "Failed to update crypto policy or restart SSH. Please check the logs for more information."
			exit 1
```

automation/src/main/java/org/greenplum/pxf/automation/components/common/cli/PivotalSshRsa.java
ashuka24: I have updated this block, please take a second look

concourse/pipelines/templates/build_pipeline-tpl.yml
ashuka24: i have added a comment

concourse/pipelines/templates/resources/greenplum-rocky9-package-tpl.yml
ashuka24: As this is a temporary template and should be removed once we can pull rocky9 rpms from pivnet, I will leave this as is...

concourse/scripts/pxf_common.bash
ashuka24: We use the `version` provided by the Concourse resource in various other scripts as well. I think for this PR, it is fine to be consistent across scripts.

concourse/scripts/install_multinode_hadoop.bash
```diff
@@ -77,8 +79,10 @@ jq <"${metadata_path}" -r ".etc_hosts" >ipa_env_files/etc_hostfile
 echo "${hadoop_namenode_1}" >ipa_env_files/nn01
 echo "${hadoop_namenode_2}" >ipa_env_files/nn02
 mkdir -p ipa_env_files/conf
-scp "${hadoop_namenode_1}:\$HADOOP_PREFIX/etc/hadoop/*-site.xml" ipa_env_files/conf/
-scp "${hadoop_namenode_2}:\$HIVE_HOME/conf/hive-site.xml" ipa_env_files/conf/
+export HADOOP_PREFIX="$(ssh ${hadoop_namenode_1} "printenv HADOOP_PREFIX")"
+rsync "${hadoop_namenode_1}:${HADOOP_PREFIX}/etc/hadoop/*-site.xml" ipa_env_files/conf/
```
yimingli-vmware, line 83: can we add come context here why we need to replace `scp` with this `rsync`?

concourse/scripts/pxf_common.bash
```diff
@@ -215,6 +215,8 @@ function install_gpdb_package() {
 
 	local gphome python_dir python_version=2.7 export_pythonpath='export PYTHONPATH=$PYTHONPATH' pkg_file version
 	gpdb_package=${PWD}/${GPDB_PKG_DIR:-gpdb_package}
+	# obtain full version name
+	gpdb_version=$(<"${gpdb_package}/version")
```
yimingli-vmware, line 219: we need the full version name only for the dev builds?

yimingli-vmware (approved):

concourse/scripts/install_multinode_hadoop.bash
ashuka24: we switched to rsync because scp is deprecated in rocky9 and was causing issues. After taking a second look, we can overcome this by pulling down the remote variables before running scp as it is only remote variable expansion that is no longer available

concourse/scripts/pxf_common.bash
ashuka24: i have moved this variable to where it is being used so hopefully it will be more clear.

PR #1060, opened 2023-12-05
Explicitly set PG_CONFIG in ~/.pxfrc

ashuka24:
This PR removes exporting the `LD_LIBRARY_PATH` variable as well as removes `$GPHOME/bin` from `PATH` in `~/.pxfrc`. Instead we explicitly set `PG_CONFIG` in `~/.pxfrc`.

Greenplum 6.26.0 and earlier vendors `zstd` v1.3.7. Starting with Rocky 8.9, `libelf` requires `zstd` v1.4. This mismatch causes errors when running commands such as `rpmbuild` which depend upon `libelf` when we have `LD_LIBRARY_PATH` pointing to the Greenplum vendored packages. 

Previously, we explicitly set `LD_LIBRARY_PATH` and `PATH` to include the GPDB vendored libraries and dependencies inside `~/.pxfrc` when building the PXF docker images. We would then source `$GPHOME/greenplum_path.sh` in the script to build PXF which would set these values again. 

However, we only need to know the location of `pg_config` when building PXF. When testing PXF, we source `$GPHOME/greenplum_path.sh` within the `Regress` class in automation. As such, explicitly exporting `LD_LIBRARY_PATH` and `PATH` to include the GPDB vendored libraries and dependencies is no longer needed in the docker images. We also do not need to explicitly source the greenplum path in the CI script either.

### DIFF
--- a/concourse/scripts/build.bash
+++ b/concourse/scripts/build.bash
@@ -50,6 +50,7 @@ function compile_pxf() {
 
     bash -c "
         source ~/.pxfrc
+        unset LD_LIBRARY_PATH
         VENDOR='${VENDOR}' LICENSE='${LICENSE}' make -C '${PWD}/pxf_src' ${MAKE_TARGET}
     "
 }
@@ -84,6 +85,7 @@ function package_pxf_fdw() {
     # build PXF FDW extension separately
     bash -c "
         source ~/.pxfrc
+        unset LD_LIBRARY_PATH
         make -C '${PWD}/pxf_src/fdw' stage
     "
     # get the filename of previously built main PXF tarball to use its full name as a suffix
--- a/concourse/scripts/build.bash
+++ b/concourse/scripts/build.bash
@@ -36,8 +36,6 @@ function install_gpdb() {
 }
 
 function compile_pxf() {
-    source "${GPHOME}/greenplum_path.sh"
-
     # CentOS releases contain a /etc/redhat-release which is symlinked to /etc/centos-release
     if [[ -f /etc/redhat-release ]]; then
         MAKE_TARGET="rpm-tar"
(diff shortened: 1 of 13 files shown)


xinzweb:
The change makes sense to me.

I don't have much context on where we need the shared library from GP to compile.
Is that only for the GP extensions?

concourse/scripts/build.bash
```diff
@@ -36,8 +36,6 @@ function install_gpdb() {
 }
 
 function compile_pxf() {
-    source "${GPHOME}/greenplum_path.sh"
```
bradfordb-vmware, line 39: This change is a little suspect to me -- when we compile the PXF extension (either external table or FDW, we should need `pg_config` in order to find the correct Greenplum headers and libraries to link against. If we don't source `greenplum_path.sh`, I don't know how our Makefile finds `pg_config`. Did we test these changes by rebuilding the container images and then running a dev pipeline using those new images?

ashuka24: Previously, we set `PATH` to explicitly include `GPHOME/bin` inside the `~/.pxfrc` file. The images were rebuilt and a dev pipeline was set. 

We have since updated this to explicitly export `PG_CONFIG` instead of sourcing `greenplum_path.sh`

bradfordb-vmware (approved):

concourse/scripts/pxf_common.bash
```diff
@@ -120,7 +120,7 @@ function run_pxf_automation() {
 		time make GROUP=${GROUP} test
 
 		# if the test is successful, create certification file
-		gpdb_build_from_sql=\$(psql -c 'select version()' | grep Greenplum | cut -d ' ' -f 6,8)
+		gpdb_build_from_sql=\$(source \$GPHOME/greenplum_path.sh && psql -c 'select version()' | grep Greenplum | cut -d ' ' -f 6,8)
```
xinzweb, line 123: If we only use the version in the certificate which is not used in the product, we can consider to remove it.

xinzweb (approved):
LGTM

denalex (approved):
Can we please add the comment in the PR summary on why this change was needed (shortly describe zstd version mismatch and how us importing LD_... was messing rpm up  ?

PR #1069, opened 2024-01-02
Add RHEL9 to the certification pipeline

ashuka24:
From PXF 6.9.0+, we now support RHEL9. As such, add it to the certification pipeline.


denalex (commented):
simple is beautiful, the fruits of the templatization

bradfordb-vmware (approved):

yimingli-vmware (approved):

PR #1072, opened 2024-01-10
Templatize TanzuNet artifacts pipeline

bradfordb-vmware:


concourse/pipelines/templates/get_pivnet_artifacts-tpl.yml
```diff
@@ -36,41 +36,40 @@ resources:
...
+    {% for p in config['platforms'] %}
+      {% set short_name = p.name.split('.')[0] %}
+- name: gpdb[[ config.gp_ver ]]_[[ short_name ]]_[[ p.artifact_type ]]_latest-[[ i ]]
   type: gcs
   icon: google-drive
   source:
     bucket: ((ud/pxf/common/pivnet-artifacts-bucket-name))
     json_key: ((ud/pxf/secrets/pxf-storage-service-account-key))
-    regexp: latest-[[i]]_gpdb6/greenplum-db-(.*)-rhel8-x86_64.rpm
+    regexp: latest-[[ i ]]_gpdb[[ config.gp_ver ]]/greenplum-db-(.*)-[[ p.name ]]-[[ p.arch ]].[[ p.artifact_type ]]
```
axue-broadcom, line 68: nit: all the other pipeline templates don't add extra spaces between `[[ ]]`. Are we adding it here now for clarity?

concourse/pipelines/templates/get_pivnet_artifacts-tpl.yml
```diff
@@ -36,41 +36,40 @@ resources:
...
-  type: gcs
-  icon: google-drive
-  source:
-    bucket: ((ud/pxf/common/pivnet-artifacts-bucket-name))
-    json_key: ((ud/pxf/secrets/pxf-storage-service-account-key))
-    regexp: latest-[[i]]_gpdb6/greenplum-db-(.*)-ubuntu18.04-amd64.deb
-{% endfor %}
+      {% endfor %}
+  {% endfor %}
+{% endfor -%}
```
axue-broadcom, line 72:
```suggestion
{% endfor %}

File: concourse/pipelines/templates/get_pivnet_artifacts-tpl.yml [hunk 3]
```diff
@@ -36,41 +36,40 @@ resources:
...
+    'platforms': [
+      {'name': 'rhel7', 'artifact_type': 'rpm', 'arch': 'x86_64'},
+      {'name': 'rhel8', 'artifact_type': 'rpm', 'arch': 'x86_64'},
+      {'name': 'ubuntu18.04', 'artifact_type': 'deb', 'arch': 'amd64'}
+    ]
+  }
+] %}
 
-- name: gpdb6_rhel8_rpm_latest-[[i]]
+{%- for config in supported_gpdb %}
```
axue-broadcom, line 58:
```suggestion
{% for config in supported_gpdb %}

File: concourse/pipelines/templates/get_pivnet_artifacts-tpl.yml [hunk 4]
```diff
@@ -36,41 +36,40 @@ resources:
...
+  {% for i in range(config.num_versions) %}
+    {% for p in config['platforms'] %}
```
axue-broadcom, line 60: nit: for consistency across template files, we have been using 4 space tabs for the jinja templating code, can we do that here too?

concourse/pipelines/templates/get_pivnet_artifacts-tpl.yml
```diff
@@ -108,96 +108,56 @@ jobs:
...
-      outputs:
-      - name: gpdb6_rhel7_rpm_latest-[[i]]
-      - name: gpdb6_rhel8_rpm_latest-[[i]]
-      - name: gpdb6_ubuntu18_deb_latest-[[i]]
-      params:
-        GPDB_VERSION: 6
-        LIST_OF_DIRS: gpdb6_rhel7_rpm_latest-[[i]],gpdb6_rhel8_rpm_latest-[[i]],gpdb6_ubuntu18_deb_latest-[[i]]
+        GPDB_VERSION: [[ config.gp_ver ]]
+        LIST_OF_DIRS: [[ x | join(',') ]]
         # yamllint disable rule:line-length
```
axue-broadcom, line 148: do we still need this yamllint

concourse/Makefile
```diff
@@ -226,21 +226,25 @@ set-cloudbuild-pipeline-base:
...
+		--config ./rendered/get_pivnet_artifacts.yml \
 		--var pxf-git-branch=$(BRANCH) \
-		${FLY_OPTION_NON-INTERACTIVE} || echo "Generated yaml has errors: check $${PIPELINE_FILE}"
+		${FLY_OPTION_NON-INTERACTIVE}
 
 	@echo using the following command to unpause the pipeline:
 	@echo -e "\t$(FLY_CMD) -t ${CONCOURSE} unpause-pipeline --pipeline ${PIVNET_PIPELINE_NAME}"
 
+rendered/get_pivnet_artifacts.yml: pipelines/templates/get_pivnet_artifacts-tpl.yml rendered
+	$(TEMPLATE_CMD) --template $(notdir $<) --output $@ --vars \
```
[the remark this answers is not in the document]
bradfordb-vmware: From [GNU Make - Functions for File Names](https://www.gnu.org/software/make/manual/html_node/File-Name-Functions.html):

> Extracts all but the directory-part of each file name

So this turns `pipelines/templates/get_pivnet_artifacts-tpl.yml` into `get_pivnet_artifacs-tpl.yml`. The `template_tool` utility doesn't take a path to a template, just the name of the template file. Instead of repeating the template name, we extract it from the name of the first prerequisite (`$<`). The Make variable `$@` is the file name of the target of the rule (in this case `rendered/get_pivnet_artifacts.yml`) (see [GNU Make - Automatic Variables](https://www.gnu.org/software/make/manual/html_node/Automatic-Variables.html)).

concourse/Makefile
```diff
@@ -226,21 +226,25 @@ set-cloudbuild-pipeline-base:
...
+		num_gpdb5_versions=$(NUM_GPDB5_VERSIONS) \
+		num_gpdb6_versions=$(NUM_GPDB6_VERSIONS)
+
+rendered:
+	install -m 0755 -d rendered
```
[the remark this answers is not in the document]
bradfordb-vmware: We are not installing the rendered file, this is ensuring that the directory (`-d`) exists with correct permissions (`-m 0755`) before we attempt to write files under that directory.

concourse/pipelines/templates/get_pivnet_artifacts-tpl.yml
bradfordb-vmware: I was looking at the official [Jinja Template Designer Documentation](https://jinja.palletsprojects.com/en/3.1.x/templates/) where they consistently use `{{ ... }}` for expressions. For some reason `template_tool` reconfigures Jinja to use `[[ ... ]]` instead of `{{ ... }}` but I think including the extra spaces aids in readability, especially when the expression includes `[` or `]`.

concourse/pipelines/templates/get_pivnet_artifacts-tpl.yml
bradfordb-vmware: This was intentional for controlling whitespace in the generate file [Jinja Template Designer Documentation - Whitespace Control](https://jinja.palletsprojects.com/en/3.1.x/templates/#whitespace-control).

concourse/pipelines/templates/get_pivnet_artifacts-tpl.yml
bradfordb-vmware: I kept this here to minimize the diff between the rendered template before and after this commit.

axue-broadcom (commented):
for the resources and resource types in the pivnet template, do we want to use the macros?
```
    [[macros.resource_type_registry_image('gcs', 'frodenas/gcs-resource', '')]]

    [[macros.resource_github_repo('pxf')]]
    
    [[macros.resource_registry_image('ubuntu18', 'gcr.io/data-gpdb-public-images/gpdb6-ubuntu18.04-build', 'latest')]]
```

concourse/Makefile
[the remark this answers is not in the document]
denalex: I thought this would be an "intemediate" PHONY target so that the file always gets generated and does not rely on Makefile mechanism to detect updated files. If we do that, we would not need to list `pipelines/templates/get_pivnet_artifacts-tpl.yml` as a pre-requisite and can just include the file name into the command.

concourse/Makefile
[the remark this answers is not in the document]
denalex: why can't this command be a part of the target above instead of being a separate target that doesn't actually do much ?

concourse/pipelines/templates/get_pivnet_artifacts-tpl.yml
```diff
@@ -36,41 +36,40 @@ resources:
...
+    {
+        'gp_ver': '6',
+        'num_versions': num_gpdb6_versions,
+        'platforms': [
+            {'name': 'rhel7', 'artifact_type': 'rpm', 'arch': 'x86_64'},
+            {'name': 'rhel8', 'artifact_type': 'rpm', 'arch': 'x86_64'},
+            {'name': 'ubuntu18.04', 'artifact_type': 'deb', 'arch': 'amd64'}
+        ]
+    }
+] %}
```
denalex, line 56: any reason to deviate from a swimlane-based configuration used in main pipelines ? while I like how this looks like structurally, you end up unfolding the structure and iterating over `platforms`, which are swimlanes in other pipelines. Again, consistency across the templates, where metadata is represented in the same (if not, similar) ways.

concourse/pipelines/templates/get_pivnet_artifacts-tpl.yml
```diff
@@ -36,41 +36,40 @@ resources:
...
 
-- name: gpdb6_rhel8_rpm_latest-[[i]]
+{%- for config in supported_gpdb %}
+    {% for i in range(config.num_versions) %}
+        {% for p in config['platforms'] %}
+            {% set short_name = p.name.split('.')[0] %}
+- name: gpdb[[ config.gp_ver ]]_[[ short_name ]]_[[ p.artifact_type ]]_latest-[[ i ]]
```
denalex, line 62: should this be a sub-template like other resources are ?

concourse/pipelines/templates/get_pivnet_artifacts-tpl.yml
denalex: for the reference -- I saw this and experimented with it, but omitted from the main pipeline templates not to overload developers with unnecessary details, since the produced templates are transient and not usually visible.

concourse/pipelines/templates/get_pivnet_artifacts-tpl.yml
```diff
@@ -108,96 +108,56 @@ jobs:
     - get: pivnet_cli
       passed: [download_latest_pivnet_cli]
       trigger: true
-{% for i in range(num_gpdb5_versions) %}
-    - get: gpdb5_rhel7_rpm_latest-[[i]]
-{% endfor %}
-{% for i in range(num_gpdb5_versions - 1, -1, -1) %}
-  - task: get_greenplum_product_files_gp5_latest-[[i]]
-    attempts: [[attempts]]
+    {% for i in range(config.num_versions) %}
```
denalex, line 111: having extra whitespace before / after `{% ..}` might lead to some unexpected consequences, since it gets rendered in the resulting tempate. that's why we always start them at the beginning of the line.

concourse/template_tool
```diff
@@ -77,6 +84,9 @@ def build_parser():
                             default="build_pipeline-tpl.yml",
                             help='Name of template to use, in concourse/pipelines/templates/')
 
+    pxf_parser.add_argument('-o', '--output',
+                            help='file to write pipeline to')
```
denalex, line 88:
```suggestion
                            help='file to write the output to')
```

concourse/pipelines/templates/get_pivnet_artifacts-tpl.yml
bradfordb-vmware: Mostly because this pipeline doesn't currently have a concept of swimlanes. I can flatten the structure to something like:

```
{'gp_ver': '5', 'num_versions': num_gpdb_5_versions, 'name': 'rhel7', 'artifact_type': 'rpm', 'arch': 'x86_64'}
{'gp_ver': '6', 'num_versions': num_gpdb_6_versions, 'name': 'rhel7', 'artifact_type': 'rpm', 'arch': 'x86_64'}
{'gp_ver': '6', 'num_versions': num_gpdb_6_versions, 'name': 'rhel8', 'artifact_type': 'rpm', 'arch': 'x86_64'}
{'gp_ver': '6', 'num_versions': num_gpdb_6_versions, 'name': 'ubuntu18.04', 'artifact_type': 'deb', 'arch': 'amd64'}
```

But this would have the result of changing from

```
for i in range():
   for p in platform:
```

into

```
for p in platform:
    for i in range():
```

which would be semantically the same but structurally different.

concourse/Makefile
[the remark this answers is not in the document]
bradfordb-vmware: Added `.PHONY` to the target.

concourse/Makefile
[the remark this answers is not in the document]
bradfordb-vmware: It can be; however if we introduce intermediate targets for other pipelines, they will all need the 'rendered' directory to exist before. In that case, we'd have to repeat this command in all those intermediate targets or extract that common logic to a separate target they all can depend on.

concourse/pipelines/templates/get_pivnet_artifacts-tpl.yml
bradfordb-vmware: Extra spaces removed from `[[]]`

concourse/pipelines/templates/get_pivnet_artifacts-tpl.yml
bradfordb-vmware: To expand on the previous point some more: the pipeline currently has two top-level jobs:

1. `get_greenplum_product_files_gp5`
1. `get_greenplum_product_files_gp6`

Updating this pipeline to use a swimlane-based configuration would involve creating separate jobs for each combination of Greenplum major version and platform. Maybe that is what we want to do but that feels like a much larger task outside the scope of templatizing the pipeline to add in GP6 RHEL9.

concourse/pipelines/templates/get_pivnet_artifacts-tpl.yml
bradfordb-vmware: I'm not sure I completely understand -- if I look at "build_pipeline-tpl.yml", I see several indented `{% %}` statements. Additionally, when I diff the generated pipeline from main with the pipeline generated from this PR, I don't see any spurious whitespace.

concourse/pipelines/templates/get_pivnet_artifacts-tpl.yml
```diff
@@ -78,7 +64,8 @@ jobs:
   - in_parallel:
     - get: twice-a-day
       trigger: true
-    - get: pxf_src
+    - get: pxf-src
```
yimingli-vmware, line 67: Should this line to be `pxf-src` or remain `pxf_src`? If I remember correctly that Alex said he want to rename `pxf_src` to be `pxf-src`, but he cannot change the name for the `get` step

bradfordb-vmware:
@denalex @axue-broadcom @yimingli-vmware I have addressed all comments and pushed up the latest set of changes for this PR. Please re-review.

yimingli-vmware (approved):

axue-broadcom (approved):

concourse/pipelines/templates/macros/macros.j2
```diff
@@ -74,20 +74,34 @@
...
+    {% for gp_ver, platforms in configs | groupby('gp_ver') %}
+        {% set x = dict({'gp_ver': gp_ver}) %}
+        {% do x.update({'platforms': []}) %}
+        {% for p in platforms %}
+            {% set y = dict() %}
+            {% if p.build_os in [ 'centos', 'rocky' ] %}
+                {% do y.update({'greenplum_platform': 'rhel' ~ p.os_ver}) %}
+                {% do y.update({'release_platform':   'rhel' ~ p.os_ver}) %}
+                {% do y.update({'pkg_type':           'rpm'}) %}
+                {% do y.update({'pkg_arch':           'x86_64'}) %}
```
denalex, line 95: maybe add `gpdb_package_resource_name` (see the logic in for_each_config() above) so that you don't have to calculate it in templates where they will just append an index to it, if needed

concourse/pipelines/templates/macros/macros.j2
```diff
@@ -74,20 +74,34 @@
...
-    {% set supported_gp_vers = configs | map(attribute='gp_ver') | list | unique %}
-    {% for gp_ver in supported_gp_vers %}
-            {# set local variables for a given loop iteration #}
-            {% set x = dict({'gp_ver': gp_ver}) %}
-        {# render the result of callback passing the resulting context specific to this loop iteration #}
+  Macro that takes an array of config dictionaries and groups them by the Greenplum version.
+  For each Greenplum version, it
+    - Creates a dictionary with the following attributes
+      - 'gp_ver'    : the Greenplum majar version
+      - 'platforms' : a list of all platforms for this Greenplum major version
```
denalex, line 81: should we describe what attributes the platform list entry dictionary has ?

concourse/pipelines/templates/resources/pivnet_cli-tpl.yml
```diff
@@ -0,0 +1,10 @@
+{# Template for PivNet CLI stored in GCS.
```
denalex, line 1:
```suggestion
{# Template for PivNet CLI resource stored in GCS.
```

concourse/Makefile
```diff
@@ -226,21 +226,24 @@ set-cloudbuild-pipeline-base:
...
-		--config "$${PIPELINE_FILE}" \
+		--config ./rendered/get_pivnet_artifacts.yml \
 		--var pxf-git-branch=$(BRANCH) \
-		${FLY_OPTION_NON-INTERACTIVE} || echo "Generated yaml has errors: check $${PIPELINE_FILE}"
+		${FLY_OPTION_NON-INTERACTIVE}
 
 	@echo using the following command to unpause the pipeline:
 	@echo -e "\t$(FLY_CMD) -t ${CONCOURSE} unpause-pipeline --pipeline ${PIVNET_PIPELINE_NAME}"
 
+.PHONY: rendered/get_pivnet_artifacts.yml
```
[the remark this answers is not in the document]
bradfordb-vmware: Ranemed the target to `generate-pivnet-pipeline` to match with `set-pivnet-pipeline`. I've also renamed "rendered" to "generated" to match with the target name.

concourse/pipelines/templates/jobs/get_greenplum_product_files.yml
```diff
@@ -0,0 +1,71 @@
...
+    - get: pxf_src
+      resource: pxf-src
+      passed: [download_latest_pivnet_cli]
+    - get: ubuntu18
+    - get: pivnet_cli
+      passed: [download_latest_pivnet_cli]
+      trigger: true
+    {% for i in range(x.gp_num_versions) %}
+        {% for p in x.platforms %}
+    - get: [[ "gpdb%s_%s_%s_latest-%d" | format(x.gp_ver, p.release_platform, p.pkg_type, i)]]
```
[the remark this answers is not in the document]
bradfordb-vmware: Removed spaces after '[[' and before ']]'.

concourse/pipelines/templates/jobs/get_greenplum_product_files.yml
```diff
@@ -0,0 +1,71 @@
...
+        {% endfor %}
+        {% if i > 0 %}
+      # save a download from pivnet if not fetching latest
+            {% for p in x.platforms %}
+      - name: [[ "gpdb%s_%s_%s_latest-%d" | format(x.gp_ver, p.release_platform, p.pkg_type, i-1)]]
+            {% endfor %}
+        {% endif %}
+      outputs:
+        {% for p in x.platforms %}
+      - name: [[ "gpdb%s_%s_%s_latest-%d" | format(x.gp_ver, p.release_platform, p.pkg_type, i)]]
```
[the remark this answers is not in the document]
bradfordb-vmware: I've moved the string creation out of this template and into macros.j2. I was originally avoiding changing the pipeline resource name from using '_' to '-' but since this require building two similar strings, adding in extra pipeline config, and the convention in other pipelines has been to use '-' instead of '_' I've simplified to just one style (with '-').

concourse/pipelines/templates/macros/macros.j2
bradfordb-vmware: Good suggestion. I've added `gpdb_package_resource_name` and `gpb_package_file_name` into this macro so the template can just reference these attributes.

concourse/pipelines/templates/macros/macros.j2
bradfordb-vmware: Added descriptions for the set of attributes expected under `platforms`.

concourse/pipelines/templates/macros/macros.j2
```diff
@@ -74,20 +74,41 @@
...
+                {% do y.update({'pkg_arch':           'x86_64'}) %}
+            {% else %}
+                {% do y.update({'greenplum_platform': 'ubuntu' ~ p.os_ver}) %}
+                {% do y.update({'release_platform':   'ubuntu' ~ p.os_ver}) %}
+                {% do y.update({'pkg_type':           'deb'}) %}
+                {% do y.update({'pkg_arch':           'amd64'}) %}
+            {% endif %}
+            {# pipelines use '-' for resource names but task inputs expect '_'#}
+            {% do y.update({'resource_name':   "gpdb%s-%s-%s-latest" | format(x.gp_ver, y.release_platform, y.pkg_type)}) %}
+            {% do y.update({'product_name':    "greenplum-db-GPDB_VERSION-%s-%s.%s" | format(y.greenplum_platform, y.pkg_arch, y.pkg_type)}) %}
```
denalex, line 108: can we name the keys with more context ?

concourse/pipelines/templates/jobs/get_greenplum_product_files.yml
```diff
@@ -0,0 +1,70 @@
...
+        PIVNET_API_TOKEN: ((ud/pxf/secrets/pivnet-uaa-refresh-token))
+        PIVNET_CLI_DIR: pivnet_cli
+        PRODUCT_SLUG: pivotal-gpdb
+        VERSIONS_BEFORE_LATEST: [[i]]
+      run:
+        path: pxf_src/concourse/scripts/get_product_files.bash
+        {% for p in x.platforms %}
+  - put: gpdb[[x.gp_ver]]_[[p.release_platform]]_[[p.pkg_type]]_latest-[[i]]
+    params:
+      file: gpdb[[x.gp_ver]]_[[p.release_platform]]_[[p.pkg_type]]_latest-[[i]]/greenplum-db-[[x.gp_ver]].*-[[p.greenplum_platform]]-[[p.pkg_arch]].[[p.pkg_type]]
```
denalex, line 68: can these be now replaced ?

concourse/pipelines/templates/jobs/get_greenplum_product_files.yml
```diff
@@ -0,0 +1,70 @@
...
+    attempts: [[attempts if attempts is not none else 1]]
+    # this task will fail on centos6
+    image: ubuntu18
+    config:
+      platform: linux
+      inputs:
+      - name: pxf_src
+      - name: pivnet_cli
+        {% for p in x.platforms %}
+      - name: [["%s-%d" | format(p.resource_name, i)]]
```
denalex, line 37: I would not have used format here, but that's a personal preference

PR #1080, opened 2024-01-29
Add support for UUID to JDBC profile

bradfordb-vmware:


tatb-vmware (approved):

automation/sqlrepo/features/multi_user/expected/query01.ans
```diff
@@ -35,14 +35,15 @@ SELECT t1, t2, num1, dub1, dec1, tm, r, bg, b, tn, sml, dt, vc1, c1, encode(bin,
  row23_varchar_null   | s_16 |   11 |   37 |    0.123456789 | 2013-07-23 21:00:00 |  7.7 | 23456789 | f | 11 | 1100 | 2015-03-26 |       | EUR | 1
  row24_char_null      | s_16 |   11 |   37 |    0.123456789 | 2013-07-23 21:00:00 |  7.7 | 23456789 | f | 11 | 1100 | 2015-03-27 | abcde |     | 1
  row25_binary_null    | s_16 |   11 |   37 |    0.123456789 | 2013-07-23 21:00:00 |  7.7 | 23456789 | f |  1 | 1100 | 2015-03-28 | abcde | USD |
+ row26_uuid_null      | s_16 |   11 |   37 |    0.123456789 | 2013-07-23 21:00:00 |  7.7 | 23456789 | f |  1 | 1100 | 2015-03-29 | abcde | USD | 1
```
denalex, line 38: any reason the new field has not been added to the SELECT clause ?

automation/sqlrepo/features/multi_user/expected/query02.ans
```diff
@@ -47,14 +47,15 @@ SELECT t1, t2, num1, dub1, dec1, tm, r, bg, b, tn, sml, dt, vc1, c1, encode(bin,
  row23_varchar_null   | s_16 |   11 |   37 |    0.123456789 | 2013-07-23 21:00:00 |  7.7 | 23456789 | f | 11 | 1100 | 2015-03-26 |       | EUR | 1
  row24_char_null      | s_16 |   11 |   37 |    0.123456789 | 2013-07-23 21:00:00 |  7.7 | 23456789 | f | 11 | 1100 | 2015-03-27 | abcde |     | 1
  row25_binary_null    | s_16 |   11 |   37 |    0.123456789 | 2013-07-23 21:00:00 |  7.7 | 23456789 | f |  1 | 1100 | 2015-03-28 | abcde | USD |
+ row26_uuid_null      | s_16 |   11 |   37 |    0.123456789 | 2013-07-23 21:00:00 |  7.7 | 23456789 | f |  1 | 1100 | 2015-03-29 | abcde | USD | 1
```
denalex, line 50: same question, `u` is missing from SELECT

automation/src/test/java/org/greenplum/pxf/automation/features/jdbc/JdbcTest.java
```diff
@@ -105,7 +108,7 @@ private void prepareTypesData() throws Exception {
         gpdbNativeTableTypes.setDistributionFields(new String[]{"t1"});
         gpdb.createTableAndVerify(gpdbNativeTableTypes);
         gpdb.copyFromFile(gpdbNativeTableTypes, new File(localDataResourcesFolder
-                + "/gpdb/" + gpdbTypesDataFileName), "E'\\t'", "E'\\\\N'", true);
+                + "/gpdb/" + gpdbTypesDataFileName), ",", "E'\\\\N'", true);
```
denalex, line 111: why was the change from tab-separated ?

automation/src/test/resources/data/gpdb/gpdb_types.txt
```diff
@@ -1 +1,26 @@
...
row19_bool_null	s_16	11	37	0.123456789	7/23/13 21:00	7.7	23456789	\N	11	1100	3/23/15	abcde	USD	1
row20_tinyint_null	s_16	11	37	0.123456789	7/23/13 21:00	7.7	23456789	FALSE	\N	1100	3/24/15	abcde	UAH	1
row21_smallint_null	s_16	11	37	0.123456789	7/23/13 21:00	7.7	23456789	FALSE	11	\N	3/25/15	abcde	USD	1
row22_date_null	s_16	11	37	0.123456789	7/23/13 21:00	7.7	23456789	FALSE	11	1100	\N	abcde	EUR	1
row23_varchar_null	s_16	11	37	0.123456789	7/23/13 21:00	7.7	23456789	FALSE	11	1100	3/26/15	\N	EUR	1
row24_char_null	s_16	11	37	0.123456789	7/23/13 21:00	7.7	23456789	FALSE	11	1100	3/27/15	abcde	\N	1
row25_binary_null	s_16	11	37	0.123456789	7/23/13 21:00	7.7	23456789	FALSE	1	1100	3/28/15	abcde	USD	\N
\ No newline at end of file
+row1,s_6,1,6,1.23456,2013-07-13 21:00:00,7.7,23456789,f,1,10,2015-03-06,abcd,USD,\x31,a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a11
+row10,s_15,10,15,45678.00002,2013-07-22 21:00:00,7.7,23456789,t,10,1000,2015-03-14,abcde,USD,\x30,a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a11
```
denalex, line 2: should we vary the uuid value (say the last 2 digits) among rows ?

automation/src/test/resources/data/gpdb/gpdb_types_ex.txt
```diff
@@ -1 +1,3 @@
-row3	s_8	3	8	-1.23456	7/15/13 21:00	9.7	23456789	FALSE	3	300	3/7/15	abcde	USD	3row6	s_11	6	11	1234.889	7/18/13 21:00	12.7	23456789	TRUE	6	600	3/10/15	abcde	USD	6row9	s_14	9	14	23457.1	7/21/13 21:00	7.7	23456789	FALSE	9	900	3/13/15	abcde	UAH	9
\ No newline at end of file
+row9,s_14,9,14,23457.1,2013-07-21 21:00:00,7.7,23456789,f,9,900,2015-03-13,abcde,UAH,\x39,a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a11
+row3,s_8,3,8,-1.23456,2013-07-15 21:00:00,9.7,23456789,f,3,300,2015-03-07,abcde,USD,\x33,a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a11
+row6,s_11,6,11,1234.889,2013-07-18 21:00:00,12.7,23456789,t,6,600,2015-03-10,abcde,USD,\x36,a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a11
```
denalex, line 3: looks like the values for the first column changed as well and binary values were added ?

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -236,6 +237,9 @@ public List<OneField> getFields(OneRow row) throws SQLException {
                                         column));
                     }
                     break;
+                case UUID:
+                    value = result.getObject(colName, java.util.UUID.class);
```
denalex, line 241: how confident are we this will be supported by JDBC drivers, as opposed to getting as a String and converting ourselves ?

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -451,6 +458,9 @@ public static void decodeOneRowToPreparedStatement(OneRow row, PreparedStatement
                         }
                     }
                     break;
+                case UUID:
+                    statement.setObject(i, field.val);
```
denalex, line 462: same question in reverse - driver support vs, setting as String. Also, if GP type is UUID but the target DB table defined the column as text, will this be able to still work ? Customers might not want to define a field as UUID in their external systems.

yimingli-vmware (approved):

automation/src/test/resources/data/gpdb/gpdb_types_ex.txt
bradfordb-vmware: I don't think the column values changed -- I think the order of the rows was changed. I can fix the order of the rows. The binary values weren't added but they were changed (unintentionally) from octal to hex. I'll fix that to minimize the changes.

automation/src/test/java/org/greenplum/pxf/automation/features/jdbc/JdbcTest.java
bradfordb-vmware: No particular reason. The original file was using only carriage-return to separate the lines so I decided to adjust the format to use line feeds so that when view the file with `less` or `cat`, it was easier to see the separate lines. I just ended up regenerating the data as CSV instead of TSV.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
bradfordb-vmware: I've tested against GPDB/Postgres and against Microsoft SQL server (manually). I can manually test/verify with sqlite and MySQL.

I had considered adding a configuration option to control whether we use `getObject` or `getString`. What are your thoughts?

automation/sqlrepo/features/multi_user/expected/query01.ans
bradfordb-vmware: I have add `u` to the `SELECT` clause.

automation/sqlrepo/features/multi_user/expected/query02.ans
bradfordb-vmware: I have added `u` to the `SELECT` clause.

automation/src/test/java/org/greenplum/pxf/automation/features/jdbc/JdbcTest.java
bradfordb-vmware: I have reverted the delimiter change back to using tabs.

automation/src/test/resources/data/gpdb/gpdb_types_ex.txt
bradfordb-vmware: I have restored the original ordering of the rows.

automation/sqlrepo/features/multi_user/expected/query01.ans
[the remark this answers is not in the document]
bradfordb-vmware: I have added `u` to the `SELECT` clause.

automation/src/test/resources/data/gpdb/gpdb_types.txt
bradfordb-vmware: I have updated the data to vary the last two characters of the UUID to be increasing from 0x10 until 0x29.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
bradfordb-vmware: - Postgres has a UUID column type and its JDBC driver supports getObject(..., UUID.class)
    - It also works with getString()
- Microsoft SQL Server does not have UUID column but its JDBC driver supports getObject(..., UUID.class)
    - It also works with getString()
- MySQL does not have a UUID column and its JDBC driver does not support getObject(..., UUID.class)
    - if the MySQL table is using char/varchar, then we can read/write a GPDB UUID column from/to MySQL by using getString()
- sqlite 3 does not have a UUID column and its JDBC driver does not support getObject(..., UUID.class)
    - if the sqlite3 table using char/varchar, then the column can be read with getString()
    - if the sqlite3 table using blob, then the GPDB table with uuid column cannot be read; it must be created with bytea
- It looks like Oracle stores UUID as STRING [1](https://docs.oracle.com/en/database/other-databases/nosql-database/23.1/sqlreferencefornosql/using-uuid-data-type.html#GUID-D2BEE2A1-BF6D-4F4A-8EF4-059A6C8FBAA6) so its probably safest to use getString() instead of getObject()

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
bradfordb-vmware: If the destination table has type UUID but we try to insert a string by calling `setString()`, the Postgres JDBC driver fails with:

> column "u" is of type uuid but expression is of type character varying

If the target database table is defined as varchar, wouldn't the user need to create the GPDB external table with a similar type?

denalex (approved):

PR #1081, opened 2024-02-05
Add automation test cases for JDBC profile DATE_WIDE_RANGE feature

tatb-vmware:
This PR adds automation tests for JDBC profile's DATE_WIDE_RANGE feature which were added in this PR: https://github.com/greenplum-db/pxf/pull/960. This increases test coverage to improve confidence that the feature works as intended and ensure that future JDBC changes don't introduce regressions.

This PR also fixed a bug with inconsistent behavior in JDBC between external and foreign tables where external tables can handle more than 4 digit years in timestamps but foreign tables cannot.


automation/sqlrepo/features/jdbc/writable/expected/query01.ans
```diff
@@ -5,36 +5,36 @@ ALTER TABLE gpdb_types_target ADD CONSTRAINT gpdb_types_target_t1_key UNIQUE (t1
 NOTICE:  ALTER TABLE / ADD UNIQUE will create implicit index "gpdb_types_target_t1_key" for table "gpdb_types_target"
 ALTER TABLE
 
-INSERT INTO pxf_jdbc_writable SELECT * FROM gpdb_types;
+INSERT INTO pxf_jdbc_writable SELECT * FROM gpdb_types_without_timestamptz;
```
yimingli-vmware, line 8: Can we still use `gpdb_types` so that the output table will have values for the `tmz` column?

automation/sqlrepo/features/multi_user/sql/query01.sql
```diff
@@ -12,12 +12,13 @@ GRANT ALL ON TABLE pxf_jdbc_readable TO PUBLIC;
 DROP ROLE IF EXISTS testuser;
 CREATE ROLE testuser LOGIN;
 
-SELECT t1, t2, num1, dub1, dec1, tm, r, bg, b, tn, sml, dt, vc1, c1, encode(bin, 'escape') FROM pxf_jdbc_readable ORDER BY t1;
+
```
yimingli-vmware, line 15: extra new line

automation/src/test/java/org/greenplum/pxf/automation/features/jdbc/JdbcTest.java
```diff
@@ -34,6 +34,23 @@ public class JdbcTest extends BaseFeature {
...
+            "r real",
+            "bg bigint",
+            "b boolean",
+            "tn smallint",
+            "sml smallint",
+            "dt date",
+            "vc1 varchar(5)",
+            "c1 char(3)",
+            "bin bytea",
+            "tmz timestamp with time zone"};
```
bradfordb-vmware, line 53:
```suggestion
    private static final String[] TYPES_TABLE_FIELDS = new String[]{
            "t1    text",
            "t2    text",
            "num1  int",
            "dub1  double precision",
            "dec1  numeric",
            "tm    timestamp",
            "r     real",
            "bg    bigint",
            "b     boolean",
            "tn    smallint",
            "sml   smallint",
            "dt    date",
            "vc1   varchar(5)",
            "c1    char(3)",
            "bin   bytea",
            "tmz   timestamp with time zone"};
```

Aligning the column type makes it a little easier to visually parse the column names and types. This is a convention in a lot of the automation tests and while this file doesn't exactly follow it currently, we can follow it with new code.

automation/src/test/resources/data/gpdb/gpdb_types_without_timestamptz.txt
```diff
@@ -0,0 +1 @@
...
row16_timestamp_null	s_16	11	37	0.123456789	\N	7.7	23456789	f	11	1100	2015-03-20	abcde	USD	1
row17_real_null	s_16	11	37	0.123456789	2013-07-23 21:00:00	\N	23456789	f	11	1100	2015-03-21	abcde	EUR	1
row18_bigint_null	s_16	11	37	0.123456789	2013-07-23 21:00:00	7.7	\N	f	11	1100	2015-03-22	abcde	USD	1
row19_bool_null	s_16	11	37	0.123456789	2013-07-23 21:00:00	7.7	23456789	\N	11	1100	2015-03-23	abcde	USD	1
row20_tinyint_null	s_16	11	37	0.123456789	2013-07-23 21:00:00	7.7	23456789	f	\N	1100	2015-03-24	abcde	UAH	1
row21_smallint_null	s_16	11	37	0.123456789	2013-07-23 21:00:00	7.7	23456789	f	11	\N	2015-03-25	abcde	USD	1
row22_date_null	s_16	11	37	0.123456789	2013-07-23 21:00:00	7.7	23456789	f	11	1100	\N	abcde	EUR	1
row23_varchar_null	s_16	11	37	0.123456789	2013-07-23 21:00:00	7.7	23456789	f	11	1100	2015-03-26	\N	EUR	1
row24_char_null	s_16	11	37	0.123456789	2013-07-23 21:00:00	7.7	23456789	f	11	1100	2015-03-27	abcde	\N	1
row25_binary_null	s_16	11	37	0.123456789	2013-07-23 21:00:00	7.7	23456789	f	1	1100	2015-03-28	abcde	USD	\N
```
bradfordb-vmware, line 1: Since we are updating this file with an additional column and row, can we reformat this to use `\n` for linebreaks?

bradfordb-vmware (commented):
isDateWideRange can effect
* DATE (reading & writing)
* TIMESTAMP (reading & writing)
* TIMESTAMP WITH TIME ZONE (reading)

It doesn't look like we've add coverage for DATE or TIMESTAMP when`DATE_WIDE_RANGE` is true and it doesn't look like we have any error cases for when `DATE_WIDE_RANGE` is false.

automation/src/main/java/org/greenplum/pxf/automation/components/regress/Regress.java
```diff
@@ -41,6 +41,7 @@ public void runSqlTest(final String sqlTestPath) throws IOException, ShellComman
         StringJoiner commandToRun = new StringJoiner(" ");
 
         commandToRun.add("PGDATABASE=" + dbName);
+        commandToRun.add("PGTZ=UTC");
```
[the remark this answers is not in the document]
tatb-vmware: The change is to ensure that our automation test is deterministic no matter which environment we run it on. We can set the top level database timezone here and then override it in individual tests if needed.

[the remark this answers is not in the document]
denalex: I think currently we can run in different timezones (PST in local, UTC in CI) and still receive the same results when we explicitly add casts to specific timezone in SELECT statement. I am not sure we would want to fix the DB timezone to UTC, this might mask bugs if they were to exist in our code.

automation/src/test/java/org/greenplum/pxf/automation/features/jdbc/JdbcTest.java
```diff
@@ -18,7 +18,7 @@ public class JdbcTest extends BaseFeature {
 
     private static final String POSTGRES_DRIVER_CLASS = "org.postgresql.Driver";
     private static final String GPDB_PXF_AUTOMATION_DB_JDBC = "jdbc:postgresql://";
-    private static final String[] TYPES_TABLE_FIELDS = new String[]{
+    private static final String[] TYPES_TABLE_FIELDS_WITHOUT_TIMESTAMPTZ = new String[]{
```
denalex, line 21: is this temporary ? If not, can we add / remove from another collection instead of copying/pasting ?

automation/src/test/java/org/greenplum/pxf/automation/features/jdbc/JdbcTest.java
```diff
@@ -381,6 +418,7 @@ private void prepareNamedQuery() throws Exception {
                 "database");
         pxfJdbcNamedQuery.setHost(pxfHost);
         pxfJdbcNamedQuery.setPort(pxfPort);
+        pxfJdbcNamedQuery.addUserParameter("date_wide_range=true");
```
denalex, line 421: I did not expect we would be adding this parameters to pretty much all the tables.

automation/src/test/java/org/greenplum/pxf/automation/features/multiuser/MultiUserTest.java
```diff
@@ -74,6 +75,7 @@ private void prepareReadable() throws Exception {
                 "database");
         pxfJdbcReadable.setHost(pxfHost);
         pxfJdbcReadable.setPort(pxfPort);
+        pxfJdbcReadable.addUserParameter("date_wide_range=true");
```
axue-broadcom, line 78: instead of modifying the existing tests to include date wide range, I think we want to have tests for both when the user parameter is present and when it is not.

automation/src/test/java/org/greenplum/pxf/automation/features/jdbc/JdbcTest.java
axue-broadcom, line 421: this looks like we are modifying the existing tests. we want to have tests that encapsulate when `date_wide_range=true` and also continue testing when it is the default value `false`. Do we still have that coverage if we are adding the user parameter to every table?

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -222,7 +223,8 @@ public List<OneField> getFields(OneRow row) throws SQLException {
                         LocalDateTime localDateTime = result.getObject(colName, LocalDateTime.class);
                         value = localDateTime != null ? localDateTime.format(LOCAL_DATE_TIME_GET_FORMATTER) : null;
                     } else {
-                        value = result.getTimestamp(colName);
+                        Timestamp timestamp = result.getTimestamp(colName);
+                        value = timestamp != null ? timestamp.toString() : null;
```
denalex: let's use a formatter here

server/pxf-jdbc/src/test/java/org/greenplum/pxf/plugins/jdbc/JdbcResolverTest.java
```diff
@@ -123,7 +123,7 @@ void getFieldDateTimeWithoutWideRangeTest() throws SQLException {
         boolean isDateWideRange = false;
         Timestamp timestamp = Timestamp.valueOf("1977-12-11 11:15:30.1234");
         OneField oneField = getOneField(timestamp, DataType.TIMESTAMP.getOID(), "timestamp");
-        assertEquals("1977-12-11 11:15:30.1234", oneField.toString());
+        assertEquals("1977-12-11 11:15:30.1234", oneField.val);
```
denalex, line 126: we might also want to `assertTrue(oneField.val instance of String)` that the class is String before checking for equality

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -35,6 +35,7 @@
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.sql.Time;
```
denalex, line 39: is this class used ?

automation/sqlrepo/features/jdbc/readable_date_wide_range/expected/query01.ans
```diff
@@ -0,0 +1,101 @@
...
+ row4                   | s_9  |    4 |    9 |    123456789.1 | 2013-07-16 21:00:00    | 10.7 | 23456789 | t |  4 |  400 | 2015-03-08 | abcde | USD | 4   | 2013-07-16 14:00:00+00
+ row5                   | s_10 |    5 |   10 | 0.000000000001 | 2013-07-17 21:00:00    | 11.7 | 23456789 | f |  5 |  500 | 2015-03-09 | abcde | USD | 5   | 2013-07-17 13:00:00+00
+ row6                   | s_11 |    6 |   11 |       1234.889 | 2013-07-18 21:00:00    | 12.7 | 23456789 | t |  6 |  600 | 2015-03-10 | abcde | USD | 6   | 2013-07-18 12:00:00+00
+ row7                   | s_12 |    7 |   12 |         0.0001 | 2013-07-19 21:00:00    |  7.7 | 23456789 | f |  7 |  700 | 2015-03-11 | abcde | USD | 7   | 2013-07-19 11:00:00+00
+ row8                   | s_13 |    8 |   13 |    45678.00002 | 2013-07-20 21:00:00    |  7.7 | 23456789 | t |  8 |  800 | 2015-03-12 | abcde | EUR | 8   | 2013-07-20 10:00:00+00
+ row9                   | s_14 |    9 |   14 |        23457.1 | 2013-07-21 21:00:00    |  7.7 | 23456789 | f |  9 |  900 | 2015-03-13 | abcde | UAH | 9   | 2013-07-21 09:00:00+00
+(26 rows)
+
+SELECT tm FROM pxf_jdbc_readable_date_wide_range_off ORDER BY t1;
```
denalex, line 65: I wonder why we need this query - what are we checking for ?

automation/src/test/resources/data/gpdb/gpdb_types_with_date_wide_range.txt
```diff
@@ -0,0 +1,26 @@
+row1	s_6	1	6	1.23456	2013-07-13 21:00:00	7.7	23456789	f	1	10	2015-03-06	abcd	USD	1	2013-07-13 12:00:00-05
+row2	s_7	2	7	1.23456	2013-07-13 21:00:00	8.7	23456789	t	2	20	2015-03-06	abcde	USD	2	2013-07-13 12:00:00-04
+row3	s_8	3	8	-1.23456	2013-07-15 21:00:00	9.7	23456789	f	3	300	2015-03-07	abcde	USD	3	2013-07-15 12:00:00-03
```
denalex, line 3: it's great that timezone offsets vary between rows. I just wonder if you have manually verified that when the results come back from a UTC-based query, the shifts to UTC are done as expected ? We hardcode the expected results in the answer SQL files, so we need to make sure they themselves are correct.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -222,7 +224,8 @@ public List<OneField> getFields(OneRow row) throws SQLException {
                         LocalDateTime localDateTime = result.getObject(colName, LocalDateTime.class);
                         value = localDateTime != null ? localDateTime.format(LOCAL_DATE_TIME_GET_FORMATTER) : null;
                     } else {
-                        value = result.getTimestamp(colName);
+                        Timestamp timestamp = result.getTimestamp(colName);
+                        value = timestamp != null ? timestamp.toLocalDateTime().format(GreenplumDateTime.DATETIME_FORMATTER) : null;
```
denalex, line 228: are we planning to do a similar adjustment for DATE type ? Also the updated test cases address only TIMESTAMP and TIMESTAMPTZ types, do we plan to adjust them for DATE types that can also have "wide-range" ? Or is it out of scope for this story ?

automation/src/test/resources/data/gpdb/gpdb_types.txt
```diff
@@ -1 +1 @@
...
row16_timestamp_null	s_16	11	37	0.123456789	\N	7.7	23456789	f	11	1100	2015-03-20	abcde	USD	1
row17_real_null	s_16	11	37	0.123456789	2013-07-23 21:00:00	\N	23456789	f	11	1100	2015-03-21	abcde	EUR	1
row18_bigint_null	s_16	11	37	0.123456789	2013-07-23 21:00:00	7.7	\N	f	11	1100	2015-03-22	abcde	USD	1
row19_bool_null	s_16	11	37	0.123456789	2013-07-23 21:00:00	7.7	23456789	\N	11	1100	2015-03-23	abcde	USD	1
row20_tinyint_null	s_16	11	37	0.123456789	2013-07-23 21:00:00	7.7	23456789	f	\N	1100	2015-03-24	abcde	UAH	1
row21_smallint_null	s_16	11	37	0.123456789	2013-07-23 21:00:00	7.7	23456789	f	11	\N	2015-03-25	abcde	USD	1
row22_date_null	s_16	11	37	0.123456789	2013-07-23 21:00:00	7.7	23456789	f	11	1100	\N	abcde	EUR	1
row23_varchar_null	s_16	11	37	0.123456789	2013-07-23 21:00:00	7.7	23456789	f	11	1100	2015-03-26	\N	EUR	1
row24_char_null	s_16	11	37	0.123456789	2013-07-23 21:00:00	7.7	23456789	f	11	1100	2015-03-27	abcde	\N	1
row25_binary_null	s_16	11	37	0.123456789	2013-07-23 21:00:00	7.7	23456789	f	1	1100	2015-03-28	abcde	USD	\N
```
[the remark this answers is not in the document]
tatb-vmware: I added a `\n` for linebreaks into this file as suggested by @bradfordb-vmware in an earlier comment. It makes it easier for developers to work on this file and does not impact how we read in the file in Postgres.

automation/sqlrepo/features/jdbc/readable_date_wide_range/expected/query01.ans
tatb-vmware: The query was meant to catch an inconsistency between how foreign tables and external tables treat timestamps.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
tatb-vmware: We should and I can add testcases as well to cover wide-range DATE types.

automation/src/test/resources/data/gpdb/gpdb_types_with_date_wide_range.txt
tatb-vmware: I manually checked by comparing the differences between an external table and internal table with the same data. They show the identical representation for the same data, so the reading of the timezone offset with PXF are consistent with Postgres.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
tatb-vmware: After speaking with @axue-broadcom, it was decided that changing DATE types would be out of scope for this PR. It can be added as a chore to do.

yimingli-vmware (approved):

automation/src/test/resources/data/gpdb/generate_gpdb_types.sql
```diff
@@ -0,0 +1,24 @@
...
+    c1 char(3),
+    bin bytea,
+    u uuid
+);
+
+--- Replace <user> with your user
+COPY gpdb_types FROM '/Users/<user>/workspace/pxf/automation/src/test/resources/data/gpdb/gpdb_types.txt';
+
+--- Make the modifications to `gpdb_types` table then run the following file to regenerate the data:
+COPY gpdb_types TO '/Users/<user>/workspace/pxf/automation/src/test/resources/data/gpdb/test_types.txt';
```
denalex, line 24: I am not sure it is a good idea to regenerate this file via Greenplum. I think this is supposed to be a "manually" crafted file. Manually crafted files can, for example have more precision for numeric types that GP allows and a test would make sure the precision is "lost" correctly. Not saying that it is what is happening with this file particularly, more of a general comment.

denalex (approved):

automation/src/test/resources/data/gpdb/README
```diff
@@ -0,0 +1,5 @@
+Data for gpdb testing `gpdb_types.txt` is generated using the SQL file called `generate_gpdb_types.sql`.
```
axue-broadcom, line 1: make sure that this is a `README.md` file

axue-broadcom (approved):

PR #1087, opened 2024-02-13
Change to using vendored gcs-resource

axue-broadcom:
The concourse pipelines are getting rate-limited by docker when trying to pull down the gcs-resource image. 

```
ERRO[0000] checking origin frodenas/gcs-resource failed: initialize transport: GET https://index.docker.io/v2/: unexpected status code 429 Too Many Requests: Too Many Requests (HAP429)
```

Instead of pulling it from Dockerhub, pull it from our vendored copy to avoid getting rate-limited.

### DIFF
--- a/concourse/pipelines/singlecluster-pipeline.yml
+++ b/concourse/pipelines/singlecluster-pipeline.yml
@@ -6,7 +6,8 @@ resource_types:
 - name: google-cloud-storage
   type: registry-image
   source:
-    repository: frodenas/gcs-resource
+    repository: gcr.io/data-gpdb-public-images/gcs-resource
+    tag: v0.6.0
 
 ## ======================================================================
 ## RESOURCES
--- a/concourse/pipelines/singlecluster-pipeline.yml
+++ b/concourse/pipelines/singlecluster-pipeline.yml
@@ -7,7 +7,6 @@ resource_types:
   type: registry-image
   source:
     repository: gcr.io/data-gpdb-public-images/gcs-resource
-    tag: v0.6.0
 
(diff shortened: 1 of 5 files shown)


tatb-vmware (approved):

concourse/pipelines/singlecluster-pipeline.yml
```diff
@@ -6,7 +6,8 @@ resource_types:
 - name: google-cloud-storage
   type: registry-image
   source:
-    repository: frodenas/gcs-resource
+    repository: gcr.io/data-gpdb-public-images/gcs-resource
+    tag: v0.6.0
```
denalex, line 10: I guess `v0.6.0` copies the DockerHub tag of that image. Since this is our repo now and we control when we refresh the image, can't we just use `latest` for our pipelines ? Basically if we tag the image in our repo with both `v0.6.0` and `latest`, we should not care about a particular version

axue-broadcom: Sounds good. I will change this to latest tag. The digest for the latest tag in our repro did not match with those in DockerHub. However, the following comment indicates that they are the same.
```
$ docker image history --no-trunc gcr.io/data-gpdb-public-images/gcs-resource
IMAGE                                                                     CREATED        CREATED BY                                                                                                       SIZE      COMMENT
sha256:f5636a3e5b49ca32fec9f65bf8176f6d3087e02f4e2ef77e8672cfa97651387b   4 months ago   /bin/sh                                                                                                          4B        same as previous v0.6.0
```

denalex (approved):

PR #1089, opened 2024-02-15
Package cluster templates into RPM/DEB

denalex:
This PR changes the directory layout of PXF_HOME such that:
* `conf` directory (and files within it) is now `templates/conf` directory
* server `*-site.xml` files previously in `templates` directory are now in `templates/server`
* new directory `templates/cluster` contains `cluster.txt` template file

Additionally, `config` specification is removed from files in RPM/DEB specs such that there are no longer any configuration files that RPM/DEB needs to recognize, since with the new design the `templates` directory is not supposed to have modifications that should persist across PXF versions.

Users will be expected to run `pxf prepare` command to prepare even a default cluster such that all configuration will be located in `PXF_BASE/clusters` (or `PXF_HOME/clusters` if users did not bother to have a separate PXF_BASE).

The resulting directory structure (for an example PXF version) is:
```
/usr/local/pxf-gp6
├── application
│   └── pxf-app-6.9.2-SNAPSHOT.jar
├── bin
│   ├── kill-pxf.sh
│   ├── merge-pxf-config.sh
│   ├── pxf
│   ├── pxf-cli
│   ├── pxf-post-gpupgrade
│   └── pxf-pre-gpupgrade
├── commit.sha
├── gpextable
│   ├── metadata
│   ├── pxf-1.0-2.0.sql
│   ├── pxf--1.0.sql
│   ├── pxf-2.0-2.1.sql
│   ├── pxf--2.0.sql
│   ├── pxf-2.1-2.0.sql
│   ├── pxf--2.1.sql
│   ├── pxf.control
│   └── pxf.so
├── keytabs
├── lib
│   └── native
├── share
│   └── pxf-hbase-6.9.2-SNAPSHOT.jar
├── templates
│   ├── cluster
│   │   └── cluster.txt
│   ├── conf
│   │   ├── pxf-application.properties
│   │   ├── pxf-env.sh
│   │   ├── pxf-log4j2.xml
│   │   └── pxf-profiles.xml
│   └── server
│       ├── adl-site.xml
│       ├── core-site.xml
│       ├── gs-site.xml
│       ├── hbase-site.xml
│       ├── hdfs-site.xml
│       ├── hive-site.xml
│       ├── jdbc-site.xml
│       ├── mapred-site.xml
│       ├── minio-site.xml
│       ├── pxf-site.xml
│       ├── s3-site.xml
│       ├── wasbs-site.xml
│       └── yarn-site.xml
└── version
```


server/pxf-service/src/templates/cluster/cluster.txt
```diff
@@ -0,0 +1,4 @@
+cluster.name=__CLUSTER_NAME__
+cluster.collocated=__CLUSTER_COLLOCATED__
```
bradfordb-vmware, line 2: Do you want to just use default values of `default` and `true` here? A convention that I have seen with distributed configuration files is to assume default values but include comments describing what the parameters controls and what allowed values are.

server/pxf-service/src/templates/server/adl-site.xml
bradfordb-vmware, line 1: For PXF 6, the directory under `$PXF_BASE` is named `servers`. I think that's because the directory structure is `servers/<server-name>/pxf-site.xml`. Here there is no named server under `templates` so I can see how naming it `server` makes sense but it wasn't obvious to me at first.

I call this out because in the PR description you state that it is `templates/server` but `templates/clusters`. However, the example directory structure is `template/cluster`.

bradfordb-vmware (approved):

server/pxf-service/src/templates/server/adl-site.xml
denalex: It should be `templates/cluster`, I fixed the description
The idea is -  those are templates for 1 "cluster" or 1 "server", while plural form "clusters" and "servers" are used as containers for the actual deployment.

server/pxf-service/src/templates/cluster/cluster.txt
denalex: I thought about it, but having placeholder tokens is better IMO, since the logic for replacement will be easy (always replace the tokens with some value). And the logic for name / collocation values will chose values ("default" vs. given by user) instead of choosing the type of operation (to replace or not to replace).

yimingli-vmware (approved):

axue-broadcom (approved):

PR #1094, opened 2024-02-22
Address test changes due to Azure migration

axue-broadcom:
With work that was completed in anticipation of the Azure migration, two changes needed to be made to automation tests. We are currently taking advantage of the Azure Gen1 Compatibility Layer so that we can still use the ADL profile.

1. HcfsGlobbingTest included a test case here a file ended in `.` This is not a valid file format with the Apache Hadoop provided ABFS when connecting to an Azure account, even through the Gen1 Compatibility Layer

2. Flakiness was introduced with some of the writable tests as the ADL profile is going through the Azure Gen1 Compatibility Layer which can cause some of the SQL statements to timeout before the command completes. As such, retries have been introduced in as much as possible to reduce this flakiness.


automation/sqlrepo/features/hcfs/globbing/match_zero_or_more_characters/expected/query02.ans
```diff
@@ -1,31 +1,29 @@
...
+ 1a a.|_row_19        |  19 |  19 | 1900000000000 | f
+ 1a a.|_row_2         |   2 |   2 |  200000000000 | t
+ 1a a.|_row_20        |  20 |  20 | 2000000000000 | t
+ 1a a.|_row_3         |   3 |   3 |  300000000000 | f
+ 1a a.|_row_4         |   4 |   4 |  400000000000 | t
+ 1a a.|_row_5         |   5 |   5 |  500000000000 | f
+ 1a a.|_row_6         |   6 |   6 |  600000000000 | t
+ 1a a.|_row_7         |   7 |   7 |  700000000000 | f
+ 1a a.|_row_8         |   8 |   8 |  800000000000 | t
+ 1a a.|_row_9         |   9 |   9 |  900000000000 | f
```
tatb-vmware, line 26: Why were the names changed?

automation/src/test/java/org/greenplum/pxf/automation/features/avro/HdfsWritableAvroTest.java
```diff
@@ -109,6 +110,8 @@ public class HdfsWritableAvroTest extends BaseWritableFeature {
     private String fullTestPath;
     private ProtocolEnum protocol;
 
+    private static final Integer NUM_RETRIES = 5;
```
tatb-vmware, line 113: What is the reason for choosing this amount or was it arbitrary?

automation/sqlrepo/features/hcfs/globbing/match_zero_or_more_characters/expected/query02.ans
axue-broadcom: The names changed because the names are created based off the input to the function that calls it (see HcfsGlobbingTest change in this PR). The input was originally `a.` and is now `a.|`

automation/src/test/java/org/greenplum/pxf/automation/features/avro/HdfsWritableAvroTest.java
axue-broadcom: Manual testing has show that it's possible for the operation to time out a couple times in a row but I have not seen it fail more than 3 times in a row. I set it to 5 for a little extra buffer beyond that.

automation/src/test/java/org/greenplum/pxf/automation/features/orc/OrcWriteTest.java
```diff
@@ -499,6 +539,36 @@ private String getRecordCSV(int row, boolean[] isNull) {
         return rowBuilder.toString();
     }
 
+    public interface ThrowingConsumer<T, E extends Exception> {
+        void accept() throws E;
+    }
+
+    private void attemptInsert(ThrowingConsumer operation, Integer numRows, String readableTableName, Integer retryCount) throws Exception {
```
denalex, line 546: seems like this is a copy from another test case. Should we extract it to some utility (or to some TEST superclass) and provide a second function for comparison/cleanup of data ?

axue-broadcom: I moved the functionality into BaseFeature

denalex (approved):

yimingli-vmware (approved):

PR #1100, opened 2024-03-05
Refactor position tracking for multi-line JSON files

axue-broadcom:
JsonRecordReader was refactored in commit 515235f (PR #879) to internally use the LineRecordReader when handling multi-line JSON files.

Previously, we were calculating the position inside JsonRecordReader by finding the length of bytes from the current index to the end of the current line that was read in by the underlying LineRecordReader. This causes performance issues for files with a lot of data condensed into very few lines.

Now, instead of calculating the number of bytes from the current index in the line to the end of the line, we calculate the length of the object we successfully read. Additionally, we take advantage of the underlying LineRecordReader to determine the true starting position of a split.

### DIFF
--- a/server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
+++ b/server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
@@ -96,8 +96,8 @@ public class JsonRecordReader implements RecordReader<LongWritable, Text> {
         // set pos and filePos to lineRecordReader's position. If the split started in the middle of a line,
         // we assume that the previous split has taken care of it, so we just need to be at the same starting
         // point as the lineRecordReader.
-        pos = lineRecordReader.getPos();
         filePos = lineRecordReader.getPos();
+        pos = filePos;
         key = lineRecordReader.createKey();
     }
 
@@ -139,29 +139,30 @@ public class JsonRecordReader implements RecordReader<LongWritable, Text> {
             }
 
             // if we have gotten here there are 2 scenarios:
-            // the object is not complete, in which case we hit end of file
-            // the object is complete, in which case, determine if we found the identifier
+            // - the object is not complete, in which case we hit end of file
+            // - the object is complete, in which case, determine if we found the identifier
             if (!isObjectComplete) {
                 // if EOF, update pos to be filePos
                 pos = filePos;
-            } else {
-                String json = parser.getCompletedObject();
+                return false;
+            }
 
-                // the object is complete so update the position
-                pos = pos + json.getBytes(StandardCharsets.UTF_8).length;
+            String json = parser.getCompletedObject();
 
-                // if we found the identifier
-                if (parser.foundObjectWithIdentifier()) {
-                    // check the char length of the json against the MAXLENGTH parameter
-                    long jsonLength = json.length();
-                    if (jsonLength > maxObjectLength) {
-                        LOG.warn("Skipped JSON object of size " + json.length());
-                    } else {
-                        // the key is set to beginning of the json object
-                        key.set(jsonLength);
-                        value.set(json);
-                        return true;
-                    }
+            // the object is complete so update the position
+            pos = pos + json.getBytes(StandardCharsets.UTF_8).length;
+
+            // if we found the identifier
+            if (parser.foundObjectWithIdentifier()) {
+                // check the char length of the json against the MAXLENGTH parameter
+                long jsonLength = json.length();
+                if (jsonLength > maxObjectLength) {
+                    LOG.warn("Skipped JSON object of size " + json.length());
+                } else {
+                    // the key is set to beginning of the json object
+                    key.set(jsonLength);
+                    value.set(json);
+                    return true;
                 }
             }
         }
--- a/server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
+++ b/server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
@@ -142,7 +142,7 @@ public class JsonRecordReader implements RecordReader<LongWritable, Text> {
             // - the object is not complete, in which case we hit end of file
             // - the object is complete, in which case, determine if we found the identifier
             if (!isObjectComplete) {
-                // if EOF, update pos to be filePos
+                // if we have gotten here, filePos is at EOF
                 pos = filePos;
                 return false;
             }
--- a/server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
+++ b/server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
@@ -283,7 +283,7 @@ public class JsonRecordReader implements RecordReader<LongWritable, Text> {
      */
     private void accountForCharacterRead(char c) {
         readValues.append(c);
-        if (readValues.length() == MAX_CHARS){
+        if (readValues.length() >= MAX_CHARS){
             pos += readValues.toString().getBytes(StandardCharsets.UTF_8).length;
             readValues.setLength(0);
         }
--- a/server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
+++ b/server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
@@ -245,27 +245,36 @@ public class JsonRecordReader implements RecordReader<LongWritable, Text> {
         // seek until we hit the first begin-object
         boolean inString = false;
         int i;
-        String readValues = "";
+        StringBuilder readValues = new StringBuilder();
         // since we have not yet found a starting object, exit if either EOF (-1) or END_OF_SPLIT (-2)
         while ((i = readNextChar()) > EOF) {
             char c = (char) i;
 
             // while scanning to start of object, we need to make sure we aren't losing our position
-            readValues += c;
+            readValues.append(c);
+            checkLength(readValues, 1024);
             // if the current value is a backslash, then ignore the next value as it's an escaped char
             if (c == BACKSLASH) {
                 int temp = readNextChar();
-                readValues += (char) temp;
+                readValues.append((char) temp);
+                checkLength(readValues, 1024);
             } else if (c == QUOTE) {
                 inString = !inString;
             } else if (c == START_BRACE && !inString) {
                 // the start brace will be accounted for later, so ignore it for now
-                pos = pos + readValues.getBytes(StandardCharsets.UTF_8).length - 1;
+                pos = pos + readValues.toString().getBytes(StandardCharsets.UTF_8).length - 1;
                 return true;
             }
         }
-        pos = pos + readValues.getBytes(StandardCharsets.UTF_8).length;
+        pos = pos + readValues.toString().getBytes(StandardCharsets.UTF_8).length;
         return false;
+    }
+
+    private void checkLength(StringBuilder sb, int maxLength) {
+        if (sb.length() >= maxLength ){
+            pos = pos + sb.toString().getBytes(StandardCharsets.UTF_8).length;
+            sb.setLength(0);
+        }
     }
 
     /**
--- a/server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
+++ b/server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
@@ -150,7 +150,7 @@ public class JsonRecordReader implements RecordReader<LongWritable, Text> {
             String json = parser.getCompletedObject();
 
             // the object is complete so update the position
-            pos = pos + json.getBytes(StandardCharsets.UTF_8).length;
+            pos += json.getBytes(StandardCharsets.UTF_8).length;
 
             // if we found the identifier
             if (parser.foundObjectWithIdentifier()) {
@@ -159,7 +159,7 @@ public class JsonRecordReader implements RecordReader<LongWritable, Text> {
                 if (jsonLength > maxObjectLength) {
                     LOG.warn("Skipped JSON object of size " + json.length());
                 } else {
-                    // the key is set to beginning of the json object
+                    // the key is set to the length of the json object
                     key.set(jsonLength);
                     value.set(json);
                     return true;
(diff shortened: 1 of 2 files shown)


server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
```diff
@@ -93,8 +93,11 @@ public JsonRecordReader(JobConf conf, FileSplit split) throws IOException {
         this.conf = conf;
         parser = new PartitionedJsonParser(jsonMemberName);
         currentLine = new Text();
-        pos = start;
-        filePos = start;
+        // set pos and filePos to lineRecordReader's position. If the split started in the middle of a line,
+        // we assume that the previous split has taken care of it, so we just need to be at the same starting
+        // point as the lineRecordReader.
+        pos = lineRecordReader.getPos();
+        filePos = lineRecordReader.getPos();
```
denalex, line 100:
```suggestion
        filePos = lineRecordReader.getPos();
        pos = filePos;
```

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
```diff
@@ -135,22 +138,30 @@ public boolean next(LongWritable key, Text value) throws IOException {
...
             }
 
-            // we've completed an object but there might still be things in the buffer. Calculate the proper
-            // position of the JsonRecordReader
-            updatePos();
-
-            if (isObjectComplete && parser.foundObjectWithIdentifier()) {
+            // if we have gotten here there are 2 scenarios:
+            // the object is not complete, in which case we hit end of file
+            // the object is complete, in which case, determine if we found the identifier
```
denalex, line 143:
```suggestion
            // if we have gotten here there are 2 scenarios:
            // - the object is not complete, in which case we hit end of file
            // - the object is complete, in which case, determine if we found the identifier
```

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
```diff
@@ -135,22 +138,30 @@ public boolean next(LongWritable key, Text value) throws IOException {
...
+            if (!isObjectComplete) {
+                // if EOF, update pos to be filePos
+                pos = filePos;
```
denalex, line 146:
```suggestion
            if (i == EOF) {
                pos = filePos;
```

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
```diff
@@ -135,22 +138,30 @@ public boolean next(LongWritable key, Text value) throws IOException {
...
+            } else {
```
denalex, line 147: can we early return from above block so that we do not need `else` block ?

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
```diff
@@ -241,18 +244,26 @@ private boolean scanToNextJsonBeginObject() throws IOException {
         // seek until we hit the first begin-object
         boolean inString = false;
         int i;
+        String readValues = "";
         // since we have not yet found a starting object, exit if either EOF (-1) or END_OF_SPLIT (-2)
         while ((i = readNextChar()) > EOF) {
             char c = (char) i;
+
+            // while scanning to start of object, we need to make sure we aren't losing our position
+            readValues += c;
```
denalex, line 253: this is expensive as it will cause string re-allocation for each character. Let's use a `StringBuilder` that we can fill up, get value and then clean and re-use.

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
```diff
@@ -241,18 +244,26 @@ private boolean scanToNextJsonBeginObject() throws IOException {
         // seek until we hit the first begin-object
         boolean inString = false;
         int i;
+        String readValues = "";
```
denalex, line 247: I wonder what is the performance difference between finding out the byte length for each character vs. accumulating a String and finding out its byte length at once. The problem with the String is that we accumulate a potentially long string in memory whereas we only need to know its running size.

tatb-vmware (approved):

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
```diff
@@ -135,20 +138,29 @@ public boolean next(LongWritable key, Text value) throws IOException {
...
             }
 
-            // we've completed an object but there might still be things in the buffer. Calculate the proper
-            // position of the JsonRecordReader
-            updatePos();
+            // if we have gotten here there are 2 scenarios:
+            // - the object is not complete, in which case we hit end of file
+            // - the object is complete, in which case, determine if we found the identifier
+            if (!isObjectComplete) {
+                // if EOF, update pos to be filePos
```
yimingli-vmware, line 145: We can only enter into this condition when we see an internal object that has a field with the matching identifier `i == EOF` and the object is not complete, is my understanding correct? Then I would not use `if` in the comment on line 145.

The comment on line 143 is a little bit confusing to me. Here I cannot find the logic for that case?

server/pxf-json/src/test/java/org/greenplum/pxf/plugins/json/parser/JsonRecordReaderTest.java
```diff
@@ -385,6 +387,14 @@ public void testStraddleSplitSmallSplitSize() throws URISyntaxException, IOExcep
...
         // 566 line 6 has the last half record -- 55 bytes
         // 567 line 7 is closing array bracket -- 1 bytes
+
+        // +1 for open bracket + 1 for new line + 2 for spaces = 4
+        // record 1 is 109 bytes = 113
+        // +2 for comma space + record 2 is 107 bytes = 224
+        // +2 for comma space + 2 for newline + record 3 is 108 bytes = 335
+        // +2 for comma space + record 4 is 124 bytes = 461
+        // +2 for comma space + record 5 is 103 bytes = 566
+        // +1 for end bracket = 567
```
yimingli-vmware, line 397:
```suggestion

        // +1 for open bracket, + 1 for new line, + 2 for spaces = 4
        // + 109 for record 1 = 113
        // +1 for comma, +1 for space, + 107 for record 2 = 224
        // +1 for comma, +1 for space, + 1 for newline, + 108 for record 3 = 336
        // +1 for comma, +1 for space, + 124 for record 4 = 462
        // +1 for comma, +1 for space, + 103 for record 5 = 567
        // +1 for end bracket = 568
```

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
axue-broadcom: since we return early, the code after this if statement covers the case listed on line 143

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
axue-broadcom: I have not found an easy way to convert from char to number of bytes without first going to a string object

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
```diff
@@ -135,20 +138,29 @@ public boolean next(LongWritable key, Text value) throws IOException {
...
                 // check the char length of the json against the MAXLENGTH parameter
                 long jsonLength = json.length();
-                long jsonStart = pos - json.getBytes(StandardCharsets.UTF_8).length;
                 if (jsonLength > maxObjectLength) {
-                    LOG.warn("Skipped JSON object of size " + json.length() + " at pos " + jsonStart);
+                    LOG.warn("Skipped JSON object of size " + json.length());
                 } else {
                     // the key is set to beginning of the json object
-                    key.set(jsonStart);
+                    key.set(jsonLength);
```
denalex, line 163: the comment does not reflect what this operation is doing now

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
```diff
@@ -241,21 +245,38 @@ private boolean scanToNextJsonBeginObject() throws IOException {
...
             char c = (char) i;
+
+            // while scanning to start of object, we need to make sure we aren't losing our position
+            readValues.append(c);
+            checkLength(readValues, 1024);
             // if the current value is a backslash, then ignore the next value as it's an escaped char
             if (c == BACKSLASH) {
-                readNextChar();
+                int temp = readNextChar();
+                readValues.append((char) temp);
```
denalex, line 259:
```suggestion
                readValues.append((char) readNextChar());
```

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
```diff
@@ -241,21 +245,38 @@ private boolean scanToNextJsonBeginObject() throws IOException {
...
         boolean inString = false;
         int i;
+        StringBuilder readValues = new StringBuilder();
         // since we have not yet found a starting object, exit if either EOF (-1) or END_OF_SPLIT (-2)
         while ((i = readNextChar()) > EOF) {
             char c = (char) i;
+
+            // while scanning to start of object, we need to make sure we aren't losing our position
+            readValues.append(c);
+            checkLength(readValues, 1024);
```
denalex, line 255: these 2 lines repeat with a magic number, can we extract into a function, take only 1 param and declare a const for a number ?
```suggestion
            accountForCharacterRead(c);
```

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
```diff
@@ -241,21 +245,38 @@ private boolean scanToNextJsonBeginObject() throws IOException {
         // seek until we hit the first begin-object
         boolean inString = false;
         int i;
+        StringBuilder readValues = new StringBuilder();
```
denalex, line 248: this function might be called multiple times, should we declare the StringBuilder once as an object member and re-use it instead of re-initing ?

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
```diff
@@ -241,21 +245,38 @@ private boolean scanToNextJsonBeginObject() throws IOException {
...
             if (c == BACKSLASH) {
-                readNextChar();
+                int temp = readNextChar();
+                readValues.append((char) temp);
+                checkLength(readValues, 1024);
             } else if (c == QUOTE) {
                 inString = !inString;
             } else if (c == START_BRACE && !inString) {
+                // the start brace will be accounted for later, so ignore it for now
+                pos = pos + readValues.toString().getBytes(StandardCharsets.UTF_8).length - 1;
```
denalex, line 265:
```suggestion
                pos += readValues.toString().getBytes(StandardCharsets.UTF_8).length - 1;
```

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
```diff
@@ -241,21 +245,38 @@ private boolean scanToNextJsonBeginObject() throws IOException {
...
                 return true;
             }
         }
+        pos = pos + readValues.toString().getBytes(StandardCharsets.UTF_8).length;
```
denalex, line 269:
```suggestion
        pos += readValues.toString().getBytes(StandardCharsets.UTF_8).length;
```

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
```diff
@@ -241,21 +245,38 @@ private boolean scanToNextJsonBeginObject() throws IOException {
...
         return false;
     }
 
+    private void checkLength(StringBuilder sb, int maxLength) {
+        if (sb.length() >= maxLength ){
+            pos = pos + sb.toString().getBytes(StandardCharsets.UTF_8).length;
```
denalex, line 275:
```suggestion
            pos += sb.toString().getBytes(StandardCharsets.UTF_8).length;
```

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
```diff
@@ -241,21 +245,38 @@ private boolean scanToNextJsonBeginObject() throws IOException {
...
+                pos = pos + readValues.toString().getBytes(StandardCharsets.UTF_8).length - 1;
                 return true;
             }
         }
+        pos = pos + readValues.toString().getBytes(StandardCharsets.UTF_8).length;
         return false;
     }
 
+    private void checkLength(StringBuilder sb, int maxLength) {
+        if (sb.length() >= maxLength ){
```
denalex, line 274:
```suggestion
        if (sb.length() >= maxLength) {
```

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
denalex, line 248: Let's set the initial capacity to max length that we'll be enforcing.

yimingli-vmware (approved):

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
axue-broadcom: i set it to max length that we're enforcing + 1 just as a precautionary

server/pxf-json/src/main/java/org/greenplum/pxf/plugins/json/JsonRecordReader.java
```diff
@@ -270,10 +274,18 @@ private boolean scanToNextJsonBeginObject() throws IOException {
...
+    /**
+     * Append the given character to the string of characters read
+     *
+     * If the number of characters read equals MAX_CHARS, then
+     * set the global pos and reset the string of characters read to empty
+     * @param c the character that was read
+     */
+    private void accountForCharacterRead(char c) {
+        readValues.append(c);
+        if (readValues.length() == MAX_CHARS){
```
denalex, line 286: I would keep `>=` comparison just for safety in case we failed to account some edge case

denalex (approved):

PR #1104, opened 2024-03-14
Add support for timestamptz to JDBC profile

tatb-vmware:
This PR adds read and write support for `timestamptz` for the JDBC profile for any profile/connector that supports the JDBC 4.2 API. As of this PR, the Hive JDBC connector does not fully support the JDBC 4.2 API and thus PXF does not support `timestamptz` when trying to connect to Hive using JDBC. 

This PR also extends functionality added in commit [e1b148a](https://github.com/greenplum-db/pxf/commit/e1b148a4472de2ed6c86342a8e6dc569a3b3d174). Wide date range `date` and `timestamp` with eras will continue to silently fail and return incorrect data if the `date_wide_range` option is turned off. Wide date range `date` and `timestamp` without eras now uses `date_wide_range` option as a hint for better performance.  If a user does not provide the `date_wide_range` option, instead of erroring out, the JDBC profile will fallback to the `date_wide_range` formatter and attempt to parse the data a second time. In such a scenario, the JDBC resolver will log a warning to suggest turning on `date_wide_range` to get better performance.


server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -123,6 +123,15 @@ public class JdbcResolver extends JdbcBasePlugin implements Resolver {
             .optionalStart().appendPattern(DATE_TIME_FORMATTER_SPECIFIER).optionalEnd()
             .toFormatter();
 
+    private static final DateTimeFormatter OFFSET_DATE_TIME_SET_FORMATTER = (new DateTimeFormatterBuilder())
```
denalex, line 126: please provide a comment like in other formatters what is this for

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -123,6 +123,15 @@ public class JdbcResolver extends JdbcBasePlugin implements Resolver {
...
             .toFormatter();
 
+    private static final DateTimeFormatter OFFSET_DATE_TIME_SET_FORMATTER = (new DateTimeFormatterBuilder())
+            .appendValue(ChronoField.YEAR_OF_ERA, 1, 9, SignStyle.NORMAL).appendLiteral('-')
+            .appendValue(ChronoField.MONTH_OF_YEAR, 1, 2, SignStyle.NORMAL).appendLiteral('-')
+            .appendValue(ChronoField.DAY_OF_MONTH, 1, 2, SignStyle.NORMAL).appendLiteral(" ")
+            .append(ISO_LOCAL_TIME)
+            .appendOffset("+HH:mm", "Z")
+            .optionalStart().appendPattern(DATE_TIME_FORMATTER_SPECIFIER).optionalEnd()
+            .toFormatter();
```
denalex, line 133: I wonder if we should move all these formatter definitions into `GreenplumDateTime` which already has `DATETIME_WITH_TIMEZONE_FORMATTER` so that a similar logic can be used in other profiles.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -230,14 +239,11 @@ public List<OneField> getFields(OneRow row) throws SQLException {
...
+                    OffsetDateTime offsetDateTime = result.getObject(colName, OffsetDateTime.class);
                     if (isDateWideRange) {
-                        OffsetDateTime offsetDateTime = result.getObject(colName, OffsetDateTime.class);
                         value = offsetDateTime != null ? offsetDateTime.format(OFFSET_DATE_TIME_GET_FORMATTER) : null;
                     } else {
-                        throw new UnsupportedOperationException(
-                                String.format("Field type '%s' (column '%s') is not supported",
-                                        DataType.get(oneField.type),
-                                        column));
+                        value = offsetDateTime != null ? offsetDateTime.format(GreenplumDateTime.DATETIME_WITH_TIMEZONE_FORMATTER) : null;
```
denalex, line 246: the logic is the same other than for a formatter, let's extract value assignment out and only determine a formatter using a ternary operator.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -230,14 +239,11 @@ public List<OneField> getFields(OneRow row) throws SQLException {
                     }
                     break;
                 case TIMESTAMP_WITH_TIME_ZONE:
+                    OffsetDateTime offsetDateTime = result.getObject(colName, OffsetDateTime.class);
```
denalex, line 242: can we put a comment above that this is the only class that JDBC drivers will most likely to respect for returning timestamps

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -461,6 +470,7 @@ public static void decodeOneRowToPreparedStatement(OneRow row, PreparedStatement
                         }
                     }
                     break;
+                case TIMESTAMP_WITH_TIME_ZONE:
```
denalex, line 473: why are we not handling `null` cases the same as other types, is  `statement.setNull()` not necessary when using `setObject()`?

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -497,4 +507,22 @@ private LocalDateTime getLocalDateTime(String rawVal) {
...
+     */
+    private OffsetDateTime getOffsetDateTime(String rawVal) {
+        try {
+            if (isDateWideRange) {
+                return OffsetDateTime.parse(rawVal, OFFSET_DATE_TIME_SET_FORMATTER);
+            } else {
+                return OffsetDateTime.parse(rawVal, GreenplumDateTime.DATETIME_WITH_TIMEZONE_FORMATTER);
+            }
+        } catch (Exception e) {
+            throw new IllegalArgumentException("Failed to convert timestamp with timezone '" + rawVal + "' to the OffsetDateTime class: " + e.getMessage(), e);
```
denalex, line 525: do we really need to wrap the original exception, would it not have enough details ?

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -497,4 +507,22 @@ private LocalDateTime getLocalDateTime(String rawVal) {
...
         }
     }
+
+    /**
+     * Convert a string to OffsetDateTime class with formatter
+     *
+     * @param rawVal the OffsetDateTime in a string format
+     * @return OffsetDateTime
+     */
+    private OffsetDateTime getOffsetDateTime(String rawVal) {
```
denalex, line 517: this function is only called from one place, should we just inline it similar to other logical blocks in that switch statement ?

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
denalex, line 126: why is it different from `OFFSET_DATE_TIME_GET_FORMATTER` ?

server/pxf-jdbc/src/test/java/org/greenplum/pxf/plugins/jdbc/JdbcResolverTest.java
```diff
@@ -23,6 +23,7 @@
 import java.time.LocalDate;
 import java.time.LocalDateTime;
 import java.time.OffsetDateTime;
+import java.time.ZoneOffset;
```
denalex, line 26: are there any unit test cases with null values ?

server/pxf-jdbc/src/test/java/org/greenplum/pxf/plugins/jdbc/JdbcResolverTest.java
```diff
@@ -402,6 +456,26 @@ void setFieldDateTimeWithEraWithoutWideRangeTest() throws ParseException {
...
+    @Test
+    void setFieldOffsetDateTimeWithWideRangeTest() throws ParseException {
+        isDateWideRange = true;
+        OffsetDateTime expectedOffsetDateTime = OffsetDateTime.of(12345, 1, 23, 11, 22, 33, 440000000, ZoneOffset.of("+12:34"));
+        String timestamptz = "12345-01-23 11:22:33.44+12:34";
+        OneField oneField = setFields(timestamptz, DataType.TIMESTAMP_WITH_TIME_ZONE.getOID(), "timestamptz");
+        assertTrue(oneField.val instanceof OffsetDateTime);
+        assertEquals(expectedOffsetDateTime, oneField.val);
+    }
+
```
denalex, line 478: are there any test cases with invalid values ? they should check the correct exception type and message

automation/src/test/java/org/greenplum/pxf/automation/features/jdbc/JdbcTest.java
```diff
@@ -297,6 +303,17 @@ private void prepareWritable() throws Exception {
...
 
+        pxfJdbcWritableWithDateWideRange = TableFactory.getPxfJdbcWritableTable(
+                "pxf_jdbc_writable_date_wide_range_on",
+                TYPES_TABLE_FIELDS,
+                gpdbWritableTargetTableWithDateWideRange.getName(),
+                POSTGRES_DRIVER_CLASS,
+                GPDB_PXF_AUTOMATION_DB_JDBC + gpdb.getMasterHost() + ":" + gpdb.getPort() + "/pxfautomation",
+                gpdb.getUserName(), null);
+        pxfJdbcWritableWithDateWideRange.setHost(pxfHost);
+        pxfJdbcWritableWithDateWideRange.setPort(pxfPort);
```
denalex, line 314: I would expect we would set `dateWideRange` option here, no ? where else is it set then ?

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
denalex, line 246: In general, what will happen if the option is not provided properly, both for read and write cases ?

We want to use different formatters for efficiency, but I think we should still thrive for being able to handle the data no matter what. So the option should be used as a hint from the user to us to use the most efficient formatter. However, if there is a parse error, we should try another formatter first (and maybe log a WARN message once per dataset, not per row) before giving up and failing the query.

automation/src/test/java/org/greenplum/pxf/automation/features/multiuser/MultiUserTest.java
```diff
@@ -28,7 +28,8 @@ public class MultiUserTest extends BaseFeature {
             "vc1   varchar(5)",
             "c1    char(3)",
             "bin   bytea",
-            "u     uuid"
+            "u     uuid",
+            "tmz timestamp with time zone"
```
denalex, line 32: ideally this column would go right after the `tm` one

automation/sqlrepo/features/jdbc/writable_date_wide_range/sql/query01.sql
```diff
@@ -0,0 +1,6 @@
+-- @description query01 for JDBC writable query
+SET timezone='America/Los_Angeles';
+
+INSERT INTO pxf_jdbc_writable_date_wide_range_on SELECT * FROM gpdb_types;
```
denalex, line 4: so gpdb_types now has wide range data ? is there a table that does not have it ?

automation/sqlrepo/features/jdbc/server_config/sql/query01.sql
```diff
@@ -1,3 +1,5 @@
 -- @description query01 for JDBC query with server config
+SET timezone='utc';
```
denalex, line 2: why do we now have to set TZ explicitly everywhere, because we include TZ sensitive column ? Why is it set to UTC in one set of cases and LA in others ?

denalex:
Is the PR description accurate ? Can we have more details in there what is the new and expected behavior and a summary of what has changed ?

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
tatb-vmware: Since these formatters are only use within the JDBC profile, we can most likely keep them here. Once other profiles start supporting wide date range timestamps with time zones, we can move them to `GreenplumDateTime` to unify them.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
tatb-vmware: I agree, I can make the change.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
tatb-vmware: `statement.setNull()` is used for setting the `java.sql.Types` to null since those were the original supported JDBC Java types. As of JDBC 4.2, timestamptz is supported by most major DB JDBC vendors so we can rely on `setObject()` to handle null properly.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
tatb-vmware: The `DateTimeParseException` does not provide sufficient information on which type of DateTime and which SQL type was failed to parsed.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
tatb-vmware: After discussion, we decided to keep the wrapped exception above so we can keep the helper function.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
tatb-vmware: The original intention of having different sets of formatters for `GET` and `SET` was because of the eras (AD/BC) being optional when we get the data from Postgres. However, we maybe able to use the same formatter for both and have Postgres figure out what to do with the data. I will test it out.

automation/src/test/java/org/greenplum/pxf/automation/features/jdbc/JdbcTest.java
tatb-vmware: Mistake when creating writable with `dateWideRange` on.

automation/sqlrepo/features/jdbc/writable_date_wide_range/sql/query01.sql
tatb-vmware: Mistake when creating writable with `dateWideRange` on.

automation/sqlrepo/features/jdbc/server_config/sql/query01.sql
tatb-vmware: I use the UTC timezone for testcases where the timezone offset is not the main purpose of the testcase since the automation testcases share the same internal gpdb_types table. For testcases where the purpose is to test if the datatypes are being properly read/written, I use a non-zero timezone offset.

automation/sqlrepo/features/jdbc/readable_date_wide_range/sql/query01.sql
```diff
@@ -27,10 +27,14 @@
 --
 -- end_matchsubs
 
-SET timezone='UTC';
```
axue-broadcom, line 30: Is there a reason why the casing here was changed?

server/pxf-jdbc/src/test/java/org/greenplum/pxf/plugins/jdbc/JdbcResolverTest.java
```diff
@@ -207,10 +247,31 @@ void getFieldOffsetDateTimeWithEraTest() throws SQLException {
...
+        isDateWideRange = false;
+        TimeZone.setDefault(TimeZone.getTimeZone("UTC"));
+        OffsetDateTime offsetDateTime = OffsetDateTime.parse("1234-11-22T11:15:30+02:00");
+        OneField oneField = getOneField(offsetDateTime, DataType.TIMESTAMP_WITH_TIME_ZONE.getOID(), "timestamptz");
+        assertTrue(oneField.val instanceof String);
+        assertEquals("1234-11-22 11:15:30+02", oneField.val);
+    }
+
+    @Test
+    void getFieldOffsetDateTimeWithHalfHourOffset() throws SQLException {
```
axue-broadcom, line 266: if someone were to provide a timestamp with timezone with an illegal offset (like `+30:00` or something), do we error out? does it go to the next day? Do we not have to worry about this case?

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -123,6 +123,20 @@ public class JdbcResolver extends JdbcBasePlugin implements Resolver {
             .optionalStart().appendPattern(DATE_TIME_FORMATTER_SPECIFIER).optionalEnd()
             .toFormatter();
 
+    /**
+     * OFFSET_DATE_TIME_SET_FORMATTER is used to transfer String to OffsetDateTime.
+     * Examples: "1980-08-10 17:10:20-07" -> 1980-08-10T17:10:20-07; "123456-10-19 11:12:13+06:30" -> +123456-10-19T11:12:13+6:30;
```
axue-broadcom, line 128: can you put each example on it's own line for easier readability?

automation/sqlrepo/features/jdbc/readable_date_wide_range/sql/query01.sql
tatb-vmware: Changed 'UTC' to 'utc' to make it more consistent with casing elsewhere in the codebase.

automation/sqlrepo/features/multi_user/sql/query01.sql
```diff
@@ -6,18 +6,20 @@
...
 --
 -- end_matchsubs
-
 GRANT ALL ON TABLE pxf_jdbc_readable TO PUBLIC;
 \set OLD_GP_USER :USER
 DROP ROLE IF EXISTS testuser;
 CREATE ROLE testuser LOGIN;
 
-SELECT t1, t2, num1, dub1, dec1, tm, r, bg, b, tn, sml, dt, vc1, c1, encode(bin, 'escape') AS bin, u FROM pxf_jdbc_readable ORDER BY t1;
+SET timezone='America/Los_Angeles';
```
yimingli-vmware, line 14: why we want to change the timezone for the multi user test?

automation/sqlrepo/features/multi_user/sql/query02.sql
```diff
@@ -12,9 +12,11 @@ DROP ROLE IF EXISTS testuser;
 CREATE ROLE testuser LOGIN;
 
 \connect - testuser
-SELECT t1, t2, num1, dub1, dec1, tm, r, bg, b, tn, sml, dt, vc1, c1, encode(bin, 'escape') AS bin, u FROM pxf_jdbc_readable_overrideddl ORDER BY t1;
+SET timezone='America/Los_Angeles';
```
yimingli-vmware, line 15: why change timezone?

automation/sqlrepo/features/multi_user/sql/query01.sql
tatb-vmware: `pxf_jdbc_readable` table now has `timestamp with timezone` data. If we do not set the timezone, it may error out depending on where which timezone the test are being run at.

automation/sqlrepo/features/multi_user/sql/query02.sql
tatb-vmware: `pxf_jdbc_readable` table now has `timestamp with timezone data`. If we do not set the timezone, it may error out depending on where which timezone the test are being run at.

automation/sqlrepo/features/multi_user/expected/query01.ans
```diff
@@ -14,79 +14,88 @@ DROP ROLE IF EXISTS testuser;
...
- row5                 | s_10 |    5 |   10 | 0.000000000001 | 2013-07-17 21:00:00 | 11.7 | 23456789 | f |  5 |  500 | 2015-03-09 | abcde | USD | 5   | a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a25
- row6                 | s_11 |    6 |   11 |       1234.889 | 2013-07-18 21:00:00 | 12.7 | 23456789 | t |  6 |  600 | 2015-03-10 | abcde | USD | 6   | a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a26
- row7                 | s_12 |    7 |   12 |         0.0001 | 2013-07-19 21:00:00 |  7.7 | 23456789 | f |  7 |  700 | 2015-03-11 | abcde | USD | 7   | a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a27
- row8                 | s_13 |    8 |   13 |    45678.00002 | 2013-07-20 21:00:00 |  7.7 | 23456789 | t |  8 |  800 | 2015-03-12 | abcde | EUR | 8   | a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a28
- row9                 | s_14 |    9 |   14 |        23457.1 | 2013-07-21 21:00:00 |  7.7 | 23456789 | f |  9 |  900 | 2015-03-13 | abcde | UAH | 9   | a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a29
-(26 rows)
+You are now connected to database "pxfautomation" as user "tatb".
```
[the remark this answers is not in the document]
tatb-vmware: We use matchsub to remove everything after the word "connected" so it does not fail in the pipeline. However, the other test cases use `gpadmin` within the answer files so I will change it to be consistent.

automation/sqlrepo/features/jdbc/readable_date_wide_range/expected/query01.ans
```diff
@@ -53,7 +54,7 @@ SELECT * FROM pxf_jdbc_readable_date_wide_range_on ORDER BY t1;
  row24_char_null        | s_16 |   11 |   37 |    0.123456789 | 2013-07-23 21:00:00    |  7.7 | 23456789 | f | 11 | 1100 | 2015-03-27    | abcde |     | 1   | a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a20 | 2013-07-23 12:00:00+00
  row25_binary_null      | s_16 |   11 |   37 |    0.123456789 | 2013-07-23 21:00:00    |  7.7 | 23456789 | f |  1 | 1100 | 2015-03-28    | abcde | USD |     | a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a21 | 2013-07-23 12:00:00+00
  row26_uuid_null        | s_16 |   11 |   37 |    0.123456789 | 2013-07-23 21:00:00    |  7.7 | 23456789 | f |  1 | 1100 | 2015-03-29    | abcde | USD | 1   |                                      | 2013-07-23 12:00:00+00
- row27_timestamptz_null | s_16 |   11 |   37 |    0.123456789 | 2013-07-23 21:00:00    |  7.7 | 23456789 | f |  1 | 1100 | 2015-03-30    | abcde | USD | 1   | a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a21 |
+ row27_timestamptz_null | s_16 |   11 |   37 |    0.123456789 | 2013-07-23 21:00:00    |  7.7 | 23456789 | f |  1 | 1100 | 2015-03-30    | abcde | USD | 1   | a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a21 |
```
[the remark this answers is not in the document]
tatb-vmware: `.out` file comes with the trailing spaces so I added them back into this `.ans` file to be consistent with the other files.

automation/sqlrepo/features/jdbc/readable_date_wide_range/expected/query01.ans
```diff
@@ -29,10 +29,11 @@
 --
 -- end_matchsubs
 
-SET timezone='UTC';
+SET timezone='utc';
 SET
+
 SELECT * FROM pxf_jdbc_readable_date_wide_range_on ORDER BY t1;
-           t1           |  t2  | num1 | dub1 |      dec1      |           tm           |  r   |    bg    | b | tn | sml  |      dt       |  vc1  | c1  | bin |                  u                   |            tmz
+           t1           |  t2  | num1 | dub1 |      dec1      |           tm           |  r   |    bg    | b | tn | sml  |      dt       |  vc1  | c1  | bin |                  u                   |            tmz
```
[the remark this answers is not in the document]
tatb-vmware: `.out` file comes with the trailing spaces so I added them back into this `.ans` file to be consistent with the other files.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
tatb-vmware: After some testing, we found that we can combine the formatters. This PR will include that change.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -471,30 +455,66 @@ public static void decodeOneRowToPreparedStatement(OneRow row, PreparedStatement
...
+
+        String value;
         try {
-            return LocalDate.parse(rawVal, LOCAL_DATE_SET_FORMATTER);
-        } catch (Exception e) {
-            throw new IllegalArgumentException("Failed to convert date '" + rawVal + "' to LocalDate class: " + e.getMessage(), e);
+            value = formatterOne.format(datetime);
+            if (value.charAt(0) == '+') {
+                throw new DateTimeParseException("year was too long", value, 0);
+            }
```
yimingli-vmware, line 476: Can you put some comments that in which case the code will throw this exception? `isDateWideRange = false` or `isDateWideRange = true`?

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -471,30 +455,66 @@ public static void decodeOneRowToPreparedStatement(OneRow row, PreparedStatement
...
+        if (datetime == null) {
+            return null;
+        }
+
+        String value;
         try {
-            return LocalDate.parse(rawVal, LOCAL_DATE_SET_FORMATTER);
-        } catch (Exception e) {
-            throw new IllegalArgumentException("Failed to convert date '" + rawVal + "' to LocalDate class: " + e.getMessage(), e);
+            value = formatterOne.format(datetime);
```
yimingli-vmware, line 473: If the this formatting failed, what is the type of the exception? If it's not `DateTimeParseException`, it will not be caught by the following catch block `catch (DateTimeParseException e)`.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -471,30 +455,66 @@ public static void decodeOneRowToPreparedStatement(OneRow row, PreparedStatement
...
     }
 
-    /**
-     * Convert a string to LocalDateTime class with formatter
-     *
-     * @param rawVal the LocalDateTime in a string format
-     * @return LocalDateTime
-     */
-    private LocalDateTime getLocalDateTime(String rawVal) {
+    private OffsetDateTime parseOffsetDateTime(String rawVal, DateTimeFormatter formatterOne, DateTimeFormatter formatterTwo) {
```
yimingli-vmware, line 485: Now we add some kind of `re-try` logic into these 4 formatting/parsing functions, do we have unit test testing that:
1. succeeded on the first formatter
2. failed on the first formatter but succeeded on the second
3. failed on both

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -32,94 +32,73 @@
...
-     * Examples: 2023-01-10 -> "2023-01-10 AD"; +12345-02-01 -> "12345-02-01 AD"; -0009-12-11 -> "0010-12-11 BC"
+     * LOCAL_DATE_FORMATTER is used to translate between String and LocalDate.
+     * Examples: "1977-12-11" <-> 1977-12-11
+     *           "456789-12-11" <-> +456789-12-11
+     *           "0010-12-11 BC" <-> -0009-12-11
      */
-    private static final DateTimeFormatter LOCAL_DATE_GET_FORMATTER = (new DateTimeFormatterBuilder())
-            .appendValue(ChronoField.YEAR_OF_ERA, 4, 9, SignStyle.NORMAL).appendLiteral("-")
+    private static final DateTimeFormatter LOCAL_DATE_FORMATTER = (new DateTimeFormatterBuilder())
+            .appendValue(ChronoField.YEAR_OF_ERA, 4, 9, SignStyle.NORMAL).appendLiteral('-')
```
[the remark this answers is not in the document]
tatb-vmware: Before in the GET case, we required years to come with a minimum of 4 digits and any year before 1000 will have leading zeroes. For example, the year 10 AD should come as `0010`. This is consistent with the ISO 8601 standard. We should have the same requirement for the SET case to avoid confusion if `01-02-03` means `January 2nd, 0003 AD` or `Feburary 3rd, 0001 AD`. There will be no formatting errors since in the SET case it gets the data from the Java object which will always be able to provide 4 digits for the year.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -32,94 +32,73 @@
...
+     *           "123456-10-19 11:12:13+06:30" <-> +123456-10-19T11:12:13+6:30
+     *           "1234-10-19 10:11:15.456+00 BC" <-> -1233-10-19T10:11:15.456+00
      */
-    private static final DateTimeFormatter LOCAL_DATE_TIME_SET_FORMATTER = (new DateTimeFormatterBuilder())
-            .appendValue(ChronoField.YEAR_OF_ERA, 1, 9, SignStyle.NORMAL).appendLiteral('-')
-            .appendValue(ChronoField.MONTH_OF_YEAR, 1, 2, SignStyle.NORMAL).appendLiteral('-')
-            .appendValue(ChronoField.DAY_OF_MONTH, 1, 2, SignStyle.NORMAL).appendLiteral(" ")
+    private static final DateTimeFormatter OFFSET_DATE_TIME_FORMATTER = (new DateTimeFormatterBuilder())
+            .appendValue(ChronoField.YEAR_OF_ERA, 4, 9, SignStyle.NORMAL).appendLiteral('-')
+            .appendValue(ChronoField.MONTH_OF_YEAR, 2, 2, SignStyle.NORMAL).appendLiteral('-')
```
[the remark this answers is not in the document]
tatb-vmware: Answer above.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -212,32 +195,36 @@ public List<OneField> getFields(OneRow row) throws SQLException {
...
                     if (isDateWideRange) {
                         LocalDate localDate = result.getObject(colName, LocalDate.class);
-                        value = localDate != null ? localDate.format(LOCAL_DATE_GET_FORMATTER) : null;
+                        value = formatDateTimeValues(localDate, LOCAL_DATE_FORMATTER, GreenplumDateTime.DATE_FORMATTER);
                     } else {
-                        Date date = result.getDate(colName);
-                        value = date != null ? date.toLocalDate().format(GreenplumDateTime.DATE_FORMATTER) : null;
+                        LocalDate localDate = result.getDate(colName) == null
+                                ? null
+                                : result.getDate(colName).toLocalDate();
```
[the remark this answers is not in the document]
tatb-vmware: I can inline it.

server/pxf-jdbc/src/test/java/org/greenplum/pxf/plugins/jdbc/JdbcResolverTest.java
djianwen-vmware: Just curious, what if the time zone different is not a whole number hour? For example India time zone IST (UTC+05:30), is the final assertEquals like "1234-11-22 11:15:30+05:30"?

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -471,30 +454,91 @@ public static void decodeOneRowToPreparedStatement(OneRow row, PreparedStatement
...
+            return null;
+        }
+
+        String value;
+        try {
+            value = formatterOne.format(datetime);
+            if (value.charAt(0) == '+') {
+                // When non-DateWideRange formatter is used first and the number of digits in the YEAR is greater than 4,
+                // it will add an unwanted '+' to the start of the string which causes an error.
+                throw new DateTimeParseException("year was too long", value, 0);
```
djianwen-vmware, line 475: Is there a test case for this case? I noticed that the test `setFieldDateWithMoreThan4digitsInYearWithoutWideRangeTest()` was removed.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -141,6 +128,12 @@ public class JdbcResolver extends JdbcBasePlugin implements Resolver {
 
     private static final Logger LOG = LoggerFactory.getLogger(JdbcResolver.class);
 
+    private boolean logWarnForDateWideRange = false;
+    private static final String dateWideRangeWarnMsg = "Please consider turning on DateWideRange for better performance.";
+    private final String dateWideRangelogMsg = isDateWideRange
```
denalex, line 133:
```suggestion
    private final String dateWideRangeLogMsg = isDateWideRange 
```
this can also be an array of constants (with 0/1 index determined by boolean conversion)

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -212,33 +206,29 @@ public List<OneField> getFields(OneRow row) throws SQLException {
                     value = result.getString(colName);
                     break;
                 case DATE:
+                    // As of JDBC 4.2, getObject API supports retrieval of LocalDate, LocalDateTime, and OffsetDateTime.
+                    // We use getDate and getTimestamp because Hive JDBC connector does not fully support JDBC 4.2 API.
+                    LocalDate localDate;
                     if (isDateWideRange) {
-                        LocalDate localDate = result.getObject(colName, LocalDate.class);
-                        value = localDate != null ? localDate.format(LOCAL_DATE_GET_FORMATTER) : null;
+                        localDate = result.getObject(colName, LocalDate.class);
```
denalex, line 213: so it will not be possible to use wideRange hint with Hive ? Should we always use non-object API then ?

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -212,33 +206,29 @@ public List<OneField> getFields(OneRow row) throws SQLException {
...
-                        OffsetDateTime offsetDateTime = result.getObject(colName, OffsetDateTime.class);
-                        value = offsetDateTime != null ? offsetDateTime.format(OFFSET_DATE_TIME_GET_FORMATTER) : null;
-                    } else {
-                        throw new UnsupportedOperationException(
-                                String.format("Field type '%s' (column '%s') is not supported",
-                                        DataType.get(oneField.type),
-                                        column));
-                    }
+                    // OffsetDateTime is the only class that JDBC drivers will most likely to respect for returning timestamptz.
+                    OffsetDateTime offsetDateTime = result.getObject(colName, OffsetDateTime.class);
```
denalex, line 230: can we add a comment that this will not work with Hive ? Is there a way to get it using a string ?

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -166,6 +159,7 @@ public JdbcResolver() {
      */
     @Override
     public List<OneField> getFields(OneRow row) throws SQLException {
+        logWarnForDateWideRange = false;
```
denalex, line 162: I thought we want to do this just once per resolver run, not for every row

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -252,6 +242,11 @@ public List<OneField> getFields(OneRow row) throws SQLException {
 
             oneField.val = result.wasNull() ? null : value;
         }
+
+        if (logWarnForDateWideRange) {
+            LOG.warn(dateWideRangeWarnMsg);
+        }
```
denalex, line 248: this seems mis-placed, we should log where we fallback the formatters

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -264,10 +259,10 @@ public List<OneField> getFields(OneRow row) throws SQLException {
      * moment, there is no way to correct the order of the fields if it is not.
      * In practice, the 'record' provided is always ordered the right way.
      * @throws UnsupportedOperationException if field of some type is not supported
-     * @throws ParseException                if the record cannot be parsed
      */
     @Override
-    public OneRow setFields(List<OneField> record) throws UnsupportedOperationException, ParseException {
+    public OneRow setFields(List<OneField> record) throws UnsupportedOperationException {
+        logWarnForDateWideRange = false;
```
denalex, line 265: same comment as for read case

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -356,6 +346,10 @@ public OneRow setFields(List<OneField> record) throws UnsupportedOperationExcept
             }
         }
 
+        if (logWarnForDateWideRange) {
+            LOG.warn(dateWideRangeWarnMsg);
```
denalex, line 350: same comment

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -471,30 +446,63 @@ public static void decodeOneRowToPreparedStatement(OneRow row, PreparedStatement
...
-        } catch (Exception e) {
-            throw new IllegalArgumentException("Failed to convert date '" + rawVal + "' to LocalDate class: " + e.getMessage(), e);
+            value = formatterOne.format(datetime);
+            if (value.charAt(0) == '+') {
+                // When non-DateWideRange formatter is used first and the number of digits in the YEAR is greater than 4,
+                // it will add an unwanted '+' to the start of the string which causes an error.
+                throw new DateTimeParseException("year was too long", value, 0);
+            }
+        } catch (DateTimeParseException e) {
+            LOG.info(dateWideRangelogMsg);
```
denalex, line 475: this should happen only once ever, with WARN level, not for every row

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -471,30 +446,63 @@ public static void decodeOneRowToPreparedStatement(OneRow row, PreparedStatement
...
+        // The remaining formatter will be used if the first chosen formatter fails.
+        DateTimeFormatter formatterOne = formatters[BooleanUtils.toInteger(isDateWideRange)];
+        DateTimeFormatter formatterTwo = formatters[BooleanUtils.toInteger(!isDateWideRange)];
         try {
-            return LocalDateTime.parse(rawVal, LOCAL_DATE_TIME_SET_FORMATTER);
-        } catch (Exception e) {
-            throw new IllegalArgumentException("Failed to convert timestamp '" + rawVal + "' to the LocalDateTime class: " + e.getMessage(), e);
+            value = formatterOne.parse(rawVal, convertToConcreteType);
+        } catch (DateTimeParseException e) {
+            LOG.info(dateWideRangelogMsg);
```
denalex, line 502: should happen only once ever, with WARN level, ideally include a column name and the value seen

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -141,6 +128,12 @@ public class JdbcResolver extends JdbcBasePlugin implements Resolver {
 
     private static final Logger LOG = LoggerFactory.getLogger(JdbcResolver.class);
 
+    private boolean logWarnForDateWideRange = false;
+    private static final String dateWideRangeWarnMsg = "Please consider turning on DateWideRange for better performance.";
+    private final String dateWideRangelogMsg = isDateWideRange 
+            ? "Date Wide Range formatter failed, try with the default formatter"
+            : "The default formatter failed, try with the Date Wide Range formatter";
```
denalex, line 135: these should be more like instructions to the users to use the hint property if we encountered some special data

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -455,65 +446,62 @@ public static void decodeOneRowToPreparedStatement(OneRow row, PreparedStatement
     }
 
     /**
-     * Formats a java.time.* datetime value using two formatters in a order depending on if DateWideRange is on.
-     * If DateWideRange is off but the value successfully parses with the DateWideRange formatter, log a warning
-     * to turn on the DateWideRange.
+     * Formats a java.time.TemporalAccessor value using two formatters in order and logs a warning if the first formatter fails.
```
axue-broadcom, line 449: Add a comment that the usage of the formatters is dependent on isDateWideRange. It is an integral aspect of this function

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -455,65 +446,62 @@ public static void decodeOneRowToPreparedStatement(OneRow row, PreparedStatement
...
+     * @return the formatted String from Temporal object
      */
-    private String formatDateTimeValues(TemporalAccessor datetime, DateTimeFormatter formatterOne, DateTimeFormatter formatterTwo) throws DateTimeParseException {
+    private String formatDateTimeValues(TemporalAccessor datetime, DateTimeFormatter[] formatters) throws DateTimeParseException {
         if (datetime == null) {
             return null;
         }
 
+        // When isDateWideRange is 0, the first formatter will be the regular formatter.
+        // When isDateWideRange is 1, the first formatter will be DateWideRange formatter.
```
axue-broadcom, line 461:
```suggestion
        // The array will always come in with the regular formatter at index 0 and the DateWideRange formatter at index 1
        // We can use this to select the first formatter to try based off the isDateWideRange boolean value
        // When isDateWideRange is 0 (false), the first formatter will be the regular formatter.
        // When isDateWideRange is 1 (true), the first formatter will be DateWideRange formatter.
```

or something along these lines

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -455,65 +446,62 @@ public static void decodeOneRowToPreparedStatement(OneRow row, PreparedStatement
...
-            LOG.trace("First formatter failed, used second formatter");
-            logDebugForDateWideRange = true;
-        }
-        return value;
-    }
-
-    private LocalDateTime parseLocalDateTime(String rawVal, DateTimeFormatter formatterOne, DateTimeFormatter formatterTwo) {
-        LocalDateTime value;
+    /**
+     * Parses a String to type T using two formatters in order and logs a warning if the first formatter fails.
```
axue-broadcom, line 483: same comment as above for the format function

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -455,65 +446,62 @@ public static void decodeOneRowToPreparedStatement(OneRow row, PreparedStatement
...
+     * @param rawVal String to parse into type T
+     * @param formatters an array of valid formatters for type T
+     * @param convertToConcreteType Lambda function to convert TemporalAccessor into type T
+     * @return the parsed LocalDate, LocalDateTime, or OffsetDateTime
+     * @param <T> one of the three possible types: LocalDate, LocalDateTime, OffsetDateTime
+     */
+    private <T extends TemporalAccessor> T parseTemporalAccessor(String rawVal, DateTimeFormatter[] formatters, TemporalQuery<T> convertToConcreteType) {
+        T value;
+
+        // When isDateWideRange is 0, the first formatter will be the regular formatter.
```
axue-broadcom, line 494: same comment as above

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
tatb-vmware: That is correct, `dateWideRange` does not work with Hive JDBC. Since the object API provides `timestamptz` and eras support, I think we should use object API when possible and needed.

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
tatb-vmware: I can add the comment. Looking into the Hive JDBC connector version that we use and some newer versions, I have not found a way to get it as a string.

yimingli-vmware (approved):

denalex (approved):

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -141,6 +128,11 @@ public class JdbcResolver extends JdbcBasePlugin implements Resolver {
 
     private static final Logger LOG = LoggerFactory.getLogger(JdbcResolver.class);
 
+    private static final String dateWideRangeWarnMsg = "Failed to use the standard formatter, so we had to fallback to the wide date range formatter. " +
+            "However, there is a performance penalty. To have better performance, specify date_wide_range=true at the table definition";
```
axue-broadcom, line 132:
```suggestion
            "However, there is a performance penalty. To have better performance, specify date_wide_range=true in the table definition";

```

they should also be able to set the value in the jdbc-site.yml (but this would affect all tables that use this profile)

server/pxf-jdbc/src/main/java/org/greenplum/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -212,33 +204,30 @@ public List<OneField> getFields(OneRow row) throws SQLException {
                     value = result.getString(colName);
                     break;
                 case DATE:
+                    // As of JDBC 4.2, getObject API supports retrieval of LocalDate, LocalDateTime, and OffsetDateTime.
+                    // We use getDate and getTimestamp because Hive JDBC connector does not fully support JDBC 4.2 API.
```
axue-broadcom, line 208: can you add a Hive Jira link that showcases this. Just so that in the future if things change we can link it back to something

axue-broadcom (approved):

PR #1105, opened 2024-03-15
Bumped Spring Framework version to 5.3.33

denalex:
This PR updates the Spring Framework (and the associated spring-mvc modules) to versions 5.3.33 to pick the fix for the async race condition (https://github.com/spring-projects/spring-framework/issues/32342). 

The race condition was manifesting itself in PXF with `partial file transfer error` that GPDB would receive from PXF when executing LIMIT queries or queries issued after LIMIT queries with hight load and concurrency. Tomcat would close the connections without writing a 0-length chunk even if there was no error reported in PXF.

### DIFF
--- a/server/build.gradle
+++ b/server/build.gradle
@@ -29,6 +29,10 @@ plugins {
     id 'io.spring.dependency-management' version '1.0.11.RELEASE'
 }
 
+// override Spring Framework version to pick the Spring MVC 5.3.33 with the async race condition fix
+// https://github.com/spring-projects/spring-framework/issues/32342
+ext['spring-framework.version'] = '5.3.33'
+
 def javaProjects = subprojects.findAll {
     it.name.startsWith("pxf-")
 }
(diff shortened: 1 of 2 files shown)


tatb-vmware (approved):

server/build.gradle
```diff
@@ -29,6 +29,10 @@ plugins {
     id 'io.spring.dependency-management' version '1.0.11.RELEASE'
 }
 
+// override Spring Framework version to pick the Spring MVC 5.3.33 with the async race condition fix
+// https://github.com/spring-projects/spring-framework/issues/32342
+ext['spring-framework.version'] = '5.3.33'
```
axue-broadcom, line 34: why do we do this dependency override here as`ext[]` but we do the postgres dependency override in a different format down on line 113. Can we not do this override with the others?

yimingli-vmware (approved):

server/build.gradle
denalex: This is the preferred way to override the versions that SpringBoot bundles via their `-bom` POM. We should also use this for Postgres and Log4j, we can do that as a separate PR.

axue-broadcom (approved):

PR #1111, opened 2024-03-20
Adjusted PXF error handling and determining client disconnects

denalex:
Spring MVC 5.3.33 now wraps ClientAbortException into its own `AsyncRequestNotUsableException` when client disconnect happens on async thread, so we need to adjust how we determine this scenario.

This PR also changes exception handling when the response is committed. It used to set an error on the response, but this was causing an IllegalStateException from Tomcat. Now we just re-throw the original exception to make Tomcat terminate the client connection abnormally without sending the closing 0-length chunk, thus informing the client that there was a processing error.

### DIFF
--- a/server/pxf-api/src/main/java/org/greenplum/pxf/api/utilities/Utilities.java
+++ b/server/pxf-api/src/main/java/org/greenplum/pxf/api/utilities/Utilities.java
@@ -389,8 +389,8 @@ public class Utilities {
     }
 
     /**
-     * Inspects the throwable and its causes to determine if the throwable due to the client disconnecting
-     * from the socket. Spring MVC since 5.3.33 is wrapping Tomcat's ClientAbortException during async processing into
+     * Inspects the throwable and its causes to determine if the throwable is due to the client disconnecting
+     * from the socket. Spring MVC 5.3.33+ wraps Tomcat's ClientAbortException during async processing into
      * AsyncRequestNotUsableException. To accommodate potential future changes in Spring MVC, the logic here
      * will traverse the exception chain to determine whether a ClientAbortException is in the chain.
      * @param e exception to analyze
--- a/server/pxf-api/src/main/java/org/greenplum/pxf/api/utilities/Utilities.java
+++ b/server/pxf-api/src/main/java/org/greenplum/pxf/api/utilities/Utilities.java
@@ -19,7 +19,9 @@ package org.greenplum.pxf.api.utilities;
  * under the License.
  */
 
+import org.apache.catalina.connector.ClientAbortException;
 import org.apache.commons.lang.StringUtils;
+import org.apache.commons.lang3.exception.ExceptionUtils;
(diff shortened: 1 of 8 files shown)


server/pxf-api/src/main/java/org/greenplum/pxf/api/utilities/Utilities.java
```diff
@@ -385,4 +387,16 @@ public static boolean parseBooleanProperty(Configuration configuration, String p
             ));
         }
     }
+
+    /**
+     * Inspects the throwable and its causes to determine if the throwable due to the client disconnecting
```
tatb-vmware, line 392:
```suggestion
     * Inspects the throwable and its causes to determine if the throwable is due to the client disconnecting
```

server/pxf-api/src/test/java/org/greenplum/pxf/api/utilities/UtilitiesTest.java
```diff
@@ -424,4 +426,13 @@ public void testParseBooleanProperty() {
 
         assertTrue(Utilities.parseBooleanProperty(configuration, "trueWithExtraWhitespace", false));
     }
+
+    @Test
+    public void isClientDisconnectException() {
```
tatb-vmware, line 431: Why does this test have multiple exceptions at the same time?

tatb-vmware (approved):

server/pxf-service/src/test/java/org/greenplum/pxf/service/spring/PxfExceptionHandlerTest.java
```diff
@@ -26,9 +30,21 @@ public void setup() {
...
 
     @Test
-    public void testHandlePxfRuntimeException() throws IOException {
+    public void testHandlePxfRuntimeException_ResponseNotCommitted() throws IOException {
+        when(mockResponse.isCommitted()).thenReturn(false);
         handler.handlePxfRuntimeException(new PxfRuntimeException("foo"), mockResponse);
         verify(mockResponse).sendError(500);
         verifyNoMoreInteractions(mockResponse);
     }
+    @Test
```
axue-broadcom, line 39: nit: new line above

server/pxf-api/src/main/java/org/greenplum/pxf/api/utilities/Utilities.java
```diff
@@ -385,4 +387,16 @@ public static boolean parseBooleanProperty(Configuration configuration, String p
...
+     * from the socket. Spring MVC since 5.3.33 is wrapping Tomcat's ClientAbortException during async processing into
```
axue-broadcom, line 393:
```suggestion
     * from the socket. Spring MVC 5.3.33+ wraps Tomcat's ClientAbortException during async processing into
```

yimingli-vmware (approved):

PR #1113, opened 2024-03-21
Modify build pipeline to make it easier to set release pipelines

axue-broadcom:
This commit removes the `build` make target in favor of 2 new make targets: `release` and `main`. 

As we no longer expect releases to be done off main, `make main` spins up the current `pxf-build` pipeline. However, shipit files for the main branch point to `${BUCKET}/shipit/noop` and should effectively be ignored.

The `release` make target is to be used for all releases going forward and expects the variable `RELEASE_BRANCH` to be set. Release branches are expected in the form `branch-<major.minor>.x`. We expect these release pipelines to be long standing and so a release for 6.10.99 would be expected to use the same pipeline as 6.10.4. 
 
Running `make release RELEASE_BRANCH=branch-6.10.x` would do the following: 
- create a pipeline with the name `pxf-build-branch-6.10.x` that points to the `branch-6.10.x` branch for deployment
- any shipit file would live in `${BUCKET}/shipit/branch-6.10.x`

### DIFF
--- a/concourse/Makefile
+++ b/concourse/Makefile
@@ -77,13 +77,13 @@ set-release-pipeline: set-build-pipeline-base
 endif
 
 .PHONY: set-main-pipeline
+set-main-pipeline: SHIPIT_REGEX=noop
 set-main-pipeline: NOTIFY=true
 set-main-pipeline: PIPELINE_NAME=$(BUILD_PIPELINE_NAME)
 set-main-pipeline: RELEASE_BRANCH=main
 set-main-pipeline: ENVIRONMENT=prod
 set-main-pipeline: BUCKET_PREFIX=prod
 set-main-pipeline: IMAGE_TAG=latest
-set-main-pipeline: SHIPIT_REGEX=noop
 set-main-pipeline: set-build-pipeline-base
 
 .PHONY: set-dev-release-pipeline
--- a/concourse/Makefile
+++ b/concourse/Makefile
@@ -69,7 +69,6 @@ set-release-pipeline:
 ifndef RELEASE_BRANCH
 	$(error RELEASE_BRANCH is not set, please set a release branch)
 else
-# substitute any / for - in the pipeline
 set-release-pipeline: SHIPIT_REGEX="$(RELEASE_BRANCH)/greenplum-pxf-(.*).txt"
 set-release-pipeline: NOTIFY=true
 set-release-pipeline: PIPELINE_NAME=$(BUILD_PIPELINE_NAME)-$(RELEASE_BRANCH)
--- a/concourse/Makefile
+++ b/concourse/Makefile
@@ -111,7 +111,6 @@ set-build-pipeline-base:
 		environment=$(ENVIRONMENT) \
 		gchat_notification=$(NOTIFY) \
 		user=$(USER) \
-		git_branch=$(BRANCH) \
 		num_gpdb5_versions=$(NUM_GPDB5_VERSIONS) \
 		num_gpdb6_versions=$(NUM_GPDB6_VERSIONS) \
 		num_gpdb7_versions=$(NUM_GPDB7_VERSIONS) >"$${PIPELINE_FILE}" && \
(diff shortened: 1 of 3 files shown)


yimingli-vmware (approved):

concourse/Makefile
```diff
@@ -50,6 +51,15 @@ ifeq ($(CHECK_CREDS), true)
...
 endif
 
+ifneq "$(RELEASE_BRANCH)" "main"
+# substitute any / for - in the pipeline
+	RELEASE_NAME:=-$(shell echo $(subst /,-,$(RELEASE_BRANCH)) | tr A-Z a-z)
+else
+# if we're on the main branch, there should be no release name
+	RELEASE_NAME:=""
+endif
+SHIPIT_REGEX="pxf-build$(RELEASE_NAME)/greenplum-pxf-(.*).txt"
```
denalex, line 61: do we need to have the pipeline name `pxf-build` in there ? And also we said we practically will no longer release from main

tatb-vmware (approved):

concourse/Makefile
axue-broadcom: I did something similar for gemfire, I think at time time, my reasoning was that it would be easier to map pipeline name, but I am ok removing the additional words

denalex: I am thinking that there is an outside tool that has to place the ship file in this bucket and the tool only knows release number (and we want to keep it that way for simplicity on user experience). So we need to derive bucket from the release number and right now we would need to hardcode `pxf-build` in the pattern and arrive to the the branch name by hardcoding `branch-` and then append a derivative of release number (use `.x` for the 3-rd digit). Nothing wrong with that, i just don't know if there is a value of having a pipeline name in there, since we will use only a single pipeline for releasing from a given branch

concourse/Makefile
```diff
@@ -71,57 +62,65 @@ dev-cloudbuild: set-dev-cloudbuild-pipeline
...
+set-release-pipeline:
+ifndef RELEASE_BRANCH
+	$(error RELEASE_BRANCH is not set, please set a release branch)
+else
+set-release-pipeline: SHIPIT_REGEX="$(RELEASE_BRANCH)/greenplum-pxf-(.*).txt"
+set-release-pipeline: NOTIFY=true
+set-release-pipeline: PIPELINE_NAME=$(BUILD_PIPELINE_NAME)-$(RELEASE_BRANCH)
+set-release-pipeline: ENVIRONMENT=prod
+set-release-pipeline: BUCKET_PREFIX=prod
+set-release-pipeline: IMAGE_TAG=latest
```
denalex, line 75: this will need to get revisited later

concourse/Makefile
```diff
@@ -71,57 +62,65 @@ dev-cloudbuild: set-dev-cloudbuild-pipeline
...
+set-main-pipeline: SHIPIT_REGEX=noop
+set-main-pipeline: set-build-pipeline-base
 
 .PHONY: set-dev-release-pipeline
 set-dev-release-pipeline: SHIPIT_REGEX="$(USER)-$(BRANCH)/greenplum-pxf-(.*).txt"
-set-dev-release-pipeline:
+set-dev-release-pipeline: PIPELINE_NAME=$(DEV_BUILD_PIPELINE_NAME)
+set-dev-release-pipeline: RELEASE_BRANCH=$(BRANCH)
+set-dev-release-pipeline: ENVIRONMENT=dev
+set-dev-release-pipeline: set-build-pipeline-base
```
denalex, line 94: are we doing it twice ? here and on line 98 ?

denalex (approved):

concourse/Makefile
```diff
@@ -51,7 +51,8 @@ SET_PIPELINE += --check-creds
 endif
 
 .PHONY: build certification dev-release dev pr cloudbuild dev-cloudbuild longevity
-build: set-build-pipeline
+release: set-release-pipeline
+main: set-main-pipeline
```
denalex, line 55: do you need to add these 2 targets to PHONY list ?

concourse/Makefile
axue-broadcom: Yes. For now, until we figure out the tagging situation for docker images, it will be pointed to latest. But in the future, we will probably need a case statement or something to look for correct release image.

yimingli-vmware (approved):

PR #1122, opened 2024-04-03
Pxf cluster config validation

djianwen-vmware:
* add structs `PxfServiceGroup`, `PxfCluster`, `PxfHost`, and `PxfDeployment`.
* add `Validation()` to  `PxfDeployment`. 
* all the validation rules can be found under `config_rules.go` as private functions.
* IsValidAddress() is public for testing purpose
  * this function contains a permissive regex to match FQDN, which means it will return error only if it's obviously an invalid format.
* add tests with 100% statement coverage for `config.go` and `config_rules.go`

Validation rules and its corresponding functions:

- [x] Attribute values should be valid for their intended types (numbers, booleans, etc): achieved by Golang `int`, `bool`, etc. Only need to handle them when parsing the config file.
- [x] Cluster::name MUST be unique: achieved by Golang `map`
- [x] Cluster::collocated MUST not be NULL: achieved by Golang `bool`
- [x] Cluster::collocated==true SHOULD NOT have a list of hosts, they WILL be ignored: `pxfClusterIgnoreHostsWhenCollocated`
- [x] Cluster::collocated==true SHOULD NOT have a Cluster::endpoint, it WILL be ignored: `pxfClusterIgnoreEndpointWhenCollocated`
- [x] Cluster::collocated==false AND Cluster::endpoint==NULL MUST have a non-empty list of hosts: `pxfClusterEndpointOrHostsMustExistWhenExternal`
- [x] Cluster::endpoint != NULL MUST have a value of a valid IPv4 or FQDN compliant syntax: `pxfClusterValidEndpoint`
- [x] Host::hostname MUST be unique (a single host must belong to only one cluster): `pxfDeploymentHostnameUnique`
- [x] Host::hostname MUST have a value with either IPv4 or FQDN compliant syntax: `pxfHostValidHostname`
- [x] ServiceGroup::name MUST NOT be null / empty: `pxfServiceGroupNameNotEmpty`
- [x] ServiceGroup::name MUST be unique across all Clusters: `pxfDeploymentGroupNameUnique`
- [x] ServiceGroup::ports MUST be a non-empty array with at least 1 element: `pxfServiceGroupAtLeastOnePort`
- [x] ServiceGroup::ports elements MUST be an integer value between 1-65535: `pxfServiceGroupValidPorts`
- [x] ServiceGroup::ports elements SHOULD be larger than 1023 and smaller than 32768 (between privileged and ephemeral ranges): `pxfServiceGroupPortsWarning`
- [x] There MUST be at least one Cluster: `pxfDeploymentAtLeastOneCluster`
- [x] There MUST be no more than one Cluster::collocated==true: `pxfDeploymentNoMoreThanOneCollocated`
- [x] There MUST be at least one ServiceGroup per Cluster: `pxfClusterAtLeastOneServiceGroup`
- [x] There must be one cluster called "default"


cli/config/config_rules.go
```diff
@@ -0,0 +1,217 @@
...
+			)
+		}
+	}
+	return nil
+}
+
+// pxfHostValidHostname
+// A PXF host MUST have a hostname which is either a valid IPv4 address or FQDN
+func pxfHostValidHostname(p PxfHost) error {
+	if p.Hostname != "" && !IsValidAddress(p.Hostname) {
```
denalex, line 192: empty hostname is a problem by itself

cli/config/config.go
```diff
@@ -0,0 +1,94 @@
...
+type pxfClusterValidator func(p PxfCluster) error
+
+type PxfHost struct {
+	Hostname string
+}
+
+type PxfDeployment struct {
+	Name     string
+	Clusters map[string]*PxfCluster
+}
```
denalex, line 26: can we re-arrange struct definitions in a logical top-down order: Deployment, Cluster, Host, ServiceGroup ? And also apply a similar order to the validators ?

cli/config/config_rules.go
```diff
@@ -0,0 +1,217 @@
...
+		return fmt.Errorf("the endpoint `%s` of the PXF cluster `%s` is not a valid IPv4 address "+
+			"or FQDN", p.Endpoint, p.Name)
+	}
+	return nil
+}
+
+// pxfClusterAtLeastOneServiceGroup
+// A PXF cluster MUST have at least one PXF service group
+func pxfClusterAtLeastOneServiceGroup(p PxfCluster) error {
+	if len(p.Groups) == 0 {
```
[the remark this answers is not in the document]
djianwen-vmware: > do we need null check for `p.Groups` (and other similar checks) ?

In Golang it's not necessary, because len() of a nil slice is also 0.

cli/config/config_rules.go
```diff
@@ -0,0 +1,217 @@
...
+		return fmt.Errorf("there should be at least one service group in PXF cluster `%s`", p.Name)
```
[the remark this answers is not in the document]
djianwen-vmware: > consistency some messages say "the PXF cluster ..", some are missing "the", not sure whats the right choice

Good catch. I will add `the` to them to be consistent.

cli/config/config_rules.go
djianwen-vmware: > empty hostname is a problem by itself

Good catch. I fixed it and added a new test case.

cli/config/config.go
```diff
@@ -0,0 +1,94 @@
...
+type pxfDeploymentValidator func(p PxfDeployment) error
+
+func (p PxfDeployment) Validate() error {
```
tatb-vmware, line 29: I see there are many `Validate` functions and the function type defined with `func(T pxfStructType) error`. Is there any benefits from creating an interface to compose all of these vs the current implementation?

cli/config/config.go
djianwen-vmware: > can we re-arrange struct definitions in a logical top-down order: Deployment, Cluster, Host, ServiceGroup ? And also apply a similar order to the validators ?

It makes sense. I re-arranged them. Thanks!

cli/config/config_rules.go
```diff
@@ -0,0 +1,217 @@
...
+import (
+	"errors"
+	"fmt"
+	"github.com/greenplum-db/gp-common-go-libs/gplog"
+	"net"
+	"regexp"
+)
+
+// pxfDeploymentNameNotEmpty
+// PXF deployment name MUST NOT be empty
```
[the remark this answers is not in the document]
djianwen-vmware: > I don't think we will capture deployment name in our configs, so this is probably not needed

Should I also remove the field from the struct, if it is not needed?

cli/config/config_test.go
```diff
@@ -0,0 +1,370 @@
...
+						"test-group-2": {
+							Name:  "test-group-4",
+							Ports: []int{7777, 8888},
+						},
+					},
+				},
+			},
+		}
+	})
+	Context("deployment name must not be empty", func() {
```
tatb-vmware, line 116: Nit: can we capitalize all the First letter of the first word of the sentence within the `Context` string?

cli/config/config_rules.go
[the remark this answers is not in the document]
djianwen-vmware: I will remove it for now. If we need it in the future, I can bring it back.

cli/config/config_rules.go
```diff
@@ -0,0 +1,217 @@
...
+	if len(p.Ports) == 0 {
+		return fmt.Errorf("there should be at least one port in PXF service group `%s`", p.Name)
+	}
+	return nil
+}
+
+// pxfServiceGroupValidPorts
+// PXF service group MUST have ports which are integers between 1-65535
+func pxfServiceGroupValidPorts(p PxfServiceGroup) error {
+	MinPortNumber := 1
```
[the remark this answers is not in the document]
djianwen-vmware: > Do we expect the min and max to be changeable and will we need to use these values elsewhere? I am thinking if we should pull these out somewhere else as constants.

The values probably won't change because they are the standard port range limit.
For now this is the only place we used them. So I just put them inside the function. I think if we need them elsewhere we can pull them out as constants at that time.

cli/config/config_rules.go
```diff
@@ -0,0 +1,217 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"github.com/greenplum-db/gp-common-go-libs/gplog"
+	"net"
+	"regexp"
+)
+
+// pxfDeploymentNameNotEmpty
```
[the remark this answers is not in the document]
djianwen-vmware: > I am curious why we have the function name repeated as a single comment above all the functions. What is the reason for this?

I think it's the Golang doc requirement. If I don't have a comment starting with its function name, the IDE gives me a warning.

cli/config/config.go
djianwen-vmware: > I see there are many `Validate` functions and the function type defined with `func(T pxfStructType) error`. Is there any benefits from creating an interface to compose all of these vs the current implementation?

I tried but unfortunately interface doesn't work perfectly here. Because the input parameters(structs) are different. Interface in Golang doesn't support dynamic inputs (you can cast them if you want, but then it brings casting-error handling to each of rules) And you will need 3 interfaces in this case.

cli/config/config_test.go
djianwen-vmware: > Nit: can we capitalize all the First letter of the first word of the sentence within the `Context` string?

I checked other test files, `pxf_test.go` is using lower case only, while `cluster_test.go` is using both lower case and upper case. Probably need an style agreement on this?

cli/config/config.go
tatb-vmware: I see, thank you for the context on this.

cli/config/config_rules.go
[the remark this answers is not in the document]
tatb-vmware: Looking at the Go docs requirements (https://go.dev/doc/comment), I found that the "idiomatic" way for function comments is to have it start with `<function name> [does something here]...`. We should have a discussion if we want to follow this.

Additional context I found here: https://stackoverflow.com/questions/35591301/function-name-in-comment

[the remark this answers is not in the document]
djianwen-vmware: > Looking at the Go docs requirements (https://go.dev/doc/comment), I found that the "idiomatic" way for function comments is to have it start with `<function name> [does something here]...`. We should have a discussion if we want to follow this.
> 
> Additional context I found here: https://stackoverflow.com/questions/35591301/function-name-in-comment

Thanks for the additional context. By the way there are existing functions following this style.  E.g. `GenerateStatusReport`, `Execute`, etc.

cli/config/config_rules.go
```diff
@@ -0,0 +1,193 @@
...
+			found = true
+		}
+	}
+	return nil
+}
+
+// PXF cluster name MUST NOT be empty
+func pxfClusterNameNotEmpty(p PxfCluster) error {
+	if p.Name == "" {
+		return errors.New("the name of the PXF cluster cannot be empty string")
```
tatb-vmware, line 72: Since this code is all new, can we change `errors.New()` to be `fmt.Errorf()`?

tatb-vmware (approved):

cli/config/config_rules.go
djianwen-vmware: > Since this code is all new, can we change `errors.New()` to be `fmt.Errorf()`?

Good point. Fixed. Thanks!

cli/config/config_rules.go
```diff
@@ -0,0 +1,192 @@
...
+			found = true
+		}
+	}
+	return nil
+}
+
+// PXF cluster name MUST NOT be empty
+func pxfClusterNameNotEmpty(p PxfCluster) error {
+	if p.Name == "" {
+		return fmt.Errorf("the name of the PXF cluster cannot be empty string")
```
axue-broadcom, line 71:
```suggestion
		return fmt.Errorf("the name of the PXF cluster cannot be an empty string")
```

cli/config/config_rules.go
```diff
@@ -0,0 +1,192 @@
...
+		gplog.Warn("the PXF endpoint `%s` of the cluster `%s` will be ignored, "+
+			"because the cluster is marked as collocated", p.Endpoint, p.Name)
+	}
+	return nil
+}
+
+// An external cluster MUST have either a non-empty list of hosts or an endpoint or both
+func pxfClusterEndpointOrHostsMustExistWhenExternal(p PxfCluster) error {
+	if !p.Collocated && p.Endpoint == "" && len(p.Hosts) == 0 {
+		return fmt.Errorf("the host list and endpoint cannot be both empty for the external PXF cluster `%s`",
```
axue-broadcom, line 99:
```suggestion
		return fmt.Errorf("either the host list or endpoint must be non-empty for the external PXF cluster `%s`",
```

cli/config/config_rules.go
```diff
@@ -0,0 +1,192 @@
...
+	if len(p.Groups) == 0 {
+		return fmt.Errorf("there should be at least one service group in the PXF cluster `%s`", p.Name)
+	}
+	return nil
+}
+
+// PXF service group name MUST NOT be empty
+func pxfServiceGroupNameNotEmpty(p PxfServiceGroup) error {
+	if p.Name == "" {
+		return fmt.Errorf("the name of the service group cannot be empty string")
```
axue-broadcom, line 125:
```suggestion
		return fmt.Errorf("the name of the service group cannot be an empty string")
```

cli/config/config_rules.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+// PXF service group MUST have ports which are integers between 1-65535
+func pxfServiceGroupValidPorts(p PxfServiceGroup) error {
+	MinPortNumber := 1
+	MaxPortNumber := 65535
+	for _, port := range p.Ports {
+		if port < MinPortNumber || port > MaxPortNumber {
+			return fmt.Errorf("invalid port number `%d` under service group `%s`, "+
+				"a port number must be between `%d` - `%d`", port, p.Name, MinPortNumber, MaxPortNumber)
```
axue-broadcom, line 145:
```suggestion
				"the port number must be between `%d` - `%d`", port, p.Name, MinPortNumber, MaxPortNumber)
```

cli/config/config_rules.go
```diff
@@ -0,0 +1,192 @@
...
+
+// PXF service group SHOULD have ports which are integers between 1023 and 32768
+// (between privileged and ephemeral ranges)
+func pxfServiceGroupPortsWarning(p PxfServiceGroup) error {
+	MinPortNumber := 1024
+	MaxPortNumber := 32767
+	for _, port := range p.Ports {
+		if port < MinPortNumber || port > MaxPortNumber {
+			gplog.Warn("recommend using port numbers between `%d` - `%d`, "+
+				"rather than `%d` under the service group `%s`", MinPortNumber, MaxPortNumber, port, p.Name,
```
axue-broadcom, line 159:
```suggestion
			gplog.Warn("the port number provided under the service group `%s` (port: `%d`) is outside the recommended range `%d` - `%d`", p.Name, port, MinPortNumber, MaxPortNumber,
```

or something similar to this sounds a little better in my opinion

yimingli-vmware (approved):

cli/config/config_test.go
```diff
@@ -0,0 +1,378 @@
...
+		fakeStdout    *bytes.Buffer
+		fakeStderr    *bytes.Buffer
+		fakeLogFile   *bytes.Buffer
+	)
+	BeforeEach(func() {
+		fakeStdout = &bytes.Buffer{}
+		fakeStderr = &bytes.Buffer{}
+		fakeLogFile = &bytes.Buffer{}
+		gplog.SetLogger(gplog.NewLogger(fakeStdout, fakeStderr, fakeLogFile,
+			"test-log-file", gplog.LOGINFO, "test-program"))
```
yimingli-vmware, line 71: I finally understand how our current codebase verify the logging (gomega). 

**Initialize the logger:**
In the `config_suite_test.go` you need to call `testhelper.SetupTestEnvironment()`, it will also call `SetupTestLogger()` to initialize the logger. See an example in `cmd_suite_test.go`.

**Verify the log message:**
```go
// the Say function already provided the regex functionality
Expect(testStdout).Should(gbytes.Say(<expected_string>))
```

denalex (approved):

axue-broadcom (approved):

cli/config/config_test.go
djianwen-vmware: > I finally understand how our current codebase verify the logging (gomega).
> 
> **Initialize the logger:** In the `config_suite_test.go` you need to call `testhelper.SetupTestEnvironment()`, it will also call `SetupTestLogger()` to initialize the logger. See an example in `cmd_suite_test.go`.
> 
> **Verify the log message:**
> 
> ```go
> // the Say function already provided the regex functionality
> Expect(testStdout).Should(gbytes.Say(<expected_string>))
> ```

Thanks! That's very helpful. I've fixed it.

PR #1125, opened 2024-04-11
Tag the images used in release

djianwen-vmware:
- Add task `tag-release-testing-image` to job `promote-pxf-artifacts`. This task will tag all the images used in PXF testing jobs.
- Add task `tag-release-repackaging-image` to job `publish-pxf-artifacts-to-gp-releng`. This task will tag all the images used in re-package jobs.
- Fix the dependency relation between jobs. Now the testing images have `passed:` condition such that all jobs will use the same set of images from beginning to the end.

Pipeline: https://ci.ud.gpdb.pivotal.io/teams/main/pipelines/dev-djianwen-release-image-tag
(I forcibly set the env "prod" and PXF version such that the tagging job will run)


concourse/scripts/tag_release_images.bash
```diff
@@ -0,0 +1,39 @@
...
+else
+    echo "PXF version ${pxf_version} detected"
+    echo "Starting tagging release image..."
+fi
+
+echo "Authenticating with Google service account..."
+gcloud auth activate-service-account --key-file=<(echo "${GOOGLE_CREDENTIALS}") >/dev/null 2>&1
+
+# TODO: remove dev
+release_tag=dev-release-${pxf_version}
```
axue-broadcom, line 28:
```suggestion 
current_release_tag=release-${pxf_version}
branch_release_tag=release-${pxf_version%.*}.x
```

concourse/scripts/tag_release_images.bash
```diff
@@ -0,0 +1,39 @@
...
+echo "New tag for images: ${release_tag}"
+
+# Assume the images to tag have folder names ending with `-image`
+image_folders=$(ls -d ./*-image)
+for folder in $image_folders; do
+    echo "Found image folder: ${folder}"
+    image=$(cat "${folder}/repository")
+    digest=$(cat "${folder}/digest")
+    echo "Tagging image ${image} with digest: ${digest}"
+    gcloud container images add-tag "${image}@${digest}" "${image}:${release_tag}" --quiet
```
axue-broadcom, line 38:
```suggestion
    gcloud container images add-tag  --quiet "${image}@${digest}" "${image}:${current_release_tag}" 
    gcloud container images add-tag --quiet "${image}@${digest}" "${image}:${branch_release_tag}"
```

concourse/scripts/tag_release_images.bash
```diff
@@ -0,0 +1,39 @@
...
+done
```
axue-broadcom, line 39: new line

concourse/scripts/tag_release_images.bash
```diff
@@ -0,0 +1,36 @@
...
+
+set -e
+
+: "${GOOGLE_CREDENTIALS:?GOOGLE_CREDENTIALS must be set}"
+
+pxf_version=$(cat pxf_src/version)
+
+# Tagging should only happen during the release
+if [[ "${pxf_version}" =~ -SNAPSHOT || "${ENVIRONMENT}" != "prod" ]]; then
+		echo "SNAPSHOT version or non-production environment detected"
```
yimingli-vmware, line 11:
```suggestion
		echo "SNAPSHOT files or non-production environment detected"
```
We've already used `SNAPSHOT files` in the promote task, let's be consistent

concourse/scripts/tag_release_images.bash
```diff
@@ -0,0 +1,36 @@
...
+
+current_release_tag=release-${pxf_version}
+branch_release_tag=release-${pxf_version%.*}.x
+
+echo "New release tag for images: ${current_release_tag}"
+echo "New branch release tag for images: ${branch_release_tag}"
+
+# Assume the images to tag have folder names ending with `-image`
+image_folders=$(ls -d ./*-image)
+for folder in $image_folders; do
```
yimingli-vmware, line 29:
```suggestion
for folder in "${image_folders}"; do
```
let's be consistent

concourse/scripts/tag_release_images.bash
```diff
@@ -0,0 +1,36 @@
...
+
+: "${GOOGLE_CREDENTIALS:?GOOGLE_CREDENTIALS must be set}"
+
+pxf_version=$(cat pxf_src/version)
+
+# Tagging should only happen during the release
+if [[ "${pxf_version}" =~ -SNAPSHOT || "${ENVIRONMENT}" != "prod" ]]; then
+		echo "SNAPSHOT version or non-production environment detected"
+		echo "Skipping tagging release image..."
+		exit 0
```
denalex, line 13: nit: alignment is too much, need to be 4 chars (or a tab)

concourse/scripts/tag_release_images.bash
```diff
@@ -0,0 +1,36 @@
...
+
+# Assume the images to tag have folder names ending with `-image`
+image_folders=$(ls -d ./*-image)
+for folder in $image_folders; do
+    echo "Found image folder: ${folder}"
+    image=$(cat "${folder}/repository")
+    digest=$(cat "${folder}/digest")
+    echo "Tagging image ${image} with digest: ${digest}"
+    gcloud container images add-tag --quiet "${image}@${digest}" "${image}:${current_release_tag}"
+    gcloud container images add-tag --quiet "${image}@${digest}" "${image}:${branch_release_tag}"
```
denalex, line 35: do we need to remove the branch tag from the previous image that held the tag before this release ?

djianwen-vmware: > do we need to remove the branch tag from the previous image that held the tag before this release ?

I think when you re-use a tag for a new image, the old image will lose the tag. From my understanding one tag can only point to one image.

denalex: I am not sure, but I think in @ashuka24 PR we are removing the tag from an old image before applying it to the new one

axue-broadcom: Technically speaking, when you re-use a tag, it should be removed from the old image. As a precautionary, we do have a precedent to explicitly remove the tag before tagging with the same name (see `check_docker_images_and_tag.bash` though the script had an error and currently does not do that. 

My PR uses the precedent that was set even though it is not required.

concourse/scripts/tag_release_images.bash
djianwen-vmware: > let's be consistent

My linter (Shellcheck) doesn't like it:
```
Since you double quoted this, it will not word split, and the loop will only run once.
See SC2066.
```
Then I tried testing locally, it didn't work. 
Bash code:
```
folders=$(ls -d ./concourse/*)
j=1
for folder in "$folders"; do
  echo "execution $j: $folder"
  ((j++))
done

```
Result:
```
% bash ./test.sh
execution 1: ./concourse/Makefile
./concourse/README.md
./concourse/ansible
./concourse/docker
./concourse/pipelines
./concourse/scripts
./concourse/settings
./concourse/tasks
./concourse/template_tool
./concourse/terraform
```
Very good point though. I learned something new.

concourse/scripts/tag_release_images.bash
djianwen-vmware: Good catch. Fixed. Thanks!

concourse/scripts/tag_release_images.bash
djianwen-vmware: I searched online and it doesn't require a removal from old image. The tag is like an identifier of an image that is assigned by the user. It is impossible to have a tag pointing to multiple images. Though one image can have multiple tags.

concourse/scripts/tag_release_images.bash
djianwen-vmware: > We've already used `SNAPSHOT files` in the promote task, let's be consistent

Good point. Fixed, thanks!

concourse/scripts/tag_release_images.bash
axue-broadcom: You should be able to do an array. For example, `"${image_folders[@]}"` should still allow word splitting and going through the array

concourse/scripts/tag_release_images.bash
axue-broadcom: That is correct. However, we have a precedent in this code base to explicitly remove tags before adding then to a different image. I would prefer us to be consistent across all our scripts.

concourse/scripts/tag_release_images.bash
yimingli-vmware: Interesting. I think that's because we are using `-d`. From `man ls`:
```
NAME
     ls – list directory contents
...
...
-d      Directories are listed as plain files (not searched recursively).
```
When we add the quotes, bash will return all the directories as one entry, but here we want them to be separate, so we should not add quotes here.

yimingli-vmware (approved):

concourse/scripts/tag_release_images.bash
djianwen-vmware: > That is correct. However, we have a precedent in this code base to explicitly remove tags before adding then to a different image. I would prefer us to be consistent across all our scripts.

I agree that we should be consistent. 
My concern about the precedent way is that, "remove old"+"add new" is a two-step operation, while in this PR it is more "atomic". 
In a scenario where the image is missing in registry, or image name is wrong for whatever reason, the first operation will succeed while the second will fail. Then we will lose the old branch tag and new branch tag is never added, which may break the other pipelines(I'm not sure). It's not a disaster though but I personally still think it's not necessary to introduce more failure cases.
If we really want to add it, I can do it for sure.

concourse/scripts/tag_release_images.bash
djianwen-vmware: I changed it to use an array instead.

tatb-vmware (approved):

concourse/scripts/tag_release_images.bash
djianwen-vmware: @axue-broadcom I've removed `untag` from file `concourse/scripts/check_docker_images_and_tag.bash` in this PR.

concourse/pipelines/templates/jobs/promote-pxf-artifacts-tpl.yml
```diff
@@ -41,6 +49,20 @@
       platform: linux
       run:
         path: pxf_src/concourse/scripts/promote_pxf_artifacts.bash
+  - task: tag-release-testing-image
```
axue-broadcom, line 52:
```suggestion
  - task: tag-release-testing-images
```

concourse/pipelines/templates/jobs/publish-pxf-artifacts-tpl.yml
```diff
@@ -58,6 +58,20 @@
       GP_VER: [[x.gp_ver]]
     {% endif %}
 {% endcall %}
+  - task: tag-release-repackaging-image
```
axue-broadcom, line 61:
```suggestion
  - task: tag-release-repackaging-images
```

concourse/scripts/tag_release_images.bash
```diff
@@ -0,0 +1,36 @@
...
+echo "New branch release tag for images: ${branch_release_tag}"
+
+# Assume the images to tag have folder names ending with `-image`
+image_folders=(./*-image)
+for folder in "${image_folders[@]}"; do
+    echo "Found image folder: ${folder}"
+    image=$(cat "${folder}/repository")
+    digest=$(cat "${folder}/digest")
+    echo "Tagging image ${image} with digest: ${digest}"
+    gcloud container images add-tag --quiet "${image}@${digest}" "${image}:${current_release_tag}"
```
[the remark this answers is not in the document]
djianwen-vmware: > let's be consistent for both of these. is there a difference between using `@` vs `:`

Fixed. Thanks!

axue-broadcom (approved):

concourse/scripts/tag_release_images.bash
axue-broadcom: Sounds good. That portion of code was not being called due to a mistype in the logic. I will remove it from the create-release-script.bash PR as well.

PR #1129, opened 2024-04-16
Use GCS to replace pivnet for GPDB artifacts

djianwen-vmware:
- Remove inputs and outputs from the pipeline because we will use `gsutil cp` to directly copy RPMs from release bucket to artifact bucket.
- Removed `Pivnet CLI` related references from pipeline
- Created a new bucket to replace the pivnet artifact bucket 
- Re-write the script to use gsutil with google service account. This simplifies the script a lot because we don't need to organize RPMs in multiple folders any more.

New pipeline:
https://ci.ud.gpdb.pivotal.io/teams/main/pipelines/dev-gpdb_gcs_artifacts-djianwen
Dev release pipeline:
https://ci.ud.gpdb.pivotal.io/teams/main/pipelines/dev-djianwen-use-gcs-for-artifacts


concourse/Makefile
```diff
@@ -15,7 +15,7 @@ CERTIFICATION_PIPELINE_NAME ?= pxf-certification
 CLOUDBUILD_PIPELINE_NAME    ?= cloudbuild
 DEV_CLOUDBUILD_PIPELINE_NAME ?= dev-cloudbuild-$(USER)-$(BRANCH)
 PERF_PIPELINE_NAME          ?= pxf_perf-$(SCALE)g-RHEL$(REDHAT_MAJOR_VERSION)
-PIVNET_PIPELINE_NAME        ?= pivnet_artifacts
+GCS_PIPELINE_NAME           ?= gpdb_gcs_artifacts
```
axue-broadcom, line 18: since we're changing the name anyway, let's switch to dashs 
```suggestion
GCS_PIPELINE_NAME           ?= gpdb-gcs-artifacts
```

concourse/pipelines/templates/get_gpdb_gcs_artifacts-tpl.yml
```diff
@@ -24,12 +23,9 @@ resources:
 
 {% call(x) macros.for_each_config(build_test_pxf_combinations) %}
     {% do x.update(gp_num_versions=gp_num_versions[x.gp_ver]) %}
-    {% include 'resources/greenplum-package-tpl.yml' %}
```
axue-broadcom, line 27: since this line was removed, we can completely remove the entire call block as this is now a no op

concourse/pipelines/templates/jobs/get_greenplum_product_files.yml
```diff
@@ -1,70 +1,36 @@
-{# Template for downloading releases from PivNet
+{# Template for copying GPDB packages from the GCS release bucket
```
axue-broadcom, line 1:
```suggestion
{# Template for copying GPDB release packages from the prod GCS release bucket
```

concourse/scripts/get_product_files.bash
```diff
@@ -2,105 +2,54 @@
...
-		"--release-version=${gpdb_version}" \
-		"--product-file-id=${id}" >/dev/null 2>&1 &
-	pids+=($!)
-done
+while IFS= read -r package_path; do
+  package_name=$(basename "${package_path}")
+  target_package_path="${target_folder}/${package_name}"
 
-wait "${pids[@]}"
+  echo "[Action] copying: ${package_path} to ${target_package_path}"
```
axue-broadcom, line 51:
```suggestion
  echo "Copying GPDB release package: ${package_path} to ${target_package_path}..."
```

concourse/pipelines/templates/jobs/get_greenplum_product_files.yml
```diff
@@ -1,70 +1,36 @@
...
  #
  # Expects a dictionary with the name 'x' to be set as defined in macros.j2 file (for_each_gp_ver)
  # Required attributes:
  #   - attempts        : number of times to attempt downloading product files before failing
  #   - gp_ver          : Greenplum major version
  #   - gp_num_versions : number of previous release of Greenplum (minor or mainteance) for this Greenplum major version to download
- #   - platforms       : list of Greenplum platforms to download for this Greenplum major version
```
axue-broadcom, line 8: Previously, we were passing this list of platforms to interpolate for the resources needed for the job. However, I see that we now hardcode the platforms in the regex string within the script itself. 

is it possible to continue using this list of platforms to dynamically generate the regex string instead of hard coding it in the script itself?

concourse/pipelines/longevity_pipeline.yml
```diff
@@ -85,7 +85,7 @@ resources:
   type: gcs
   icon: google-drive
   source:
-    bucket: ((ud/pxf/common/pivnet-artifacts-bucket-name))
+    bucket: ((ud/pxf/common/gpdb-artifacts-bucket-name))
```
axue-broadcom, line 88: This pipeline, as well as the `external-table-tpl`, `perf-pipeline-tpl`, `pr_pipeline-tpl` are no longer used. Previous refactors of the pipelines have not touched these files and they are very likely extremely out of date. As such, I'm not sure that we need to include the changes in these files as well. There may be work lined up in the future to completely remove these pipelines from the repo.

concourse/Makefile
denalex, line 18:
```suggestion
GPDB_ARTIFACTS_PIPELINE_NAME           ?= gpdb-artifacts
```
GCS is too generic of a name for a pipeline that is supposed to fetch GPDB artifacts. Also if we remove GCS from the name, we can change implementation without changing the name / build history in the future

concourse/Makefile
```diff
@@ -240,28 +240,25 @@ set-cloudbuild-pipeline-base:
 		--pipeline $(PIPELINE_NAME) \
 		${FLY_OPTION_NON-INTERACTIVE} || echo "pipelines/cloudbuild_pipeline.yml has errors"
 
-.PHONY: pivnet
-pivnet: set-pivnet-pipeline
-
-.PHONY: set-pivnet-pipeline
-set-pivnet-pipeline: generate-pivnet-pipeline
+.PHONY: set-gcs-pipeline
```
denalex, line 243:
```suggestion
.PHONY: set-gpdb-artifacts-pipeline
```

concourse/Makefile
```diff
@@ -240,28 +240,25 @@ set-cloudbuild-pipeline-base:
...
+set-gcs-pipeline: generate-gcs-pipeline
```
denalex, line 244: this new line should probably go in the common section of short goals on line 64.
```
gpdb-artifacts: set-gpdb-artifacts-pipeline
```

```suggestion
set-gpdb-artifacts-pipeline: generate-gpdb-artifacts-pipeline
```

concourse/Makefile
```diff
@@ -240,28 +240,25 @@ set-cloudbuild-pipeline-base:
...
 	$(FLY_CMD) --target=$(CONCOURSE) \
 		$(SET_PIPELINE) \
-		--pipeline=${PIVNET_PIPELINE_NAME} \
-		--config ./generated/get_pivnet_artifacts.yml \
+		--pipeline=${GCS_PIPELINE_NAME} \
+		--config ./generated/get_gpdb_gcs_artifacts.yml \
```
denalex, line 248:
```suggestion
		--config ./generated/get_gpdb_artifacts.yml \
```

concourse/pipelines/longevity_pipeline.yml
denalex: this pipeline will likely be resurrected in some form for the Kerberos token refresh story

concourse/pipelines/templates/get_gpdb_gcs_artifacts-tpl.yml
```diff
@@ -7,8 +7,7 @@ resource_types:
 resources:
     [[macros.resource_timer('twice-a-day', '12h')]]
     [[macros.resource_github_repo('pxf')]]
-    [[macros.resource_registry_image('ubuntu18', 'gcr.io/data-gpdb-public-images/gpdb6-ubuntu18.04-build', 'latest')]]
-    [[macros.resource_pivnet_cli()]]
+    [[macros.resource_registry_image('google-cloud-sdk-slim-image', 'google/cloud-sdk', '402.0.0-slim')]]
```
denalex, line 10: where did ubuntu18 go ?

denalex: never mind, it seems it is removed from this file only and the new slim image is used instead

concourse/pipelines/templates/jobs/get_greenplum_product_files.yml
djianwen-vmware: In the past , it is also hardcoded in the script itself(old file line 84). This `platforms` is used for pipeline construction and never passed into the script.
I can try and see if it works.

concourse/pipelines/longevity_pipeline.yml
djianwen-vmware: Do you suggest to keep the `pivnet` references in the repo for a while? I'm deleting them because pivnet references should be gone now.

concourse/Makefile
djianwen-vmware: > GCS is too generic of a name for a pipeline that is supposed to fetch GPDB artifacts. Also if we remove GCS from the name, we can change implementation without changing the name / build history in the future

Good suggestion. I removed GCS from the name.

concourse/pipelines/templates/jobs/get_greenplum_product_files.yml
djianwen-vmware: I passed in the variable as a string where each platform is separated by a space. Now it's working. Thanks for the advice.

concourse/scripts/get_product_files.bash
```diff
@@ -2,105 +2,63 @@
...
-		"--release-version=${gpdb_version}" \
-		"--product-file-id=${id}" >/dev/null 2>&1 &
-	pids+=($!)
-done
+while IFS= read -r package_path; do
+  package_name=$(basename "${package_path}")
+  target_package_path="${target_folder}/${package_name}"
 
-wait "${pids[@]}"
+  echo "Copying GPDB release package: ${package_path} to ${target_package_path}..."
```
yimingli-vmware, line 60:
```suggestion
  echo "Copying GPDB release package from ${package_path} to ${target_package_path}..."
```

tatb-vmware (approved):

concourse/pipelines/longevity_pipeline.yml
denalex: no, I was just commenting that the pipeline is likely to be used soon. It might be already broken, but let's not make it even more broken. So we should apply the changes needed here to migrate off pivnet.

concourse/Makefile
```diff
@@ -61,6 +61,7 @@ pr: set-pr-build-pipeline
 cloudbuild: set-cloudbuild-pipeline
 dev-cloudbuild: set-dev-cloudbuild-pipeline
 longevity: set-longevity-pipeline
+gpdb-artifacts: set-gpdb-artifacts-pipeline
```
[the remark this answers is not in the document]
djianwen-vmware: > please add `gpdb-artifacts` goal to the list of PHONY on line 54

Fixed. Thanks!

denalex (approved):

yimingli-vmware (approved):

concourse/pipelines/templates/jobs/get_greenplum_product_files.yml
```diff
@@ -29,6 +30,8 @@
           SOURCE_BUCKET: ((ud/pxf/common/gpdb-concourse-resources-prod-bucket-name))
           TARGET_BUCKET: ((ud/pxf/common/gpdb-artifacts-bucket-name))
           GPDB_VERSION: [[x.gp_ver]]
+          PLATFORMS: [[x.platforms | map(attribute='greenplum_platform') | unique | join(' ')]]
```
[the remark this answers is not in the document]
djianwen-vmware: > is there a reason you join with a space here, but then immediately change the space to a pipe in the script where we use it?

I think using `|` can simplify the script. 
However I found an existing example which is using spaces to pass a list: https://github.com/greenplum-db/pxf/blob/main/concourse/pipelines/cloudbuild_pipeline.yml#L87
This was why I followed this convention in the pipeline. If this is not a convention or it doesn't matter, I can change mine to use pipes.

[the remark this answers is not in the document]
djianwen-vmware: I've updated the script/pipeline to use `|`.

concourse/scripts/get_product_files.bash
```diff
@@ -2,105 +2,60 @@
...
-	# if the version doesn't exist in either i-1, i-th bucket or the SHAs don't match, download it from pivnet.
-	id=$(jq <<<"${product_files_json}" -r --arg object_key "${file}" '.[] | select(.aws_object_key == $object_key).id')
+# get the `VERSIONS_BEFORE_LATEST`-th (zero-based index) latest version
+target_version=$(echo "${versions}" | awk -v i=${VERSIONS_BEFORE_LATEST} 'NR == i+1')
+echo "The ${VERSIONS_BEFORE_LATEST}-th latest version is: ${target_version}"
 
-	if [[ -z "${id}" ]]; then
-		echo "Did not find '${file}' in product files for GPDB '${gpdb_version}'"
+# escape the dots in the version number for regex
+regex_version="${target_version//./\\.}"
```
axue-broadcom, line 47: here we have `regex_version` but on line 23 we use the syntax `package_regex` and below on line 50 we use `versioned_package_regex`, can we make this `version_regex` here for consistency?

axue-broadcom (approved):
In the commit message, can you also include the pipeline rename from pivnet_artifacts to gpdb-artifacts. Maybe something like:

```
This commit does the following:
- Renames the `pivnet_artifacts` pipeline to `gpdb-artifacts`. 
- Removes `Pivnet CLI` related references from pipeline.
- Re-writes the script to use `gsutil` with google service account. This simplifies the script a lot because we don't need to organize RPMs in multiple folders any more.
- Removes inputs and outputs from the pipeline because we will use `gsutil cp` to directly copy RPMs from release bucket to artifact bucket.

As part of this work, we also
- Created a new bucket to replace the pivnet artifact bucket in GCP
- Added a `gpdb-artifacts` target to the concourse Makefile

PR #1134, opened 2024-04-18
Create PxfCluster and connect it to ClusterData

tatb-vmware:
This PR does the following 3 things: 

1. It adds library functions to parse the configuration file into the `config.PxfCluster` struct. 
2. Introduces the fields `GreenplumCluster` and `PxfCluster` to the `cluster.ClusterData` struct. 
3. Adds functionality to  `cluster.GetClusterData()` to properly instantiate both the `GreenplumCluster` and the `PxfCluster`. In the case of a collocated PXF cluster, `PxfCluster` is equal to the `GreenplumCluster`. Otherwise, we create the `PxfCluster` based off the `config.PxfCluster` struct.


cli/cmd/cluster.go
```diff
@@ -138,18 +145,71 @@ func GenerateOutput(cmd *command, clusterData *ClusterData) error {
...
 	return clusterData, nil
 }
 
+func convertPxfClusterInfo(pxfCluster *config.PxfCluster) *cluster.Cluster {
+	fakeSegConfigList := make([]cluster.SegConfig, len(pxfCluster.Hosts))
+	for i, pxfHost := range pxfCluster.Hosts {
+		fakeSegConfigList[i] = cluster.SegConfig{
+			DbID:          i + 1,
+			ContentID:     i,
+			Role:          "p",
```
djianwen-vmware, line 180: Are these fields just placeholders for PXF cluster struct, because we use the same struct for both GPDB and PXF?

cli/cmd/cluster.go
```diff
@@ -138,18 +145,71 @@ func GenerateOutput(cmd *command, clusterData *ClusterData) error {
...
+	if baseDirName == "" {
+		baseDirName = os.Getenv("PXF_HOME")
+		usedEnvVar = "PXF_HOME"
+	}
+	if baseDirName == "" {
+		return "", fmt.Errorf("please set PXF_BASE or PXF_HOME")
+	}
+	if fileInfo, err := os.Stat(baseDirName); os.IsNotExist(err) || !fileInfo.IsDir() {
+		return "", fmt.Errorf("%s is not a valid directory: %s\n", usedEnvVar, baseDirName)
+	}
```
djianwen-vmware, line 208:
```suggestion
	} else if err != nil {
	   // return other errors when accessing the file, such as IO error and permission error
	}
```

cli/cmd/cluster_test.go
```diff
@@ -599,15 +601,48 @@ var _ = Describe("GetClusterData()", func() {
...
+			_ = os.Unsetenv("PXF_BASE")
+			_ = os.Setenv("PXF_HOME", "foobar")
+			fakeSegConfigs = sqlmock.NewRows(header).AddRow(coordinator...).AddRow(localSegOne...).AddRow(localSegTwo...)
+			mock.ExpectQuery("SELECT (.*)").WillReturnRows(fakeSegConfigs)
+
+			clusterData, err := cmd.GetClusterData()
+			Expect(clusterData).To(BeNil())
+			Expect(err.Error()).To(Equal("PXF_HOME is not a valid directory: foobar\n"))
+		})
+
```
djianwen-vmware, line 643: Consider another test case where the folder exists but no permission to access

cli/cmd/cluster_test.go
```diff
@@ -662,7 +698,7 @@ var _ = Describe("GetClusterData()", func() {
 		})
 
 		It("returns an error indicating that the command is not running on the coordinator", func() {
-			fakeSegConfigs := sqlmock.NewRows(header).AddRow(coordinator...).AddRow(localSegOne...).AddRow(localSegTwo...)
+			fakeSegConfigs = sqlmock.NewRows(header).AddRow(coordinator...).AddRow(localSegOne...).AddRow(localSegTwo...)
 			mock.ExpectQuery("SELECT (.*)").WillReturnRows(fakeSegConfigs)
```
djianwen-vmware, line 702: I notice these two lines appeared in several test cases. Maybe move to `BeforeEach`?

cli/cmd/pxf.go
```diff
@@ -52,6 +53,7 @@ func (cmd *command) GetFunctionToExecute() (func(string) string, error) {
 	if err != nil {
 		return nil, err
 	}
+	cmd.envVarMap = inputs
```
djianwen-vmware, line 56: Is there a reason for this new field?

cli/config/config.go
```diff
@@ -93,3 +101,168 @@ func (p PxfServiceGroup) validate() error {
...
 	return pxfHostValidHostname(p)
 }
+
+var ReadFile = readFile
+
+func ParsePxfCluster(pxfBase string, clusterName string) (*PxfCluster, error) {
+	pxfDeployment := PxfDeployment{Clusters: make(map[string]*PxfCluster)}
+	clusterNames := []string{"default"}
+
+	fileContents, err := ReadFile(fmt.Sprintf("%s/clusters/%s/conf/cluster.txt", pxfBase, clusterName))
```
djianwen-vmware, line 111: Consider using `filepath.Join()`, for example, `filePath := filepath.Join(pxfBase, "clusters", clusterName, "conf", "cluster.txt")`

cli/config/config.go
```diff
@@ -93,3 +101,168 @@ func (p PxfServiceGroup) validate() error {
 func (p PxfHost) validate() error {
 	return pxfHostValidHostname(p)
 }
+
+var ReadFile = readFile
+
+func ParsePxfCluster(pxfBase string, clusterName string) (*PxfCluster, error) {
+	pxfDeployment := PxfDeployment{Clusters: make(map[string]*PxfCluster)}
+	clusterNames := []string{"default"}
```
djianwen-vmware, line 109: This means we will have another function to get list of cluster names from config file in the future? So it's a two rounds of config file parsing?

cli/config/config.go
```diff
@@ -93,3 +101,168 @@ func (p PxfServiceGroup) validate() error {
...
+			return nil, err
+		}
+		pxfDeployment.Clusters[currentClusterName] = pxfCluster
+	}
+
+	if err := pxfDeployment.Validate(); err != nil {
+		return nil, err
+	}
+
+	return pxfDeployment.Clusters[clusterName], nil
```
djianwen-vmware, line 134: I guess is that in the future we will have a function `ParsePxfDeployment()`, then this function is not necessary any more. Just call the map element access from the PXF deployment struct.

cli/config/config.go
```diff
@@ -93,3 +101,168 @@ func (p PxfServiceGroup) validate() error {
...
+
+	for lineNum, line := range fileContents {
+		line = strings.ToLower(line)
+		line = strings.TrimSpace(line)
+		if line == "" || line[0] == '#' { // Remove any line that is a TOML comment
+			continue
+		}
+		lineKeyVal := strings.Split(line, "=")
+		if len(lineKeyVal) != 2 { // lineKeyVal[0] = key, lineKeyVal[1] = value, anything else is invalid
+			return nil, fmt.Errorf("Line must have only one '=' on line %d: %s\n", lineNum+1, line)
```
djianwen-vmware, line 157: In this repo our error messages all start with lower cases. Capitalized error message will cause IDE linter to give warning. Here is what I found relevant: https://go.dev/wiki/CodeReviewComments#error-strings

cli/config/config.go
```diff
@@ -93,3 +101,168 @@ func (p PxfServiceGroup) validate() error {
...
+		return fmt.Errorf("invalid key: %s\n", key)
+	}
+	switch tokens[1] {
+	case "name":
+		err := assertNotEmpty(val, "cluster name")
+		if err != nil {
+			return err
+		}
+		if val != clusterName {
+			return fmt.Errorf("cluster name does not equal directory name of cluster")
```
djianwen-vmware, line 205:
```suggestion
			return fmt.Errorf("cluster name does not match directory name of cluster")
```

cli/cmd/cluster.go
denalex: After another review of `pxf.sh` which currently calls this script, it seems that PXF_BASE will always be setup, so we can only check for it and not for PXF_HOME, and not do the fallback. Also, please use existing function `validateEnvVar` for validation.

cli/cmd/cluster.go
```diff
@@ -14,10 +15,16 @@ import (
...
-	Output     *cluster.RemoteOutput
-	NumHosts   int
-	connection *dbconn.DBConn
+	GreenplumCluster       *cluster.Cluster
+	PxfCluster             *cluster.Cluster
+	GreenplumClusterOutput *cluster.RemoteOutput
+	PxfClusterOutput       *cluster.RemoteOutput
+	NumGreenplumHosts      int
+	NumPxfHosts            int
+	// TODO: Remove the following three fields after refactoring
```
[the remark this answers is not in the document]
denalex: each command will need to get updated to use the right cluster, there will be a lot for further changes. Once all of them are updated, these will be removed, they are left right now only for the purposes of not breaking the compilation.

cli/cmd/cluster.go
```diff
@@ -138,18 +145,71 @@ func GenerateOutput(cmd *command, clusterData *ClusterData) error {
...
 // and makes sure this script is running on the Greenplum coordinator host
 func GetClusterData() (*ClusterData, error) {
-	clusterData, err := retrieveClusterData()
+	clusterData := &ClusterData{}
+
+	segConfigs, err := retrieveGreenplumClusterInfo()
 	if err != nil {
 		return nil, err
 	}
+	clusterData.Cluster = cluster.NewCluster(segConfigs)
```
denalex, line 154:
```suggestion
	clusterData.GreenplumCluster = cluster.NewCluster(segConfigs)
	// TODO: remove the line below
	clusterData.Cluster = cluster.NewCluster(segConfigs)
```

cli/cmd/cluster.go
```diff
@@ -138,18 +145,71 @@ func GenerateOutput(cmd *command, clusterData *ClusterData) error {
...
 		return nil, err
 	}
 
+	pxfCluster, err := retrievePxfClusterInfo("default")
+	if err != nil {
+		return nil, err
+	}
+
+	if pxfCluster.Collocated {
+		clusterData.PxfCluster = clusterData.Cluster
```
denalex, line 166:
```suggestion
		clusterData.PxfCluster = clusterData.GreenplumCluster
```

cli/cmd/cluster.go
```diff
@@ -138,18 +145,71 @@ func GenerateOutput(cmd *command, clusterData *ClusterData) error {
...
+			Status:        "u",
+			Port:          0,
+			Hostname:      pxfHost.Hostname,
+			Address:       pxfHost.Hostname,
+			DataDir:       "",
+		}
+	}
+	newCluster := cluster.NewCluster(fakeSegConfigList)
+
+	return newCluster
```
denalex, line 192:
```suggestion
	return cluster.NewCluster(fakeSegConfigList)
```

cli/cmd/cluster_test.go
```diff
@@ -583,9 +583,11 @@ var _ = Describe("GetClusterData()", func() {
 		connection            *dbconn.DBConn
 		mock                  sqlmock.Sqlmock
 		oldConnectionProvider func() (*dbconn.DBConn, error)
+		fakeSegConfigs        *sqlmock.Rows
```
denalex, line 586:
```suggestion
		mockSegConfigs        *sqlmock.Rows
```

cli/cmd/cluster_test.go
```diff
@@ -599,15 +601,48 @@ var _ = Describe("GetClusterData()", func() {
...
 			_ = os.MkdirAll(coordinatorDataDir, 0777)
 		})
 
 		AfterEach(func() {
 			_ = os.RemoveAll(coordinatorDataDir)
 		})
 
+		It("returns an error when failing to get PXF_BASE or PXF_HOME", func() {
+			fakeSegConfigs = sqlmock.NewRows(header).AddRow(coordinator...).AddRow(localSegOne...).AddRow(localSegTwo...)
+			mock.ExpectQuery("SELECT (.*)").WillReturnRows(fakeSegConfigs)
```
denalex, line 614: lets rename this mock since we have other mocks as well
```suggestion
			mockConnection.ExpectQuery("SELECT (.*)").WillReturnRows(fakeSegConfigs)
```

cli/cmd/cluster_test.go
denalex: this is a bit tricky, for `pxf cluster prepare` command only, we do not even expect the directory to exist

cli/cmd/cluster_test.go
```diff
@@ -599,15 +601,48 @@ var _ = Describe("GetClusterData()", func() {
...
+
+			clusterData, err := cmd.GetClusterData()
+			Expect(clusterData).To(BeNil())
+			Expect(err.Error()).To(Equal("PXF_HOME is not a valid directory: foobar\n"))
+		})
+
 		It("returns the cluster data with no error", func() {
-			fakeSegConfigs := sqlmock.NewRows(header).AddRow(coordinator...).AddRow(localSegOne...).AddRow(localSegTwo...)
+			fakeSegConfigs = sqlmock.NewRows(header).AddRow(coordinator...).AddRow(localSegOne...).AddRow(localSegTwo...)
```
denalex, line 645: this seems to be repeating a lot, can it either be placed into `BeforeEach` or into a private helper function ?

cli/config/config.go
denalex: there will be logic in the future to get the list of cluster names, but that logic will get the list form the directory layout of `PXF_BASE/clusters/`, and then for every directory inside it will call this function.

cli/config/config.go
denalex: we envisaged that `parseDeployment()` function will iterate over directories and for each directory it will call this function to parse a cluster (since cluster.txt is per-cluster only), then it will wire everything up and validate the deployment before returning the single cluster info back.

cli/config/config.go
```diff
@@ -93,3 +101,168 @@ func (p PxfServiceGroup) validate() error {
...
+		}
+		boolVal, err := strconv.ParseBool(val)
+		if err != nil {
+			return fmt.Errorf("cluster collocated must be a boolean (true/false): %s\n", val)
+		}
+		cluster.Collocated = boolVal
+	case "endpoint":
+		cluster.Endpoint = val
+	default:
+		gplog.Warn("Ignoring %s because it is not a valid key for cluster\n", key)
```
denalex, line 221:
```suggestion
		gplog.Warn("Ignoring %s entry because it is not a valid property for cluster definition\n", key)
```

cli/config/config.go
```diff
@@ -93,3 +101,168 @@ func (p PxfServiceGroup) validate() error {
...
+	}
+
+	return nil
+}
+
+func fillServiceGroupValue(groups map[string]*PxfServiceGroup, tokens []string, key, val string) error {
```
denalex, line 227: to make it more "object-oriented"
```suggestion
func setGroupProperty (cluster *PxfCluster) (tokens []string, key, val string) error {
```

cli/config/config_test.go
```diff
@@ -367,3 +367,189 @@ var _ = Describe("PXF config validation rules", func() {
...
+	BeforeEach(func() {
+		fileContents = []string{}
+
+		config.ReadFile = func(string) ([]string, error) {
+			return fileContents, nil
+		}
+	})
+	It("returns an error when failing to read file", func() {
+		config.ReadFile = func(string) ([]string, error) {
+			return fileContents, errors.New("failed to read file")
```
denalex, line 384: should we be using `fmt.Errorf(..)` ?

cli/config/config_test.go
```diff
@@ -367,3 +367,189 @@ var _ = Describe("PXF config validation rules", func() {
...
+		Expect(pxfCluster).To(BeNil())
+		Expect(err).To(Equal(errors.New("failed to read file")))
+	})
+	Context("parseFile errors", func() {
+		It("returns an error when parsing a line with more than one = sign", func() {
+			fileContents = []string{"cluster.name = foo = bar"}
+
+			pxfCluster, err := config.ParsePxfCluster("/test/base/path/", "default")
+			Expect(pxfCluster).To(BeNil())
+			Expect(err).To(Equal(errors.New("Line must have only one '=' on line 1: cluster.name = foo = bar\n")))
```
denalex, line 397:
```suggestion
			Expect(err).To(Equal(errors.New("There should be only one character '=' on line 1: cluster.name = foo = bar\n")))
```

cli/config/config_test.go
```diff
@@ -367,3 +367,189 @@ var _ = Describe("PXF config validation rules", func() {
...
+			pxfCluster, err := config.ParsePxfCluster("/test/base/path/", "default")
+			Expect(pxfCluster).To(BeNil())
+			Expect(err).To(Equal(errors.New("invalid key: cluster.invalid.key\n")))
+		})
+		It("returns an error when cluster.name is empty", func() {
+			fileContents = []string{"cluster.name="}
+
+			pxfCluster, err := config.ParsePxfCluster("/test/base/path/", "default")
+			Expect(pxfCluster).To(BeNil())
+			Expect(err).To(Equal(errors.New("cluster name cannot be empty")))
```
denalex, line 430:
```suggestion
			Expect(err).To(Equal(errors.New("'cluster.name' property cannot be empty")))
```

cli/config/config_test.go
```diff
@@ -367,3 +367,189 @@ var _ = Describe("PXF config validation rules", func() {
...
+		})
+		It("returns an error when cluster.name is different from directory cluster name", func() {
+			fileContents = []string{"cluster.name="}
+
+			pxfCluster, err := config.ParsePxfCluster("/test/base/path/", "default")
+			Expect(pxfCluster).To(BeNil())
+			Expect(err).To(Equal(errors.New("cluster name cannot be empty")))
```
denalex, line 437: what is the general approach to an error message ? Choices:
1. State the rule (and implicitly assume it is violated): "File should exist" (assumes that we looked for file and it did not exist)
2. State the fact observed (and implicitly assume the rule is the opposite of the stated fact): "File not found" (assumes that the rule is that the file should exists)

cli/config/config_test.go
```diff
@@ -367,3 +367,189 @@ var _ = Describe("PXF config validation rules", func() {
...
+			Expect(err).To(Equal(errors.New("cluster collocated must be a boolean (true/false): not-a-boolean\n")))
+		})
+	})
+	Context("ParseCluster errors", func() {
+		It("returns an error when getting an invalid key for group", func() {
+			fileContents = []string{"group.foo.invalid.key=bar"}
+
+			pxfCluster, err := config.ParsePxfCluster("/test/base/path/", "default")
+			Expect(pxfCluster).To(BeNil())
+			Expect(err).To(Equal(errors.New("Invalid key: group.foo.invalid.key\n")))
```
denalex, line 460: we need consistency on:
- naming "key" vs. "property"
- enclosing keys and values in separators - single or double quote
- adding . between tokens of a property name

cli/cmd/cluster.go
tatb-vmware: After discussion, we have decided to only check if it exists. If it does exists, check if that directory is writable. If it doesn't exist, it's fine.

cli/cmd/cluster_test.go
```diff
@@ -606,9 +612,36 @@ var _ = Describe("GetClusterData()", func() {
 			_ = os.RemoveAll(coordinatorDataDir)
 		})
 
+		It("returns an error when failing to get PXF_BASE or PXF_HOME", func() {
+			mockConnection.ExpectQuery("SELECT (.*)").WillReturnRows(mockSegConfigs)
+			_ = os.Unsetenv("PXF_BASE")
+
+			clusterData, err := cmd.GetClusterData()
+			Expect(clusterData).To(BeNil())
+			Expect(err).To(MatchError("please set PXF_BASE or PXF_HOME"))
```
axue-broadcom, line 621: in this error situation, there is no \n at the end of the string, however there is if it is incorretly set (see line 630 and 641) can we be consistent?

cli/config/config.go
```diff
@@ -93,3 +102,157 @@ func (p PxfServiceGroup) validate() error {
...
+
+func parseFile(fileContents []string) (map[string]string, error) {
+	parsedFileKeyVals := make(map[string]string)
+	for lineNum, line := range fileContents {
+		line = strings.ToLower(line)
+		line = strings.TrimSpace(line)
+		if line == "" || line[0] == '#' { // Remove any line that is a TOML comment
+			continue
+		}
+		lineKeyVal := strings.Split(line, "=")
```
axue-broadcom, line 149: Do we want to check if the `=` is escaped or inside quotes?

Is it possible have a situation where we can set some cluster level gucs where we might something like 
`pxf.cluster.options='date_wide_range=true'`

In a situation like this, we do not want to error out but we would want to parse this as 
`pxf.cluster.options` as the key and `date_wide_range=true`. I'm not entirely sure there would ever be such a scenario, but is this something we would potentially need to account for in the future?

cli/cmd/cluster_test.go
yimingli-vmware: But I think we can still have a unit test about this case by just setting `PXF_BASE` with a non-writable directory

cli/config/config.go
```diff
@@ -93,3 +102,157 @@ func (p PxfServiceGroup) validate() error {
...
+		p.Groups[groupName] = group
+	}
+	switch propertyName {
+	case "ports":
+		valList := strings.Split(val, ",")
+		for _, v := range valList {
+			v = strings.TrimSpace(v)
+			parsedInt, err := strconv.ParseInt(v, 10, 32)
+			if err != nil {
+				return fmt.Errorf("ServiceGroup.ports property must be int: %s\n", v)
```
axue-broadcom, line 237: are these properties going to be case sensitive? should this be `servicegroup.ports` alternatively, if it should be the name of the service group itself, we should either print the actual value here

yimingli-vmware:
Could you provide some high level summary about this PR in the PR description? That will be the commit message for this PR

cli/config/config.go
```diff
@@ -93,3 +101,168 @@ func (p PxfServiceGroup) validate() error {
...
+	return strings.Split(string(data), "\n"), nil
+}
+
+func parseFile(fileContents []string) (map[string]string, error) {
+	parsedFileKeyVals := make(map[string]string)
+
+	for lineNum, line := range fileContents {
+		line = strings.ToLower(line)
+		line = strings.TrimSpace(line)
+		if line == "" || line[0] == '#' { // Remove any line that is a TOML comment
```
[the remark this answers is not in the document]
tatb-vmware: `strings.TrimSpace` will remove all leading and trailing white space so when we check for `#` it should be at the beginning of the string.

cli/config/config_test.go
tatb-vmware: As per team discussion, we will go with the second choice. We will state the fact observed then how to resolve it.

cli/config/config.go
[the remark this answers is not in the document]
tatb-vmware: Part of the code is to account for when we support multiple clusters for a single PXF deployment. Currently in MVP, we only support a single cluster so we return the first item in the list.

cli/config/config.go
```diff
@@ -93,3 +101,168 @@ func (p PxfServiceGroup) validate() error {
...
+		group = &PxfServiceGroup{}
+		group.Name = groupName
+		groups[groupName] = group
+	}
+	switch tokens[2] {
+	case "ports":
+		valList := strings.Split(val, ",")
+		for _, v := range valList {
+			v = strings.TrimSpace(v)
+			parsedInt, err := strconv.ParseInt(v, 10, 32)
```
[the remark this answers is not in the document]
tatb-vmware: The function signature is `func ParseInt(s string, base int, bitSize int) (i int64, err error)` so we use `base=10` and `bit=32`. The bitSize must be 0, 8, 16, 32, 64 according to the docs. We chose 32 because it fits the range we support for ports.

More context here: https://pkg.go.dev/strconv#ParseInt

cli/config/config.go
```diff
@@ -93,3 +101,168 @@ func (p PxfServiceGroup) validate() error {
...
+			if err != nil {
+				return fmt.Errorf("ServiceGroup ports must be int: %s\n", v)
+			}
+			group.Ports = append(group.Ports, int(parsedInt))
```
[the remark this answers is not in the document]
tatb-vmware: `strconv.ParseInt` returns an `int64` and we want to use the`int` type within our code to keep it generic so we need the type conversion.

cli/cmd/cluster.go
```diff
@@ -138,18 +145,71 @@ func GenerateOutput(cmd *command, clusterData *ClusterData) error {
...
+	var baseDirName string
+	baseDirName = os.Getenv("PXF_BASE")
+	if baseDirName == "" {
+		baseDirName = os.Getenv("PXF_HOME")
+		usedEnvVar = "PXF_HOME"
+	}
+	if baseDirName == "" {
+		return "", fmt.Errorf("please set PXF_BASE or PXF_HOME")
+	}
+	if fileInfo, err := os.Stat(baseDirName); os.IsNotExist(err) || !fileInfo.IsDir() {
```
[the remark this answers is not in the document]
tatb-vmware: No, it only checks the if the fileInfo has the ModeDir bit is set to 1.

More context here: https://pkg.go.dev/io/fs#FileMode.IsDir

cli/cmd/cluster_test.go
tatb-vmware: As per discussion with Alex, we have decided to keep it simple and only check if PXF_BASE is set or not.

cli/config/config.go
```diff
@@ -93,3 +101,168 @@ func (p PxfServiceGroup) validate() error {
...
+	}
+
+	if err := pxfDeployment.Validate(); err != nil {
+		return nil, err
+	}
+
+	return pxfDeployment.Clusters[clusterName], nil
+}
+
+func readFile(path string) ([]string, error) {
```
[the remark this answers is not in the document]
tatb-vmware: `ReadFile` is a function pointer that allows us to inject mocked fileContents in our unit test without having to read from a file. As per discussion, I can rename the function pointer and add a comment

cli/config/config.go
tatb-vmware: I have updated all of these to be `group.groupName.propertyName` to make it more consistent with what is inside the file.

cli/config/config.go
tatb-vmware: This situation is something we do not expect to support right now. Let's just error out if we see more than 1 `=`. If we need to support it in the future, we will add it.

cli/cmd/cluster.go
```diff
@@ -172,14 +235,25 @@ func retrieveClusterData() (*ClusterData, error) {
...
 	}
-	clusterData := &ClusterData{Cluster: cluster.NewCluster(segConfigs), connection: connection}
 
-	return clusterData, nil
+	return segConfigs, nil
 }
 
-func clusterRun(cmd *command, clusterData *ClusterData) error {
-	defer clusterData.connection.Close()
+func retrievePxfClusterInfo(name string) (*config.PxfCluster, error) {
```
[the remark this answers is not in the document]
tatb-vmware: I'm not sure if we can do this because we need `pxfCluster.Collocated` in `GetClusterData()`. Since that determines whether or not  `clusterData.PxfCluster` is the same as the Greenplum cluster or if it is needs to have different data to communicate with the external PXF cluster. If we move the construction and assignment of the PXF cluster to `convertPxfClusterInfo`, it will also need to be passed the `ClusterData` which might make the API for it more murky.

cli/cmd/pxf.go
```diff
@@ -38,6 +38,7 @@ type command struct {
 	whereToRun cluster.Scope
 	envVars    []envVar
 	warn       bool // whether the command requires a warning/prompt
+	envVarMap  map[envVar]string
```
denalex, line 41: do we still need this ?

denalex (approved):

cli/cmd/cluster.go
```diff
@@ -138,18 +145,63 @@ func GenerateOutput(cmd *command, clusterData *ClusterData) error {
...
+		clusterData.PxfCluster = clusterData.GreenplumCluster
+	} else {
+		clusterData.PxfCluster = convertPxfClusterInfo(pxfCluster)
+	}
+
 	return clusterData, nil
 }
 
+func convertPxfClusterInfo(pxfCluster *config.PxfCluster) *cluster.Cluster {
+	fakeSegConfigList := make([]cluster.SegConfig, len(pxfCluster.Hosts))
```
axue-broadcom, line 177: can we move the comment from line 179 to up here so we know why we need a fakeSegConfigList

cli/cmd/cluster.go
```diff
@@ -138,18 +145,63 @@ func GenerateOutput(cmd *command, clusterData *ClusterData) error {
...
+	for i, pxfHost := range pxfCluster.Hosts {
+		// We only need Port, Hostname, and Address to create a Cluster from PxfCluster
+		fakeSegConfigList[i] = cluster.SegConfig{
+			DbID:          i + 1,
+			ContentID:     i,
+			Role:          "p",
+			PreferredRole: "p",
+			Mode:          "n",
+			Status:        "u",
+			Port:          5888,
```
axue-broadcom: do we want this hard coded? users can update the port using `PXF_PORT`

cli/cmd/cluster.go
```diff
@@ -172,14 +227,24 @@ func retrieveClusterData() (*ClusterData, error) {
...
 	}
-	clusterData := &ClusterData{Cluster: cluster.NewCluster(segConfigs), connection: connection}
 
-	return clusterData, nil
+	return segConfigs, nil
 }
 
-func clusterRun(cmd *command, clusterData *ClusterData) error {
-	defer clusterData.connection.Close()
+func retrievePxfClusterInfo(name string) (*config.PxfCluster, error) {
```
axue-broadcom, line 234: the signatures for getting the Greenplum cluster is different from the PXF cluster. Would it make sense to make these consistent? (either both ask for `[]cluster.SegConfig` or both as for `*config.XXXCluster`)

If not, should the Greenplum function be `[]GreenplumCluster.SegConfig` since the other one is `config.PxfCluster`?

cli/cmd/cluster_test.go
```diff
@@ -581,19 +581,25 @@ var _ = Describe("GetClusterData()", func() {
...
 		connection            *dbconn.DBConn
-		mock                  sqlmock.Sqlmock
+		mockConnection        sqlmock.Sqlmock
+		mockSegConfigs        *sqlmock.Rows
 		oldConnectionProvider func() (*dbconn.DBConn, error)
 	)
 
 	BeforeEach(func() {
+		_ = os.Setenv("PXF_BASE", "../testdata")
+		_ = os.Unsetenv("PXF_HOME")
```
axue-broadcom, line 591: does this mean we no longer have a test case when the config is the default value of `PXF_HOME`?

cli/cmd/cluster.go
```diff
@@ -138,18 +145,71 @@ func GenerateOutput(cmd *command, clusterData *ClusterData) error {
...
+	fakeSegConfigList := make([]cluster.SegConfig, len(pxfCluster.Hosts))
+	for i, pxfHost := range pxfCluster.Hosts {
+		fakeSegConfigList[i] = cluster.SegConfig{
+			DbID:          i + 1,
+			ContentID:     i,
+			Role:          "p",
+			PreferredRole: "p",
+			Mode:          "n",
+			Status:        "u",
+			Port:          0,
```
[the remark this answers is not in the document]
tatb-vmware: 5888 is the PXF default port but this SegConfig port is the port that the database is using. Since this PxfCluster is not suppose to interact with the database, we can use a port value of 0

cli/cmd/pxf.go
tatb-vmware: No, it can be removed.

cli/cmd/cluster.go
tatb-vmware: SegConfig port value refers to the database port value and not the PXF cluster, so this value should never get used. We should set it to 0.

cli/cmd/cluster.go
```diff
@@ -159,8 +211,11 @@ func connectToGPDB() (*dbconn.DBConn, error) {
 	return connection, nil
 }
 
-func retrieveClusterData() (*ClusterData, error) {
+func retrieveGreenplumClusterInfo() ([]cluster.SegConfig, error) {
 	connection, err := ConnectionProvider()
+	if connection != nil {
```
djianwen-vmware, line 216: If only one of `connection` and `err` will be nil, is it better to use `else` rather than two `if`?

cli/cmd/cluster.go
```diff
@@ -138,18 +145,63 @@ func GenerateOutput(cmd *command, clusterData *ClusterData) error {
...
+		clusterData.PxfCluster = convertPxfClusterInfo(pxfCluster)
+	}
+
 	return clusterData, nil
 }
 
+func convertPxfClusterInfo(pxfCluster *config.PxfCluster) *cluster.Cluster {
+	fakeSegConfigList := make([]cluster.SegConfig, len(pxfCluster.Hosts))
+	for i, pxfHost := range pxfCluster.Hosts {
+		// We only need Port, Hostname, and Address to create a Cluster from PxfCluster
```
axue-broadcom: please update this comment since we no longer need port, can you also add a comment why it is ok for address to also be set as Hostname

cli/cmd/cluster.go
```diff
@@ -159,8 +211,11 @@ func connectToGPDB() (*dbconn.DBConn, error) {
 	return connection, nil
 }
 
-func retrieveClusterData() (*ClusterData, error) {
+func retrieveGreenplumClusterInfo() ([]cluster.SegConfig, error) {
```
axue-broadcom, line 214:
```suggestion
func retrieveGreenplumSegConfig() ([]cluster.SegConfig, error) {
```

cli/cmd/cluster.go
axue-broadcom, line 234:
```suggestion
func retrievePxfCluster(name string) (*config.PxfCluster, error) {
```

cli/cmd/cluster.go
```diff
@@ -138,18 +145,63 @@ func GenerateOutput(cmd *command, clusterData *ClusterData) error {
 // GetClusterData fetches information about Greenplum segments from Greenplum coordinator
 // and makes sure this script is running on the Greenplum coordinator host
 func GetClusterData() (*ClusterData, error) {
-	clusterData, err := retrieveClusterData()
+	clusterData := &ClusterData{}
+
+	segConfigs, err := retrieveGreenplumClusterInfo()
```
axue-broadcom, line 150:
```suggestion
	segConfigs, err := retrieveGreenplumSegConfig()
```

cli/cmd/cluster.go
```diff
@@ -138,18 +145,63 @@ func GenerateOutput(cmd *command, clusterData *ClusterData) error {
...
 	if err != nil {
 		return nil, err
 	}
+	clusterData.GreenplumCluster = cluster.NewCluster(segConfigs)
+	// TODO: remove the line below
```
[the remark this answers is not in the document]
tatb-vmware: After refactoring, it will not be done in this PR.

cli/cmd/cluster.go
tatb-vmware: After looking at the underlying implementation of `ConnectionProvider()`, it can return either a `connection` or `err` so a if-else statement would work here. I can add it.

cli/cmd/cluster.go
```diff
@@ -138,18 +145,63 @@ func GenerateOutput(cmd *command, clusterData *ClusterData) error {
...
+		}
+	}
+	return cluster.NewCluster(fakeSegConfigList)
+}
+
+func getPxfBase() (string, error) {
+	var baseDir string
+	baseDir, err := validateEnvVar("PXF_BASE")
+	if err != nil {
+		return "", fmt.Errorf("PXF_BASE is not set")
```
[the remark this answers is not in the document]
tatb-vmware: I wrap it at this level to be consistent with `stating the fact` opposed to `stating the violated rule`. Since `validateEnvVar()` is one of our functions, we can refactor it in the future to be consistent with our new error statements.

djianwen-vmware (approved):

cli/cmd/cluster.go
```diff
@@ -142,39 +142,45 @@ func GenerateOutput(cmd *command, clusterData *ClusterData) error {
...
 	}
 
 	return clusterData, nil
 }
 
-func convertPxfClusterInfo(pxfCluster *config.PxfCluster) *cluster.Cluster {
-	// We only need Port, Hostname, and Address to create a Cluster from PxfCluster
+// convertPxfCluster converts our internal representation of the PXF cluster into the `gp-common-go-libs`
+// representation of the cluster. Since `gp-common-go-libs` does not have a generic cluster definition, we need to fake
+// out our PxfCluster into a fake Greenplum cluster to use their functionality.
```
axue-broadcom, line 180:
```suggestion
// representation of the cluster. Since `gp-common-go-libs` does not have a generic cluster definition, we need to coerce the PxfCluster type into the gp-common-go-libs representation 
So that we can use the cluster functionality it provides to communicate with the external PXF cluster.
```

or something.

cli/cmd/cluster.go
```diff
@@ -142,39 +142,45 @@ func GenerateOutput(cmd *command, clusterData *ClusterData) error {
...
-// and makes sure this script is running on the Greenplum coordinator host
+// GetClusterData fetches information about Greenplum segments from Greenplum coordinator and PXF cluster from PXF
+// cluster config file. It makes sure that this script is running on the Greenplum coordinator host. Then, it will
+// see if the PXF cluster is collocated with the Greenplum cluster. If PXF cluster is external, convert the PxfCluster
+// to a `gp-common-go-lib` Cluster such that we can use its functionality on the PXF cluster.
 func GetClusterData() (*ClusterData, error) {
 	clusterData := &ClusterData{}
 
-	segConfigs, err := retrieveGreenplumClusterInfo()
+	greenplumSegConfigs, err := retrieveGreenplumSegConfig()
```
axue-broadcom, line 152: either both `SegConfigs` or both `SegConfig`. Consistency please.

axue-broadcom (approved):

axue-broadcom (approved):

PR #1135, opened 2024-04-18
Add go tool coverage as a Make command

tatb-vmware:
This PR adds a Make command to easily generate code coverage report when running test and add `ginkgo --randomize-all` to help reduce test spec pollution.

More info here: 
https://pkg.go.dev/cmd/cover
https://onsi.github.io/ginkgo/#spec-randomization


cli/Makefile
```diff
@@ -56,5 +60,6 @@ install-tools:
 clean:
 	rm -rf $(PXF_CLI_DIR)/build
 	go clean -i
+	rm -rf cover.*
```
denalex, line 63: can we create these files in some other directory and add that directory to `.gitignore` ?

denalex (approved):

yimingli-vmware (approved):

cli/Makefile
tatb-vmware: `go tool cover --coverprofile` does not allow writing to a file outside the current directory. The workaround is creating the `coverage.out` file in the current directory, then moving it to `coverage` dir before covering it to an HTML file.

tatb-vmware: After some discussion, we found that `go tool cover` runs per Go package. Therefore, we need a `coverage` directory within each package before it aggregates all the results into our top level `coverage` directory.

cli/Makefile
```diff
@@ -46,15 +46,24 @@ install: build
 	cp build/$(TARGET) $(PXF_HOME)/bin
 
 test: build install-tools
-	${GOBIN}/ginkgo cmd end_to_end
+	${GOBIN}/ginkgo --randomize-all cmd config end_to_end
+
+coverage: build install-tools
+	mkdir -p $(PXF_CLI_DIR)/{,cmd,config,end_to_end}/coverage
+	${GOBIN}/ginkgo --coverprofile=coverage/coverage.out --randomize-all cmd config end_to_end
+	go tool cover -html=coverage/coverage.out -o coverage/coverage.html
```
axue-broadcom, line 54:
```suggestion
        // ginkgo/go tool creates a coverage report per package and then consolidates them all into one report 
        // at the parent directory. 
	${GOBIN}/ginkgo --coverprofile=coverage/coverage.out --randomize-all cmd config end_to_end
	go tool cover -html=coverage/coverage.out -o coverage/coverage.html
```
 
or something that explains why we need coverage folders at every package level. either here or above the mkdir

cli/Makefile
```diff
@@ -46,15 +46,24 @@ install: build
 	cp build/$(TARGET) $(PXF_HOME)/bin
 
 test: build install-tools
-	${GOBIN}/ginkgo cmd end_to_end
+	${GOBIN}/ginkgo --randomize-all cmd config end_to_end
+
+coverage: build install-tools
```
axue-broadcom, line 51: please include all new make targets in the `help` target

cli/README.md
```diff
@@ -37,6 +37,14 @@
    ```
    See more in [Ginkgo Focused Specs](https://onsi.github.io/ginkgo/#focused-specs).
 
+1. Run unit tests with coverage
+
+   This will provide an html page showing which lines of code are covered or not.
+   ```
+   make coverage
+   open cover.html
```
axue-broadcom, line 45:
```suggestion
   open coverage/coverage.html
```

axue-broadcom (approved):

PR #1143, opened 2024-05-07
Parse cluster.txt in C extension

djianwen-vmware:
PXF 7 will support multi-cluster feature and thus the C extension needs the functionality to know which PXF host and port to communicate with.
This PR does the following:
- Add `pxf_deployment`, `pxf_cluster`, `pxf_service_group` and `pxf_host` structs
- Add functions to parse `cluster.txt` for `default` cluster into a `pxf_deployment` struct instance
  - For now, there is only one PXF cluster: `default` and it is hardcoded when fetching the `cluster.txt` file.
- Add cleanup functions to free the memory used by `pxf_deployment`


extension-common/Makefile
```diff
@@ -0,0 +1,8 @@
+GPDB_REPO_DIR ?= ${HOME}/workspace/gpdb/
+top_builddir=$(GPDB_REPO_DIR)
+
+include $(GPDB_REPO_DIR)/src/Makefile.global
+
+TARGETS=pxfconfigparser
+
+include $(GPDB_REPO_DIR)/src/backend/mock.mk
```
denalex, line 8: we explicitly did not want to do that, this introduces dependency on GPDB source code to be able to build PXF extension, which is not a dependency that is required for other Postgres / GP extensions. Only Greenplum installation is required, not the source code. We should ideally keep it that way, potentially copying over whatever pieces of mock testing framework we might need.

djianwen-vmware: > we explicitly did not want to do that, this introduces dependency on GPDB source code to be able to build PXF extension, which is not a dependency that is required for other Postgres / GP extensions. Only Greenplum installation is required, not the source code. We should ideally keep it that way, potentially copying over whatever pieces of mock testing framework we might need.

That's a valid concern. However, this doesn't introduce GPDB as dependency to compile PXF(pxf.so, compiled by `external-table/Makefile`). It's only a dependency to easily compile the test(pxfconfigparser.t, compiled by `extension-common/Makefile`).

More specifically, the dependency is Postgres server lib. It's already used before this PR. (see `external-table/Makefile` about `pg_config`). The problem is that our C extension uses the types and functions defined in Postgres server source code(You can see `#include "utils/guc.h"`, `#include "utils/builtins.h"` in our code), and they are not in the PXF repo. We either copy all the source code of Postgres server, or "link" it from somewhere.

The existing solution for building `pxf.so` is to refer to to makefile of Postgres(inside GPDB, in our case), using `pg_config --pgxs` command to find its folder. However, I tried the same way but couldn't get it compile into a binary for testing. I can explore further on this.

extension-common/pxfconfigparser.c
```diff
@@ -0,0 +1,538 @@
...
+ * Will error out and exit if key or value are not valid
+ * Expect the key to be a string separated by '.'
+ * Both key and value should be in lowercases and shorter than MAX_LENGTH_FILE_LINE
+ */
+static void
+populate_config_property(pxf_cluster *cluster, const char *key, const char *value)
+{
+    // Copy of key and value for tokenizing
+    char key_copy[MAX_LENGTH_FILE_LINE];
+    strlcpy(key_copy, key, MAX_LENGTH_FILE_LINE);
```
denalex, line 357: could use `pstrdup` that will malloc the new memory as well instead of allocating on the stack, not sure what's better

extension-common/pxfconfigparser.c
```diff
@@ -0,0 +1,538 @@
...
+    return (int)num;
+}
+
+/*
+ * Parse a string into a bool
+ * key is used for error message only
+ */
+static bool parse_bool(const char *str, const char *key)
+{
+    if (strcmp(str, "true") == 0 || strcmp(str, "1") == 0)
```
denalex, line 241: I don't think 1 and 0 are valid representation of boolean values for our properties
```suggestion
    if (strcmp(str, "true") == 0)
```

extension-common/pxfconfigparser.c
```diff
@@ -0,0 +1,538 @@
...
+    {
+        return true;
+    }
+    else if (strcmp(str, "false") == 0 || strcmp(str, "0") == 0)
```
denalex, line 245:
```suggestion
    else if (strcmp(str, "false") == 0)
```

extension-common/pxfconfigparser.c
```diff
@@ -0,0 +1,538 @@
...
+    pxf_service_group *group = hash_search(groups, group_name,
+                                           HASH_ENTER, &found);
+
+    if (!found)
+    {
+        // The first time to see this group, set all fields to zero
+        memset(group, 0, sizeof(pxf_service_group));
+    }
+
+    strlcpy(group->name, group_name, MAX_LENGTH_PXF_SERVICE_GROUP_NAME);
```
denalex, line 279: should this only be done once ? if yes, move it inside the if above

extension-common/pxfconfigparser.c
```diff
@@ -0,0 +1,538 @@
+#include "pxfconfigparser.h"
```
denalex, line 1: During the review I had to go from the bottom of the file up, since I start with the main function and then dive into helper functions it calls, so some comments have more context at the end of the file. I wonder if we should re-arrange the order of functions in the reverse order in this file.

extension-common/pxfconfigparser_test.c
```diff
@@ -0,0 +1,304 @@
...
+        ListCell *cell;
+        foreach (cell, entry->ports)
+        {
+            printf("group.port=%d\n", cell->data.int_value);
+        }
+    }
+}
+
+/*
+ * Verify that the pxf_cluster struct content is consistentwith fixture config file
```
denalex, line 65:
```suggestion
 * Verify that the pxf_cluster struct content is consistent with fixture config file
```

extension-common/pxfconfigparser_test.c
```diff
@@ -0,0 +1,304 @@
...
+ */
+static void verify_default_pxf_cluster_fixture(pxf_cluster *default_cluster)
```
denalex, line 67: interestingly, I rarely encountered a word `fixture` before, even though it is a thing for testing, apparently

extension-common/pxfconfigparser_test.c
```diff
@@ -0,0 +1,304 @@
...
+    PG_END_TRY();
+
+    test_reset();
+}
+
+static void
+test_parse_line_pxf_cluster_error_invalid_key(void **state)
+{
+    test_parse_line_pxf_cluster_fail_with_error("foo=value",
+                                                "invalid cluster.txt for 'test' cluster: please make sure key 'foo' is valid");
```
denalex, line 144: aren't we supposed to ignore unknown properties ?

extension-common/pxfconfigparser_test.c
```diff
@@ -0,0 +1,304 @@
...
+{
+    test_parse_line_pxf_cluster_fail_with_error("foo=value=value2",
+                                                "invalid cluster.txt for 'test' cluster: please make sure each key-value pair only has one '='");
+}
+
+static void
+test_parse_line_pxf_cluster_error_key_too_many_tokens(void **state)
+{
+    test_parse_line_pxf_cluster_fail_with_error("group.default.ssl.extra=value",
+                                                "invalid cluster.txt for 'test' cluster: please make sure key 'group.default.ssl.extra' is valid");
```
denalex, line 158: it is not clear what is wrong with the key `group.default.ssl.extra` and how to make it valid. ideally we should have  a better error message

extension-common/pxfconfigparser_test.c
```diff
@@ -0,0 +1,304 @@
...
+{
+    test_parse_line_pxf_cluster_fail_with_error("group.default.ssl=foo",
+                                                "configuration property 'group.default.ssl' is set to a non-boolean value 'foo'");
+}
+
+static void
+test_parse_line_pxf_cluster_error_invalid_ports(void **state)
+{
+    test_parse_line_pxf_cluster_fail_with_error("group.default.ports=foo",
+                                                "configuration property 'group.default.ports' is set to an invalid integer value");
```
denalex, line 193: should we include the value that is invalid ?

extension-common/pxfconfigparser.c
```diff
@@ -0,0 +1,538 @@
...
+         token = strtok_r(NULL, delim, &save_ptr))
+    {
+        trim_space(token);
+        if (strlen(token) == 0)
+        {
+            // Empty token is not allowed
+            return NULL;
+        }
+        tokens = lappend(tokens, token);
+    }
```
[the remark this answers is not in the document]
djianwen-vmware: > I wonder why `strtok_r` is used vs `strtok` - for thread safety or otherwise ? I don't think we have multiple threads in the backend process.

Yes, I'm not sure about the memory model in terms of extension, so I'm conservative here. If we know it's not a matter, I will change it back to `strtok`.

extension-common/pxfconfigparser_test.c
djianwen-vmware: > aren't we supposed to ignore unknown properties ?

This errors out because the key only has one token. We expect number of tokens in key to be either 2 or 3. I notice in Golang we also check the number of tokens and error out, so I did the same way.

extension-common/pxfconfigparser_test.c
djianwen-vmware: > it is not clear what is wrong with the key `group.default.ssl.extra` and how to make it valid. ideally we should have a better error message

This is the same error as above, where the number of tokens in key is wrong. I can improve the error message or skip the unknown property. Do you have a preference?

extension-common/pxfconfigparser.h
```diff
@@ -0,0 +1,72 @@
...
+#define parser_terminate(...)     \
+    do                            \
+    {                             \
+        deployment_cleanup();     \
+        elog(ERROR, __VA_ARGS__); \
+    } while (0)
+
+typedef struct
+{
+    char name[MAX_LENGTH_PXF_SERVICE_GROUP_NAME];
```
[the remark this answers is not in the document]
djianwen-vmware: > I wonder why "strings" are defined as `char[]` and not as `char*` ?

I wanted to avoid memory allocation/free when possible. If we already know the length limit(here the filename max length is 256, so group name won't exceed it), then we can avoid null pointer check, length check, and memory management everywhere. Allocate/free the struct will also allocate/free the string inside.
I can change it to char* if we want, just wondering do we have a convention on this?

extension-common/pxfconfigparser_test.c
```diff
@@ -0,0 +1,304 @@
+#include <stdarg.h>
```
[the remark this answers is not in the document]
djianwen-vmware: > I wonder if we can have a test directory and place tests there (Java style) vs having tests in the same directory as the source code (Go style). This is how existing (non-functional) tests are structured.

I can try that. I put them in the same folder because `external-table/test` folders already have "oudated" tests and makefiles. I didn't want to remove them so I put in the same level. Common folder refactoring was done later.

extension-common/pxfconfigparser.c
```diff
@@ -0,0 +1,538 @@
...
+    if (str[start] == '\0')
+    {
+        // The string only has spaces
+        str[0] = '\0';
+        return;
+    }
+
+    // Find the index of the last non-space char
+    int end = (int)strlen(str) - 1;
+    while (end >= start && isspace(str[end]))
```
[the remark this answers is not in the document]
djianwen-vmware: > why do we need to check the character that start points to again ?

Good catch. It's actually not necessary. I will fix it.

extension-common/test-fixture-pxf-base/clusters/default/conf/cluster.txt
```diff
@@ -0,0 +1,9 @@
+
```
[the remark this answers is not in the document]
djianwen-vmware: > I wonder if this can go into `extension-common/test/resources/clusters/default/conf/cluster.txt` to be more like Maven style project setup, unless C code has specific conventions

I will do it. Thanks!

extension-common/pxfconfigparser.c
```diff
@@ -0,0 +1,538 @@
...
+
+    for (char *token = strtok_r(str, delim, &save_ptr);
+         token != NULL;
+         token = strtok_r(NULL, delim, &save_ptr))
+    {
+        trim_space(token);
+        if (strlen(token) == 0)
+        {
+            // Empty token is not allowed
+            return NULL;
```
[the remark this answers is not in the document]
djianwen-vmware: > endpoint is an optional property and might not have a value, so we should allow for empty tokens at this level and if this is the case, the loop can be re-written as a while loop

Endpoint will not go through this function, because it's a single value. Is there a case we allow an array of empty strings in value?

extension-common/pxfconfigparser.h
```diff
@@ -0,0 +1,72 @@
...
+    char hostname[MAX_LENGTH_HOSTNAME];
+} pxf_host;
+
+typedef struct
+{
+    char name[MAX_LENGTH_PXF_CLUSTER_NAME];
+    bool collocated;
+    char endpoint[MAX_LENGTH_HOSTNAME];
+    List *hosts;
+    HTAB *groups;
```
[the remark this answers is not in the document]
djianwen-vmware: > I am not sure if a hashtable is really warranted here (have not seen it in many C structs) -- we will have at most a few clusters, so having a list and iterating over it until we find the one needed would have worked fine as well

I saw we already used "List" struct in our codebase, then I found hash table is also available there. So I followed our Golang struct to keep consistent.

extension-common/pxfconfigparser.c
```diff
@@ -0,0 +1,538 @@
...
+    }
+
+    fclose(cluster->config_file);
+    cluster->config_file = NULL;
+}
+
+static void
+add_pxf_cluster(HTAB *clusters, const char *cluster_name)
+{
+    pxf_cluster *cluster = hash_search(clusters, cluster_name,
```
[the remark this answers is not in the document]
djianwen-vmware: > can we add a comment on what this is doing ? does it allocate memory for the cluster struct and hooks it into the hashtable if it does not exist ?

That's correct. I will add a comment here.

extension-common/pxfconfigparser.c
```diff
@@ -0,0 +1,538 @@
...
+
+    // Parse each of the cluster
+    HASH_SEQ_STATUS status;
+    pxf_cluster *cluster;
+    char filepath[MAX_LENGTH_FILEPATH];
+    hash_seq_init(&status, deployment->clusters);
+    while ((cluster = (pxf_cluster *)hash_seq_search(&status)) != NULL)
+    {
+        // TODO(MULTI-CLUSTER): fetch PXF_BASE from GPDB GUC
+        snprintf(filepath, MAX_LENGTH_FILEPATH, "%s/clusters/%s/conf/cluster.txt", pxf_base, cluster->name);
```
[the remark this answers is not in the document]
djianwen-vmware: > I know GP uses `StringInfo` for string content manipulation, should we use it here ?

I looked at the header containing `StringInfo` and it doesn't provide much useful functionalities in our case. It brings a drawback that we need to remember to free its data after using it, which is not detectable by testing framework(because function is not defined by us). There is a provided `initStringInfo` function but no `cleanup` function for it. So I hesitated to use it.

extension-common/pxfconfigparser.c
djianwen-vmware: > could use `pstrdup` that will malloc the new memory as well instead of allocating on the stack, not sure what's better

Stack memory will be cleaned up after function returns, while malloc needs to free before return or before any error happens. For example, 
```
char* str1 = strdup(str);
if (error){
  free(str1);
  print_error();
}
char* str2 = strdup(str1);
if (error) {
  free(str1);
  free(str2);
  print_error();
}
```
The more malloc we have, the more complex the error handling will be in a function. My concern is the risk of memory leaks and code readability issues if we introduce more heap memory operations. Let me know if you feel heap is better, I can modify the code and show you what it will look like.

extension-common/pxfconfigparser_test.c
djianwen-vmware: Update: I will ignore all unknown properties and log to DEBUG1 level.

extension-common/pxfconfigparser_test.c
djianwen-vmware: Update: will ignore unknown property and log to DEBUG1 level

extension-common/Makefile
djianwen-vmware: Update on this: some dependency functions we used in C extensions are from GPDB(or Postgres) source code. Even though those functions are declared in `/usr/local/gpdb/include/**` header files, they are not defined in the `/usr/local/gpdb/lib/**` lib files. 
We either use GPDB repo to link into an executable, or build a `.so` without dependency and load it into a running GPDB where functions exist in memory. The second option is how we compile the `pxf.so`, and the first option is for the current unit tests. Let's keep it before there is a better solution.

extension-common/pxfconfigparser.c
```diff
@@ -0,0 +1,538 @@
...
+
+    if (cluster->config_file == NULL)
+    {
+        parser_terminate("cannot open the PXF cluster config file: '%s', "
+                         "please make sure the file exists and you have access to it",
+                         filepath);
+    }
+
+    char line[MAX_LENGTH_FILE_LINE]; // Buffer to hold each line
+    while (fgets(line, sizeof(line), cluster->config_file))
```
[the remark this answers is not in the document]
djianwen-vmware: > what will happen if the line is longer than `MAX_LENGTH_FILE_LINE` ?

The line will be split into two or more lines. Each new line contains at most `MAX_LENGTH_FILE_LINE-1` char and with a `\0` at the end.

extension-common/pxfconfigparser.h
```diff
@@ -0,0 +1,72 @@
...
+#define PXF_DEFAULT_SERVICE_GROUP "default"
+
+#define MAX_LENGTH_FILEPATH 4096
+#define MAX_LENGTH_FILE_LINE 256
+
+#define MAX_NUM_PXF_CLUSTER 128
+#define MAX_NUM_PXF_SERVICE_GROUP 128
+
+#define MAX_LENGTH_PXF_SERVICE_GROUP_NAME 256
+#define MAX_LENGTH_PXF_CLUSTER_NAME 256
```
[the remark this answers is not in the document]
djianwen-vmware: > We don't have the length limitation for these variables. If we define the variable as a pointer not an array, we can get rid of these.

I've removed most of them. However, the `MAX_LENGTH_FILE_LINE ` can simplify the handling of line reading. I will updated the code to remove the limit and see if you feel it's better.

extension-common/pxfconfigparser.c
```diff
@@ -0,0 +1,538 @@
...
+    {
+        fclose(cluster->config_file);
+        cluster->config_file = NULL;
+    }
+    if (cluster->groups != NULL)
+    {
+        HASH_SEQ_STATUS status;
+        pxf_service_group *entry;
+        hash_seq_init(&status, cluster->groups);
+        while ((entry = (pxf_service_group *)hash_seq_search(&status)) != NULL)
```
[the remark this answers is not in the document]
djianwen-vmware: > what does `hash_seq_init` check, and why this `while` is needed? Can we just directly use the `if-else` block on line 76?

I've removed hash table from our code base

extension-common/pxfconfigparser.c
```diff
@@ -0,0 +1,538 @@
...
+
+static void parse_file_pxf_cluster(pxf_cluster *cluster, const char *filepath);
+
+static void add_pxf_cluster(HTAB *clusters, const char *cluster_name);
+
+/*
+ * Helper function called in deployment_cleanup to clean up a pxf_cluster struct
+ */
+static void
+cluster_cleanup(pxf_cluster *cluster)
```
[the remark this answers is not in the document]
djianwen-vmware: > Is this the convention having these 2 on separate lines?

Yes

extension-common/pxfconfigparser.c
```diff
@@ -0,0 +1,538 @@
...
+ */
+static List *
+get_tokens(char *str, char *delim)
+{
+    List *tokens = NULL;
+    char *save_ptr = NULL; // Save the progress of the tokenizer
+
+    for (char *token = strtok_r(str, delim, &save_ptr);
+         token != NULL;
+         token = strtok_r(NULL, delim, &save_ptr))
```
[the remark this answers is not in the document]
djianwen-vmware: > why is the first parameter `NULL` when we want to update the `token` for next iteration? Is that because `save_ptr` will be increased inside `strtok_r`, and you don't want `save_ptr` to be update when calling `token = strtok_r(NULL, delim, &save_ptr)`? Can we replace the `for` loop with a `while` loop?

```
On a first call, the function expects a C string as argument for str, whose first character is used as the starting location to scan for tokens. In subsequent calls, the function expects a null pointer and uses the position right after the end of the last token as the new starting location for scanning.
```
https://cplusplus.com/reference/cstring/strtok/
Yes, we can change it to a while-loop. Then we need to call it for the first time before the while-loop, because the parameters are different.

extension-common/pxfconfigparser.c
```diff
@@ -0,0 +1,538 @@
...
+        List *ports = get_tokens(value, ",");
+        if (ports == NULL)
+        {
+            // Some ports are empty strings
+            parser_terminate("configuration property '%s' contains invalid value, "
+                             "please make sure the value is valid port numbers separated by ','",
+                             key);
+        }
+
+        ListCell *cell;
```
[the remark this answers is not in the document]
djianwen-vmware: > `ListCell` is a linkedlist in postgres right? Why we need a linkedlist and use `foreach` instead of having a for loop over the `ports`?

Because we need to know the number of ports before we parse it if we use an array or pointer
(```int* ports; ports=palloc(MAX_NUM_PORTS*sizeof(int));``` 
or
```int ports[MAX_NUM_PORTS];```), and need an additional variable to save the length. In C there is no built-in array object where we can append flexibly and easily. We can re-build the wheel where we double the size if exceeding capacity, but I think it's too much for our repo.

File: extension-common/pxfconfigparser.c [hunk 27]
```diff
@@ -0,0 +1,538 @@
...
+static void
+populate_config_property(pxf_cluster *cluster, const char *key, const char *value)
+{
+    // Copy of key and value for tokenizing
+    char key_copy[MAX_LENGTH_FILE_LINE];
+    strlcpy(key_copy, key, MAX_LENGTH_FILE_LINE);
+    char value_copy[MAX_LENGTH_FILE_LINE];
+    strlcpy(value_copy, value, MAX_LENGTH_FILE_LINE);
+
+    List *tokens = get_tokens(key_copy, ".");
```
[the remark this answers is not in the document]
djianwen-vmware: > I'm wondering if we can pass the pointer value into `get_tokens`, then we would not need to do the manual copy? But the function implementation needs to be changed since the first variable would be a value not a pointer.

There is no way to pass a string in value. Because there is no `string` type in C. In C language, string is an array of `char` where its symbol `key` is the pointer to its first `char`.

extension-common/pxfconfigparser.c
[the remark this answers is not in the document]
djianwen-vmware: Update: I've removed the line length limit by adding additional logic to check the last `char` in the buffer. It's a little bit more complicated than I thought. I will need to write new tests for it.

extension-common/pxfconfigparser.c
djianwen-vmware: > During the review I had to go from the bottom of the file up, since I start with the main function and then dive into helper functions it calls, so some comments have more context at the end of the file. I wonder if we should re-arrange the order of functions in the reverse order in this file.

I've changed it to be top-down. Thanks!

extension-common/pxfconfigparser.c
```diff
@@ -0,0 +1,538 @@
...
+    }
+
+    const char delim[] = "=";
+    char *save_ptr = NULL; // Save the progress of the tokenizer
+    char *key = strtok_r(line_copy, delim, &save_ptr);
+    char *value = strtok_r(NULL, delim, &save_ptr);
+    if (value == NULL)
+    {
+        // No '=' found in the line
+        parser_terminate("invalid cluster.txt for '%s' cluster: please make sure each key-value pair has '=' in between",
```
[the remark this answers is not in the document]
djianwen-vmware: > it would be best to give them the full path to the file and the line number (and not the cluster name)

Updated. In order to include filepath and line number in all functions, I introduced the `parser_line_context `struct ` as a parameter.

extension-common/pxfconfigparser.c
```diff
@@ -0,0 +1,694 @@
...
+         * While a buffer of 4 bytes containing 'a''b''\n''\0' means it's the end of line
+         */
+        if (read_count == (buffer_size - 1) && buffer[buffer_size - 2] != '\n')
+        {
+            // The current line is longer than buffer size, continue reading and appending
+            line_not_finished = true;
+            continue;
+        }
+
+        // Already the end of line (with a '\n' at the end)
```
denalex, line 506: so where are we removing the trailing `\n` since we do not need it ?

extension-common/pxfconfigparser.h
```diff
@@ -0,0 +1,67 @@
...
+ * Make a copy of a field if modification is needed
+ */
+typedef struct
+{
+    const int line_number;
+    const char *filepath;
+    const char *key;
+    const char *value;
+    const char *line;
+} parser_line_context;
```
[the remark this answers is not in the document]
djianwen-vmware: > this seems like an internal structure for parsing, should we take it out of `.h` file (which is like a public interface for this module)?

Very good point. I will move it.

extension-common/pxfconfigparser.c
```diff
@@ -0,0 +1,694 @@
...
+
+#define malloc(size) _test_malloc(size, __FILE__, __LINE__)
+#define palloc(size) _test_malloc(size, __FILE__, __LINE__)
+
+#define palloc0(size) _test_palloc0(size, __FILE__, __LINE__)
+#define pstrdup(str) _test_pstrdup(str, __FILE__, __LINE__)
+#define calloc(num, size) _test_calloc(num, size, __FILE__, __LINE__)
+#define initStringInfo(str) _test_initStringInfo(str, __FILE__, __LINE__)
+#define free(ptr) _test_free(ptr, __FILE__, __LINE__)
+#define pfree(ptr) _test_free(ptr, __FILE__, __LINE__)
```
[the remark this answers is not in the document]
djianwen-vmware: > why all these definitions needed for testing ? why not use the real stuff ?

`_test_calloc()` and `_test_free()` are the functions provided by the testing framework. It injects the memory allocation by tracking all the addresses allocated and freed. At the end of the test, if any allocated memory is not freed, it will error out. If you double-free an allocated address, it will also error out. 

When we want to check if we have memory leaks, we can enable the macros MEMORY_TESTING flag and see if the tests report unexpected leaks. When the flag is not enabled (be default), it will use the real functions.

extension-common/pxfconfigparser.c
djianwen-vmware: > so where are we removing the trailing `\n` since we do not need it ?

Good question. It's deleted in `trim_space()` function. All the following characters will be identified as spaces by the built-in C standard function `isspace()`

' ' | space
'\n' | newline
'\t' | horizontal tab
'\v' | vertical tab
'\f' | form feed
'\r' | Carraige return

extension-common/pxfconfigparser.c
```diff
@@ -0,0 +1,694 @@
...
+ */
+static void init_pxf_cluster(pxf_cluster *cluster, const char *cluster_name)
+{
+    // Initialize all fields as zeros
+    memset(cluster, 0, sizeof(pxf_cluster));
+
+    initStringInfo(&cluster->name);
+    initStringInfo(&cluster->endpoint);
+    appendStringInfo(&cluster->name, "%s", cluster_name);
+}
```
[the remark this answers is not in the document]
djianwen-vmware: > I'd combine these into 1 function that does palloc, init, hook into parent also I think `palloc0` already fills the memory with 0, why another `memset` ? cluster name will not be composed, there is no reason to use StringInfo, we can just `pstrdup` existing string into a new string and save the pointer.
> 
> I'd have endpoint as a char* and leave it as null at this point.Same for the endpoint, if it is a char*

Using char* in struct may simplify the code. My concern is that the if the returned struct contains `NULL`, it's more error-prone. My initial idea is to make sure the returned struct contains at least "valid" values in all the fields, including empty string and empty array, but not NULL. Though the whole deployment is not validated in C.

If the config file doesn't contain a line for the string, it will be NULL if we use char*. If we use StringInfoData, it's initialized to an empty string which is always safe to use and can error out properly if empty string is wrong to use.

extension-common/pxfconfigparser.c
```diff
@@ -0,0 +1,694 @@
...
+    return (int)num;
+}
+
+/*
+ * Parse a string into a bool
+ * ctx is used for error message only
+ */
+static bool
+parse_bool(const char *str, const parser_line_context *ctx)
+{
```
yimingli-vmware, line 640: Do we need to trim space or convert to lower case before doing the compare? Are we expecting the only valid inputs are `true` and `false`?

extension-common/pxfconfigparser.c
```diff
@@ -0,0 +1,694 @@
...
+
+    // Get the new length of the trimmed line
+    int len = strlen(line_copy);
+    if (len == 0 || line_copy[0] == '#')
+    {
+        // Skip empty lines or comments
+        pfree(line_copy);
+        return;
+    }
+
```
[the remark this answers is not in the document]
djianwen-vmware: > would it make sense to move lowercasing here after we made sure to trim and analyze for empty/comment ?

Good point. Will fix it.

extension-common/pxfconfigparser.c
```diff
@@ -0,0 +1,694 @@
...
+    const char delim[] = "=";
+
+    char *key = strtok(line_copy, delim);
+    char *value = strtok(NULL, delim);
+    if (value == NULL)
+    {
+        // No '=' found in the line
+        ereport(ERROR, (errcode(ERRCODE_CONFIG_FILE_ERROR),
+                        errmsg("invalid config file '%s': please make sure each line is a valid key-value pair", filepath),
+                        errhint("Is there a '=' between the key and value?")));
```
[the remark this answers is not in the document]
djianwen-vmware: > 

Do we want to have `.` at the end of all error and hint messages? In the repo most of them don't have a `.`. Maybe keep all of them without `.` such that in this file all messages are consistent?

extension-common/pxfconfigparser.c
```diff
@@ -0,0 +1,694 @@
...
+{
+    if (strcmp(property_name, "name") == 0)
+    {
+        if (strcmp(ctx->value, cluster->name.data) != 0)
+        {
+            // cluster.name should already be set. Error out if not consistent.
+            ereport(ERROR, (errcode(ERRCODE_CONFIG_FILE_ERROR),
+                            errmsg("configuration property 'cluster.name' is set to '%s', does not match cluster directory name '%s', at line %d of file '%s'",
+                                   ctx->value, cluster->name.data, ctx->line_number, ctx->filepath),
+                            errhint("the value of 'cluster.name' should be the same as [cluster-name] "
```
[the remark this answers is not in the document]
djianwen-vmware: > how does it work with 2 strings with no operation between them ?

The `cluster->name` should already be set when we initialize the pxf_cluster inside the pxf_deployment (see where we call `add_pxf_cluster(deployment, PXF_DEFAULT_CLUSTER);`). 

We should have a list of cluster names. However right now we we only have `default` cluster. In the future we should read the folder names and fetch them as a list and support multi-cluster.

extension-common/pxfconfigparser.c
```diff
@@ -0,0 +1,694 @@
...
+                                    "in its config file path '${PXF_BASE}/clusters/[cluster-name]/conf/cluster.txt'")));
+        }
+        // if consistent, do nothing
+    }
+    else if (strcmp(property_name, "collocated") == 0)
+    {
+        cluster->collocated = parse_bool(ctx->value, ctx);
```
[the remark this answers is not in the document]
djianwen-vmware: > the value is optional, might be empty
From the design doc, the collocated should not be NULL. So I think it must be set?
Even though a field is optional, we should not have a incomplete line like `key=` in config file. Just do not specify I think it will skip setting it.

extension-common/pxfconfigparser.c
djianwen-vmware: > Do we need to trim space or convert to lower case before doing the compare? Are we expecting the only valid inputs are `true` and `false`?

All the tokens are already trimmed and converted into lowercase, so only `true` and `false` are valid. I can add a comment.

extension-common/pxfconfigparser.h
```diff
@@ -0,0 +1,67 @@
...
+
+pxf_deployment *get_pxf_deployment(const char *);
+
+void deployment_cleanup(pxf_deployment *deployment);
+
+// Helper functions to search for an entry
+pxf_cluster *get_pxf_cluster(pxf_deployment *deployment, const char *cluster_name);
+pxf_service_group *get_pxf_service_group(pxf_cluster *cluster, const char *group_name);
```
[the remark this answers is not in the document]
djianwen-vmware: > in the future when the workflow rules are implemented, the rule will target a service group by name only, it will not know which particular cluster the group will reside on, so this API should not take in a cluster name or a pointer, just a deployment pointer and the group name. Service group names are supposed to be unique within a deployment.

That's a good point. 
Updated the code to use deployment for fetching a group in the public function. 

By the way I think it's still good to have a private function for fetching group from cluster. So I created one there. It benefits the test and internal function logic.

extension-common/pxfconfigparser.c
```diff
@@ -0,0 +1,694 @@
...
+{
+    pxf_deployment *deployment = (pxf_deployment *)palloc0(sizeof(pxf_deployment));
+
+    // Create an entry for the default cluster
+    // TODO(MULTI-CLUSTER): support multi-cluster in the future. Need to fetch the list of the clusters and add them here
+    add_pxf_cluster(deployment, PXF_DEFAULT_CLUSTER);
+
+    // Buffer string for constructing filepath
+    StringInfoData filepath;
+    initStringInfo(&filepath);
```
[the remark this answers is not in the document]
djianwen-vmware: > 

After taking look at other examples, if we use it in this way, we need to free twice because StringInfo and its data are both pointers to heap memory:
```
pfree(filepath->data);
pfree(filepath);
filepath=NULL;
```
And in memory test I need to inject this `makeStringInfo` to track its address. Thanks for bringing up this idea though.

extension-common/pxfconfigparser.c
```diff
@@ -0,0 +1,694 @@
...
+    }
+
+    if (cluster->hosts != NULL)
+    {
+        pxf_host *host;
+        ListCell *cell;
+        foreach (cell, cluster->hosts)
+        {
+            host = cell->data.ptr_value;
+            pfree(host->hostname.data);
```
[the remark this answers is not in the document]
djianwen-vmware: > if this becomes char* you can use `list_free_deep`

I updated it to use `list_free_deep`. However, `list_free_deep()` can only help free the `host` struct, but not the `hostname` string.  We still need to loop through and free strings manually.

extension-common/pxfconfigparser.c
```diff
@@ -0,0 +1,694 @@
...
+
+    if (config_file == NULL)
+    {
+        ereport(ERROR, (errcode(ERRCODE_CONFIG_FILE_ERROR),
+                        errmsg("cannot open the PXF cluster config file: '%s', please make sure the file exists and you have access to it", filepath),
+                        errhint("Did you run 'pxf cluster prepare' and 'pxf cluster sync' to define cluster metadata?")));
+    }
+
+    StringInfoData line;
+    initStringInfo(&line);
```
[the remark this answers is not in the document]
djianwen-vmware: > 

Same here, if we use `makeStringInfo`, we need to free twice, which doesn't save any code line for us.
Reference in `stringinfo.h`:
```
 * StringInfo stringptr = makeStringInfo();
 *		Both the StringInfoData and the data buffer are palloc'd.
 *
 * StringInfoData string;
 * initStringInfo(&string);
 *		The data buffer is palloc'd but the StringInfoData is just local.
 *		This is the easiest approach for a StringInfo object that will
 *		only live as long as the current routine.
 *
 * To destroy a StringInfo, pfree() the data buffer, and then pfree() the
 * StringInfoData if it was palloc'd.  There's no special support for this.
 ```

File: extension-common/pxfconfigparser.c (hunk 31 above)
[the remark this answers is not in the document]
Comment (djianwen-vmware) [reply extension-common/pxfconfigparser.c]: > > why all these definitions needed for testing ? why not use the real stuff ?
> 
> `_test_calloc()` and `_test_free()` are the functions provided by the testing framework. It injects the memory allocation by tracking all the addresses allocated and freed. At the end of the test, if any allocated memory is not freed, it will error out. If you double-free an allocated address, it will also error out.
> 
> When we want to check if we have memory leaks, we can enable the macros MEMORY_TESTING flag and see if the tests report unexpected leaks. When the flag is not enabled (be default), it will use the real functions.

Update: I've removed all the mocks from unit tests.

File: extension-common/pxfconfigparser.c [hunk 42]
```diff
@@ -0,0 +1,661 @@
...
+        char *token3 = lthird(tokens);
+
+        populate_group_property(cluster, token2, token3, ctx);
+    }
+    else
+    {
+        ereport(DEBUG1,
+                (errmsg("ignoring unknown configuration property '%s' in config file '%s' at line %d", ctx->key, ctx->filepath, ctx->line_number)));
+    }
+    pfree(key_copy);
```
denalex, line 226: do we need to `list_free(tokens)` ?

extension-common/pxfconfigparser.c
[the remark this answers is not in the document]
denalex: I meant syntactically, there are 2 lines of quoted text, how does it become 1 line without any operands ?:
```
errhint("the value of 'cluster.name' should be the same as [cluster-name] "
"in its config file path '${PXF_BASE}/clusters/[cluster-name]/conf/cluster.txt'")
```

extension-common/pxfconfigparser.c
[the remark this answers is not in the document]
denalex: I would agree, we need to check on the Go side whether we allow for an empty value. This migrated from the previous iteration where the file was in YAML and it was possible to have an non-specified value for the key.

extension-common/pxfconfigparser.c
```diff
@@ -0,0 +1,661 @@
...
+                                   ctx->key, ctx->value, ctx->line_number, ctx->filepath),
+                            errhint("please make sure the value is valid port numbers separated by ','")));
+        }
+
+        ListCell *lc;
+        foreach (lc, ports)
+        {
+            int port = parse_int(lfirst(lc), ctx);
+            group->ports = lappend_int(group->ports, port);
+        }
```
denalex, line 306: do we need to `list_free(ports)` ?

extension-common/pxfconfigparser.c
```diff
@@ -0,0 +1,661 @@
...
+    if (deployment->clusters != NULL)
+    {
+        ListCell *lc;
+        foreach (lc, deployment->clusters)
+        {
+            pxf_cluster *cluster = lfirst(lc);
+            clean_pxf_cluster(cluster);
+            pfree(cluster);
+        }
+        list_free(deployment->clusters);
```
denalex, line 355:
```suggestion
        }
        list_free_deep(deployment->clusters);
```
I think you can do a deep clean of the list to clean cluster pointers

extension-common/pxfconfigparser.c
```diff
@@ -0,0 +1,661 @@
...
+        deployment->clusters = NULL;
```
denalex, line 356: while this is generally a good practice, you're pfree-ing the whole deployment right after, so there is not much point of setting this pointer to NULL.

extension-common/pxfconfigparser.c
```diff
@@ -0,0 +1,661 @@
...
+ */
+static void
+clean_pxf_cluster(pxf_cluster *cluster)
+{
+    if (cluster == NULL)
+    {
+        return;
+    }
+
+    if (cluster->name)
```
denalex, line 372: it's unfortunate `pfree` wants a not-NULL pointer since regular `free(NULL)` is ok: https://stackoverflow.com/questions/6084218/is-it-good-practice-to-free-a-null-pointer-in-c

extension-common/pxfconfigparser.c
```diff
@@ -0,0 +1,661 @@
...
+        list_free_deep(cluster->groups);
+        cluster->groups = NULL;
+    }
+
+    if (cluster->hosts != NULL)
+    {
+        ListCell *lc;
+        foreach (lc, cluster->hosts)
+        {
+            pxf_host *host = lfirst(lc);
```
denalex, line 406:
```suggestion
            char *host = lfirst(lc);
```
why not make it `char*` since there are no other attributes for the host ? Then you can just do `list_clean_deep` here instead of iterating over hosts list.

extension-common/pxfconfigparser.c
```diff
@@ -0,0 +1,661 @@
...
+pxf_cluster *get_pxf_cluster(const pxf_deployment *deployment, const char *cluster_name)
+{
+    pxf_cluster *found = NULL;
+    ListCell *lc;
+    foreach (lc, deployment->clusters)
+    {
+        pxf_cluster *cluster_entry = lfirst(lc);
+        if (strcmp(cluster_entry->name, cluster_name) == 0)
+        {
+            found = cluster_entry;
```
denalex, line 620:
```suggestion
        pxf_cluster *cluster = lfirst(lc);
        if (strcmp(cluster->name, cluster_name) == 0)
        {
            found = cluster;
```

extension-common/pxfconfigparser.c
```diff
@@ -0,0 +1,661 @@
...
+                        errmsg("configuration property '%s' is set to a non-boolean value '%s' in the config file '%s' at line %d", ctx->key, str, ctx->filepath, ctx->line_number),
+                        errhint("Please make sure the value is either 'true' or 'false'")));
+    }
+}
+
+/*
+ * A helper function to get a pxf_cluster from a pxf_deployment struct
+ */
+pxf_cluster *get_pxf_cluster(const pxf_deployment *deployment, const char *cluster_name)
+{
```
denalex, line 612: if this is a public function, we should check for nulls on both params.

extension-common/pxfconfigparser.c
```diff
@@ -0,0 +1,661 @@
...
+        }
+    }
+    return found;
+}
+
+/*
+ * A helper function to get a pxf_service_group from a pxf_deployment struct
+ */
+pxf_service_group *get_pxf_service_group(const pxf_deployment *deployment, const char *group_name)
+{
```
denalex, line 631: if this is a public function, we should check for nulls on both params.

extension-common/pxfconfigparser.c
```diff
@@ -0,0 +1,661 @@
...
+    ListCell *lc_cluster;
```
denalex, line 632:
```suggestion
    ListCell *lc;
```

extension-common/pxfconfigparser.c
```diff
@@ -0,0 +1,661 @@
...
+ */
+static pxf_service_group *get_pxf_service_group_in_cluster(pxf_cluster *cluster, const char *group_name)
+{
+    ListCell *lc;
+    foreach (lc, cluster->groups)
+    {
+        pxf_service_group *group_entry = lfirst(lc);
+        if (strcmp(group_entry->name, group_name) == 0)
+        {
+            return group_entry;
```
denalex, line 656:
```suggestion
        pxf_service_group *group = lfirst(lc);
        if (strcmp(group->name, group_name) == 0)
        {
            return group;
```

extension-common/pxfconfigparser.c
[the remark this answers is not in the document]
djianwen-vmware: > I meant syntactically, there are 2 lines of quoted text, how does it become 1 line without any operands ?:
> 
> ```
> errhint("the value of 'cluster.name' should be the same as [cluster-name] "
> "in its config file path '${PXF_BASE}/clusters/[cluster-name]/conf/cluster.txt'")
> ```

I think this is a supported syntax in C to split a string into multi lines. There are two ways of doing it: https://stackoverflow.com/questions/1135841/how-to-write-a-multi-line-string-literal

extension-common/pxfconfigparser.c
```diff
@@ -0,0 +1,661 @@
...
+populate_config_property(pxf_cluster *cluster, const parser_line_context *ctx)
+{
+    // Create a copy of the original key for tokenizing, since tokenizing modifies the key
+    char *key_copy = pstrdup(ctx->key);
+
+    List *tokens = get_tokens(key_copy, ".");
+    if (list_length(tokens) > MAX_NUM_CONFIG_KEY_TOKENS ||
+        list_length(tokens) < MIN_NUM_CONFIG_KEY_TOKENS)
+    {
+        pfree(key_copy);
```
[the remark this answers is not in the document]
djianwen-vmware: > do we need to `list_free(tokens)` ?

Good catch. Thanks!

extension-common/pxfconfigparser.c
djianwen-vmware: > it's unfortunate `pfree` wants a not-NULL pointer since regular `free(NULL)` is ok: https://stackoverflow.com/questions/6084218/is-it-good-practice-to-free-a-null-pointer-in-c

Yes, `free(NULL)` is allowed but `pfree(NULL)` isn't. I tried it and got a segment fault.

extension-common/pxfconfigparser.c
djianwen-vmware: > why not make it `char*` since there are no other attributes for the host ? Then you can just do `list_clean_deep` here instead of iterating over hosts list.

I followed the same data struct in Golang. Initially in Golang code I made it a string but the team gave me a feedback to make it a separate struct for consistency and extendability.

extension-common/pxfconfigparser.c
djianwen-vmware: > while this is generally a good practice, you're pfree-ing the whole deployment right after, so there is not much point of setting this pointer to NULL.

Yes, deployment will be freed, but it still accessible until some other struct takes its place in memory. I followed the convention in other cleanup functions in the repo. Let me know if you have a preference here.

extension-common/pxfconfigparser.c
djianwen-vmware: > if this is a public function, we should check for nulls on both params.

Good point. Thanks!

PR #1144, opened 2024-05-09
Made 'pxf (cluster) prepare' command work in a multi-cluster architecture

denalex:
This PR makes the PXF CLI and the pxf bash script support a multi-cluster architecture. In this first phase the functionality is limited to having only 1 cluster with the name "default". Therefore, a "cluster name" parameter is not added at this time.

Changes include:
- pxf bash script `prepare` command now creates `$PXF_BASE/clusters/default` directory and its subdirectories and copies templates from `$PXF_HOME/templates/cluster` and `$PXF_HOME/templates/conf` into respective subdirectories.
- `$PXF_BASE` value set by user is now allowed to be the same as `$PXF_HOME` value
- PXF CLI go tool now requires both `PXF_BASE and `PXF_HOME` environment variables to be set, which they are if it is invoked from the pxf bash script
- internal refactoring of the PXF CLI go tool and unit tests
- PXF CLI go tool now supports running commands across both Greenplum and PXF external clusters with the reporting of status indicating the count of PXF external hosts the command will run on. The outcome reporting also include the count of PXF external hosts as well as any errors that might have occurred there.


cli/cmd/cluster.go
```diff
@@ -47,76 +45,190 @@ func createCobraCommand(use string, short string, cmd *PxfCommand) *cobra.Comman
...
-		return
-	}
-	standbyMsg := ""
-	numHosts := clusterData.NumHosts
-	if cmd.runOnMaster() {
-		numHosts--
+	// the cmd.message.status will be of the form "Doing something on"
+	// so we need to fill out the host categories and the number of hosts for each category
+	willRunOnGreenplumCluster := cmd.runOnGreenplumCluster(clusterData)
+	willRunOnCoordinator := willRunOnGreenplumCluster && cmd.runOnCoordinator()
```
djianwen-vmware, line 138: When will `cmd.runOnGreenplumCluster` be false but `cmd.runOnCoordinator` be true? 
Logically, commands running on coordinator should be a subset of running on greenplum, so checking `runOnCoordinator` should be enough?

cli/cmd/cluster.go
```diff
@@ -47,76 +45,190 @@ func createCobraCommand(use string, short string, cmd *PxfCommand) *cobra.Comman
...
+	willRunOnStandbyCoordinator := willRunOnGreenplumCluster && cmd.runOnStandbyCoordinator() && isStandbyAloneOnHost(clusterData)
+	willRunOnPxfExternalHosts := cmd.runOnPxfExternalCluster(clusterData)
+	// calculate how many Greenplum segment hosts are there by considering all hosts, but the coordinator
+	// the sync command will already have discounted the coordinator, but others will account for it
+	numSegmentHosts := clusterData.NumGreenplumHostsToRun - 1
+	if cmd.name == sync {
+		numSegmentHosts++
```
djianwen-vmware, line 145: This part is a little bit confusing for me. Just to clarify, if it is `sync`, do we want to include coordinator in the num of hosts? We have special computation on `NumGreenplumHostsToRun`, is this because `clusterData.NumGreenplumHostsToRun = len(greenplumCommandList)` is not an accurate assignment?

cli/cmd/pxf.go
```diff
@@ -29,253 +48,177 @@ type messages struct {
...
+// A command might need to run on the hosts where PXF is installed, but not on the hosts where Greenplum is installed.
+// In this case of a collocated PXF cluster, the command will actually run on the Greenplum hosts, so the boolean
+// flags `runOnGreenplum` and `runOnPxf` are prescriptive and not prohibitive.
+// If a flag is set to false, the command does not need to run on the hosts that are part of the respective cluster.
+// But if it ends up running there because the clusters are collocated, that is ok.
+type PxfCommand struct {
+	name           commandName
+	messages       messages
+	GreenplumScope cluster.Scope
+	runOnGreenplum bool // whether the command needs to run where Greenplum is installed
```
djianwen-vmware, line 64: Just a thought, if we can update `cluster.Scope` to include one more case: `none`, some of our code can be simplified, for example`runOnGreenplum` and `runOnPxf` are not necessary then. We just need a `GreenplumScope` and a `PxfScope`. Some if conditions can also simpler.

cli/cmd/cluster.go
```diff
@@ -47,76 +45,190 @@ func createCobraCommand(use string, short string, cmd *PxfCommand) *cobra.Comman
...
+	}
+	var greenplumCommandList, pxfCommandList []cluster.ShellCommand
+	// if the PXF cluster is collocated, or if the command needs to run on the Greenplum cluster anyway
+	if cmd.runOnGreenplumCluster(clusterData) {
+		greenplumCommandList = clusterData.GreenplumCluster.GenerateSSHCommandList(cmd.GreenplumScope, functionToExecute)
+		clusterData.NumGreenplumHostsToRun = len(greenplumCommandList)
+	}
+	// if the PXF cluster is external and the command needs to run there
+	if cmd.runOnPxfExternalCluster(clusterData) {
+		pxfCommandList = clusterData.PxfExternalCluster.GenerateSSHCommandList(ALL_PXF_HOSTS, functionToExecute)
```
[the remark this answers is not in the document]
denalex: for now yes, all PXF external hosts are the same and if a command needs to run on any external PXF host, it needs to run on all of them.

cli/cmd/cluster.go
denalex: I think you're correct. There is some special case with the `sync` command that excludes coordinator, so it will not run on coordinator (it will not target coordinator, to be precise, it will still run on the coordinator host, though). The semantics are that `cmd.runOnGreenplumCluster` / `cmd.runOnPxfExternalCluster` are derived from the facts of presence of the respective cluster and boolean flags for commands based on out tooling design. The `runOnCoordinator` is derived from Greenplum Scope for Greenplum cluster only. While intuitively your argument is correct, I am filling more confident with the double check here as an extra protection from future potential changes on these 2 disjoined ways to express metadata.

cli/cmd/cluster.go
```diff
@@ -47,76 +45,190 @@ func createCobraCommand(use string, short string, cmd *PxfCommand) *cobra.Comman
...
-	if cmd.runOnMaster() {
-		numHosts--
+	// the cmd.message.status will be of the form "Doing something on"
+	// so we need to fill out the host categories and the number of hosts for each category
+	willRunOnGreenplumCluster := cmd.runOnGreenplumCluster(clusterData)
+	willRunOnCoordinator := willRunOnGreenplumCluster && cmd.runOnCoordinator()
+	willRunOnStandbyCoordinator := willRunOnGreenplumCluster && cmd.runOnStandbyCoordinator() && isStandbyAloneOnHost(clusterData)
+	willRunOnPxfExternalHosts := cmd.runOnPxfExternalCluster(clusterData)
+	// calculate how many Greenplum segment hosts are there by considering all hosts, but the coordinator
+	// the sync command will already have discounted the coordinator, but others will account for it
```
axue-broadcom, line 142: why does the sync command already discount the coordinator? can we make them all consistent?

cli/cmd/cluster.go
```diff
@@ -135,8 +247,46 @@ func (cmd *PxfCommand) GenerateOutput(clusterData *ClusterData) error {
...
 
-	return fmt.Errorf(response)
+func logCommandOutput(output *cluster.RemoteOutput) {
+	if isDebugEnabled() {
+		gplog.Debug("Remote command output - scope=%d", output.Scope)
+		for _, command := range output.Commands {
+			hasFailed := command.Error != nil
+			gplog.Debug("host=%s, command=%q, failed=%t, stdout=%q", command.Host, command.CommandString, hasFailed, command.Stdout)
+			if hasFailed {
+				gplog.Debug("    stderr=%q", command.Stderr)
```
axue-broadcom, line 260: why do we have so many spaces in front? 
```suggestion
				gplog.Debug(" stderr=%q", command.Stderr)
```

cli/cmd/cluster.go
denalex: hmm, I can't explain this logic at this point either :) Maybe it was some intermediate thing I had to do, but it passed the unit tests.

I can try changing it back closer to the original logic, which surprisingly also passes the unit tests:
```
	numSegmentHosts := clusterData.NumGreenplumHostsToRun
	if willRunOnCoordinator {
		numSegmentHosts--
	}
``` 
This newer (original) logic says that the SSH commands will be generated for all remote nodes that are targeted for the run. If the coordinator is included, then there will be an SSH command for it as well I believe, even though it runs on the local machine, so maybe the library will not generate `ssh` part but instead use `bash`. But there will be an entry for this command for the coordinator, so we have to decrease the count by one to treat the other hosts as segment hosts.

cli/cmd/cluster.go
```diff
@@ -135,8 +247,46 @@ func (cmd *PxfCommand) GenerateOutput(clusterData *ClusterData) error {
...
+
+func (cmd *PxfCommand) runOnGreenplumCluster(clusterData *ClusterData) bool {
+	// the command will run on the Greenplum cluster if the PXF cluster is collocated
+	// or if the command is marked as needed to run on Greenplum cluster even if the PXF cluster is external
+	return clusterData.PxfExternalCluster == nil || cmd.runOnGreenplum
+}
+
+func (cmd *PxfCommand) runOnPxfExternalCluster(clusterData *ClusterData) bool {
+	// the command will run on the PXF cluster if the PXF cluster is external (not collocated)
+	// and if the command is marked as needed to run on the PXF cluster even if the PXF cluster is external
```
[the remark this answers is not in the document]
denalex: no, `cmd.runOnPxf` means run on PXF if the PXF is EXTERNAL. For example the `register` command is marked as false, meaning if the PXF cluster is external, it will not run there.

cli/cmd/pxf.go
denalex: we cannot easily change `cluster.Scope` as it is a bitmap from another package. and even if we were able to do, I don't think it is that trivial to represent the current boolean flags using scopes as the meaning is a bit different -- scope means which types of hosts to run on while booleans mean where to run if the cluster is external.

cli/cmd/pxf.go
```diff
@@ -29,253 +48,177 @@ type messages struct {
...
-}
-
-func (cmd *PxfCommand) GetFunctionToExecute() (func(string) string, error) {
-	inputs, err := cmd.makeValidCliInputs()
-	if err != nil {
-		return nil, err
-	}
-	cmd.envVarMap = inputs
+// if messages.standby is defined, then messages.status is a printf-style format string that takes three arguments
+// otherwise the format string takes two arguments
```
axue-broadcom, line 52: should these comments be part of lines 54-59 instead of hovering above

cli/cmd/cluster.go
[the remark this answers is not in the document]
djianwen-vmware: > no, `cmd.runOnPxf` means run on PXF if the PXF is EXTERNAL. For example the `register` command is marked as false, meaning if the PXF cluster is external, it will not run there.

I see. Sorry I missed the "&" symbol.

cli/cmd/cluster.go
djianwen-vmware: > I think you're correct. There is some special case with the `sync` command that excludes coordinator, so it will not run on coordinator (it will not target coordinator, to be precise, it will still run on the coordinator host, though). The semantics are that `cmd.runOnGreenplumCluster` / `cmd.runOnPxfExternalCluster` are derived from the facts of presence of the respective cluster and boolean flags for commands based on our tooling design. The `runOnCoordinator` is derived from Greenplum Scope for Greenplum cluster only. While intuitively your argument is correct, I am feeling more confident with the double check here as an extra protection from future potential changes on these 2 disjoined ways to express metadata.

Yes, I understand it is better to be defensive in function. However, my concern is that, as we add complexity to the codebase, we are breaking some assumptions on the functions, and it may become less sustainable with more features added. For example, `cmd.runOnCoordinator()` was designed to check "if the command should be run on coordinator". But today, after introducing external clusters, it only checks "if the command should be run on coordinator `when it is greenplum scope`".  But the name of the function didn't change. This kind of breaks the encapsulation philosophy, because now if I want to check if the command is to run on coordinator, I need to read the underlining implementation detail of `runOnCoordinator` to see what extra conditions to check, other than calling the function itself.
My proposal is to hide the extra condition inside the function as implementation detail, or create a new function to do it, rather than double check everything whenever I want to use it. What do you think?

cli/cmd/cluster.go
djianwen-vmware: > hmm, I can't explain this logic at this point either :) Maybe it was some intermediate thing I had to do, but it passed the unit tests.
> 
> I can try changing it back closer to the original logic, which surprisingly also passes the unit tests:
> 
> ```
> 	numSegmentHosts := clusterData.NumGreenplumHostsToRun
> 	if willRunOnCoordinator {
> 		numSegmentHosts--
> 	}
> ```
> 
> This newer (original) logic says that the SSH commands will be generated for all remote nodes that are targeted for the run. If the coordinator is included, then there will be an SSH command for it as well I believe, even though it runs on the local machine, so maybe the library will not generate `ssh` part but instead use `bash`. But there will be an entry for this command for the coordinator, so we have to decrease the count by one to treat the other hosts as segment hosts.

Thanks for the clarification. So `numSegmentHosts ` actually means `numSegmentHostsToRunSSH`?

cli/cmd/pxf.go
djianwen-vmware: > we cannot easily change `cluster.Scope` as it is a bitmap from another package. and even if we were able to do, I don't think it is that trivial to represent the current boolean flags using scopes as the meaning is a bit different -- scope means which types of hosts to run on while booleans mean where to run if the cluster is external. I will change the comment for line 65 to express that more clearly.

Yes, I saw the bitmap. I was just thinking about adding a new "none" flag to the bitmap which represents do not run on any host. Then the bool=false can be represented as flag=none. bool=true can be represented as flag=anything else. The flag itself can play as an bool. But if it is difficult to change the `cluster.Scope`, we don't need to worry about it.

cli/cmd/cluster.go
```diff
@@ -47,76 +45,190 @@ func createCobraCommand(use string, short string, cmd *PxfCommand) *cobra.Comman
...
+		for _, c := range greenplumCommandList {
+			hostList = append(hostList, c.Host)
+		}
+		for _, c := range pxfCommandList {
+			hostList = append(hostList, c.Host)
+		}
+		gplog.Verbose("Running '%s' on hosts: %s", functionToExecute("HOSTNAME"), strings.Join(hostList, ", "))
+	}
+	// execute the command on the Greenplum cluster first and then on the PXF cluster, if applicable
+	if clusterData.NumGreenplumHostsToRun > 0 {
```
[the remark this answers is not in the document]
denalex: They cannot be less than 0 since it gets populated as `len(SSHCommandsToRun)` which cannot be negative. It can be 0 if we only have an external cluster and we are doing start / stop, no action on Greenplum nodes will be taken. So I believe nothing needs to change here.

cli/cmd/cluster.go
```diff
@@ -47,76 +45,190 @@ func createCobraCommand(use string, short string, cmd *PxfCommand) *cobra.Comman
...
-	}
-	standbyMsg := ""
-	numHosts := clusterData.NumHosts
-	if cmd.runOnMaster() {
-		numHosts--
+	// the cmd.message.status will be of the form "Doing something on"
+	// so we need to fill out the host categories and the number of hosts for each category
+	willRunOnGreenplumCluster := cmd.runOnGreenplumCluster(clusterData)
+	willRunOnCoordinator := willRunOnGreenplumCluster && cmd.runOnCoordinator()
+	willRunOnStandbyCoordinator := willRunOnGreenplumCluster && cmd.runOnStandbyCoordinator() && isStandbyAloneOnHost(clusterData)
```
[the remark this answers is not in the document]
denalex: this is an old function and I think the name actually matches exactly what it does -- check if there is a standby and if there is nothing else (segments or primary coordinator) on its host.

cli/cmd/cluster.go
denalex: I will remove this sentence from the comments, its confusing. the sync command is the only one that does not target the coordinator as we are copying filed from the coordinator to the other hosts.

cli/cmd/cluster.go
denalex: this is intentional, since this is a continuation of the previous message, but it goes on a separate line, it is indented with spaces.

cli/cmd/pxf.go
denalex: thanks, I removed these lines, they are no longer valid and were leftover from the previous code.

cli/cmd/cluster.go
denalex: yes

cli/cmd/pxf.go
denalex: discussed with @djianwen-vmware and we decided to leave this as is as changing `cluster.Scope` is not feasible

cli/cmd/cluster.go
denalex: discussed with @djianwen-vmware and refactored similar to the proposal here

cli/cmd/cluster.go
```diff
@@ -47,76 +45,190 @@ func createCobraCommand(use string, short string, cmd *PxfCommand) *cobra.Comman
...
+	// generate and display verbose and debug details about commands to be run
+	if isDebugEnabled() {
+		for _, c := range greenplumCommandList {
+			gplog.Debug("Running '%s' on '%s'", c.CommandString, c.Host)
+		}
+		for _, c := range pxfCommandList {
+			gplog.Debug("Running '%s' on '%s'", c.CommandString, c.Host)
+		}
+	} else {
+		hostList := make([]string, clusterData.NumGreenplumHostsToRun, clusterData.NumGreenplumHostsToRun+clusterData.NumPxfExternalHostsToRun)
```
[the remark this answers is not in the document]
denalex: good catch, it should be length 0 to start with

tatb-vmware (approved):

axue-broadcom (approved):

cli/cmd/pxf.go
```diff
@@ -29,253 +48,174 @@ type messages struct {
...
-			status:  "Checking status of PXF servers on coordinator host%s and %d segment host%s...\n",
+			status:  "Checking status of PXF servers on",
 			standby: ", standby coordinator host,",
 			err:     "PXF is not running on %d out of %d hosts\n",
 		},
-		warn:       false,
-		envVars:    []envVar{pxfHome, pxfBase},
-		whereToRun: cluster.ON_REMOTE | cluster.ON_HOSTS | cluster.INCLUDE_MASTER | cluster.INCLUDE_MIRRORS,
+		GreenplumScope: ALL_GREENPLUM_HOSTS,
+		runOnGreenplum: false, // stop should run only on the hosts where PXF is installed
```
yimingli-vmware:
```suggestion
		runOnGreenplum: false, // status should run only on the hosts where PXF is installed
```

cli/cmd/cluster.go
```diff
@@ -47,76 +45,189 @@ func createCobraCommand(use string, short string, cmd *PxfCommand) *cobra.Comman
...
+		return err
 	}
-	return "s"
+	// ask the PxfCommand for a function that can be used to produce an actual command to run
+	functionToExecute, err := pxfCommand.GetFunctionToExecute()
+	if err != nil {
+		return err
+	}
+	var greenplumCommandList, pxfCommandList []cluster.ShellCommand
+	// if the PXF cluster is collocated, or if the command needs to run on the Greenplum cluster anyway
```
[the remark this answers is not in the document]
denalex: `command needs to run on the Greenplum cluster anyway` means that it needs to run there even if PXF is external, for example the `register` command will run on Greenplum even if the PXF cluster is external.

djianwen-vmware (approved):

yimingli-vmware (approved):

PR #1145, opened 2024-05-10
Custom guc

axue-broadcom:
This PR adds a custom PXF guc called pxf.pxf_base. The default value for this GUC is set to empty string and can only be
set via `gpconfig --change pxf.pxf_base --value <PXF_BASE_DIR> --skipvalidation`. Setting this guc requires a configuration reload `gpstop --u`.

Upon first use of the extension in a new session, the GUC will be loaded and a directory validation check will occur. A warning will be printed if the directory does not exist.


external-table/src/pxfprotocol.c
```diff
@@ -46,6 +48,70 @@ static gphadoop_context *create_context(PG_FUNCTION_ARGS, bool is_import);
...
+		gettext_noop("PXF_BASE directory containing PXF configuration files"),
+		NULL,
+		&PxfBase,
+		"",
+		PGC_SIGHUP,
+		0,
+		check_pxf_base, NULL, NULL);
+}
+
+static bool
```
denalex, line 75: can we add a function comment here explaining what it checks and when it runs

external-table/src/pxfprotocol.c
```diff
@@ -46,6 +48,70 @@ static gphadoop_context *create_context(PG_FUNCTION_ARGS, bool is_import);
...
+		 * will not be set. the check function runs even for the initialization of the default value
+		 * the assumption is that pxf register will take care of setting this guc
+		 */
+		if (source == PGC_S_DEFAULT)
+			return true;
+		else
+		{
+			GUC_check_errcode(ERRCODE_CONFIG_FILE_ERROR);
+			GUC_check_errmsg("invalid value for parameter \"pxf.pxf_base\": parameter cannot be empty");
+			GUC_check_errhint("pxf.pxf_base cannot be empty. Please run `pxf register` to set it.");
```
denalex, line 92:
```suggestion
			GUC_check_errhint("Please run \"pxf register\" command to set \"pxf.pxf_base\" parameter.");
```
I don't think we need to repeat "pxf.pxf_base cannot be empty" since this is what the error message says already.

external-table/src/pxfprotocol.c
```diff
@@ -46,6 +48,70 @@ static gphadoop_context *create_context(PG_FUNCTION_ARGS, bool is_import);
...
+			return false;
+		}
+	}
+	else
```
denalex, line 96: can we remove "else" since both branches under "if" end with return statements

external-table/src/pxfprotocol.c
```diff
@@ -46,6 +48,70 @@ static gphadoop_context *create_context(PG_FUNCTION_ARGS, bool is_import);
...
+	{
+		struct stat statbuf;
+		if (stat(*newval, &statbuf) != 0)
+		{
+			if (errno == ENOENT)
+			{
+				GUC_check_errcode(ERRCODE_CONFIG_FILE_ERROR);
+				GUC_check_errmsg("invalid value for parameter \"pxf.pxf_base\": directory \"%s\" does not exist", *newval);
+				GUC_check_errhint("Please ensure the directory exists. Please re-run `pxf register`.");
```
denalex, line 105:
```suggestion
				GUC_check_errhint("Please ensure the directory exists and re-run \"pxf register\" command.");
```

external-table/src/pxfprotocol.c
```diff
@@ -46,6 +48,70 @@ static gphadoop_context *create_context(PG_FUNCTION_ARGS, bool is_import);
...
+			GUC_check_errcode(ERRCODE_CONFIG_FILE_ERROR);
+			GUC_check_errmsg("invalid value for parameter \"pxf.pxf_base\": parameter cannot be empty");
+			GUC_check_errhint("pxf.pxf_base cannot be empty. Please run `pxf register` to set it.");
+			return false;
+		}
+	}
+	else
+	{
+		struct stat statbuf;
+		if (stat(*newval, &statbuf) != 0)
```
denalex, line 99: would be nice to have a comment to explain what is happening here

concourse/scripts/pxf_common.bash
```diff
@@ -115,6 +115,7 @@ function run_pxf_automation() {
 		export PXF_HOME=${PXF_HOME}
 		export PGPORT=${PGPORT}
 		export USE_FDW=${USE_FDW}
+		export MASTER_DATA_DIRECTORY=/home/gpadmin/data/master/gpseg-1
```
denalex, line 118: I wonder why this was needed

automation/src/test/java/org/greenplum/pxf/automation/features/custom_guc/CustomGucTest.java
```diff
@@ -0,0 +1,82 @@
...
+    };
+
+    private ReadableExternalTable externalTable;
+
+    @Override
+    protected void beforeMethod() throws Exception {
+        super.beforeMethod();
+        cluster.runCommand("source $GPHOME/greenplum_path.sh");
+        String PXF_HOME = System.getenv("PXF_HOME");
+        // set the dynamic library path since we use an absolute path in the contorl file
```
denalex, line 28:
```suggestion
        // set the dynamic library path since we use an absolute path in the control file
```

external-table/src/pxfprotocol.c
```diff
@@ -46,6 +48,70 @@ static gphadoop_context *create_context(PG_FUNCTION_ARGS, bool is_import);
...
+
+/*
+ * _PG_init gets called when the extension is loaded.
+ */
+void
+_PG_init(void)
+{
+	ereport(DEBUG1, (errmsg("PXF extension loaded.")));
+	DefineCustomStringVariable(
+		"pxf.pxf_base",
```
denalex, line 65: we would need to have the `pxf-fdw` extension to support this as well. can the same GUC be used by both extensions ?

automation/src/test/java/org/greenplum/pxf/automation/features/custom_guc/CustomGucTest.java
```diff
@@ -0,0 +1,82 @@
...
+
+    @Test(groups = {"features", "gpdb"})
+    public void testCustomGucPXFBaseNotExist() throws Exception {
+        cluster.runCommand("gpconfig -c pxf.pxf_base  -v \"'/tmp/pxf-base-nonexistent'\" --skipvalidation; gpstop -u");
+        sleep(1000);
+
+        runSqlTest("features/custom_guc/guc_non_existent");
+    }
+
+    private String prepareData() throws Exception {
```
denalex, line 66: you don't have to create files anymore, there is now a `system:sample` demo profile that will generate data on the fly.

automation/sqlrepo/features/custom_guc/guc_valid_value/expected/query01.ans
```diff
@@ -0,0 +1,54 @@
...
+(1 row)
+
+SET client_min_messages = 'DEBUG1';
+SET
+
+LOAD 'pxf';
+DEBUG1:  "pxf.pxf_base" value is set to /tmp/pxf-base
+DEBUG1:  "pxf.pxf_base" value is set to /tmp/pxf-base
+DEBUG1:  "pxf.pxf_base" value is set to /tmp/pxf-base
+DEBUG1:  "pxf.pxf_base" value is set to /tmp/pxf-base
```
denalex, line 31: why is this shown 4 times ? If due to the cluster size, will the test fail on clusters of different sizes ?

denalex (commented):
need to add similar behavior to the pxf-fdw extension

concourse/scripts/pxf_common.bash
axue-broadcom: The automation test requires `gpconfig` commands to be run and `gpconfig` requiresthis to be set.

automation/sqlrepo/features/custom_guc/guc_non_existent/expected/query01.ans
```diff
@@ -0,0 +1,42 @@
+-- @description query01 for PXF guc nonexistent directory test on small data
+-- start_matchsubs
+--
+-- m{.*WARNING:  invalid value for parameter "pxf.pxf_base".*}
+-- s{.*WARNING:  invalid value for parameter "pxf.pxf_base"}{WARNING:  invalid value for parameter "pxf.pxf_base"}
```
tatb-vmware, line 5: Can you explain how this syntax works? I am confused on what it is doing. I thought the sub uses `/` to denote what to substitute.

external-table/src/pxfprotocol.c
```diff
@@ -46,6 +48,70 @@ static gphadoop_context *create_context(PG_FUNCTION_ARGS, bool is_import);
...
+		gettext_noop("PXF_BASE directory containing PXF configuration files"),
+		NULL,
+		&PxfBase,
+		"",
+		PGC_SIGHUP,
```
tatb-vmware, line 70: What is the purpose of this value, `PGC_SIGHUP`?

external-table/src/pxfprotocol.c
```diff
@@ -46,6 +48,70 @@ static gphadoop_context *create_context(PG_FUNCTION_ARGS, bool is_import);
...
+{
+	if (newval && strcmp(*newval, "") == 0)
+	{
+		/*
+		 * don't do anything if we have the default value
+		 * if we add something here, there will be an err and the guc value
+		 * will not be set. the check function runs even for the initialization of the default value
+		 * the assumption is that pxf register will take care of setting this guc
+		 */
+		if (source == PGC_S_DEFAULT)
```
tatb-vmware, line 86: Similar question, what is `PGC_S_DEFAULT`?

automation/src/test/java/org/greenplum/pxf/automation/features/custom_guc/CustomGucTest.java
```diff
@@ -0,0 +1,82 @@
...
+        String location = prepareData();
+        createReadablePxfTable("default", location);
+
+        runSqlTest("features/custom_guc/guc_valid_value");
+    }
+
+    @Test(groups = {"features", "gpdb"})
+    public void testCustomGucPXFBaseNotExist() throws Exception {
+        cluster.runCommand("gpconfig -c pxf.pxf_base  -v \"'/tmp/pxf-base-nonexistent'\" --skipvalidation; gpstop -u");
+        sleep(1000);
```
tatb-vmware, line 61: Why are we sleeping for a second here?

external-table/src/pxfprotocol.c
```diff
@@ -46,6 +48,70 @@ static gphadoop_context *create_context(PG_FUNCTION_ARGS, bool is_import);
...
+/* GUC variables */
+char *PxfBase = NULL;
+
+/*
+ * _PG_init gets called when the extension is loaded.
+ */
+void
+_PG_init(void)
+{
+	ereport(DEBUG1, (errmsg("PXF extension loaded.")));
```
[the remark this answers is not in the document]
axue-broadcom: Yes. There are 3 different debug levels and this is the most basic.

external-table/src/pxfprotocol.c
axue-broadcom: `PGC_SIGHUP` signifies that this guc can only be loaded via config file and requires a configuration reload but not a full cluster restart. It cannot be set in a session and is a cluster level guc

external-table/src/pxfprotocol.c
axue-broadcom: `PGC_S_DEFAULT` refers to the guc's default value. If the source of the guc (where the value is coming from) is default, then it is ok for the guc value to be empty string. If the source of the guc is not default -- aka it came from a config file `PCG_S_FILE` or something else (see guc.c in gpdb code) -- then we want to error out.

automation/sqlrepo/features/custom_guc/guc_valid_value/expected/query01.ans
axue-broadcom: Correct. These debug prints happen per segment.

automation/src/test/java/org/greenplum/pxf/automation/features/custom_guc/CustomGucTest.java
```diff
@@ -0,0 +1,82 @@
...
+        if (gpdb != null && externalTable != null) {
+            gpdb.dropTable(externalTable, true);
+        }
+        super.afterMethod();
+    }
+
+    @Test(groups = {"features", "gpdb"})
+    public void testCustomGucValidPXFBase() throws Exception {
+        cluster.runCommand("mkdir -p /tmp/pxf-base");
+        cluster.runCommand("gpconfig -c pxf.pxf_base  -v \"'/tmp/pxf-base'\" --skipvalidation; gpstop -u");
```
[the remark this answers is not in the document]
axue-broadcom: This is because we have to skip validation. We define this guc as a custom string guc. However, as a custom guc, thedefinition lives as part of an extension. `gpconfig` has no way of validating the guc value and we must have `--skipvalidation` in order to be able to set the guc. As such, we need to do the work to make sure that it makes it into the gpconfig file as the proper type. So we have both the single quotes and the double quotes. On line 29, the guc being set (`dynamic_library_path`) is a known guc that gpdb validates (there is no `--skipvalidation`. So we do not need to worry about the having both double and single quotes

automation/sqlrepo/features/custom_guc/guc_non_existent/expected/query01.ans
axue-broadcom: The syntax here is 
```
m {text}
s{text}{replacement} 
```
And is equivalent to 
```
m /text/
s /text/replacement/
```

match in items in {}
substitute things in first {} with things in second {}

automation/src/test/java/org/greenplum/pxf/automation/features/custom_guc/CustomGucTest.java
axue-broadcom: we want to make sure that there was enough time for the config to reload before we run any tests

external-table/src/pxfprotocol.c
axue-broadcom: The same guc cannot be used by both extensions

fdw/pxf_fdw.c
```diff
@@ -105,6 +107,81 @@ static CopyState BeginCopyTo(Relation forrel, List *options);
 static void PxfBeginScanErrorCallback(void *arg);
 static void PxfCopyFromErrorCallback(void *arg);
 
+/*
+ * GUC related declarations
+ */
+void _PG_init(void);
+static bool check_pxf_base(char **newval, void **extra, GucSource source);
+char *PxfBase = NULL;
```
denalex, line 115: will this be a global variable and if yes, would it conflict with the similar one from the external table extension ?

denalex (approved):

yimingli-vmware (approved):

tatb-vmware (approved):

fdw/pxf_fdw.c
axue-broadcom: preliminary testing shows that they are treated separately. I think is global in the scope of the extension.

djianwen-vmware (approved):

PR #1147, opened 2024-05-11
Provision external pxf ccp cluster

yimingli-vmware:
- Provide 2 extra nodes for GP6 multi-node tasks with impersonation
- Add another standalone CLI test `test-pxf-gp6-cli-rocky8`.  Now both `test-pxf-gp6-cli-rocky8` and `test-pxf-gp7-cli-rocky8` will run cli test for collocated PXF and external PXF. Nested cli tests will only test collocated pxf.


concourse/pipelines/templates/dev_build_pipeline-tpl.yml
```diff
@@ -234,10 +234,9 @@ jobs:
     {% endif %}
 {% endcall %}
 
-## ---------- GP7 PXF CLI tests ----------
+## ---------- PXF CLI tests ----------
 {% call(x) macros.for_each_config(build_test_pxf_combinations) %}
-    {# for now Makefile only defines the switch for GP7 #}
-    {% if x.test_cli is sameas true and x.gp_ver == '7' and gp7_cli %}
+    {% if x.test_cli is sameas true and gp6_cli and gp7_cli %}
```
denalex, line 239: this requires both `gp6_cli` and `gp7_cli` to be true. Can we make it such that it is generates for gp6 or gp7 or both depending on the flags set ? You might need to create a new dictionary similar to `requested_multi` on line 13.

concourse/pipelines/templates/jobs/test-multinode-tpl.yml
```diff
@@ -67,6 +68,9 @@
             WITH_MIRRORS: [[ccp_default_conf]]
             number_of_nodes: ((number_of_gpdb_nodes))
             extra_nodes: 1
+{% if x.with_external_pxf == 'true' %}
+            extra_nodes: 3
+{% endif %}
```
denalex, line 73: will `extra_nodes: 3` silently override `extra_nodes: 1` set just before it ? If yes, would it be better to include it in the else clause to be more obvious ?

yimingli-vmware: I tried to use `if-else` at first, but it complains `Key 'extra_nodes' is duplicated`. `extra_nodes: 3` will silently override the previous one.

denalex: it should've worked, we have an example here: https://github.com/greenplum-db/pxf/blob/main/concourse/pipelines/templates/jobs/compatibility-gate-tpl.yml#L28

you can also use a temporary variable like here: https://github.com/greenplum-db/pxf/blob/main/concourse/pipelines/templates/jobs/test-tpl.yml#L24

Ideally we would want to avoid a silent override and be explicit in our intentions

yimingli-vmware: Why the `if-else` is working inside `compatibility-gate-tpl.yml` is because it is nested in the for loop. I defined another metadata `extra_nodes` in this file and set it by different conditions, and use `extra_nodes: [[extra_nodes]]`

concourse/scripts/install_pxf.bash
```diff
@@ -246,7 +246,7 @@ function run_pxf_installer_scripts() {
 			gpconfig -c gp_hadoop_target_version -v hdp
 			gpstop -u
 		fi &&
-		sed -i '/edw/d' hostfile_all &&
+		sed -i '/edw0/d' hostfile_all &&
```
axue-broadcom, line 249: why is this change necessary?

yimingli-vmware: Before when there is no nodes for external pxf, the `hostfile_all` contained `cdw, sdw1, sdw2, edw0`. `edw0` is reserved for singlecluster and no pxf should be run there. Therefore before copying files to nodes, we want to make sure the `hostfile_all` only contains gp segment hosts.

Now we add external pxf, then the `hostfile_all` contains `cdw, sdw1, sdw2, edw0, edw1, edw2`. Now we need to include `edw1, edw2`, so here we only delete the `edw0` for singlecluster.

yimingli-vmware: From PR 1150, we decide to make `pxf cluster prepare` only runs on the collocated pxf, so I'll revert the change back to make the `hostfile_all` only contains gpdb segment hosts even if there are external pxf nodes.

concourse/pipelines/templates/build_pipeline-tpl.yml
```diff
@@ -282,10 +282,9 @@ jobs:
     {% endif %}
 {% endcall %}
 
-## ---------- GP7 PXF CLI tests ----------
-{# include a CLI test, for now Makefile only defines the switch for GP7 #}
+## ---------- PXF CLI tests ----------
```
denalex, line 285:
```suggestion
## ---------- PXF CLI tests with external PXF cluster ----------
```

concourse/pipelines/templates/dev_build_pipeline-tpl.yml
```diff
@@ -234,10 +237,9 @@ jobs:
     {% endif %}
 {% endcall %}
 
-## ---------- GP7 PXF CLI tests ----------
+## ---------- PXF CLI tests ----------
```
denalex, line 240:
```suggestion
## ---------- PXF CLI tests with external PXF cluster ----------
```

concourse/Makefile
```diff
@@ -39,6 +39,7 @@ GS                          ?= false
 MINIO                       ?= false
 OEL7                        ?= false
 FILE                        ?= false
+GP6_CLI                     ?= false
 GP7_CLI                     ?= false
```
denalex, line 43:
```suggestion
CLI_GP6                     ?= false
CLI_GP7                     ?= false
```
should we switch such that the name of the feature comes first and the version specifier last ?

concourse/pipelines/templates/dev_build_pipeline-tpl.yml
```diff
@@ -12,6 +12,9 @@
 {# define whether multinode testing is requested for given EL OS versions #}
 {% set requested_multi = {'7': multinode_el7, '8': multinode_el8, '9': multinode_el9} %}
 
+{# define whether external pxf testing is requested for given Greenplum versions #}
+{% set requested_external_pxf = {'6': gp6_cli, '7': gp7_cli} %}
```
denalex, line 16:
```suggestion
{% set requested_cli = {'6': cli_gp6, '7': cli_gp7} %}
```
here we are requesting CLI tests for given GP versions as semantically specified by variable names (GP6_CLI and GP7_CLI). The fact that they will be run with external PXF or not is decided internally.

Also, maybe we should switch the order in the names to list the feature first and then version it applies to last

denalex (approved):

djianwen-vmware (approved):

tatb-vmware (approved):

PR #1148, opened 2024-05-14
Add hosts file parsing

tatb-vmware:
This PR adds parsing of the optional `hosts.txt` file for a given PXF cluster.

Changes include:
- Add implementation for parsing and creating the list of `PxfHosts`
- Refactor parsing and creating both cluster and hosts into respective functions called `createCluster()` and `createHosts()`
- Consolidate `PxfCluster` parsing and creating logic. Before, the file parsing was outside the loop and cluster creation was inside the loop. Although we currently support a single PXF cluster, this change will be needed once we have support multiple clusters.
- Rename file related variables and functions to differentiate between the cluster file and hosts file
- Add godocs comments for `parser.go`
- Add test helper function to easily set the errors for `ReadFileAndClose()`


cli/config/parser.go
```diff
@@ -165,10 +192,25 @@ func (c *PxfCluster) setGroupProperty(groupName, propertyName, val string) error
...
+// parseAndCreateHosts creates a slice of PxfHost given the hosts file contents.
+func parseAndCreateHosts(fileContents []string) []*PxfHost {
+	parsedHosts := make([]*PxfHost, 0, len(fileContents))
+	for _, line := range fileContents {
+		line = strings.ToLower(line)
+		line = strings.TrimSpace(line)
+		// Remove any line that is a TOML comment
+		if line == "" || line[0] == '#' {
+			continue
+		}
```
tatb-vmware, line 212: I'm debating whether or not I should refactor this part into `readFile()` since it is repeated here and within `parseClusterFile()`.

Thoughts?

cli/config/parser_test.go
```diff
@@ -79,70 +105,73 @@ var _ = Describe("ParsePxfCluster tests", func() {
...
+			clusterFileContents = []string{"group.default.ssl=not-a-boolean"}
 			expectClusterParseErrorWithClusterHint(
 				"failed to parse cluster configuration file : configuration property 'group.default.ssl' is set to a non-boolean value 'not-a-boolean'")
 		})
 	})
 	Context("log warnings", func() {
 		It("logs a warning for an unknown property", func() {
-			fileContents = addLineToNormalClusterFileContents("unknown-key.foo=bar")
+			clusterFileContents = addLineToNormalClusterFileContents("unknown-key.foo=bar")
+			hostsFileContents = normalHostsFileContents
```
denalex, line 153: this line should go into `@ BeforeEach` for the given context

cli/config/parser.go
```diff
@@ -1,37 +1,57 @@
 package config
 
 import (
+	"errors"
```
denalex, line 4: why do we need this package now ?

cli/config/parser.go
```diff
@@ -1,37 +1,57 @@
...
+		hostsFileContents, err := HostFileProvider(hostsPath)
+		// Since the hosts file is optional, we don't care if the error is "file does not exist".
+		if err != nil && !errors.Is(err, os.ErrNotExist) {
+			return nil, NewPxfError(
+				// TODO: in the future add cluster name to the hint message 'pxf cluster prepare <cluster-name>'
+				"please check your PXF_BASE variable. Did you run 'pxf cluster prepare' command?",
+				"could not read configuration file %s : %w", hostsPath, err)
 		}
+
+		pxfCluster.Hosts = parseAndCreateHosts(hostsFileContents)
```
denalex, line 54: there is some asymmetry between:
- parseCluster and then createCluster
- parseAndCreateHosts
can we merge `parseCluster` and `createCluster` into 1 function `parseCluster` ? Maybe the function also takes a  "reader" interface that will actually provide contents and can be mocked ? That will eliminate the concept of "contents" at this level and will be internal to the `parseXYZ` function ? Also look at gp-common-go-lib.

cli/config/parser.go
```diff
@@ -1,37 +1,57 @@
...
-	}
-	contents, err := parseFile(fileContents)
-	if err != nil {
-		return nil, NewClusterConfigError(clusterName,"failed to parse cluster configuration file : %w", err)
-	}
 	var pxfCluster *PxfCluster
 	for _, currentClusterName := range clusterNames {
-		pxfCluster, err = parseCluster(currentClusterName, contents)
+		clusterPath := GetPathForClusterFile(currentClusterName)
+		clusterFileContents, err := ClusterFileProvider(clusterPath)
```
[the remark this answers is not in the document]
tatb-vmware: Let's go with `content` since it's a single file whose content we've extracted out.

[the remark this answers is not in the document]
tatb-vmware: Since we're referring to the countable objects within the file, I think it's `contents`.

(Source: https://english.stackexchange.com/questions/56831/file-content-vs-file-contents)

cli/config/parser.go
tatb-vmware: `errors.Is()` is now used to inspect the error returned from `HostFileProvider()` to see if it's a `s.ErrNotExist`. This may change once I switch to using gp-common-go-libs for reading the file.

cli/config/parser.go
tatb-vmware: I agree. I ended up combining the parsing and creating of hosts because each function was very small, but now it introduces asymmetry.

We can merge `parseCluster()` and `createCluster()`, but one concern I have is that the function will be long and complex. I can test it out and see what it ends up looking like.

cli/config/parser.go
```diff
@@ -1,37 +1,57 @@
...
+// HostFileProvider allows us to inject host fileContents in our unit test.
+var HostFileProvider = readFile
+
+// ParsePxfCluster creates a PxfCluster from the given name.
+// It assumes that each clusters' configuration files are stored within $PXF_BASE/clusters/[clusterName]/conf
+// The hosts file is optional.
+// Currently, it only supports a single PxfCluster with the name "default".
 func ParsePxfCluster(clusterName string) (*PxfCluster, error) {
 	pxfDeployment := PxfDeployment{Clusters: make(map[string]*PxfCluster)}
 	clusterNames := []string{"default"}
```
yimingli-vmware, line 24: Right now we hard code `clusterNames` to be only one value, if we start to support more clusters, where do we get this `clusterNames`? From the function input?

cli/config/parser.go
```diff
@@ -1,37 +1,57 @@
...
+		clusterFileContents, err := ClusterFileProvider(clusterPath)
+		if err != nil {
+			return nil, NewPxfError(
+				// TODO: in the future add cluster name to the hint message 'pxf cluster prepare <cluster-name>'
+				"please check your PXF_BASE variable. Did you run 'pxf cluster prepare' command?",
+				"could not read configuration file %s : %w", clusterPath, err)
+		}
+		clusterContents, err := parseClusterFileContents(clusterFileContents)
+		if err != nil {
+			return nil, NewClusterConfigError(clusterName, "failed to parse cluster configuration file : %w", err)
```
yimingli-vmware, line 37:
```suggestion
			return nil, NewClusterConfigError(currentClusterName, "failed to parse cluster configuration file : %w", err)
```
We are using `currentClusterName` for each iteration

cli/config/parser.go
```diff
@@ -40,6 +60,7 @@ func ParsePxfCluster(clusterName string) (*PxfCluster, error) {
 	return pxfDeployment.Clusters[clusterName], nil
```
yimingli-vmware, line 60: Inside the for loop we make `currentClusterName` to be hardcoded `default`. For the testing if we call `ParsePxfCluster("foo")`, `pxfDeployment.Clusters["foo"]` will be returned but it is never being set.

cli/config/parser_test.go
```diff
@@ -38,38 +50,52 @@ var normalPxfCluster = config.PxfCluster{
 
 var _ = Describe("ParsePxfCluster tests", func() {
```
yimingli-vmware, line 51: Can we also test these cases:
1. hosts.txt is not provided 
2. hosts.txt is empty
3. user try to parse a cluster name which should not exist

cli/config/parser.go
```diff
@@ -40,6 +60,7 @@ func ParsePxfCluster(clusterName string) (*PxfCluster, error) {
...
 }
 
+// readFile creates a slice of lines from the file within the given path.
```
axue-broadcom, line 63:
```suggestion
// readFile creates a slice containing all the lines from the file within the given path.
```

cli/config/parser.go
```diff
@@ -165,10 +191,24 @@ func (c *PxfCluster) setGroupProperty(groupName, propertyName, val string) error
...
 
+// assertNotEmpty checks if a given value is empty and returns the formatted error.
 func assertNotEmpty(val string, errorMsg string) error {
 	if val == "" {
 		return fmt.Errorf("configuration property '%s' has an empty value", errorMsg)
 	}
 	return nil
 }
 
+// parseAndCreateHosts creates a slice of PxfHost given the hosts file contents.
```
axue-broadcom, line 202:
```suggestion
// parseAndCreateHosts creates a slice of PxfHost objects given the hosts file contents.
```

cli/config/parser_test.go
```diff
@@ -38,38 +50,52 @@ var normalPxfCluster = config.PxfCluster{
...
 				"HINT: please check your PXF_BASE variable. Did you run 'pxf cluster prepare' command?")
 		})
+		It("returns an error when failing to read hosts file", func() {
+			clusterFileContents = normalClusterFileContents
+			config.HostFileProvider = func(string) ([]string, error) {
+				return nil, fmt.Errorf("test error")
+			}
+			expectClusterParseError(
+				"could not read configuration file /test/base/path/clusters/default/conf/hosts.txt : test error",
+				"HINT: please check your PXF_BASE variable. Did you run 'pxf cluster prepare' command?")
```
axue-broadcom, line 83: a read err could also be a permissions thing right? not just that the directory is missing. do we need to separate out these 2 cases?

cli/config/parser.go
```diff
@@ -1,37 +1,57 @@
...
+				"could not read configuration file %s : %w", clusterPath, err)
+		}
+		clusterContents, err := parseClusterFileContents(clusterFileContents)
+		if err != nil {
+			return nil, NewClusterConfigError(clusterName, "failed to parse cluster configuration file : %w", err)
+		}
+		pxfCluster, err = createCluster(currentClusterName, clusterContents)
 		if err != nil {
-			return nil, NewClusterConfigError(clusterName,"failed to parse cluster configuration file : %w", err)
+			return nil, NewClusterConfigError(clusterName, "failed to parse cluster configuration file : %w", err)
```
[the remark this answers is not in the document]
tatb-vmware: I'm not sure if that is the correct error message we want to display to the user. The user does not have a concept of a `PxfCluster` and we want to provide feedback that makes them from their perspective. Additionally, the error that is getting wrapped may mention information that relates to the text file. example: `fmt.Errorf("found more than one character '=' on line %d: %s", lineNum+1, line)`

cli/config/parser.go
```diff
@@ -1,37 +1,57 @@
...
 	"strings"
 )
 
-// ReadFileProvider allows us to inject fileContents in our unit test
-var ReadFileProvider = readFile
+// ClusterFileProvider allows us to inject cluster fileContents in our unit test.
+var ClusterFileProvider = readFile
 
+// HostFileProvider allows us to inject host fileContents in our unit test.
+var HostFileProvider = readFile
```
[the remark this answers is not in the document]
tatb-vmware: With our current implementation, it would be difficult. As suggested by @denalex, I am looking into using `gp-common-go-libs` to see what they provide for mocking file IO.

cli/config/parser.go
```diff
@@ -1,37 +1,57 @@
...
+		if err != nil {
+			return nil, NewClusterConfigError(clusterName, "failed to parse cluster configuration file : %w", err)
+		}
+		pxfCluster, err = createCluster(currentClusterName, clusterContents)
 		if err != nil {
-			return nil, NewClusterConfigError(clusterName,"failed to parse cluster configuration file : %w", err)
+			return nil, NewClusterConfigError(clusterName, "failed to parse cluster configuration file : %w", err)
 		}
+
+		hostsPath := GetPathForHostsFile(currentClusterName)
```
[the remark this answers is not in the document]
tatb-vmware: They are both defined in `config.go`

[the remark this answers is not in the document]
tatb-vmware: We don't have logic to skip it. What will happen is that the returned `fileContent` will be a slice with no elements, hence making the loop a no-op. I'm not sure if an explicit conditional to check would be needed.

cli/config/parser.go
tatb-vmware: You're correct. In the future when we have support for multi-cluster, we plan to scan the filesystem and find all the cluster directories to populate this `clusterNames` list. For now, I can add a TODO to make note of that.

cli/config/parser.go
tatb-vmware: That's correct, I will add a test for that since we currently only support a cluster name with "default".

cli/config/parser_test.go
tatb-vmware: Good idea, I will add those in.

cli/config/parser.go
```diff
@@ -69,53 +63,39 @@ func readFile(path string) ([]string, error) {
...
-	for key, val := range parsedKeyVal {
+	for key, val := range clusterMap {
 		tokens := strings.Split(key, ".")
 		switch tokens[0] {
 		case "cluster":
 			if len(tokens) != 2 {
 				return nil, fmt.Errorf("invalid configuration property '%s'", key)
 			}
-			if err := cluster.setClusterProperty(tokens[1], val, clusterName); err != nil {
+			if err = cluster.setClusterProperty(tokens[1], val, clusterName); err != nil {
```
axue-broadcom, line 88: why are we no longer doing a declaration and assignment (`:=`) here and the lines below?

cli/config/parser.go
```diff
@@ -128,6 +108,32 @@ func createCluster(clusterName string, parsedKeyVal map[string]string) (*PxfClus
...
+	for lineNum, line := range fileContents {
+		line = strings.TrimSpace(strings.ToLower(line))
+		// Remove any line that is a TOML comment
+		if line == "" || line[0] == '#' {
+			continue
+		}
+		lineKeyVal := strings.Split(line, "=")
+		// lineKeyVal[0] = key, lineKeyVal[1] = value, anything else is invalid
+		if len(lineKeyVal) != 2 {
+			return nil, fmt.Errorf("found more than one character '=' on line %d: %s", lineNum+1, line)
```
axue-broadcom, line 123:
```suggestion
			return nil, fmt.Errorf("found more than one character '=' on line %d: %s", lineNum + 1, line)
```

nit: consistency of spaces between operators here and lines below

tatb-vmware: `lineNum+1` is enforced by `gofmt` because they want to show operator precedence (sorta) and enforce consistency.

Here is the Go source code that handles formatting: https://github.com/golang/go/blob/master/src/go/printer/nodes.go#L734

Email group discussion:
https://groups.google.com/g/golang-nuts/c/Uqu6IVLDSeI/m/7xpPLrPxMHAJ

cli/config/parser.go
tatb-vmware: Because the old `createCluster` and `parseClusterFileContents` has been combined into the new `createCluster`, there is an `err` variable that is instantiated earlier in the code. Now when we hit this line, we just reassign the value of the `err` instead of creating a new one.

cli/config/parser.go
```diff
@@ -40,60 +53,39 @@ func ParsePxfCluster(clusterName string) (*PxfCluster, error) {
...
-		}
-		key := strings.TrimSpace(lineKeyVal[0])
-		if key == "" {
-			return nil, fmt.Errorf("found empty property on line %d: %s", lineNum+1, line)
-		}
-		if _, exist := parsedFileKeyVals[key]; exist {
-			return nil, fmt.Errorf("found duplicate property on line %d: %s", lineNum+1, key)
-		}
-		parsedFileKeyVals[key] = strings.TrimSpace(lineKeyVal[1])
+	clusterMap, err := createClusterMap(fileContents)
```
denalex, line 66:
```suggestion
	clusterPropertiesMap, err := createClusterPropertiesMap(fileContents)
```

cli/config/parser.go
```diff
@@ -40,60 +53,39 @@ func ParsePxfCluster(clusterName string) (*PxfCluster, error) {
 	return pxfDeployment.Clusters[clusterName], nil
 }
 
-func readFile(path string) ([]string, error) {
-	data, err := os.ReadFile(path)
+// createCluster creates a PxfCluster given the cluster name and file contents of the cluster file.
+func createCluster(clusterName string) (*PxfCluster, error) {
+	clusterPath := GetPathForClusterFile(clusterName)
+	fileContents, err := FileProvider(clusterPath)
```
denalex, line 59:
```suggestion
	fileContents, err := ReadFileAndClose(clusterPath)
```
this name would be more intuitive to read, it indicates an action rather than a noun

cli/config/parser.go
```diff
@@ -165,10 +187,33 @@ func (c *PxfCluster) setGroupProperty(groupName, propertyName, val string) error
...
+// assertNotEmpty checks if a given value is empty and returns the formatted error.
 func assertNotEmpty(val string, errorMsg string) error {
 	if val == "" {
 		return fmt.Errorf("configuration property '%s' has an empty value", errorMsg)
 	}
 	return nil
 }
 
+// createHosts creates a slice of PxfHosts given the hosts file contents.
+func createHosts(clusterName string) ([]*PxfHost, error) {
```
denalex, line 199:
```suggestion
func createHosts(clusterName string) ([]*PxfHost, error) {
```
I wonder why pointers: `[]*PxfHost` and not `[]PxfHost`

denalex (approved):

axue-broadcom (approved):

cli/config/parser.go
```diff
@@ -1,37 +1,50 @@
...
-	var pxfCluster *PxfCluster
 	for _, currentClusterName := range clusterNames {
-		pxfCluster, err = parseCluster(currentClusterName, contents)
+		pxfCluster, err := createCluster(currentClusterName)
+		if err != nil {
+			// errors.As takes in a pointer to an interface and does reflection to see if
+			// the err (concrete error type) is assignable to t (pointer to custom error interface)
+			// to check if err already has a hint
+			var t ErrorWithHint
+			if errors.As(err, &t) {
```
yimingli-vmware, line 37: Is this the only place we need to check whether the error already has a hint?

yimingli-vmware (approved):

cli/config/parser.go
```diff
@@ -1,37 +1,50 @@
...
+			"Cluster '%s' not found", clusterName)
 	}
-	var pxfCluster *PxfCluster
 	for _, currentClusterName := range clusterNames {
-		pxfCluster, err = parseCluster(currentClusterName, contents)
+		pxfCluster, err := createCluster(currentClusterName)
+		if err != nil {
+			// errors.As takes in a pointer to an interface and does reflection to see if
+			// the err (concrete error type) is assignable to t (pointer to custom error interface)
+			// to check if err already has a hint
```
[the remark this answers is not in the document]
tatb-vmware: You're correct, it is documented in `errors.As`.  I err on the side of caution and added the comment just in case. I can trim it down to explain what we're using it for but leave the implementation details out.

cli/config/parser.go
tatb-vmware: Yes because there is a special error case within `createCluster()` where we want a different hint compared to the other error cases. Hence, we wrap that particular error within `createCluster()` then check at this level to see if it has already been wrapped.

cli/config/parser.go
tatb-vmware: We use a slice of pointers because we wanted uniformity with the maps we hold within `PxfDeployment` and `PxfCluster` struct definitions. Maps have a constraint where if you want to modify the reference within, you need the map to hold pointers. Slices don't have the same constraint but we still use pointers to make it consistent.

PR #1150, opened 2024-05-17
Made pxf cluster prepare run on coordinator only

denalex:
The PR makes the following changes:
- `pxf cluster prepare` command now runs on coordinator only
- it and potentially other commands do not validate existence of the cluster configuration
- the full success output from the command that ran on coordinator only is now shown to the user
- added unit tests
- renamed receiver variable for pxfCommand
- removed installing some directories into PXF_HOME as they now belong in a cluster-specific directory


cli/cmd/cluster.go
```diff
@@ -129,42 +129,58 @@ func clusterRun(pxfCommand *PxfCommand, clusterData *ClusterData) error {
 	return pxfCommand.GenerateOutput(clusterData)
 }
 
+// GenerateGreenplumCommands creates commands to be run on the Greenplum cluster
+func GenerateGreenplumCommands(pxfCommand *PxfCommand, clusterData *ClusterData, functionToExecute func(string) string) []cluster.ShellCommand {
+	// if the command needs to run on coordinator only, we need to do further processing.
+	// since Scope assumes the command will always be run on segment hosts,
+	// we need to create a command for the coordinator to run.
+	if pxfCommand.runOnCoordinatorOnly {
```
djianwen-vmware, line 137: It feels like the greenplum scope defined by `gp-common-go-lib` doesn't meet our needs today very well. We are creating more and more flags now. Do you think it's worth it to build our own bitmap `PxfScope`? Such as coordinator=1, standbyCoordinator=2,GpSegments=4,externalHost=8, then we create extra constants which are combinations of them.

djianwen-vmware (approved):

cli/cmd/cluster.go
denalex: yes, we could do this. I am not sure it will end up simpler since we will have to turn on individual bits and then convert to `gp-common-go-lib` one when we call their APIs. In either case, it will be out of scope of this PR.

cli/cmd/cluster_test.go
```diff
@@ -120,7 +120,7 @@ var _ = Describe("GenerateStatusReport()", func() {
 				},
 				Entry("When syncing"     , cmd.SyncCommand,     2, "Syncing PXF configuration files from coordinator host to 2 segment hosts..."),
 				Entry("When registering" , cmd.RegisterCommand, 3, "Installing PXF extension on coordinator host and 2 segment hosts..."),
-				Entry("When preparing"   , cmd.PrepareCommand,  3, "Preparing PXF on coordinator host and 2 segment hosts..."),
+				Entry("When preparing"   , cmd.PrepareCommand,  1, "Preparing PXF on coordinator host..."),
```
yimingli-vmware, line 123: Why the number is changed to `1`? I know now `pxf cluster prepare` should only be run on the coordinator, but I think it can still prepare the pxf for the other gp segment hosts? Is there any piece I missed here?

cli/cmd/cluster.go
```diff
@@ -210,10 +226,18 @@ func (cmd *PxfCommand) GenerateOutput(clusterData *ClusterData) error {
 	totalErrors := numGreenplumErrors + numPxfErrors
 	totalHostsToRun := clusterData.NumGreenplumHostsToRun + clusterData.NumPxfExternalHostsToRun
 	if totalErrors == 0 {
-		gplog.Success(cmd.messages.success, totalHostsToRun, totalHostsToRun)
+		gplog.Success(pxfCommand.messages.success, totalHostsToRun, totalHostsToRun)
+		if pxfCommand.runOnCoordinatorOnly && totalHostsToRun == 1 &&
```
tatb-vmware, line 230: I'm curious why `totalHostsToRun==1` is being checked here and what situation would cause this. Does that mean Greenplum and PXF are collocated and we are only running the `pxfCommand` on the shared coordinator?

tatb-vmware (approved):

server/pxf-service/src/scripts/pxf
```diff
@@ -317,10 +317,10 @@ function doPrepare() {
     mkdir -p "${PXF_BASE}" || fail "Unable to create the \$PXF_BASE directory '${PXF_BASE}'"
 
     # verify the cluster directory is empty
-    if [[ -d ${cluster_dir} ]] && [[ "$(ls -A ${cluster_dir})" ]]; then
-        echoRed "ERROR: The cluster directory '$cluster_dir' is not empty."
-        echoRed "       Did you already run 'pxf prepare'? Is your \$PXF_BASE environment variable set correctly?"
-        echoRed "       Ensure that the \$PXF_BASE/clusters/${cluster_name} directory is empty."
+    if [[ -d ${cluster_dir} ]] && [[ $(ls -A "${cluster_dir}") ]]; then
+        echoRed "ERROR: The cluster directory '$cluster_dir' is not empty. Did you already run 'pxf prepare'?"
+        echoRed "       Your PXF_BASE is '${PXF_BASE}', if incorrect, set PXF_BASE environment variable and re-run the command."
```
axue-broadcom, line 322:
```suggestion
        echoRed "       Your PXF_BASE is '${PXF_BASE}'. If this is incorrect, set the PXF_BASE environment variable and re-run the command."
```

cli/cmd/pxf.go
```diff
@@ -103,80 +101,76 @@ var (
 		GreenplumScope:               SYNC_GREENPLUM_HOSTS,
```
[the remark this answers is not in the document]
denalex: no, even though it runs on Coordinator, it targets all other hosts. Again, perhaps better names could be used, but I can't think of any :)

cli/cmd/cluster_test.go
denalex: it could've, but this PR changed it such that `prepare` runs on coordinator host only so that any GP and external host have the same configuration.

cli/cmd/cluster.go
```diff
@@ -210,10 +226,18 @@ func (cmd *PxfCommand) GenerateOutput(clusterData *ClusterData) error {
...
+			clusterData.GreenplumClusterOutput != nil && len(clusterData.GreenplumClusterOutput.Commands) == 1 {
```
axue-broadcom, line 231: there's a lot of AND's in this conditional. can we add a comment here why we all of these are needed?

cli/cmd/cluster.go
denalex: this is super-extra precaution to make sure this is only done when the command SHOULD have run on coordinator (`pxfCommand.runOnCoordinatorOnly`) and actually only 1 command has been generated as the result of that (`totalHostsToRun==1`). Technically not necessary, but serves as an extra safety check.

yimingli-vmware (approved):

